PR #8, opened 2020-02-19
Add basic distributed evaluation

danieljanes:
Please note: The seemingly non-DRY code in module `server.py` (compare functions `fit_clients` and `eval_clients`) is intentional. This part is likely to change quite significantly and keeping those implementations slightly non-DRY will make those changes easier.


danieljanes:
Also, this initial PR only covers loss aggregation, but no other metrics such as accuracy.

PR #24, opened 2020-02-21
Clearly specify derived class instead of base class as type

tanertopal:
When providing an instance of the derived class to the Server on instantiation mypy will otherwise complain.


tanertopal:
Closed as I was wrong and mypy was right to complain. Derived class was not so derived after all. Fixed in next PR.

PR #58, opened 2020-03-11
Add script to update gh-pages branch

tanertopal:


tanertopal:
We decided to host the docs outside of GitHub therefore this is obsolete.

PR #78, opened 2020-03-18
Add possibility to enable faulthandler

danieljanes:


tanertopal:
Using cmd argument for now.

PR #160, opened 2020-04-20
Train Fashion-MNIST benchmark using a custom training loop

danieljanes:


danieljanes:
@tanertopal thanks :)

PR #380, opened 2020-08-10
Add missing `-> None` type annotations, auto-format docstrings

### LINKED ISSUE #372: Replace individual mypy flags with `strict = True`
The mypy configuration in `mypy.ini` uses several flags to increase strictness.

To simplify this setup and automatically opt-in to upcoming strictness flags in future mypy versions, it would be preferable to replace those individual flags with the single `strict = True` setting.

danieljanes:
Fixes #372


danieljanes:
@tanertopal could you take a look at `src/py/flwr_experimental/ops/cluster.py`, line 77?

PR #405, opened 2020-09-23
Create Pytorch help functions for CIFAR 10

pedropgusmao:
The code augments the original dataset and partitions it into files having the same number of samples per
class. The original dataset is contained in the augmented one. 
Optionally, it generates thumbnails with the generated images.


src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
```diff
@@ -0,0 +1,218 @@
...
+    nb_partitions : int
+        Total number of partitions to be generated.
+    data_dir: str
+        Path to directory where the partition dataset files will be stored.
+    """
+
+    X, Y = dataset
+
+    X, Y = shuffle(X, Y)
+    X, Y = sort_by_label(X, Y)
```
tanertopal, line 136: I think this line is not required because of the next line the sorting order will be overridden anyway.

src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
```diff
@@ -0,0 +1,218 @@
...
+            augmented_imgs[idx, j, :, :, :] = combined_transform(img)
+        if isdir(save_thumbnails_dir):
+            tmp = [*augmented_imgs[idx, :, :, :, :]]
+            tmp2 = np.concatenate(tmp, axis=-1)
+            save_image(
+                from_numpy(tmp2), join(save_thumbnails_dir, f"cifar10_{idx}.jpg")
+            )
+
+    # Rearrange dataset. Every multiple of augment_factor is an original image.
+    X = augmented_imgs.reshape(-1, 3, 32, 32)
```
tanertopal, line 108: Not sure but shouldn't this be: `X = augmented_imgs.reshape(-1, (3, 32, 32))` or am I getting something wrong?

pedropgusmao: Tbh, I'm not sure if there is a practical difference. I expect it to generate a final tensor that is `(total number of images, 3, 32, 32)`.

src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
pedropgusmao: I guess it depends on what we want to do and if I understood the `sort_by_label` correctly. I think `sort_by_label` will not shuffle before reordering, right? `dataset` is highly correlated, meaning that for each original image found in the dataset, the following ones will be augmented version of those. I'm not sure, but if `sort_by_label` works as I expect, then the first client would have all the original images. This would actually be periodic. Every N clients would have only original images. On the other hand this could be useful because we could launch only clients with those `id`s and see if augmentation has a bad effect.

tanertopal: I might have been a bit unclear. The shuffle is totally fine and reasonable. What I mean is:

> line 136 => X, Y = sort_by_label(X, Y)
> line 137 => X, Y = sort_by_label_repeating(X, Y)

The `sort_by_label_repeating` will internally anyway call `sort_by_label` first thing.
https://github.com/adap/flower/blob/main/src/py/flwr_experimental/baseline/dataset/dataset.py#L46

Therefore the line 136 is not needed/redundant.

src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
tanertopal: It might be backwards compatibility why its working although I might be wrong. Based on the docs

* https://pytorch.org/docs/stable/tensors.html#torch.Tensor.reshape
* https://numpy.org/doc/stable/reference/generated/numpy.reshape.html#numpy.reshape

I would suggest we follow the documented behavior and put the desired shape into a tuple.

src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
pedropgusmao: Oh, I see. Okay. I'll change this

pedropgusmao: I see. I missed the connection with 137. I'm removing 136

src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
pedropgusmao: I see, the tuple, right? I'll change it to  `X = augmented_imgs.reshape((-1, 3, 32, 32))`

src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
pedropgusmao: Solved in https://github.com/adap/flower/commit/5235c38e8259f44f47f609a2a1ee36bbda7c1864

src/py/flwr_experimental/baseline/dataset/pytorch_cifar_partitioned.py
pedropgusmao: solved in https://github.com/adap/flower/commit/5235c38e8259f44f47f609a2a1ee36bbda7c1864

PR #417, opened 2020-10-08
Fix and improve docker container

tanertopal:


docker/sshd.Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN apt-get clean && \
 
 WORKDIR /root
 
-RUN pip install tensorflow-cpu==2.2.0 numpy==1.18.3
+RUN pip install tensorflow-cpu==2.2.0 torch==1.6.0 numpy==1.18.3
```
danieljanes, line 22: Perhaps better `RUN python3.7 -m pip ...`?
Also add `torchvision==0.7.0`

danieljanes (changes requested):

docker/sshd.Dockerfile
tanertopal: Great suggestion. Done! 👍

danieljanes (approved):

PR #427, opened 2020-10-09
Add MobileNetV2 model for CIFAR-10 dataset

tanertopal:
@danieljanes should we keep this code as is (copy/paste with reference to the original source) or rather extend it also with types instead of disabling type checking in the file.


tanertopal:
@danieljanes should we keep this code as is (copy/paste with reference to the original source) or rather extend it also with types instead of disabling type checking in the file.

PR #442, opened 2020-10-28
Upgrading torch to 1.7.0 and trochvision to 0.8.1

pedropgusmao:
Updating torch

### DIFF
--- a/docker/sshd.Dockerfile
+++ b/docker/sshd.Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get clean && \
 
 WORKDIR /root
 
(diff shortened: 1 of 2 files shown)


danieljanes (changes requested):
Both `torch` and `torchvision` should be marked as optional (as they were before)

pedropgusmao:
Okay, for my future reference, `poetry add` and `poetry remove` will cause the above behavior.

danieljanes:
Oh, I forgot to mention that `docker/sshd.Dockerfile` needs to be updated as well. Unfortunately, we don't have a unified way of setting this yet.

danieljanes (approved):

PR #506, opened 2020-12-04
Update README.md

jafermarq:
Removing the ` -b flower_example` from the shell snippet showing how to clone the repo and checkout the example.


tanertopal (approved):
Good catch! Thank you!

PR #534, opened 2020-12-14
Update examples/pytorch readme

Rubiel1:
I updated the example command, since this example is for pytorch, we are moving to pytorch directory not to tf.


danieljanes (approved):

danieljanes:
Good catch, thanks for the fix @Rubiel1 !

PR #543, opened 2020-12-28
Fix bug Model device

Riccardinho22:
Fixed the issue related to the right use of the device on the Net() model class.


danieljanes (approved):
Thanks for the fix @Riccardinho22!

Issue #583, opened 2021-01-12
Add clang-format to the respective format and test dev scripts


tanertopal:
Currently, we don't have autoformatting for the protobuf files. We need to add clang-format to the format.sh and test.sh scripts.

PR #584, opened 2021-01-12
Proto Linter | Add clang-format to Flower

CanTuerk:
This PR will add clang-format to `test.sh` and `format.sh` in the Flower repository. Refers to #583 .

### DIFF
--- a/.github/workflows/flower.yml
+++ b/.github/workflows/flower.yml
@@ -22,6 +22,7 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Install build tools
         run: |
+          sudo apt-get install clang-format-10
           python -m pip install -U pip==20.2.4
           python -m pip install -U setuptools==49.3.1
           python -m pip install -U poetry==1.1.4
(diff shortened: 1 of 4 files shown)


danieljanes (approved):

danieljanes (changes requested):
Looks good, we need to install clang-format on the CI though

danieljanes (approved):

tanertopal (approved):

pedropgusmao:
clang-format needs to be installed on the user side as well. I'm getting `xargs: clang-format-10: No such file or directory` when trying to run `dev/format.sh`

danieljanes:
Oh right, thanks for mentioning it. Should we add it to `bootstrap.sh`? Or should we run it optionally depending on whether the user has it installed or not?
@pedropgusmao @tanertopal

pedropgusmao:
I would run it optionally, but if the CI job fails, then I guess the developer would not be able to fix it?

danieljanes:
One other option would be to install it via pip: https://github.com/mgevaert/clang-format-wheel/

pedropgusmao:
Yes, I think that's a better solution it is well maintained.

danieljanes:
Follow-up issue #594

PR #596, opened 2021-01-18
Add dirichlet partition

pedropgusmao:


src/py/flwr/dataset/utils/common.py
```diff
@@ -204,3 +204,107 @@ def adjust_y_shape(nda: np.ndarray) -> np.ndarray:
...
+
+
+def create_dla_partitions(
+    dataset: XY,
+    dirichlet_dist: np.ndarray = np.empty(0),
+    num_partitions: int = 100,
+    concentration: float = 0.5,
+) -> Tuple[np.ndarray, XYList]:
+    """Create ibalanced non-iid partitions using Dirichlet Latent
+    Allocation(LDA) without resampling.
```
tanertopal, line 217:
```suggestion
    """Create imbalanced non-iid partitions using Latent Dirichlet
    Allocation (LDA) without resampling.
```

src/py/flwr/dataset/utils/common.py
```diff
@@ -204,3 +204,107 @@ def adjust_y_shape(nda: np.ndarray) -> np.ndarray:
...
+
+    Args:
+        dataset (XY): Datasets containing samples X
+            and labels Y.
```
tanertopal, line 221:
```suggestion
        dataset (XY): Dataset containing samples X and labels Y.
```

src/py/flwr/dataset/utils/common.py
```diff
@@ -204,3 +204,107 @@ def adjust_y_shape(nda: np.ndarray) -> np.ndarray:
...
+        dirichlet_dist (numpy.ndarray, optional): previously generated distribution to
+            be used. This s useful when applying the same distribution for train and
```
tanertopal, line 223:
```suggestion
            be used. This is useful when applying the same distribution for train and
```

src/py/flwr/dataset/utils/common.py
```diff
@@ -204,3 +204,107 @@ def adjust_y_shape(nda: np.ndarray) -> np.ndarray:
...
+            validation sets.
+        num_partitions (int, optional): Number of partitions to be created.
+            Defaults to 100.
+        concentration (float, optional): Dirichlet Concentration (:math:`\\alpha`)
+            parameter.
+            An :math:`\\alpha \\to \\Inf` generates uniform distributions over classes.
+            An :math:`\\alpha \\to 0.0` generates on class per client. Defaults to 0.5.
```
tanertopal, line 230:
```suggestion
            An :math:`\\alpha \\to 0.0` generates one class per client. Defaults to 0.5.
```

src/py/flwr/dataset/utils/common.py
```diff
@@ -205,3 +205,100 @@ def adjust_y_shape(nda: np.ndarray) -> np.ndarray:
...
+
+    Returns:
+        Tuple[numpy.ndarray, XYList]: List of XYList containing partitions
```
tanertopal, line 233: The return type does not match what is actually returned.

src/py/flwr/dataset/utils/common.py
```diff
@@ -205,3 +205,100 @@ def adjust_y_shape(nda: np.ndarray) -> np.ndarray:
...
+            sample_class: int = np.where(
+                np.random.multinomial(1, dirichlet_dist[partition_id]) == 1
+            )[0][0]
+            sample: np.ndarray = list_samples_per_class[sample_class].pop()
+
+            data[partition_id].append(sample)
+            target[partition_id].append(sample_class)
+
+            # If last sample of the class was drawn,
+            # then set pdf to zero for that class.
```
tanertopal, line 286:
```suggestion
            # If last sample of the class was drawn, then set the 
            #  probability density function (PDF) to zero for that class.
```

src/py/flwr/dataset/utils/common.py
```diff
@@ -205,3 +205,100 @@ def adjust_y_shape(nda: np.ndarray) -> np.ndarray:
     """Turn shape (x, 1) into (x)."""
     nda_adjusted = np.reshape(nda, (nda.shape[0]))
     return cast(np.ndarray, nda_adjusted)
+
+
+def create_dla_partitions(
```
tanertopal, line 210: Shouldn't this function rather be named `create_lda_partitions`?

tanertopal (changes requested):

src/py/flwr/dataset/utils/common.py
pedropgusmao: Yes, I got an error from Mypy. It's inverted

src/py/flwr/dataset/utils/common.py
pedropgusmao: Yep, changed to reflect the correct name.

pedropgusmao (commented):
Issues have been addressed

tanertopal (approved):

tanertopal (approved):

PR #602, opened 2021-01-19
Clang Format | Installation Shellscript

### LINKED ISSUE #594: Include clang-format in dev dependencies
`clang-format` is required to format `.proto` files. To prevent user users from having to install it via `sudo apt ...` we can include it in our dev dependencies (it's available via `pip`):

https://github.com/mgevaert/clang-format-wheel/

CanTuerk:
This PR will add `install-clang-format.sh` this will be useful to install and use `clang-format-10` which is also included in `test.sh`.
Will close  #594.


danieljanes (approved):

dev/install-clang-format.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/../
+
+# Check if Clang-Format is already installed
+if clang-format-10 --version | grep -q 'clang-format version 10.0.0-4ubuntu1'; 
+then
+   echo "Clang-Format is already installed on your Environment"
+else
+    # Install Clang-Format if it is not installed already
+    sudo apt install clang-format-10
```
tanertopal, line 11: Alternative proposal:

```suggestion
if ! command -v clang-format-10 &> /dev/null
then
    sudo apt install clang-format-10
    exit
fi
```

tanertopal (approved):

Issue #601, opened 2021-01-19
E2E test for examples


tanertopal:
# Description
Not all examples currently have a simple E2E test as seen in https://github.com/adap/flower/blob/main/.github/workflows/examples.yml

# Task
- [x] Create CI test for `examples/pytorch_from_centralized_to_federated`
- [x] Create CI test for `examples/simulation`

danieljanes:
We should definitely add this, but please make sure that both tests only use the minimal amount of CI resources possible.

PR #603, opened 2021-01-19
examples.yml | additional Tests

CanTuerk:
This PR will add additional tests to `example.yml` as described in #601 .


.github/workflows/examples.yml
```diff
@@ -117,3 +117,89 @@ jobs:
...
+          python -m poetry install
+      - name: Import flwr
+        run: |
+          python3 -c "import flwr"
+      - name: Test
+        run: |
+          cd examples/simulation
+
+          python3 simulation.py &
+          sleep 3
```
tanertopal, line 200:
```suggestion
          sleep 10
```
Let's give this slightly longer time as it will start the server and various clients.

tanertopal (changes requested):

tanertopal (approved):

tanertopal (approved):

Issue #599, opened 2021-01-19
FL Checkpoint generation


jafermarq:
It would be great to have a checkpointing mechanism built-in in the Flower server. This would serve a dual purpose: keep backups of the last N rounds (global model parameters, round number, strategy status, etc); and, make it easy to replicate the experiment and run the final evaluation of the global model at a later stage. 

I think as a starting point the following elements should go in the checkpoint:

- Model parameters.
- Current round number (so `server.fit()` starts from the round number stored in the checkpoint).
- Strategy-specific parameters. Some strategies might keep track of certain parameters during the training process (?) but others don't (e.g. FedAvg). 
- Anything else that changes over time.

By means of an additional flag (e.g. `--checkpoint <path_to_checkpoint>`) when launching the server, we can indicate that we are in fact resuming an experiment. 

To make the implementation of this functionality easier, it might be a good time to replace the use of `_get_initial_weights()` from the first client that connects to the server and, instead, use the model that lives in the server to initialise all the clients. This will make things easier when loading the model weights from a checkpoint.

pedropgusmao:
This should include handling SIGINT.

jafermarq:
Checkpoint handling would likely be framework-specific so maybe a good way to add this functionality is by adding a new server argument (e.g. `on_init`) that can be passed when creating the server object. This could be just a user-define function that gets executed right at the end of the server object constructor.

tanertopal:
@pedropgusmao @jafermarq @danieljanes is this kind of resolve with this: https://flower.dev/docs/saving-progress.html ?

danieljanes:
@tanertopal it's possible with the approach described in the docs, but my understanding is that the issue is about providing a built-in way to save checkpoints automatically

WilliamLindskog:
Seeing this is available https://flower.ai/docs/framework/how-to-save-and-load-model-checkpoints.html
Closing this.

PR #606, opened 2021-01-23
Extra server funcs

jafermarq:
This is a partial solution to issue #599. This change to `server.py` proposes using two (optional) arguments in the form of normal functions. These will be converted into methods upon server construction. These functions (then methods) are:

-  `on_init()`: which is called at the end of `__init__()`: Could be used when resuming training from a checkpoint (as motivated in #599)
- `on_round_end()`: which is called once a FL round is completed: Could be used to store a checkpoint (as motivated in #599)

Then, it would be up to the user to define exactly what should happen (if something at all) in these methods. Below I provide a simple example:

```python
# in standard python script configuring and launching the server
.... 
server = fl.server.Server(client_manager=fl.server.SimpleClientManager(),
                          strategy=strategy,
                          on_init_fn=init_fn(),
                          on_round_end_fn=end_round_fn())

fl.server.start_server(args.server_address, server, config={...},strategy=strategy)


def my_init_fn() -> Callable[[None], None]:
    def f(self) -> None:
        # Server ended __init__(), DO something?
        # (e.g. load weights from a checkpoint and assign to server.weights)
        pass
    return f


def my_end_round_fn() -> Callable[[Dict], None]:
    def f(self, args: Dict) -> None:
        # Round ended, DO something? (e.g. write a checkpoint)
        pass
    return f
```

Transforming these functions into server methods should give all the flexibility needed to implement the checkpoints storing/resuming functionality. However, this might feel a bit hacky. Would it be better to use normal functions? If so, we'll need to define what arguments should be passed to these. The down side of this approach is that the user might be restricted on what can be done (unless the user modifies the source code)


tanertopal:
@danieljanes I think we can reassign this to Javier? Not sure if its still relevant though.

jafermarq:
This is no longer relevant. These ideas can be better supported with the new driver api

PR #616, opened 2021-02-03
Single-process simulation using InMemoryClientProxy

danieljanes:


tanertopal:
@danieljanes closing this as we have now the approach with ray

PR #633, opened 2021-02-13
Make deprecated return values optional

danieljanes:


examples/advanced_tensorflow/client.py
```diff
@@ -56,14 +57,16 @@ def evaluate(self, parameters, config):
...
-
-        # Populate the 3rd return value with a 0.0 float, it'll be removed in the future
-        return len(self.x_test), loss, 0.0, {"accuracy": accuracy}
+        num_examples_test = len(self.x_test)
+        return loss, num_examples_test, {"accuracy": accuracy}
 
 
 def main() -> None:
+    # Sleep for 2s to give the server enough time to start
+    time.sleep(2)
```
tanertopal, line 68: This will impact each client start in the example. Is that intentional?

src/py/flwr/client/numpy_client.py
```diff
@@ -50,7 +50,10 @@ def get_parameters(self) -> List[np.ndarray]:
     @abstractmethod
     def fit(
         self, parameters: List[np.ndarray], config: Dict[str, Scalar]
-    ) -> Union[Tuple[List[np.ndarray], int], Tuple[List[np.ndarray], int, Metrics]]:
+    ) -> Union[
+        Tuple[List[np.ndarray], int, Dict[str, Scalar]],
```
tanertopal, line 54:
```suggestion
        Tuple[List[np.ndarray], int, Metrics],
```

src/py/flwr/client/numpy_client.py
```diff
@@ -65,15 +68,20 @@ def fit(
...
+                int, or str. It can be used to communicate arbitrary values
+                back to the server.
         """
 
     @abstractmethod
     def evaluate(
         self, parameters: List[np.ndarray], config: Dict[str, Scalar]
-    ) -> Union[Tuple[int, float, float], Tuple[int, float, float, Metrics]]:
+    ) -> Union[
+        Tuple[float, int, Dict[str, Scalar]],
```
tanertopal, line 81:
```suggestion
        Tuple[float, int, Metrics],
```

src/py/flwr/server/strategy/fast_and_slow.py
```diff
@@ -320,6 +320,7 @@ def aggregate_fit(
             # Track contributions to the global model
             for client, fit_res in results:
                 cid = client.cid
+                assert fit_res.num_examples_ceil is not None
```
tanertopal, line 323: Intentional to keep the asserts here?

examples/advanced_tensorflow/client.py
danieljanes: Yes, that's intentional - I had several issues with the clients trying to connect before the server was ready. Unfortunately the logs do not explain this well, so it can be a mystery for users to understand what's happening. We need to follow up with improved logging at a later point.

src/py/flwr/client/numpy_client.py
danieljanes: Discarding this - we've had this in the earlier version of the feature, but I don't want to narrow down the naming too much for now. We can make this more specific at a later stage.

src/py/flwr/client/numpy_client.py
danieljanes: Same as above.

src/py/flwr/server/strategy/fast_and_slow.py
danieljanes: Yes - this is not ideal, but part of the migration away from the deprecated values inside FitRes and EvaluateRes. The next step (i.e., one of the upcoming releases) will remove these properties and migrate to custom key/value pairs in the `dict` returned by `fit`/`evaluate`, which means that the strategies need to be (a) refactored to pull the necessary values from the `dict`, and (b) communicate what kind of key/value pairs each strategy expected to do their job.

tanertopal (approved):

tanertopal (approved):

PR #667, opened 2021-03-08
(de)-serialization for android

akhilmathurs:
Added serialization and de-serialization for tensors from java clients


danieljanes:
Thanks for the PR @akhilmathurs :) closing this since we have a different solution in place now

PR #673, opened 2021-03-11
Fix Mxnet loss and accuracy calculation

mariaboerner1987:
- change Loss to CrossEntropyLoss in "from_centralized_to_federated"
- create eval metric and add Loss and Accuracy
- adjust return values to eval metric


mariaboerner1987:
@danieljanes number of samples are updated now

danieljanes (approved):

PR #686, opened 2021-03-25
Add new example for upcoming blog post

tanertopal:


tanertopal:
I am closing this in favor of current developments. I think I need to update the first part of the blog series and write a subsequent part slightly different.

PR #696, opened 2021-04-09
Implementation of fedbn on client side

MeiruiJiang:
Implementation of FedBN
change model with BN layers added
modified fuctions: get/set_prameters


danieljanes:
@MeiruiJiang thanks for the PR, this looks great!
I'll run some regression tests and add a flag that enables users to enable/disable FedBN, then we can merge

MeiruiJiang:
> @MeiruiJiang thanks for the PR, this looks great!
> I'll run some regression tests and add a flag that enables users to enable/disable FedBN, then we can merge

Great! Thanks and look forward to the Flower with FedBN.

danieljanes (approved):
Will add the flag to enable/disable FedBN in a follow-up PR.

danieljanes:
Merged, thank again for the PR @MeiruiJiang!

PR #701, opened 2021-04-16
Leaf shakespeare

pedropgusmao:


baselines/flwr_baselines/models/leaf/shakespeare.py
```diff
@@ -0,0 +1,57 @@
...
+        super().__init__()
+        self.dict_size = len(chars)
+        self.seq_len = seq_len
+        self.hidden_size = hidden_size
+
+        self.encoder = nn.Embedding(self.dict_size, embedding_dim)
+        self.lstm = nn.LSTM(
+            input_size=embedding_dim, hidden_size=hidden_size, num_layers=2
+        )
+        self.decoder = nn.Linear(hidden_size * seq_len, self.dict_size)
```
jafermarq, line 47: This should be `nn.Linear(num_hidden, num_classes)`, we are predicting the next character in the play.

baselines/flwr_baselines/models/leaf/shakespeare.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class ShakespeareLeafNet(nn.Module):
+    def __init__(
+        self,
+        chars: str = LEAF_CHARACTERS,
+        seq_len: int = 80,
+        hidden_size: int = 256,
+        embedding_dim=8,
+    ):
```
jafermarq, line 29: need to pass the number of classes (i.e. number of potential next characters)

jafermarq: https://github.com/TalwalkarLab/leaf/blob/master/models/shakespeare/stacked_lstm.py

baselines/flwr_baselines/models/leaf/shakespeare.py
pedropgusmao: I'm getting it from the dictionary used by LEAF. So if the dictionary should change, then num_classes changes right after. What do you think?

jafermarq: sounds good 👍

baselines/flwr_baselines/scripts/leaf/shakespeare/create_datasets_and_splits.sh
```diff
@@ -0,0 +1,31 @@
...
+./preprocess.sh -s niid --sf 1.0 -k 2 -t sample --tf 0.8
+
+# Format for Flower experiments. Val/train fraction set to 0.25 so validation/total=0.20
+cd ${FLOWER_ROOT}/baselines/flwr_baselines/scripts/leaf/shakespeare
+python split_json_data.py \
+--save_root ${SAVE_ROOT}/shakespeare \
+--leaf_train_json ${LEAF_ROOT}/data/shakespeare/data/train/all_data_niid_0_keep_2_train_8.json \
+--val_frac 0.25 \
+--leaf_test_json ${LEAF_ROOT}/data/shakespeare/data/test/all_data_niid_0_keep_2_test_8.json
+echo 'Done'
```
tanertopal, line 31:
```suggestion
echo 'Done'

```

baselines/flwr_baselines/scripts/leaf/shakespeare/create_datasets_and_splits.sh
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+# Process Shakespeare for non-iid, 20% testset, minimum two lines
+# Delete previous partitions if needed then extract the entire dataset
+echo 'Deleting previous dataset split.'
+cd ${LEAF_ROOT}/data/shakespeare
```
tanertopal, line 19: I think it would be great to check if the env variable is set and otherwise warn and exit. Maybe something like this?:

```suggestion
if [[ -z "${LEAF_ROOT}" ]]; then
  echo 'The environment variable LEAF_ROOT is not set. Exiting...'
  exit 1
fi

cd ${LEAF_ROOT}/data/shakespeare
```

tanertopal (changes requested):

baselines/flwr_baselines/scripts/leaf/shakespeare/create_datasets_and_splits.sh
pedropgusmao: Nice

tanertopal (approved):

jafermarq (commented):
Looks good!!

PR #741, opened 2021-05-28
Update examples/embedded_devices

jafermarq:
Updating the example for RPi and NVIDIA-Jetson devices. Changes in the code involve:

- Updating to the latest Flower version
- Making use of `Metrics` return type for `fit()` and `evaluate()`
- Updated Docker images for Jetson devices (adding JetPack4.5 support)
- Other minor changes.


examples/embedded_devices/base_images/gpu/Dockerfile
```diff
@@ -2,25 +2,12 @@
...
+FROM nvcr.io/nvidia/l4t-pytorch:r32.5.0-pth1.6-py3
 
 RUN mkdir /app
 WORKDIR /app
 
 # update pip
-RUN python3 -m pip install --upgrade pip
-
-# Installing flower
-RUN pip3 install flwr==0.13.0
```
tanertopal, line 14: Is `flwr` installed in a later step instead anyway?

tanertopal (commented):
I tried to build the docker images but unfortunately, it fails. It seems to be an issue with the base images. Would it be possible to commit the base images also? Alternatively, extend the current docker image and use one of the public docker base images? e.g. ubuntu:20.04 or some debian etc.?

EDIT: My mistake. The images are already committed and it works!

examples/embedded_devices/base_images/gpu/Dockerfile
jafermarq: makes sense 👍

tanertopal (approved):

PR #742, opened 2021-06-04
Formatting: Fixing offending files 

pedropgusmao:


baselines/flwr_baselines/scripts/leaf/reddit/split_json_data.py
```diff
@@ -0,0 +1,134 @@
...
+            f"""Invalid partition fraction {fvalue}. This must be between 0 and 1."""
+        )
+    return fvalue
+
+
+def split_json_and_save(
+    list_datasets: List[Tuple[str, float]],
+    path_to_json: PathLike,
+    save_root: PathLike,
+    users_list: Optional[List[str]] = [],
```
tanertopal, line 37: Using mutable default arguments in Python is usually [discouraged](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

Better:

```python
users_list: Optional[List[str]] = Null
```
and in the function body
```python
if users_list is Null
  users_list = []
```

tanertopal (changes requested):

tanertopal (approved):

PR #753, opened 2021-06-08
Add option to skip final distributed evaluation

joshua-sterner:
This PR adds an optional argument `force_final_distributed_eval` to start_server so that the final distributed evaluation step can be skipped. Note that this currently defaults to true in order to avoid changing the current behavior of start_server.


src/py/flwr/server/app.py
```diff
@@ -76,7 +77,8 @@ def start_server(
         initialized_config["num_rounds"],
     )
 
-    _fl(server=initialized_server, config=initialized_config)
+    _fl(server=initialized_server, config=initialized_config,
+        force_final_distributed_eval=force_final_distributed_eval)
```
danieljanes, line 81: This will fail the formatting check - you can fix it by running `./dev/format.sh` (make sure to have the dev dependencies installed beforehand)

danieljanes (changes requested):
@joshua-sterner thanks for the contribution! This is useful. I left a comment about formatting.

danieljanes:
Hi @joshua-sterner , the CI fails because Pylint complains about there being too many arguments for `start_server` - in this particular case we can ignore this warning. Can you add a Pylint disable statement (`# pylint: disable=...`) with the right message code?

danieljanes:
@joshua-sterner friendly ping, we can merge this once the CI is green

danieljanes:
The changes in this PR were merged in #793

PR #781, opened 2021-07-22
Virtual Client Simulation

danieljanes:


src/py/flwr/simulation/app.py
```diff
@@ -0,0 +1,106 @@
...
+) -> None:
+    """Start a Ray-based Flower simulation server.
+
+    Parameters
+    ----------
+    client_fn : Callable[[str], Client]
+        A function creating client instances. The function must take a single
+        str argument called cid. It should return a single client instance
+        containing the local dataset and model of the client with the matching
+        client id (i.e., cid).
```
tanertopal, line 56:
```suggestion
        A function creating client instances. The function must take a single
        str argument called `cid`. It should return a single client instance.
        Each time a client method is called the client will be instantiated and
        after the method has been executed the instance will be destroyed.
        The `cid` serves as the only form of identity for a specific client.
        The client has to load on instantiation or method call any external state
        e.g. hyperparameters or the dataset using only the `cid`.
```

src/py/flwr/simulation/app.py
```diff
@@ -0,0 +1,106 @@
...
+    num_clients : int
+        The total number of clients in this simulation.
+    client_resources : Optional[Dict[str, int]] (default: None)
+        CPU and GPU resources for a single client. Supported keys are
+        `num_cpus` and `num_gpus`. Example: `{"num_cpus": 4, "num_gpus": 1}`
```
tanertopal, line 61:
```suggestion
        `num_cpus` and `num_gpus`. Example: `{"num_cpus": 4, "num_gpus": 1}`
        With regards to actual GPU utilization see the Ray documentation:
        https://docs.ray.io/en/master/using-ray-with-gpus.html
```

tanertopal (changes requested):
Left few comments. One more thing:

`src/py/flwr/simulation/ray_simulation/` seems a bit redundant so I would propose to rename it to:
`src/py/flwr/simulation/ray/`

danieljanes:
> Left few comments. One more thing:
> 
> `src/py/flwr/simulation/ray_simulation/` seems a bit redundant so I would propose to rename it to:
> `src/py/flwr/simulation/ray/`

Renamed to `ray_transport` and pushed a commit with docstrings inspired by your suggestions

tanertopal (approved):

tanertopal (approved):

PR #789, opened 2021-07-26
Tensorboard Strategy.

MakGulati:
For incorporating tensorboard, we need to finalize two things:

- should i put in strategy folder so that i can directly include weighted_loss_avg function from .aggregate?

- regarding the indexing of round: 0-indexed or 1-indexed and notion of -1 round

Thank you.


examples/simulation/TensorBoardStrategy.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    weighted_losses = [num_examples * loss for num_examples, loss, _ in results]
+    return sum(weighted_losses) / num_total_evaluation_examples
+
+
+class TensorBoardStrategy(fl.server.strategy.FedAvg):
+    def aggregate_evaluate(
+        self, rnd: int, results, failures,
+    ) -> Tuple[Optional[float], Dict[str, Scalar]]:
+        """Aggregate evaluation losses using weighted average."""
```
tanertopal, line 22: After thinking about it a while I was wondering if this could be written as a wrapper for a Strategy instance hooking into the `aggregate_evaluate` or as a simple function which everyone can use in their `aggregate_evaluate` and just pass the `aggregate_evaluate` arguments to it.

examples/simulation/TensorBoardStrategy.py
```diff
@@ -0,0 +1,58 @@
...
+        writer_distributed = tf.summary.create_file_writer("mylogs/distributed")
+
+        if rnd != -1:
+            step = rnd
+        else:
+            step = len(
+                [
+                    name
+                    for name in os.listdir("mylogs/distributed")
+                    if os.path.isfile(os.path.join("mylogs/distributed", name))
```
tanertopal, line 33: Might be cool if this is configurable and defaults to e.g. `$(cwd)/tensorboard`

examples/simulation/TensorBoardStrategy.py
```diff
@@ -0,0 +1,58 @@
...
+                ]
+            )
+
+        with writer_distributed.as_default():
+            for client_idx, (_, evaluate_res) in enumerate(results):
+                tf.summary.scalar(
+                    f"num_examples_client_{client_idx+1}",
```
tanertopal, line 40: Maybe use the client cid in the filename? Although need to think a bit about it. Wdyt?

tanertopal (commented):
Shared some thoughts. Looking forward to your answer.

examples/simulation/TensorBoardStrategy.py
MakGulati: @tanertopal I was also thinking about writing a utils file for a function and put in common folder.

examples/simulation/TensorBoardStrategy.py
MakGulati: yes this was just a hack since I was not sure how to consider rnd=-1. We can definitely do it in  a cleaner way :)

examples/simulation/TensorBoardStrategy.py
MakGulati: I am not sure how to extract that (client cid) but certainly a better option.

tanertopal (approved):
@MakGulati thank you for the PR. We decided to merge this now. We will iterate over the API in the upcoming weeks and keep you updated on Slack :) Thank you for starting this!

PR #795, opened 2021-07-27
Create get properties msg

pedropgusmao:


src/proto/flwr/proto/transport.proto
```diff
@@ -69,7 +69,7 @@ message ClientMessage {
     float accuracy = 3 [ deprecated = true ];
     map<string, Scalar> metrics = 4;
   }
-
+  message GetInfo { map<string, Scalar> info = 1; }
```
danieljanes, line 72: `GetProperties`

src/py/flwr/server/client_proxy.py
```diff
@@ -33,11 +36,15 @@ class ClientProxy(ABC):
 
     def __init__(self, cid: str):
         self.cid = cid
+        self.properties: Properties = {}
 
     @abstractmethod
     def get_parameters(self) -> ParametersRes:
         """Return the current local model parameters."""
 
+    def get_properties(self, ins: PropertiesIns) -> PropertiesRes:
```
danieljanes, line 45: This needs an `@abstractmethod`

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -46,6 +46,17 @@ def get_parameters(self) -> common.ParametersRes:
             res,
         )
 
+    def get_properties(self, ins: common.PropertiesIns) -> common.PropertiesRes:
+        """Returns client's properties."""
+        future_get_properties_res = launch_and_get_properties.options(
```
danieljanes, line 51:
```suggestion
        future_properties_res = launch_and_get_properties.options(
```

src/proto/flwr/proto/transport.proto
```diff
@@ -44,12 +44,13 @@ message ServerMessage {
     Parameters parameters = 1;
     map<string, Scalar> config = 2;
   }
-
+  message PropertiesIns { map<string, Scalar> properties = 1; }
```
danieljanes, line 47:
```suggestion
  message PropertiesIns { map<string, Scalar> config = 1; }
```

danieljanes: @pedropgusmao not that this renaming probably requires a few downstream renamings as well

danieljanes (changes requested):

src/py/flwr/client/numpy_client.py
```diff
@@ -86,6 +89,17 @@ def get_parameters(self) -> List[np.ndarray]:
             The local model parameters as a list of NumPy ndarrays.
         """
 
+    # @abstractmethod
+    def get_properties(self, ins: PropertiesIns) -> PropertiesRes:
```
danieljanes, line 93: The argument should be called `config` and the type should be the same as in `fit/evaluate`

danieljanes: Some background: `Ins/Res` types are only used in `Client` whereas `NumPyClient` uses higher-level / more Pythonic / easier-to-use types

danieljanes (changes requested):

src/py/flwr/client/numpy_client.py
pedropgusmao: As in `Ins/Res` use gRPC messages, and `NumpyClient` uses dictionary?

src/proto/flwr/proto/transport.proto
pedropgusmao: @danieljanes same for `PropertiesRes` ?

src/py/flwr/client/numpy_client.py
```diff
@@ -175,6 +189,10 @@ class NumPyClientWrapper(Client):
 
     def __init__(self, numpy_client: NumPyClient) -> None:
         self.numpy_client = numpy_client
+        self.properties: Properties = {"tensor_type": "numpy.ndarray"}
+
+    def get_properties(self, ins: PropertiesIns) -> PropertiesRes:
+        return PropertiesRes(properties=self.properties)
```
danieljanes, line 195: Using self.properties here means that `NumPyClient.get_properties` never gets called. My recommendation is to keep `NumPyClientWrapper` stateless (apart from the `numpy_client` property). The most flexible approach would be to leave setting ` {"tensor_type": "numpy.ndarray"}` up to the user.

src/py/flwr/client/keras_client.py
```diff
@@ -108,13 +111,17 @@ class KerasClientWrapper(Client):
...
+        self.properties: Properties = {"tensor_type": "numpy.ndarray"}
 
     def get_parameters(self) -> ParametersRes:
         """Return the current local model parameters."""
         weights = self.keras_client.get_weights()
         parameters = weights_to_parameters(weights)
         return ParametersRes(parameters=parameters)
 
+    def get_properties(self, ins: PropertiesIns) -> PropertiesRes:
+        return PropertiesRes(properties=self.properties)
```
danieljanes, line 123: `KerasClientWrapper` should ideally remain stateless (apart from the `keras_client` property, see comments on `NumPyClientWrapper` for comparison)

src/py/flwr/client/keras_client.py
```diff
@@ -108,13 +111,17 @@ class KerasClientWrapper(Client):
 
     def __init__(self, keras_client: KerasClient) -> None:
         self.keras_client = keras_client
+        self.properties: Properties = {"tensor_type": "numpy.ndarray"}
```
danieljanes, line 114: This should be removed, see comment below

danieljanes (changes requested):
Thanks for the updates @pedropgusmao. Just a few minor things left, almost there!

src/py/flwr/client/numpy_client.py
pedropgusmao: I removed `self.properties` from wrapper and forwarded the `get_properties` call to `numpy_client` attribute.

danieljanes (approved):

PR #796, opened 2021-07-27
Adding logging of exceptions when clients are called asynchronously

amm299:


tanertopal:
@amm299 could you explain the motivation behind the PR and detail what it does?

tanertopal:
@amm299 Thank you for opening the PR. Closing it for now due to inactivity, but I am happy to reopen it if you want to continue working on it.  Feel free to ping me on Slack with a reference to this PR.

amm299:
Hi Taner,
It's been 2 (3?) years since I worked on Flower, so I've forgotten all of the details. The essential problem, as I recall, was that when you spawned a bunch of processes with Ray (I think that's what it was called), so you could perform federated learning, these processes weren't tied to the command line of the main process. Thus, any logs or error messages would not be displayed, and so the code could fail silently. I made some changes to the code so that these logs/error messages would be displayed, which made debugging much easier. I don't know if any of this is relevant at this point, but I thought I should explain the problem I was trying to solve.

Best,
Angus
------- Original Message -------

PR #814, opened 2021-08-10
Activate gRPC retry mechanism

tanertopal:
This will enable the clients to retry in case the server can not be reached when establishing the initial stream connection. It should reduce errors where the server started up slowly or the network connection was unstable for few seconds.


tanertopal:
Closing for now as it seems this will not work with the way we use gRPC streams. Need to analyze it further

PR #826, opened 2021-08-30
Fixing script to work with Ray ^1.6.0 without mypy `--no-implicit-reexport` issues.

pedropgusmao:
This took a long time. Ray changed the way it structures the internal imports. Now one should call directly the submodule `ray.worker` to access `get`.


pedropgusmao:
The justification for the changes can be found here: https://github.com/ray-project/ray/blob/7916500c43de46721c51e6b95fb51cfa2c6078ba/python/ray/__init__.py#L89

tanertopal (commented):
@pedropgusmao looks good to me! Great you found this! Merging it.

tanertopal (approved):

PR #838, opened 2021-09-10
Cpp v2

pedropgusmao:
Final tests with the CPP implementation


scylj1:
Libtorch example finished

tanertopal:
@scylj1 Thank you again for this great work. A few todo's as discussed previously:

- [ ] Add client code for secure connection which was merged here: #844 also we can think of adding this as a subsequent PR if it increases complexity too much. Beware that there where some additional PR's after #844 so ideally checkout the code in main.
- [ ] Use clang-format-10 for C++ code. It is currently used for the protobufs. See [here](https://github.com/adap/flower/blob/main/dev/format.sh#L12)
- [ ] Add the default copyright header to all files in `src` as can be seen [here](https://github.com/adap/flower/blob/main/src/py/flwr/__init__.py#L1-L14)

I will do a code review sometime today and give a bit more feedback.

src/cc/flwr/include/typing.h
```diff
@@ -0,0 +1,316 @@
...
+			return parameters;
+		}
+		int getNum_example()
+		{
+			return num_examples;
+		}
+		/*std::optional<int> getNum_examples_ceil()
+		{
+			return num_examples_ceil;
+		}*/
```
tanertopal, line 198: Do we need this here? If its not used I would say lets remove it. Its deprecated in the Python code anyway.

src/cc/flwr/include/typing.h
```diff
@@ -0,0 +1,316 @@
...
+		std::optional<float> getFit_duration()
+		{
+			return fit_duration;
+		}
```
tanertopal, line 202: This is deprecated in the Python code so lets remove it here too.

tanertopal (commented):
Left a few remarks and thoughts..

pedropgusmao:
I believe we do not have a `get_properties()` implementation yet.

PR #846, opened 2021-09-15
Client dropout example

hei411:
This is a simple example for users to simulate client dropouts for FL experiments


mariaboerner1987:
Hi @hei411 ,
can you please merge the newest main into your PR. 
@danieljanes can we merge this PR?

hei411:
Hi @mariaboerner1987 
I have merged the newest main into my branch.
The main purpose of the changes is to simulate client dropouts realistically. Before, when a client crashes, it does not recover. However in real life, a client might just crash temporarily and can recover for future rounds. To simulate this, I use a try-catch block to catch the raised exception, and return an error message back to the server. The server will be able to detect the error message, but it can still include the client in the next round of fit.

Happy to discuss this in more detail anytime

PR #845, opened 2021-09-15
Document how to enable SSL in public API

tanertopal:
This will document how to start the Flower server with SSL enabled.

## Important
This PR requires #842 and #844 to be merged first.

PR #847, opened 2021-09-15
Modify example to start secure server

tanertopal:
This PR depends on #845 before it can be merged. It modifies the existing Flower Advanced TensorFlow example to include a secure server. In the next PR the documentation will be updated.

### DIFF
--- a/examples/advanced_tensorflow/README.md
+++ b/examples/advanced_tensorflow/README.md
@@ -42,7 +42,7 @@ poetry run python3 -c "import flwr"
 
 If you don't see any errors you're good to go!
 
-# Run Federated Learning with TensorFlow/Keras and Flower
+## Run Federated Learning with TensorFlow/Keras and Flower
 
 The included `run.sh` will start the Flower server (using `server.py`), sleep for 2 seconds to ensure the the server is up, and then start 10 Flower clients (using `client.py`). You can simply start everything in a terminal as follows:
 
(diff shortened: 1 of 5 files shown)


examples/advanced_tensorflow/README.md
```diff
@@ -42,12 +42,16 @@ poetry run python3 -c "import flwr"
...
+The `run.sh` script starts processes in the background so that you don't have to open eleven terminal windows. If you experiment with the code example and something goes wrong, simply using `CTRL + C` on Linux (or `CMD + C` on macOS) wouldn't normally kill all these processes, which is why the script ends with `trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT` and `wait`. This simply allows you to stop the experiment using `CTRL + C` (or `CMD + C`). If you change the script and anything goes wrong you can still use `killall python` (or `killall python3`) to kill all background processes (or a more specific command if you have other Python processes running that you don't want to kill).
+
+## Important / Warning
+
+The approach how the SSL certificates are generated in this example can serve as an inspiration and starting point but should not be taken as complete for production environments. Please refer to other sources regarding the issue of correctly generating certificates for production environments.
```
danieljanes, line 57:
```suggestion
The approach used to generate SSL certificates can serve as an inspiration and starting point, but it should not be considered as viable for production environments. Please refer to other sources regarding the issue of correctly generating certificates for production environments.
```

examples/advanced_tensorflow/run.sh
```diff
@@ -1,10 +1,16 @@
...
 echo "Starting server"
+
 python server.py &
 sleep 3  # Sleep for 3s to give the server enough time to start
 
-for i in `seq 0 9`; do
+# Ensure that the Keras dataset used in client.py is already cached.
+python -c "import tensorflow as tf; tf.keras.datasets.cifar10.load_data()"
+
+for i in `seq 0 10`; do
```
danieljanes, line 13: Shouldn't this remain unchanged?
```suggestion
for i in `seq 0 9`; do
```

examples/advanced_tensorflow/server.py
```diff
@@ -26,8 +27,17 @@ def main() -> None:
         initial_parameters=fl.common.weights_to_parameters(model.get_weights()),
     )
 
-    # Start Flower server for four rounds of federated learning
-    fl.server.start_server("[::]:8080", config={"num_rounds": 4}, strategy=strategy)
+    # Start Flower SSL-enabled server for three rounds of federated learning
+    fl.server.start_server(
+        "0.0.0.0:8080",
```
danieljanes, line 32: Shouldn't this align with the value set in `client.py`?
```suggestion
        "localhost:8080",
```

danieljanes (changes requested):
@tanertopal looks good, just a few minor comments

examples/advanced_tensorflow/server.py
tanertopal: The server should imo listen for all hosts. So I think 0.0.0.0 is better as otherwise only clients on the same machine would be able to connect.

danieljanes (approved):

Issue #762, opened 2021-06-24
Migrate mypy configuration to pyproject.toml


danieljanes:
Parts of our mypy configuration live in `mypy.ini`. Support for configuration via `pyproject.toml` was recently added to mypy (https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file), so we should migrate our configuration to follow a consistent approach over all Python tooling

architjen:
Hi @danieljanes, I'm looking for my first contribution, can I take this issue? 
Also, could you brief me about the issue a little bit?

danieljanes:
Hi @architjen , great to hear! Absolutely, feel free to take the issue. The general idea is to replicate the mypy config (that we currently have in mypy.ini) using pyproject.toml (and then get rid of mypy.ini altogether). Do you have any specific questions? Happy to provide more details

architjen:
Ok, I'm on it! 
I'll get back to you if I have any questions.

architjen:
Hi @danieljanes I'm trying to wrap my head around the issue rather than just solving it, correct me if I'm wrong, also, any additional remarks/suggestions from you would be much appreciated 

The idea here is: Instead of using `mypy.ini` it's preferable to use `myproject.toml` (according to PEP 518) so that tools like pip can read what tools are specified in `myproject.toml` install them to a virtual env to build our project.  

For solving the issue: We simply migrate the content from the `mypy.ini` file located at the root of the repo to the `myproject.toml` file located at the repo by converting it into the desired format given in mypy docs, and then removing the file `mypy.ini` from the repo. 

Sorry if part of my comment seems like a repetition from above.

ronnie-llamado:
@architjen Is there an update on this? I have the changes on a local branch queued up for a PR if you decide to pass on this one. 

If I can offer a quick remark, try not to overthink it - it sounds like you have the idea.

tanertopal:
@architjen thank you for your contribution. This issue is resolved in #853

PR #853, opened 2021-09-18
Migrate mypy

architjen:


pyproject.toml
```diff
@@ -115,3 +115,20 @@ testpaths = [
...
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "flwr_experimental.*"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "torch.*"
+follow_imports = skip
+follow_imports_for_stubs = true
```
tanertopal, line 134:
```suggestion
follow_imports_for_stubs = true

```

tanertopal (changes requested):
Hi @architjen, thanks for this PR. Did not know this is already possible. Could you delete the `mypy.ini` so we can see how the CI will react? To be sure the new config is taken from the `pyproject.toml`

pyproject.toml
```diff
@@ -115,3 +115,21 @@ testpaths = [
...
+module = "flwr_example.*"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "flwr_experimental.*"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "torch.*"
+follow_imports = skip
```
tanertopal, line 133: CI is failing and I think its because of this line. Did you try running `dev/test.sh` locally?

tanertopal:
```
follow_imports = "skip"
```
might work but I am not sure.

architjen: Ok, I'll try `follow_imports = "skip"`

No, I didn't run `dev/test.sh` locally

tanertopal: My recommendation would be try running the mypy part in "dev/test.sh" or generally all tests. This makes it easier to detect errors before it hits the CI.

architjen: Thanks for the recommendation @tanertopal 

Here's what I did:

-  tried running `dev/test.sh` It basically stops after printing this: "17 files would be reformatted, 51 files would be left unchanged." It broke somewhere in between or else the message should have been "All Python checks passed"

- tried running `python -m mypy src/py ` as you also suggested that running just mypy part would also do. Again, I had lots of errors revolving around "Library stubs not installed for "google.protobuf.*******"

Now, I have few questions to ask:

To make sure that I have the right environment to run the test, how can I create one? What I'm looking for is something like requirements.txt, which I directly pip install to get all the packages.

Also, does the operating system plays any part here? 

Also, if you could share any links or references to create the environment for flower that'd be highly appreciated. 

Thanks again )

tanertopal: @architjen so to create a correct environment you can run sequentially the following scripts:

* venv-create.sh
* bootstrap.sh

The first one you may skip if you take care of a venv yourself. The second script will install everything including poetry into your environment.

Afterwards you can use mostly all other scripts. Does this help?

architjen: Hi @tanertopal, nope it doesn't seem to help. I even tried running `poetry install` so that it creates a virtual env with all dependencies. 

I tried testing on different branches as well.

I don't know what else can be done: Maybe deleting my local repo and re-git cloning?

tanertopal: @architjen I checked your PR and it looks good so I am going to merge it so #762 is resolved. Thank you for your contribution. We'll create a contributor guide to explain the dev setup in a bit more details in the foreseeable future.  If you are still interested to setup the repo so you can contribute further let us continue the discussion on Slack?

tanertopal (approved):
Resolves #762

pyproject.toml
architjen: @tanertopal thanks for the merge. 

Yeah, it'll be nice to have a guide on dev setup, maybe I can help there as well?

Sure, we can take it to Slack, I'm looking forward to more contributions, having a setup would be requisite.

PR #860, opened 2021-09-28
Adding possibility of sending a `List[str]` instead of just the number of clients.

pedropgusmao:
This is useful when your dataloading is not based on integer.

### DIFF
--- a/src/py/flwr/simulation/app.py
+++ b/src/py/flwr/simulation/app.py
@@ -108,7 +108,7 @@ def start_simulation(  # pylint: disable=too-many-arguments
         An empty dictionary can be used (ray_init_args={}) to prevent any
         arguments from being passed to ray.init.
     """
-    list_clients: List[str]
+    cids: List[str]
 
     # clients_ids takes precedence
     if clients_ids is not None:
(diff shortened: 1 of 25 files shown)


src/py/flwr/simulation/app.py
```diff
@@ -26,10 +27,37 @@
 from flwr.server.strategy import Strategy
 from flwr.simulation.ray_transport.ray_client_proxy import RayClientProxy
 
+INVALID_ARGUMENTS_START_SIMULATION = """
+INVALID ARGUMENTS ERROR
+
+Invalid Arguments in method:
+
+`start_simulation(  # pylint: disable=too-many-arguments
```
danieljanes, line 35:
```suggestion
`start_simulation(
```

src/py/flwr/simulation/app.py
```diff
@@ -75,6 +108,21 @@ def start_simulation(  # pylint: disable=too-many-arguments
         An empty dictionary can be used (ray_init_args={}) to prevent any
         arguments from being passed to ray.init.
     """
+    list_clients: List[str]
```
danieljanes, line 111: How about calling this `cids`?
```suggestion
    cids: List[str]
```

danieljanes (changes requested):
@pedropgusmao thanks, this looks good - I suggested two ideas below

danieljanes (approved):

PR #869, opened 2021-10-12
Remove flwr_example and flwr_experimental from packages

danieljanes:
Tasks:

- [x] Remove `flwr_example` and `flwr_experimental` from included packages (in `pyproject.toml`)
- [x] Remove tool configurations for `flwr_example` and `flwr_experimental`
- [x] Remove related CI jobs
- [x] Make related install extras no-op (and remove them in an upcoming release)
- [x] Remove CI-related /dev scripts
- [x] Add release notes


pyproject.toml
```diff
@@ -56,24 +54,15 @@ importlib-metadata = { version = "^1.0", python = "<3.8" }
...
-baseline = ["tensorflow-cpu", "boto3", "boto3_type_annotations", "paramiko", "docker", "matplotlib"]
-examples-pytorch = ["torch", "torchvision", "tqdm"]
-examples-tensorflow = ["tensorflow-cpu"]
-http-logger = ["boto3", "boto3_type_annotations", "tensorflow-cpu", "boto3", "boto3_type_annotations", "paramiko", "docker", "matplotlib"]
-ops = ["boto3", "boto3_type_annotations", "paramiko", "docker"]
+baseline = []
+examples-pytorch = []
+examples-tensorflow = []
+http-logger = []
+ops = []
```
tanertopal, line 65: I think these can also be removed

danieljanes: They could be removed, but it'd be a breaking change. Keeping them as an empty list makes the extras a no-op, they can be completely removed in a later release.

tanertopal: Could you explain? Removing the packages and dependencies seems to be a breaking change in itself.

tanertopal (approved):

PR #882, opened 2021-10-17
Support for Object type numpy arrays

Chris-george-anil:
Added support,  weights that  are passed as Numpy arrays with dtype as objects


danieljanes:
Thanks for the PR @Chris-george-anil !

Could you elaborate on what the effects of this change will be? I'm not that familiar with the `allow_pickle` argument, so it's difficult to understand the impact of this change. What kind of use cases does it enable? Is it incompatible with any prior behavior?

tanertopal:
@Chris-george-anil Any update or thoughts form your side?

Chris-george-anil:
> @Chris-george-anil Any update or thoughts form your side?

Nothing else @tanertopal , it works fine on my device.

mariaboerner1987:
@tanertopal and @Chris-george-anil 
how do we gonna proceed with this PR?

tanertopal:
@Chris-george-anil thank you for your effort and this PR! All contributions are highly appreciated. Out of security considerations we are not going to merge this as it would allow clients the execution of arbitrary code on the server. Further this would mean each client can likewise execute any code on any other client using the captured server as a proxy. We are looking forward to your future work in with Flower :)

PR #886, opened 2021-10-18
Changed tensor convertor from torch.Tensor() to torch.tensor()

PratikGarai:
So I was working with this repo : https://github.com/amirhosseinh77/UNet-AerialSegmentation
I cloned it and refactored it to train the model over flower. I used the code provided in the flower pytorch quickstart example.

However, the load_state_dict method threw the following error :
```IndexError: index 0 is out of bounds for dimension 0 with size 0```

Seemed like an issue of shape mismatch. So I changed the set_parameters method to the following to analyse the situation:
```py
        def set_parameters(self, parameters):
            params_dict = zip(model.state_dict().keys(), parameters)
            state_dict = OrderedDict({k: torch.Tensor(v) for k, v in params_dict})

            it1 = model.state_dict().items()
            it2 = state_dict.items()

            l1 = len(it1)
            l2 = len(it2)

            if l1!=l2 :
                print(f"{l1} : {l2} length do not match")
            else :
                for i in model.state_dict() :
                    if not model.state_dict()[i].shape == state_dict[i].shape:
                        print(i, model.state_dict()[i].shape, state_dict[i].shape,"Different")

            model.load_state_dict(state_dict, strict=False)
```

And I get the following output from the if-else block :
```
inc.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
inc.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down1.maxpool_conv.1.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down1.maxpool_conv.1.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down2.maxpool_conv.1.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down2.maxpool_conv.1.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down3.maxpool_conv.1.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down3.maxpool_conv.1.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down4.maxpool_conv.1.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
down4.maxpool_conv.1.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up1.conv.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up1.conv.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up2.conv.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up2.conv.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up3.conv.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up3.conv.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up4.conv.double_conv.1.num_batches_tracked torch.Size([]) torch.Size([0]) Different
up4.conv.double_conv.4.num_batches_tracked torch.Size([]) torch.Size([0]) Different
``` 

So, it turns out to be a case of mismatch when the client side expects torch.Size([]) but the server side message is parsed to torch.Size([0]). On checking online, I found this link : https://discuss.pytorch.org/t/what-does-torch-size-0-means/21559

and tried to come up with a solution, and turn out it was just changing the wrapper : 
image

 Here you can see that torch.tensor() preserves the shape properly in both cases. But torch.Tensor() is unable to handle empty tensors. So, when I changed my code to : 
```py
        def set_parameters(self, parameters):
            params_dict = zip(model.state_dict().keys(), parameters)
            state_dict = OrderedDict({k: torch.tensor(v) for k, v in params_dict})

            it1 = model.state_dict().items()
            it2 = state_dict.items()

            l1 = len(it1)
            l2 = len(it2)

            if l1!=l2 :
                print(f"{l1} : {l2} length do not match")
            else :
                for i in model.state_dict() :
                    if not model.state_dict()[i].shape == state_dict[i].shape:
                        print(i, model.state_dict()[i].shape, state_dict[i].shape,"Different")

            model.load_state_dict(state_dict, strict=False)
```
The code executed without any errors.

I believe making this small change in the example will make the it more deployment ready and error free.

### DIFF
--- a/doc/source/quickstart_pytorch.rst
+++ b/doc/source/quickstart_pytorch.rst
@@ -163,7 +163,7 @@ which can be implemented in the following way:
 
         def set_parameters(self, parameters):
             params_dict = zip(net.state_dict().keys(), parameters)
-            state_dict = OrderedDict({k: torch.Tensor(v) for k, v in params_dict})
+            state_dict = OrderedDict({k: torch.tensor(v) for k, v in params_dict})
             net.load_state_dict(state_dict, strict=True)
 
         def fit(self, parameters, config):


anindyaspaul:
I faced the same issue with VGG16 where the `num_batches_tracked` tensor is of shape `torch.Size([])`. When it is converted to numpy array and then loaded back using `torch.Tensor()`, it changes the shape to `torch.Size([0])`.

I believe a better solution is to use `torch.from_numpy(ndarray)` since we are extracting the model parameters into numpy array during in the `NumPyClient.get_parameters()` method. This seems semantically more aligned given the usage of `NumPyClient`.

tanertopal:
@anindyaspaul could you maybe create a PR where you show what you would change exactly?

I am going to merge this now and open a PR for the other places we have this. Thanks again @PratikGarai

tanertopal (approved):

danieljanes:
@anindyaspaul @tanertopal the CI failure is not related to this PR, it might be caused by a regression in TensorFlow 2.6.0 (https://github.com/tensorflow/tensorflow/issues/52951#issuecomment-962082712).

I've opened https://github.com/adap/flower/pull/896, let's hope it fixes the issue.

danieljanes:
Merged - thanks again for the contribution @anindyaspaul !

PR #893, opened 2021-10-27
Fix FedAdam and FedYogi

kuihao:
When I run the Fedadam and Fedyogi provided by Flower (I use a simple CNN model and the original cifar-10, 2 clients), the accuracy rate drops drastically whenever I run the second round and it seems hard to improve the accuracy no matter how many local epochs are set to the client-side model. And the Fedyogi is even worse, the accuracy rate seemingly remaining at 0.01% forever.

After comparing the algorithm on **page 23** of the original reference paper **"Adaptive Federated Optimization" (https://arxiv.org/abs/2003.00295)**, I found that the original code provided by Flower **mistook delta_t (gradients) for m_t ( momentum),** which causes the final step of "new_weights = [x + self.eta * y / (np.sqrt(z) + self.tau) for x, y, z in zip(self.current_weights, self.m_t, self.v_t)]" **to be incorrectly calculated.** 

**With this fix**, Flower server and client are now **working correctly, with a significant improvement in accuracy.**


pedropgusmao:
Hi @kuihao , I pushed a modification in the branch `update_fedopt`. The PR is #895 . Would you like to run your experiments with that version and let us know if everything is ok?

danieljanes:
@pedropgusmao CI is still failing, seems like a formatting thing:

```
would reformat src/py/flwr/server/strategy/fedadam.py
would reformat src/py/flwr/server/strategy/fedyogi.py
Oh no! 💥 💔 💥
2 files would be reformatted, 67 files would be left unchanged.
```

pedropgusmao:
Thanks @danieljanes , @kuihao , since this is your PR, could you please run `dev/fnt.sh` to format and test your update and then push them again please? This will make sure CI accepts the differences.

danieljanes:
@kuihao thank you for the contribution. We've decided to merge your changes into PR https://github.com/adap/flower/pull/895 (which includes a few other FedOpt improvements) and close this PR.

We just merged PR https://github.com/adap/flower/pull/895 a few minutes ago, which means your changes will become available to all Flower users (first in tonights nightly build, and then in the upcoming 0.18 release). Thanks again for the contribution @kuihao!

PR #900, opened 2021-11-13
Revert "adding testing result"

cozek:
This reverts commit 892e3705c4c939c6c3250712ace4c1c3f044cb9f.
This is because the commit was made using a private email by mistake.


tanertopal (approved):
Looks good. Merging as previously discussed.

PR #916, opened 2021-11-28
Implemented Federated Byzantine Averaging Strategy

brandon-fremin:


mariaboerner1987:
Hi  @brandon-fremin ,
can you please add a description to this PR. Just write down what files are you creating/changing/adding and the goal of the PR. 
Thank you very much in advance.

pedropgusmao:
Hi, @brandon-fremin ! Thanks for your contribution.
Would you mind keeping the original implementation of `aggregate` as it was and removing the logs/prints from the code, please? Also, the brief descriptions suggested by @mariaboerner1987 would help us understand better the code.

PR #918, opened 2021-11-30
Update dataset.py

MakGulati:
convert uint8 to float32 used to compute gradients.


danieljanes (approved):

danieljanes:
Thanks for the PR @MakGulati !

PR #934, opened 2021-12-09
Add original data download link and pre-process code for FedBN

MeiruiJiang:
- Update readme with download link
- Add data preprocess file in utils

### DIFF
--- a/baselines/flwr_baselines/experiments/fedbn_convergence_rate/README.md
+++ b/baselines/flwr_baselines/experiments/fedbn_convergence_rate/README.md
@@ -33,6 +33,8 @@ A more detailed explanation of the datasets is given in the following table.
 The Research team from the [FedBN paper](https://arxiv.org/pdf/2102.07623.pdf) prepared a pre-processed dataset on their GitHub repository that is available [here](https://github.com/med-air/FedBN). Please download their data, save it in a `/data` directory and unzip afterward. 
 The training data contains only 7438 samples and is split into 10 files but only one file is used for **FedBN: Convergence Rate** baseline. Therefore, 743 samples are used for the local training. 
 
+For the original data, please find the download link in FedBN Github repository [here](https://github.com/med-air/FedBN). After downloading the data, save it in a `/data` directory and unzip afterward. Please refer to `utils/data_preprocess.py` to process the original data.
+
 ## Training Setup ##
 
 ### CNN Architecture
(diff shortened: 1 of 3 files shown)


jafermarq:
Thank you for adding this. Certainly the `data_preprocessing.py` seems needed to run the FedBN example. Could you add maybe a simple script that downloads and extracts MNIST and SVHN into the `data/` directory (maybe even creates it). Then the user can simply download the other three datasets following the GDrive links. 

also probably you want to update the `project.toml` to include the dependencies added in your `data_preprocessing.py` (It seems Matplotlib is not needed)

jafermarq:
Also, it might be useful if in addition to the `data_preprocessing.py` script we could have a "data_download" script that downloads all the original MNIST, MNIST_M, SVHN, SynthDigits and USPS datasets. Then, `data_preprocessing.py` could be expanded to also create the partitions for each dataset. This would make the FedBN baseline code in Flower fully self-contained.

MeiruiJiang:
> Also, it might be useful if in addition to the `data_preprocessing.py` script we could have a "data_download" script that downloads all the original MNIST, MNIST_M, SVHN, SynthDigits and USPS datasets. Then, `data_preprocessing.py` could be expanded to also create the partitions for each dataset. This would make the FedBN baseline code in Flower fully self-contained.

Hi Javier,
Many thanks for  your suggestions, I have updated and made a new commit including the data download script, please check~

MeiruiJiang:
> Thank you for adding this. Adding the `data_preprocessing.py` script would require you to update the `project.toml` to include the dependencies added (`sklearn` and `scipy`) in your `data_preprocessing.py` (It seems Matplotlib is not needed).

I am not familiar with the toml regulations, would you mind giving some guides on how to include the right dependencies? I am not sure how to insert and which version should be used.

mariaboerner1987:
> > Thank you for adding this. Adding the `data_preprocessing.py` script would require you to update the `project.toml` to include the dependencies added (`sklearn` and `scipy`) in your `data_preprocessing.py` (It seems Matplotlib is not needed).
> 
> I am not familiar with the toml regulations, would you mind giving some guides on how to include the right dependencies? I am not sure how to insert and which version should be used.

Hi @MeiruiJiang ,
thank you for your contribution. I will have a look and also modify the` pyproject.toml` file. 
FYI @jafermarq (and also thanks for having a first look into it)

mariaboerner1987:
@danieljanes can you please merge this PR to my baseline branch

PR #951, opened 2021-12-20
Renamed rnd to fl_round

sisco0:
Renaming `rnd` to `fl_round` ensures that this variable is self-explanatory and is not mistaken with random. Solves https://github.com/adap/flower/issues/448.

The test suite has been run using `pytest` from the root folder and a screenshot of its tail is attached below as evidence.
image


danieljanes:
Thanks for the PR @sisco0 ! Since this will be a breaking change, we'll defer review until we released the 0.18 (we're planning to do a release with several breaking changes after that).

danieljanes:
@sisco0 we're (finally!) ready to merge this, but we've decided to that the parameter should be called `server_round` instead of `fl_round` (to signal that Flower can be used for things beyond Federated Learning, for example, Federated Evaluation). Would you be open to updating this PR?

danieljanes:
@sisco0 seeing the list of conflicting files I'd guess that it might be easier to create a new PR based on `main` - up to you

danieljanes:
Closing in favour of #1321 (added @sisco0 as a co-author)

PR #952, opened 2021-12-20
Added __version__ to Flower

sisco0:
Added `__version__` using `importlib.metadata` to retrieve installed package information. This pull request accomplishes https://github.com/adap/flower/issues/883 duty.

A screenshot of the testing process is attached below.
image

### DIFF
--- a/src/py/flwr/__init__.py
+++ b/src/py/flwr/__init__.py
@@ -24,9 +24,11 @@ __all__ = [
     "simulation",
 ]
 
+# pylint: disable=import-error, no-name-in-module
 if sys.version_info < (3, 8):
-    import importlib_metadata  # pylint: disable=import-error, no-name-in-module
+    import importlib_metadata
 else:
-    import importlib.metadata as importlib_metadata  # pylint: disable=import-error, no-name-in-module
+    import importlib.metadata as importlib_metadata
+# pylint: enable=import-error, no-name-in-module
 
 __version__ = importlib_metadata.version(__name__)
--- a/src/py/flwr/__init__.py
+++ b/src/py/flwr/__init__.py
@@ -25,8 +25,8 @@ __all__ = [
 ]
 
 if sys.version_info < (3, 8):
-    import importlib_metadata # pylint: disable=import-error
+    import importlib_metadata  # pylint: disable=import-error
 else:
-    import importlib.metadata as importlib_metadata # pylint: disable=import-error
+    import importlib.metadata as importlib_metadata  # pylint: disable=import-error
 
 __version__ = importlib_metadata.version(__name__)
--- a/src/py/flwr/__init__.py
+++ b/src/py/flwr/__init__.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Flower main package."""
 
-from importlib.metadata import version
+import sys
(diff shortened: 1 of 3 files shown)


tanertopal:
@sisco0 Many thanks for your contributions. I can recommend to run the `dev/test.sh` script locally. It should allow you to see most errors which otherwise are discovered in the CI.

src/py/flwr/__init__.py
```diff
@@ -22,3 +23,8 @@
     "server",
     "simulation",
 ]
+
+try:
+    __version__ = version(__name__)
+except:
+    pass
```
tanertopal, line 30: @sisco0 what is the reason for the try/except? Can't we just use `__version__ = version(__name__)`? Are there cases in which this would fail?

tanertopal (commented):
@sisco0 Thanks for opening this PR! Left a question.

tanertopal:
@sisco0 just did some digging and a good solution seems to be: https://github.com/python-poetry/poetry/pull/2366#issuecomment-652418094 as `importlib.metadata` is only available after Python 3.8. We still support Python 3.7 though so we would need a solution which can handle it too.

sisco0:
I would do it without inconvenience by today by using setup tools instead of importlib.

sisco0:
Help is needed as I could accomplish this task locally by using 3.7.9 but for some reason, the same is result not obtained when using the _Github Actions_ test, we are going to the incorrect branch of the comparison.

danieljanes:
@sisco0 I took the liberty to update the PR with the latest changes from `main`, but Pylint is still failing: https://github.com/adap/flower/runs/4636866489?check_suite_focus=true

`test_core` runs on Ubuntu 20.04, but it uses GitHub Actions to test against Python 3.6 - 3.9 (https://github.com/adap/flower/blob/main/.github/workflows/flower.yml). I don't think the Python version is an issue here given that `test_core` also fails on Python 3.8.3.

sisco0:
We are not using an env in the Github Actions environment, I am trying configurations locally with `act` but I could not find a way yet.
On that path, I am trying to use another image as a base and see what happens.

sisco0:
image
Seems promising locally, let me check what happens on the sorting.

sisco0:
The good news is that the tests seem to pass now by the `import` statements, the bad news is that we had a line length violation. The new good news is that we already solved this by using sectioning for _PyLint_ directives. Double good news is fine 🚀.

tanertopal (approved):
@sisco0 Amazing! Thank you a lot for this! We can merge it as is but one more improvement I would suggest would be to add a test for this.

My suggestions would be to have a `__init___test.py` with e.g. this test:

```
def test_version() -> None:
    """Tests if version is correctly imported."""
    # Execute
    from flwr import __version__  # pylint: disable=import-outside-toplevel

    # Assert
    assert isinstance(__version__, str)
```

wdyt?

tanertopal (approved):

tanertopal:
@sisco0 Merged! Thank you!

sisco0:
Adding the test would be a good point for sure. I would try this during tomorrow but if someone wants to continue on this before me I would create an issue.

PR #959, opened 2021-12-25
Add PoC for request/response based transport layer

tanertopal:
This PR is only a draft / PoC and only open so some people can have a look. Will probably change substantially over the next few days or weeks.

**Current State:**
The `quickstart_pytorch` example can be used to test the PoC for async communication. It works for the happy case as of now. No error handling or crash handling is done yet. The basic idea behind the approach is that each ServerMessage has a unique secure identifier which the client has to set in a `reply_to` field in the ClientMessage.


MakGulati:
Hi @tanertopal
could you please update the status of this PR. Would it be possible to trigger fit method using some external event meaning making it `async` subjected to some external trigger?
Thanks :)

tanertopal:
@MakGulati, this effort is superseded by the Flower Next effort. Feel free to check that out. Let me know if you have questions.

PR #962, opened 2021-12-26
Bumped mypy and tensorflow-cpu to newer versions

sisco0:
The changes present in the `pyproject.tml` _Poetry_ file were required to overcome the issues when running `./dev/test.sh` in a clean new _Poetry_ environment.

A wider definition of these changes is presented in https://github.com/adap/flower/issues/961.


sisco0:
This seems to take infinite time to build, so this action should be canceled.

sisco0:
Rolled back to mypy 0.910. How could we get into 0.930? The previous _Github Action_ run needs to be canceled.

danieljanes:
@sisco0 what issues did you see with mypy 0.930? Always good to bump dependencies! We usually pin `dev-dependencies` to a fixed version (`==`, not sure why this isn't the case for those `types-*` deps, we should pin them as well).

sisco0:
The issues could be seen as `poetry install` ends up in an infinite loop, this could be reviewed by checkout out the commit https://github.com/adap/flower/pull/962/commits/0d6a9ef7e8efdcfa25c95b50242d21a3c080a016 locally and running `poetry install -vvv`. I am unsure of how could we get mypy 0.930 into our current _development dependencies_ of Flower.

danieljanes:
Thanks for the update @sisco0 , I was able to reproduce the described behaviour. As a first step, we've updated pip/setuptools/poetry to the latest possible versions in https://github.com/adap/flower/pull/970 Let's see what else we need to make to update to 0.930

danieljanes:
@sisco0 we dropped quite a few optional dependencies in https://github.com/adap/flower/pull/869 - `tensorflow-cpu` amongst them. Could you try to update to mypy 0.930 again? I could do it in a separate PR, but I don't want to steal your contribution :) Ideally try it without the Python version range.

sisco0:
The issue must be related to not compatible _Numpy_ packages for `Python <= 3.7`. It should be tested if setting `python >=3.7,<3.10` could do the trick in 0d6a9ef.

sisco0:
I tried newly with `python` version specifier set to `">=3.6.2,<3.10"` and we got no luck with `numpy`. The output for `poetry install -vvv` is shown below:

```text
   1: fact: numpy (1.21.5) requires Python >=3.7,<3.11
   1: derived: not numpy (==1.21.5)
   1: fact: numpy (1.21.4) requires Python >=3.7,<3.11
   1: derived: not numpy (==1.21.4)
   1: fact: numpy (1.21.3) requires Python >=3.7,<3.11
   1: derived: not numpy (==1.21.3)
   1: fact: numpy (1.21.2) requires Python >=3.7,<3.11
   1: derived: not numpy (==1.21.2)
   1: fact: numpy (1.21.1) requires Python >=3.7
   1: derived: not numpy (==1.21.1)
   1: fact: numpy (1.21.0) requires Python >=3.7
   1: derived: not numpy (==1.21.0)
   1: fact: numpy (1.20.3) requires Python >=3.7
   1: derived: not numpy (==1.20.3)
   1: fact: numpy (1.20.2) requires Python >=3.7
   1: derived: not numpy (==1.20.2)
```

If we go on with `python = ">=3.7.0,<3.10"` it works as expected.

sisco0:
It is `mypy` in conjunction with `numpy` version that is causing some errors when running the `./dev/test.sh` tests. Maybe some new typing behavior is included in `numpy` that was not in previous versions. Is this only seen by my side and should I provide further details on my configuration?

pedropgusmao:
Hi @sisco0 , setting `mypy==0.931` and `python = "^3.6.2"` worked for me. i.e. the `dev/test.sh` works and I end up with `numpy` version `'1.19.5`. Could try this version of mypy, please?

sisco0:
The solution has been tested locally with _success_ results. Thank you for that @pedropgusmao, as the `0.931` version seems to fix the dependency chain issue that was had before with the `0.930` one ⭐.

pedropgusmao:
Thanks, @sisco0 ! 
@danieljanes or @tanertopal , could you please merge this PR? I'm not code-owner ...

danieljanes (approved):

PR #963, opened 2021-12-26
Update CI Python patch versions to latest

tanertopal:
Some tests in current Python patch versions failed in a recent PR because the GitHub CI wasn't able to install our current version. We had this issue a while ago and updating fixed it at that time. I thought its a good idea to update the patch versions again.

Example from #952 => https://github.com/adap/flower/runs/4638198208

Merging this PR should fix that issue.

### DIFF
--- a/.github/workflows/baselines.yml
+++ b/.github/workflows/baselines.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8.3
+          python-version: 3.8.12
       - name: Install build tools
         run: |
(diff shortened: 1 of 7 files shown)


tanertopal:
@danieljanes As soon as you are fine with this PR I will adjust the required tests for main so this can become green.

sisco0:
https://www.githubstatus.com/history It seems that last two days we had issues with Github Actions. Maybe it is not related or not.

danieljanes (approved):

danieljanes:
Great, let's hope this resolves the issues with GitHub Actions.

PR #967, opened 2021-12-28
Test if __version__ is correctly importable

tanertopal:


danieljanes:
@tanertopal should we perhaps test if `__version__` is a valid semver string?

tanertopal:
@danieljanes adjusted it!

danieljanes (approved):

PR #974, opened 2021-12-30
Returning history object rounds

pedropgusmao:
At the end of the experiment is it good to have access to the collected data. This PR makes sure the experiments return a History.


tanertopal:
Good idea. I was just thinking what if we return a dict which has the history object in it. This will enable us to extend that object in a backwards compatible way in the future. Wdyt @danieljanes @pedropgusmao

danieljanes:
@pedropgusmao thanks, at first glance it looks great. Could you add release notes / docs to this PR as well?

danieljanes:
@tanertopal sure, returning a dict could also be an option. Question is, would we add future metrics to the dict or would we add them to the `History` object? The dict would be more general purpose, but it'd require users to do one more step of unpacking. `History` would be a bit narrower, but easier to use.

pedropgusmao:
@danieljanes @tanertopal , How about giving `History` a `to_dict()` method?

tanertopal:
@danieljanes I think the benefits are outweighing the effort for the user. We might want to allow more fine grained control over the server in the future which would even be a good motivation to return a e.g. fl_server instance which has e.g. a history method. History is one of many things which will be relevant for users in the future. I think just returning the history object will require breaking changes in the foreseeable near future.

pedropgusmao:
@danieljanes  #976 was that you meant?

sisco0:
I think that it would be good that we could include the history in a _dictionary_ if `History` is intended to be an input for other functions and we plan to return other _non-history-related_ (e.g., configuration?) outputs.

doc/source/changelog.rst
```diff
@@ -17,7 +17,8 @@ What's new?
 
 * **Minor updates**
     * Update num_examples caclulation in PyTorch code examples in (`#909 <https://github.com/adap/flower/pull/909>`_)
-    * Expose Flower version through "flwr.__version__" (`#952 <https://github.com/adap/flower/pull/952>`_)
+    * Expose Flower version through "flwr.__version__" (`#952 <https://github.com/adap/flower/pull/952>`_)#974
```
sisco0, line 20: Why is there a #974 at the end of this line?

danieljanes: Good catch @sisco0 , @pedropgusmao could you remove it?

danieljanes:
> @danieljanes #976 was that you meant?

Exactly @pedropgusmao  :)

doc/source/changelog.rst
pedropgusmao: will change, thanks!

danieljanes:
@tanertopal @sisco0 @pedropgusmao I'm not sure tbh - I understand the appeal of returning a dict, but I'm a bit hesitant because of two reasons:

(1) the argument can be made for all functions, but we wouldn't want to always return a dict just because we might want to add something in the future
(2) the fl_server example given isn't very convincing, if we want to return a running server then it doesn't make sense to also return `History` because the call would immediately return before there's a `History` to return - wouldn't it be better to have `start_server` which executes a certain recipe (as it does now) and a second function that starts and returns a running server?
(3) additional metrics/configs that we'd like to return could be added to `History` as well - after all, those metrics/configs are part of the execution history of a particular workload, so wouldn't `History` be a good place to put them?

Again, I'm not against returning a dict, I'm just a bit hesitant because of the given reasons. Curious to hear your thoughts.

sisco0:
(1) Solves by the solution at (3)
(2) Isn't the history more like a log that gets live updates while the server is running?
(3) We should not return a dictionary but only an History Object which we could subclass for other methods if required. Would it make sense?

I am a bit lost about the (2) capabilities.

pedropgusmao:
Tbh I guess for now a single `History` object could suffice. However, we could consider using hooks here. 
Since the main goal is to give users access to results (maybe even intermediate ones), we could consider allowing users to register functions to be called at the beginning and end of each round. This could eventually replace `eval_fn` in the future.

tanertopal:
Generally we are always very reluctant to make breaking changes. A few methods which fit IMO into a return object of `start_server`:

* Clean shutdowns of the server before training ends for various reasons => shutdown method
* Be able to get information about the server state for monitoring => various methods for e.g. number of clients, amount data transfered, etc.

So if we make these things available through a returned object from the `start_server` function somehow we will need to make breaking changes if we limit now to a single object. Generally @danieljanes is right that we could make that argument about everything so I am also fine to merge it as is and make a breaking change when required later with a better concept.

pedropgusmao:
Could we merge this then? It will help with the baselines.

danieljanes (approved):
@pedropgusmao changes look good, thanks!

PR #979, opened 2022-01-01
Fixed max_workers mypy related error

sisco0:
Related to https://github.com/adap/flower/pull/978.
There was an error about the default value for the optional `max_workers` parameter when running `mypy` as pointed out by our _Continuous Integration_ system.
The proposed fix has been tested successfully in a local environment without inconvenience.

### DIFF
--- a/src/py/flwr/server/server.py
+++ b/src/py/flwr/server/server.py
@@ -357,7 +357,7 @@ def reconnect_client(
 
 def fit_clients(
     client_instructions: List[Tuple[ClientProxy, FitIns]],
-    max_workers: Optional[int],
+    max_workers: Optional[int] = None,
 ) -> FitResultsAndFailures:
     """Refine parameters concurrently on all selected clients."""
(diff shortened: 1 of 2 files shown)


danieljanes:
Thanks for the fix @sisco0 ! We try to reduce the usage of default arguments wherever we can because they can lead to subtle failures, it usually improves readability if the caller has to explicitly state new args. The biggest exception is user-facing API's where new args would lead to breaking changes.

Looking at the error, it could also be fixed by passing `None` to the function calls in `server_test`:

```
src/py/flwr/server/server_test.py:94: error: Missing positional argument "max_workers" in call to "fit_clients"
src/py/flwr/server/server_test.py:118: error: Missing positional argument "max_workers" in call to "evaluate_clients"
``` 

Happy to merge the PR with that change.

sisco0:
I added the corresponding required changes @danieljanes. Should we check up the number of elements in the `results` variable to match `max_workers`?

danieljanes:
`max_workers` only controls the concurrency, the number of elements in `results` should always stay the same.

PR #982, opened 2022-01-03
Fix missing type error thrown by mypy

tanertopal:
This PR fixes the typing error by using `numpy.typing` types. To ensure they are available we have to set the minimum required numpy version to `^1.20.0`

@pedropgusmao please have a look at the dtype I have set. I think you can judge best if its correct.


danieljanes:
@tanertopal PR lgtm, but make sure to also migrate/remove what's left in `baselines/mypy.ini`

tanertopal:
Missed that thanks! Will do that.

danieljanes (approved):

PR #997, opened 2022-01-07
[Do not merge] Created weighted_avg function and tensorflow_from_centralized_to_federated_example

sisco0:
The current _PR_ fixes https://github.com/adap/flower/issues/983 in a way that the `weighted_losses_avg` function is changed by `weighted_avg` generic function and a new example is created for decentralized federated learning where no server evaluation is performed.

Please take care that the current new example uses the development source code of the flower package as the new changes in the `weighted_avg` function would break this example if we use flower 0.17.

A fix on aggregation tests is proposed as well.

Then, should we split this _PR_ into two or three?


sisco0:
This _PR_ is not ready for three reasons:
1. Contains 3 changes that should be splitted into 3 Pull Requests
2. The new example is set for using the development source code of flower
3. Changes performed on the averaging function and related testing refactoring should be studied

danieljanes:
@sisco0 we decided to close this, but feel free to open separate PRs if there's something that we should consider merging.

Issue #1016, opened 2022-01-20
Upgrade Ray to the latest stable version


danieljanes:
### Describe the type of feature and its functionality.

The `ray` dependency is currently pinned to `==1.6` because `ray==1.7.0` introduced backwards-incompatible changes which broke the Virtual Client Engine. The current Ray release is 1.9, so Flower should (a) update to the latest version and (b) declare the dependency using `^` so that users can use Flower with future `1.x` Ray releases (`1.10` and above).

### Describe step by step what files and adjustments are you planning to include.

- Update the Ray dependency to the latest version
- Update `RayClientProxy` to be compatible with the latest Ray release
- Declare the dependency as `^`
- Test the change thoroughly

jafermarq:
I was able to run the `simulation_pytorch` example with PyTorch 1.8.2 LTS, Ray 1.9.2 (the latests) and the latest flwr. I had no errors. I also run other of my projects that make use of the `RayClientProxy` without issues. 

Maybe the only thing required is to raise the Ray version in the .toml?

danieljanes:
Thanks for giving it a shot @jafermarq , let's declare the Ray dependency with `^1.9.2` then.

danieljanes:
Moved to Ray ~1.13 with Flower 1.0

PR #1036, opened 2022-01-29
Upgrading Ray and updating simulation_pytorch example

jafermarq:
This PR updates the Ray version, as dicussed in Issue #1016. I have also used this opportunity to slightly update the `simulation_pytorch` example by:

- Updating the main to explain what the example is about.
- Exposed a couple of hyperparamters via `argparse`. I think it is quite relevant to do so for `num_client_cpus` which controls the number of threads assigned to each client.
- Moved boilerplate code to a new script, `utils.py`. This is something suggested by @mariaboerner1987 .


jafermarq:
One point I wanted to ask: do we need a `pyproject.toml` for this example? One was added a while back.

danieljanes (dismissed):

danieljanes:
Thanks for the update @jafermarq , looks good! I've approved the PR, but one question: is the switch from 5 to 20 local epochs intentional?

Re `pyproject.toml`, yes it's needed. We're following the general idea that each example is a self-contained project, `pyproject.toml` is mainly there to list dependencies. One thing we might do in the future is to have a `requirements.txt` (in addition to `pyproject.toml`) to make the examples more approachable for users who do not want to use Poetry.

jafermarq:
@danieljanes , I see now what you mean about the epochs. I've reverted down to 5 local epochs. Good catch.

danieljanes (approved):

PR #1064, opened 2022-02-12
Use clang-format instead of clang-format-10

danieljanes:
The latest version of `clang-format` is `13.0.0`, this PR migrates scripts to call `clang-format` in a version-independent manner.


tanertopal:
@danieljanes Doesn't this go against the principle that we are fixing the version of all dev dependencies? If we are going to upgrade than lets use `clang-format-13`

danieljanes:
@tanertopal tricky one: yes and no. Yes, we fix the version of dev dependencies and we want to continue to do so. But we usually fix the version by making our install scrips install a fixed version (e.g., through poetry), not by changing the scripts that use those tools (e.g., we just call `python` - not `python3.7` - and rely on `pyenv` to provide the right version of the `python` binary).

On macOS / brew, there is no `clang-format-10`, only `clang-format` which brew can install in different versions. So to achieve the desired outcome, we'd have to make brew install `clang-format` at a fixed version, and then invoke `clang-format` in other scripts (e.g., `format.sh`). On Ubuntu, we'd have to change the install script to install a specific version (or to use brew as well). If `clang-format` refers to an old version then we could install `clang-format-13` and also create a `clang-format` alias. Wdyt?

danieljanes:
@tanertopal yes, exactly

tanertopal (approved):

PR #1068, opened 2022-02-14
Make get_properties optional

danieljanes:


tanertopal (dismissed):
Looks awesome! Approved. Just one failing linter error.

src/py/flwr/client/app.py
```diff
@@ -160,6 +162,12 @@ class `flwr.client.NumPyClient`.
     # Wrap the NumPyClient
     flower_client = NumPyClientWrapper(client)
 
+    # Remove NumPyClientWrapper.get_properties with the base class version
+    # Client.get_properties if method was not overridden by user-provided
+    # NumPyClient instance
```
tanertopal, line 167: Can you elaborate this further or rephrase. I don't think I understand whats happening here.

src/py/flwr/client/app.py
```diff
@@ -207,6 +215,12 @@ class `flwr.client.KerasClient`.
     # Wrap the Keras client
     flower_client = KerasClientWrapper(client)
 
+    # Remove KerasClientWrapper.get_properties with the base class version
+    # Client.get_properties if method was not overridden by user-provided
+    # KerasClient instance
```
tanertopal, line 220: Same as above.

src/py/flwr/client/app.py
danieljanes: @tanertopal updated the comment, is it better now?

src/py/flwr/client/app.py
danieljanes: Same as above.

src/py/flwr/client/app.py
tanertopal: Yes, thx!

tanertopal (approved):

Issue #1078, opened 2022-02-19
Dirichlet distribution ignores random seed


jafermarq:
### Describe the bug

The utility function `create_lda_partitions()` can be used to partition an existing dataset given its number of classes and a user-defined concentration parameter. Ideally this partitioning would be deterministic, because this reason the script where that function lives fixes the Numpy random seed. However, because of the way the Dirichlet distribution is being created, that random seed is ignored.

The problematic bit of code is [this one](https://github.com/adap/flower/blob/7473520012bbea9b805bf564b11b590fd6394672/src/py/flwr/dataset/utils/common.py#L454). Using `default_rng()` ignores any seed that has been manually setup.
```
    if dirichlet_dist is None:
        dirichlet_dist = np.random.default_rng().dirichlet(
            alpha=concentration, size=num_partitions
        )
```

### Steps/Code to Reproduce

This simple example shows that every time we generate `direchlet_dist`, the result is different even though we've fixed a seed at the top.

```
import numpy as np

np.random.seed(2020)

num_classes = 5
concentration = [1]*num_classes
num_partitions = 3

for i in range(4):
    print(f"Call #{i+1}")
    dirichlet_dist = np.random.default_rng().dirichlet(
                alpha=concentration, size=num_partitions
            )
    print(dirichlet_dist)

```

This would print 4 different distributions (see Actual Results section below)

### Expected Results

The expected result is to generate 4 equal distributions. This can be achieved by replacing `np.random.default_rng().dirichlet(...)` with `np.random.dirichlet(...)`. The update example below shows the result is the same:

```
import numpy as np

np.random.seed(2020)

num_classes = 5
concentration = [1]*num_classes
num_partitions = 3

for i in range(4):
    print(f"Call #{i+1}")
    dirichlet_dist = np.random.dirichlet(
                alpha=concentration, size=num_partitions
            )
    print(dirichlet_dist)
```

This will print 4 identical distributions. This is the desired behavior.
```
Call #1
[[0.5500942  0.26508389 0.09143256 0.04068992 0.05269943]
 [0.07786622 0.10303762 0.13389849 0.63093419 0.05426349]
 [0.0404233  0.3766745  0.35484996 0.11700229 0.11104996]]
Call #2
[[0.5500942  0.26508389 0.09143256 0.04068992 0.05269943]
 [0.07786622 0.10303762 0.13389849 0.63093419 0.05426349]
 [0.0404233  0.3766745  0.35484996 0.11700229 0.11104996]]
[...]

PR #1079, opened 2022-02-20
`create_lda_partitions ` now to receive a `seed` or RNG

pedropgusmao:
#### Reference Issues/PRs
Proposes a solution to #1078.

This PR changes `create_lda_partitions` so it can receive either a `seed` or a `Random Number Generator` as input, allowing for reproducibility. 

#### Any other comments?

This PR keeps the use of a `Generator` and does not use `np.random.dirichlet` since Numpy recommends using the newer approach. See first Note in https://numpy.org/doc/stable/reference/random/generated/numpy.random.dirichlet.html  .


jafermarq:
Looks good to me!
I put together this short example to test it. Seems to work as expected. (note i only tested passing the seed as an `int`)

```
from flwr.dataset.utils.common import create_lda_partitions
from torchvision.datasets import MNIST
import numpy as np

trainset = MNIST('./data', train=True, download=True)

dataset = [trainset.data.numpy(), np.array(trainset.targets, dtype=np.int64)]
num_classes = len(trainset.classes)

partitions, b = create_lda_partitions(dataset, num_partitions=4,
                                      concentration=[1.0]*num_classes,
                                      seed=1234)

for p in range(4):
    print(f"Labels partition #{p}: {partitions[p][1][:20]}")
```
Which should always print:
```
Labels partition #0: [1 0 2 1 4 1 0 2 4 5 4 1 4 8 2 8 7 4 7 4]
Labels partition #1: [9 6 1 2 0 2 1 1 6 1 0 2 1 6 9 2 4 6 1 3]
Labels partition #2: [5 1 2 9 3 6 2 2 1 5 1 3 7 1 2 1 1 2 1 6]
Labels partition #3: [7 5 3 0 8 7 0 0 3 5 0 5 7 0 3 5 0 7 4 5]
```

pedropgusmao:
@danieljanes / @tanertopal  This can be approved

PR #1081, opened 2022-02-21
Increasing version of Numpy to ^1.20.0 so we can use numpy.typing

pedropgusmao:
#### Reference Issues/PRs
NA

Allows the use `numpy.typing` which is available in version 1.20 
https://numpy.org/doc/stable/reference/typing.html 

#### Any other comments?
This will help with code maintenance.


pedropgusmao:
I'm guessing this hasn't been done before because we still need to support python 3.6?
Should we create a poll to see how many people are using Python 3.6? @tanertopal @danieljanes

PR #1091, opened 2022-02-24
Perform centralized and federated evaluation

sandracl72:

When giving a centralized evaluation function, decentralized evaluation is not performed. 
This gives the option to perform both centralized and federated evaluation by passing {"fed_eval": 1} to on_fit_config_fn() in the strategy.


src/py/flwr/server/strategy/fedavg.py
```diff
@@ -211,16 +211,17 @@ def configure_evaluate(
         self, rnd: int, parameters: Parameters, client_manager: ClientManager
     ) -> List[Tuple[ClientProxy, EvaluateIns]]:
         """Configure the next round of evaluation."""
-        # Do not configure federated evaluation if a centralized evaluation
-        # function is provided
-        if self.eval_fn is not None:
```
danieljanes, line 216: We could simply change this line to check if `fraction_eval` is `0.0` (and return an empty list if it is). That'd allow both distributed and federated evaluation, but it'd enable users to disable federated evaluation by simply setting `fraction_eval` to `0.0`.

danieljanes (commented):
Thanks for the PR @sandracl72 ! I've added two comments below.

Once we converge to a solution we should make the same changes in `qfedavg.py` and `fedavg_android.py` to keep different strategies consistent.

src/py/flwr/server/strategy/fedavg.py
```diff
@@ -211,11 +211,10 @@ def configure_evaluate(
...
+        # Do not configure federated evaluation if fraction eval is 0.
+        if self.fraction_eval == 0.0:
             return []
-
+
```
danieljanes, line 217: @sandracl72 this added whitespace is probably the reason for the CI failure, everything else looks great 👍

danieljanes (changes requested):

danieljanes (approved):

PR #1092, opened 2022-02-24
Log decentralized metrics

sancarlim:

The decentralized metrics ( _metrics_distributed_ -  https://github.com/adap/flower/blob/main/src/py/flwr/server/app.py#L152) are not logged since _metrics_aggregated_ is always empty (https://github.com/adap/flower/blob/main/src/py/flwr/server/server.py#L231-L248, https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/qfedavg.py#L214-L239).

What I propose here is just a quick workaround to output the decentralized metrics, because I am not sure what the role of _metrics_aggregated_ is here and if there is a reason for this.

#### Comments
The log message should probably be changed, as the client {i} would be wrong in case not all clients are evaluated.

### DIFF
--- a/src/py/flwr/server/server.py
+++ b/src/py/flwr/server/server.py
@@ -169,13 +169,12 @@ class Server:
             # Evaluate model on a sample of available clients
             res_fed = self.evaluate_round(rnd=current_round)
             if res_fed:
-                loss_fed, evaluate_metrics_fed, _ = res_fed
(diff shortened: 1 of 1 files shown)


danieljanes:
We decided to close this PR because users can achieve the same behavior by customizing the `Strategy`.

PR #1115, opened 2022-03-05
Expose gRPC parameters in start_client() and start_numpy_client()

g-pichler:
Arguments of the call to class grpc.StreamStreamMultiCallable can be passed in grpc_args and grpc_kwargs. These are timeout, metadata, credentials, wait_for_ready and compression (https://grpc.github.io/grpc/python/grpc.html?highlight=wait_for_ready#grpc.StreamStreamMultiCallable).

This change exposes the parameters for gRPC (used by flower under the hood) to the user. I use this to set wait_for_ready=True to ensure that a client waits until a server is up.


src/py/flwr/client/app.py
```diff
@@ -34,6 +34,8 @@ def start_client(
     client: Client,
     grpc_max_message_length: int = GRPC_MAX_MESSAGE_LENGTH,
     root_certificates: Optional[bytes] = None,
+    grpc_args: Optional[Iterable[Any]] = None,
```
orlandohohmeier, line 37: I'd only add specific arguments here instead of transparently passing all gRPC arguments. The client app should abstract connection handling for the user. For the specific use case you're addressing I'd only add an argument to toggle gRPC `wait_for_reay` defaulting to `True`.

orlandohohmeier (changes requested):
Love the introduction of gRPC `wait_for_ready` to improve connection robustness, but would change the interface to not leak the client abstraction.

src/py/flwr/client/app.py
g-pichler: Great point. Adding only `wait_for_ready` would do just fine. IMO the parameter should default to `False` to preserve established behavior. I believe that by default, users expect the client to fail if a connection to the server cannot be established.

orlandohohmeier: Fair point! I was mostly considering the benefits and would have argued that these outdo the change in behavior but thinking about it again, I agree with you and it should default to `False`.

danieljanes:
Thanks for the PR @g-pichler !

Would there be any downsides to generally setting `wait_for_ready=True` @g-pichler @orlandohohmeier @tanertopal and not exposing the flag to the user?

danieljanes:
The current PR looks good to me though, I wouldn't mind exposing this to the user if it provides value.

orlandohohmeier:
@danieljanes the client currently fails if a connection to the server cannot be established. With `wait_for_ready` enabled the client would wait for the server till the RPC’s deadline is reached breaking with the current behavior.

orlandohohmeier:
> I wouldn't mind exposing this to the user if it provides value.

@danieljanes Not sure what you're alluding to? Exposing the `wait_for_ready` option?

g-pichler:
I think that in general users expect a client to terminate if a connection to the server cannot be established. Setting `wait_for_ready=True` breaks this assumption and current behavior. I think `wait_for_ready=True` by default is a bad idea. For example, if the client port is misconfigured, this behavior could be very frustrating.

I am currently dealing with a situation where I need to spawn many clients and servers in separate threads to test many runs of a federated learning system. Typically the client starts quicker than the server process. So, exposing this flag provides a lot of value.

danieljanes:
> > I wouldn't mind exposing this to the user if it provides value.
> 
> @danieljanes Not sure what you're alluding to? Exposing the `wait_for_ready` option?

@orlandohohmeier [3761f5e](https://github.com/adap/flower/pull/1115/commits/3761f5e8a5d3086e9c11c0ab2c3b7c523e04a433) changed the user-facing API to expose `wait_for_ready` to the user.

@g-pichler thanks for the context, makes sense

g-pichler:
To follow up: Are the more changes necessary or are you planning to merge this PR?

PR #1116, opened 2022-03-06
Update sim.ipynb

gubertoli:
Simple changes to correct validation dataset references on FlowerClient class and min_eval_clients to be in accordance with the comments.


danieljanes:
Good catch @gubertoli , thanks for the PR!

danieljanes (approved):

danieljanes:
CI still fails, but nothing to do here - CI failure is being worked on in #1126

PR #1124, opened 2022-03-11
Quickstart Pytorch with ResNet18

pedropgusmao:
Changes basic example to use less-verbose ResNet18


danieljanes:
@pedropgusmao thanks, looks great! Could you take a look at the `test_baselines` failure?

danieljanes (approved):

PR #1130, opened 2022-03-16
Upgrade importlib-metadata to ^4.4

danieljanes:
Some libraries are starting to depend on `importlib-metadata>=4.4`, which leads to unresolvable situations when installing Flower. This PR updates our dependency to 4.4 or later, which is the level that's equivalent to `importlib-metadata` in Python 3.10

See, for example, the following log output that can be reproduced by running the first code cell in `sim.ipynb`:
image


tanertopal (dismissed):
Lgtm but some linter is still failing.

tanertopal (approved):

Issue #1113, opened 2022-03-05
Error when invoking start_numpy_client() more than once


g-pichler:
### Describe the bug

When invoking start_numpy_client() more than once within the same python interperter, this will yield an error when the clients passed to it do not have a get_parameters() method.

The culprit seems to be line 168 of `src/py/flwr/client/app.py`:
```
   del NumPyClientWrapper.get_properties
```
If this deletion is invoked a second time, an AttributeError is raised. Probably there is a more graceful way of dealing with get_properties() inside NumPyClientWrapper, e.g., by returning an empty dictionary.

### Steps/Code to Reproduce

In `quickstart_pytorch` place clients.py with the following contents:
```
import time
from client import main as client_main

def main():
    for _ in range(2):
        time.sleep(5)
        client_main()

if __name__ == '__main__':
    main()
```
Then run the followin shell script:
```
#!/bin/bash

for i in `seq 0 1`; do
    echo "Starting client $i"
    python clients.py &
done

for i in `seq 0 1`; do
    echo "Starting Server $i ..."
    python server.py
done
```

### Expected Results

I expect two runs of training to complete within several minutes all processes should exit with exit code 0.

### Actual Results

The first training (and server process) completes without issues. In the second run, the clients fail with the following backtrace:
```
Traceback (most recent call last):
  File "./clients.py", line 12, in <module>
    main()
  File "./clients.py", line 8, in main
    client_main()
  File "./client.py", line 122, in main
    fl.client.start_numpy_client("[::]:8080", client=CifarClient())
  File "flwr/client/app.py", line 168, in start_numpy_client
    del NumPyClientWrapper.get_properties
AttributeError: get_properties
```

danieljanes:
Thanks for the report @g-pichler. The current implementation is indeed not ideal, one additional drawback of the solution is a setup in which different client implementations - some with, some without implementing `get_properties` - are started in the same Python process. Because `get_properties` gets deleted from the class, it won't be available for those client that actually implement it.

The current solution was chosen over returning an empty dict (as in #1114) because a `NumPyClient` wrapped in a `NumPyClientWrapper` behaves exactly the same as / is indistinguishable from a regular `Client` that does not implement `get_properties`. Reducing the differences between a user-defined `Client` and a user-defined `NumPyClient` wrapped in a `NumPyClientWrapper` is important for consistency. The `start_client` implementation needs a way to check if `get_properties` was implemented *without* calling it (the function is user-provided, so it can potentially run for a long while).

[...]

PR #1138, opened 2022-03-21
Taking care of get_properties() by catching an AttributeError.

g-pichler:
#### Reference Issues/PRs

This is meant to address #1113.

It is an improved attempt at dealing with the issue appropriately dealing with `get_properties` not being defined by a user implementation of `Client` or `NumPyClient`. In essence, the default implementation raises an `AttributeError`, which is caught and an appropriate response sent to the server.

#### Any other comments?

This should allow a user to run multiple `Client`s in the same python process.


g-pichler:
Just, to follow up: What do you think of this idea? Is there an issue with this solution?
Best, Georg

g-pichler:
I see. I'll close this PR then.

PR #1140, opened 2022-03-22
Fix an error in the quickstart_pytorch example on GPU.

g-pichler:
Fix a problem where the label tensor was not properly converted to a CUDA tensor leading to an error.

Running the `quickstart_pytorch/run.sh` with a CUDA-enabled GPU yields to an exception 

```
[...]
  File "/examples/quickstart_pytorch/client.py", line 42, in test
    correct += (torch.max(outputs.data, 1)[1] == labels).sum().item()
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```
This is due to `labels` not being on the CUDA device. This PR fixes this issue.

### DIFF
--- a/examples/quickstart_pytorch/client.py
+++ b/examples/quickstart_pytorch/client.py
@@ -37,7 +37,8 @@ def test(net, testloader):
     with torch.no_grad():
         for images, labels in tqdm(testloader):
             outputs = net(images.to(DEVICE))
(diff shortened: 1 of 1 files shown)


tanertopal:
@danieljanes lgtm. wdyt?

danieljanes (approved):
Thanks for the fix @g-pichler !

PR #1144, opened 2022-03-23
Enable metric aggregation through user-provided functions

danieljanes:
This PR enables users to customize metric aggregation in built-in strategies. This prevents users from having to customize the strategy just to customize the aggregation of metrics dicts.

Currently, users have to do this:

image

With this PR, users can instead do this:

image


tanertopal (approved):
Some nitpicking but overall lgtm.

tanertopal (approved):

PR #1146, opened 2022-03-23
Add timeout option to GRPCBridge request method

tanertopal:

Enables passing a timeout to the GRPCBridge.request method which will throw after given number of seconds.


tanertopal:
Closing for now in favor of #1162

PR #1162, opened 2022-03-27
Enable request timeouts

tanertopal:
#### Reference Issues/PRs

None

When merged this PR will enable the server to set a round timeout which will be applied to all requests and enforced on the server side by dropping the client. It does not modify the client side. The user will need to handle the reconnect mechanism. 

#### Any other comments?

Only the fit method in the ClientProxy is as of now adjusted as we first need to discuss if we are going with this.


src/py/flwr/server/grpc_server/flower_service_servicer.py
```diff
@@ -102,8 +132,29 @@ def Join(  # pylint: disable=invalid-name
...
+                    client_message = next_with_timeout(
+                        iterator=client_message_iterator,
+                        timeout=ins_wrapper.timeout,
+                    )
+                    if client_message is None:
+                        # Important: calling `context.abort` in gRPC always
+                        # raises an exception so that all code after the call to
+                        # `context.abort` will not run. If subsequent code should
+                        # be executed, the `rpc_termination_callback` can be used
+                        # (as shown in the `register_client` function).
```
tanertopal, line 146:
```suggestion
                    if client_message is "None":
                        # Important: calling `context.abort` in gRPC always
                        # raises an exception so that all code after the call to
                        # `context.abort` will not run. If subsequent code should
                        # be executed, the `rpc_termination_callback` can be used
                        # (as shown in the `register_client` function).
```

danieljanes (approved):

PR #1177, opened 2022-04-07
option to not shutdown ray if already initialised

jafermarq:
This a short PR enabling the user to decide whether `start_simulation` should do `ray.shutdown()` if an existing Ray runtime is already initialized. In some settings, we might want to have Ray already running before calling `start_simulation` if, for example, some other tasks have to run first before FL.


danieljanes (approved):

danieljanes:
Good idea, thanks for the PR @jafermarq ! I've approve it, there just a lint issue (`too-many-locals`) that needs to be resolved. In this particular case I'd say it's ok to just ignore it.

PR #1208, opened 2022-05-09
experimental fedhenn

MakGulati:
This PR(oriented towards FedHeNN algorithm) is incomplete and throws error as it is not possible to combine parameters from different NN architectures. In this initial draft PR I need help to make sure I can supply different model architectures (for federated personalisation) to respective clients and also this condition is fulfilled on server side during weights initialization.

### DIFF
--- a/examples/experimental_FedHeNN/README.md
+++ b/examples/experimental_FedHeNN/README.md
@@ -1,0 +1,22 @@
+# FL example on MNIST through shell script
+## Prerequisites
+---
+- python: this code is tested on python 3.8
+- poetry: for python installation
+
+```bash
+pip install poetry
+```
+
+To install all the dependencies from `.toml` file and inialize poetry environment:
+```bash
+poetry install
+poetry shell
+```
+## Code execution
+---
+Finally, for code execution:
+```bash
+sh run.sh
+```
+
--- a/examples/experimental_FedHeNN/client.py
+++ b/examples/experimental_FedHeNN/client.py
@@ -1,0 +1,133 @@
+from collections import OrderedDict
+import warnings
+
+import flwr as fl
+import torch
+import torch.nn.functional as F
+from model_mnist import Net0, Net1, Net2, Net3
+from dataset import load_mnist_data_partition
+import argparse
+
+warnings.filterwarnings("ignore", category=UserWarning)
+
+
+# #############################################################################
+# 1. PyTorch pipeline: model/train/test/dataloader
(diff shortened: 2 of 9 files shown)


pedropgusmao:
Hello @MakGulati, 
When sending back the model parameters to the server, you might want to change the  [ tensor_type: str](https://github.com/adap/flower/blob/1bc9649204821a204e25436034c955f5fc54c126/src/py/flwr/common/typing.py#L59) of your `Parameters` before returning them in `fit`. You could include a tag that defines which model you are sending back, e.g. `model_a` .

To deal with receiving different NN architectures, you will need to create a new ` Strategy`  (possibly deriving from an existing one like ` FedAvg` ). You should override [aggregate_fit](https://github.com/adap/flower/blob/1bc9649204821a204e25436034c955f5fc54c126/src/py/flwr/server/strategy/fedavg.py#L211)  so that it will check the `tensor_type` for each `FitRes` in results, and process them accordingly (if `model_a` then... , if `model_b` then...). 

When sending different parameters to different clients, you will need to override the [configure_fit](https://github.com/adap/flower/blob/1bc9649204821a204e25436034c955f5fc54c126/src/py/flwr/server/strategy/fedavg.py#L161). Here you can either change the `tensor_type` of pass the model information using the `config` dictionary.  

Does that make sense?

MakGulati:
In order to include tag in `Parameters` data type, one needs to update it in proto files, please see the error of the updated code.
I was wondering if there is some workaround. Would be nice if you could share some code snippet to incorporate different model types in `Parameters`. @pedropgusmao @danieljanes 

>   File "/home/mayag88/Documents/Code/fork/flower/src/py/flwr/common/serde.py", line 107, in parameters_res_from_proto
>     parameters = parameters_from_proto(msg.parameters)
>   File "/home/mayag88/Documents/Code/fork/flower/src/py/flwr/common/serde.py", line 43, in parameters_from_proto
>     return typing.Parameters(tensors=tensors, tensor_type=msg.tensor_type)
> TypeError: __init__() missing 1 required positional argument: 'tag'

pedropgusmao:
Hi @MakGulati ,
Sorry, I means using the `tensor_type` as the tag:
`parameters.tensor_type='model_a'`
after having created the `parameters` object

pedropgusmao:
Hi @MakGulati ,
We are currently looking in to how to allow for a Strategy to have multiple Parameters. Will get back to you soon.

PR #1241, opened 2022-06-01
Dp

RaghavNaswa:


danieljanes:
Thanks for the PR @RaghavNaswa ! I see it's still in draft mode, so you'll probably continue to iterate on it, but here are two first impressions:
- All the new diagrams in the project root could be moved into the docs
- A description in the PR outlining the changes would be helpful
But wow, the PR looks like a lot of effort went into it, I look forward to reviewing it properly!

PR #1277, opened 2022-07-11
Make client methods optional

danieljanes:


src/py/flwr/client/app.py
```diff
@@ -181,5 +220,103 @@ class `flwr.client.NumPyClient`.
...
+        member_dict["get_parameters"] = _get_parameters
+
+    if numpyclient_has_fit(client=client):
+        member_dict["fit"] = _fit
+
+    if numpyclient_has_evaluate(client=client):
+        member_dict["evaluate"] = _evaluate
+
+    # Create wrapper class with a random class name
+    wrapper_class_name = "NumPyClientWrapper" + uuid4().hex[:8]
```
tanertopal, line 318: @danieljanes why do we do this? Can you explain in the comment. I am unclear why random.

tanertopal (changes requested):
Apart from one question I have it looks good to me.

src/py/flwr/client/app.py
danieljanes: @tanertopal I looked deeper into it, it's not really necessary, so I removed it

tanertopal (approved):
lgtm!

PR #1296, opened 2022-07-16
Rename Weights to NDArrays and use it more consistently

danieljanes:
The PR
1. Renames `Weights` to `NDArrays`
2. Uses it more consistently (instead of mixing `Weights` and `List[np.ndarray` across the codebase)


tanertopal (approved):
lgtm!!

PR #1324, opened 2022-07-24
 Update advanced_tensorflow example to Flower 1.0

tanertopal:

Apply all changes required for this example to work with Flower 1.0.


examples/advanced_tensorflow/pyproject.toml
```diff
@@ -12,6 +12,6 @@ authors = ["The Flower Authors <hello@flower.dev>"]
 
 [tool.poetry.dependencies]
 python = ">=3.7,<3.11"
-flwr = "^0.19.0"
+flwr = "^1.0.0"
 # flwr = { path = "../../", develop = true }  # Development
```
danieljanes, line 16: Should we remove this line from all the examples? I'm in favor of removing

tanertopal: I am fine either way. I think its useful but we could also move it into the developer docs. Wdyt?

danieljanes: Cool, let's move it to the contributor docs then

danieljanes (approved):

PR #1338, opened 2022-07-27
Use keyword-only arguments in public API

danieljanes:


tanertopal (approved):

alexkyllo:
Could you please explain what is the reason for this change? I found that a lot of code I wrote just a few months ago using Flower v0.19.0 is broken by this when I try to run it with v1.0.0. I do understand what I need to do to make my code forward compatible, but what I don't understand is the justification for why this change was made, because no reason was given here or in the release notes--appreciate if you could shed some light on the decision.

> All arguments must be passed as keyword arguments ([#1338](https://github.com/adap/flower/pull/1338))
> Pass all arguments as keyword arguments, positional arguments are not longer supported. Code that uses positional arguments (e.g., start_client("127.0.0.1:8080", FlowerClient())) must add the keyword for each positional argument (e.g., start_client(server_address="127.0.0.1:8080", client=FlowerClient())).

PR #1351, opened 2022-08-01
Flower Profiler

Ryan0v0:


.gitignore
```diff
@@ -15,6 +15,7 @@ baselines/datasets/leaf
 
 # macOS
 .DS_Store
+.idea/
```
danieljanes, line 18: This should go into the `# Editor` section

examples/simulation_pytorch/main.py
```diff
@@ -11,6 +11,8 @@
 from dataset_utils import get_cifar_10, do_fl_partitioning, get_dataloader
 from utils import Net, train, test
 
+import os
+from src.ProfilingThread import UtilMonitor
```
danieljanes, line 15: Does this work? It might need to change to something like `from flwr.profiling_thread import ...`

src/py/flwr/server/strategy/__init__.py
```diff
@@ -24,6 +24,7 @@
 from .fedopt import FedOpt as FedOpt
 from .fedyogi import FedYogi as FedYogi
 from .qfedavg import QFedAvg as QFedAvg
+from .ResourceAwareFedAvg import ResourceAwareFedAvg as ResourceAwareFedAvg
```
danieljanes, line 27: This needs to be adjusted according to the recent renaming

src/py/flwr/simulation/app.py
```diff
@@ -63,7 +63,7 @@ def start_simulation(  # pylint: disable=too-many-arguments
     client_fn: Callable[[str], Client],
     num_clients: Optional[int] = None,
     clients_ids: Optional[List[str]] = None,
-    client_resources: Optional[Dict[str, int]] = None,
+    client_resources: Optional[Dict[str, Any]] = None,
```
danieljanes, line 66: What value types are we expecting here? We're generally trying to avoid `Any`.

danieljanes (changes requested):
Thanks for the update @Ryan0v0, I've added a few more comments. Apart from these comments it'd be good to add some basic documentation guide that explains how users can use these new capabilities step by step.

To add new documentation, create a new `.rst` file under `doc/source`, link to the `.rst` file from `index.rst`, add content to the `.rst` file, and then build the docs (`cd docs`, then `make html`) to see if the content is formatted correctly.

Here's an example for such a page: https://github.com/adap/flower/blob/main/doc/source/saving-progress.rst

src/py/flwr/profiling_thread.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def _read_utilization(self):
+        with self.lock:
+            if psutil is not None:
+                self.values["cpu_util_percent"].append(
+                    float(psutil.Process(self.pid).cpu_percent(interval=None))
+                )
+                # self.values["ram_util_percent"].append(
+                #     float(getattr(psutil.Process(self.pid).virtual_memory(), "percent"))
+                # )
```
tanertopal, line 67:
```suggestion
```
Let's remove commented code.

src/py/flwr/profiling_thread.py
```diff
@@ -0,0 +1,100 @@
...
+            if nvsmi is not None:
+                pros = nvsmi.get_gpu_processes()
+                gpus = GPUtil.getGPUs()
+                for g in gpus:
+                    total_memory = g.memoryTotal
+                for pro in pros:
+                    if pro.pid == self.pid:
+                        self.values["vram_util_percent" + str(pro.gpu_id)].append(
+                            float(float(pro.used_memory)/(total_memory-1024)) #(11019-1024))
```
tanertopal, line 76: Could you explain the comment? What does `#(11019-1024))` mean here?

tanertopal (commented):
While reviewing (see comments), I realized that we depend on three libraries.

## psutils
This absolutely makes sense

## GPUtil
Very small single file library without any dependencies

## nvsmi
Contains the part of GPUtil we are using so that we can probably remove GPUtil

See lines in libs: [GPUtil](https://github.com/anderskm/gputil/blob/master/GPUtil/GPUtil.py#L68)
vs. [nvsmi](https://github.com/pmav99/nvsmi/blob/master/nvsmi.py#L126)

Additionally, `nvsmi` is a single file library without any dependencies (except `nvsmi` installed as system dependency) that we could think of vendoring it.

@danieljanes @Ryan0v0 thoughts?

I'll continue the review after our discussion.

src/py/flwr/simulation/app.py
Ryan0v0: We are expecting a float number to represent the fractional GPUs rather than the original int. I changed it to `float` in the latest version.

src/py/flwr/simulation/app.py
```diff
@@ -63,7 +63,7 @@ def start_simulation(  # pylint: disable=too-many-arguments
     client_fn: Callable[[str], Client],
     num_clients: Optional[int] = None,
     clients_ids: Optional[List[str]] = None,
-    client_resources: Optional[Dict[str, int]] = None,
+    client_resources: Optional[Dict[str, float]] = None,
```
danieljanes, line 66: Do we also need to support `int` values? If so, we might need to update the type to `Union[float, int]`

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -29,7 +29,7 @@
 class RayClientProxy(ClientProxy):
     """Flower client proxy which delegates work using Ray."""
 
-    def __init__(self, client_fn: ClientFn, cid: str, resources: Dict[str, int]):
+    def __init__(self, client_fn: ClientFn, cid: str, resources: Dict[str, Any]):
```
danieljanes, line 32: The type of `resources` needs to be the same as the type of `client_resources` in `src/py/flwr/simulation/app.py`.

danieljanes (changes requested):

Ryan0v0:
Thanks @danieljanes and @tanertopal for the comments and suggestions! I think I have resolved most of them except for the package vendoring. 

Regarding the discussion about packages, the reason why I chose to use both `GPUtil` and `psutil` is that, they are "on the same level". The former is the library for retrieving information on running processes and system utilization on CPU, while the latter does the same thing but on GPU, which makes the code clear for readers. However, `GPUtil` cannot get metrics for one particular GPU. That’s why I imported the lower-level package `nvsmi` to get a specific `pid`.

Actually we didn't use the metrics we retrieved from `psutil` in our resource re-allocation. If we are going to remove `GPUtil`, should we still keep the `psutil` package and related codes to get CPU utilization percentage as well? 

Looking forward to hearing your thoughts @tanertopal @danieljanes @pedropgusmao!  Please let me know if there are any other questions you have/changed you would suggest.

PR #1357, opened 2022-08-05
Add wrappers for DP-FedAvg

VasundharaAgarwal:


src/py/flwr/server/strategy/dp_adaptive_clip_strategy.py
```diff
@@ -0,0 +1,80 @@
...
+            strategy, num_sampled_clients, noise_multiplier, init_clip_norm
+        )
+        self.clip_norm_lr = clip_norm_lr
+        self.clip_norm_target_quantile = clip_norm_target_quantile
+        if not clip_count_stddev:
+            self.clip_count_stddev = (
+                self.num_sampled_clients / 20.0 if noise_multiplier else 0
+            )
+        else:
+            self.clip_count_stddev = clip_count_stddev
```
tanertopal, line 43:
```suggestion
        self.clip_count_stddev = clip_count_stddev

        if self.clip_count_stddev is None:
            self.clip_count_stddev = 0
            if noise_multiplier > 0:
                 self.clip_count_stddev = self.num_sampled_clients / 20.0
```

Reducing branching and making sure we don't let Python implicitly convert the float into a boolean but rather do it explicitly by using `noise_multiplier > 0`

tanertopal: Also, I'd rather use `x is None` instead of `not x` as I feel that's faster to grasp.

src/py/flwr/server/strategy/dp_fixed_clip_strategy.py
```diff
@@ -0,0 +1,99 @@
...
+    def configure_fit(
+        self, server_round: int, parameters: Parameters, client_manager: ClientManager
+    ) -> List[Tuple[ClientProxy, FitIns]]:
+        """Configure the next round of training."""
+
+        client_instructions = self.strategy.configure_fit(
+            server_round, parameters, client_manager
+        )
+
+        noise_stddev = self.__calc_client_noise_stddev()  # type: ignore
```
danieljanes, line 59:
```suggestion
        noise_stddev = self._calc_client_noise_stddev()  # type: ignore
```

src/py/flwr/server/strategy/dp_adaptive_clip_strategy.py
```diff
@@ -0,0 +1,80 @@
...
+    ) -> None:
+
+        super().__init__(
+            strategy, num_sampled_clients, noise_multiplier, init_clip_norm
+        )
+        self.clip_norm_lr = clip_norm_lr
+        self.clip_norm_target_quantile = clip_norm_target_quantile
+        if not clip_count_stddev:
+            self.clip_count_stddev = (
+                self.num_sampled_clients / 20.0 if noise_multiplier else 0
```
danieljanes, line 40: Where does the `20.0` come from?

src/py/flwr/server/strategy/dp_adaptive_clip_strategy.py
```diff
@@ -0,0 +1,80 @@
...
+        if noise_multiplier:
+            self.noise_multiplier = (
+                self.noise_multiplier ** (-2) - (2 * self.clip_count_stddev) ** (-2)
+            ) ** (-0.5)
+
+    def __repr__(self) -> str:
+        rep = "Strategy with DP with Adaptive Clipping enabled."
+        return rep
+
+    def __update_clip_norm(self, results: List[Tuple[ClientProxy, FitRes]]) -> None:
```
danieljanes, line 53:
```suggestion
    def _update_clip_norm(self, results: List[Tuple[ClientProxy, FitRes]]) -> None:
```

src/py/flwr/server/strategy/dp_adaptive_clip_strategy.py
```diff
@@ -0,0 +1,80 @@
...
+    def aggregate_fit(
+        self,
+        server_round: int,
+        results: List[Tuple[ClientProxy, FitRes]],
+        failures: List[Union[Tuple[ClientProxy, FitRes], BaseException]],
+    ) -> Tuple[Optional[Parameters], Dict[str, Scalar]]:
+        if failures:
+            return None, {}
+        new_global_model = super().aggregate_fit(server_round, results, failures)
+        self.__update_clip_norm(results)
```
danieljanes, line 79:
```suggestion
        self._update_clip_norm(results)
```

danieljanes (changes requested):
Thanks for the PR @VasundharaAgarwal !

The general structure is quite nice, I quite like the idea of making this "pluggable" and the approach you chose makes this possible without requiring any changes to the Flower core framework abstractions, which is nice.

We left a few review comments, but they are mainly of a cosmetic nature to align the code style  with the rest of the codebase. The only thing missing would be a new page in the docs (under `doc/source`) that explains how these new features can be used, perhaps accompanied by a new code example (under `examples`, could be done in a different PR).

Again, thanks for the PR, this will be a major contribution.

src/py/flwr/server/strategy/dp_fixed_clip_strategy.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class DPFixedClipStrategy(Strategy):
+    """Wrapper for configuring a Strategy for DP with Fixed Clipping."""
+
+    # pylint: disable=too-many-arguments,too-many-instance-attributes
+    def __init__(
+        self,
+        strategy: Strategy,
+        num_sampled_clients: int,
```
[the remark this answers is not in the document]
VasundharaAgarwal: Yes, unfortunately most DP algorithms for FL right now don't really seem to be tolerant of client failures or drop-outs. :(

src/py/flwr/server/strategy/dp_adaptive_clip_strategy.py
VasundharaAgarwal: It's a recommendation from the [paper](https://arxiv.org/pdf/1905.03871.pdf) that the algorithm is based on.

VasundharaAgarwal:
Thanks @danieljanes and @tanertopal for the comments and suggestions! I think I have addressed them all now -- please let me know if there are any other questions you have/changed you would suggest.

Getting started on the docs now! :)

src/py/flwr/server/strategy/dp_fixed_clip_strategy.py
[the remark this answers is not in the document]
danieljanes: Fair enough, should be ok if we mention the implications in our docs

src/py/flwr/server/strategy/dp_adaptive_clip_strategy.py
danieljanes: Would it make sense to expose this as a configurable parameter?

VasundharaAgarwal: `clip_count_stddev` is already exposed as a configurable parameter that can be passed in to the constructor, am i misunderstanding something?

danieljanes: Ah, sorry, you're obviously right, I was just looking at the `20.0` and I totally ignored the larger context.

tanertopal (approved):

tanertopal:
Thank you for the great work @VasundharaAgarwal !

VasundharaAgarwal:
Thanks @tanertopal! I've run the reformatter and test scripts once more  -- looks fine now!

danieljanes (approved):

PR #1377, opened 2022-08-18
Add 'py.typed' to provide type information to packages using flower

nfnt:

As described in [PEP 561](https://peps.python.org/pep-0561/), packages that want to provide type information need to indicate that by including an empy `py.typed` file. This file has been added for the `flwr` package.

#### Any other comments?

Without this change, type-checking code that is using flwr will assume the `Any` type for every object provided by flwr. By adding `py.typed`, type-checking is working as intended with flwr objects.


danieljanes:
Thanks for the PR @nfnt! Is there anything we need to do to ensure that Poetry includes it in the `whl`, or does it happen automatically?

nfnt:
> Thanks for the PR @nfnt! Is there anything we need to do to ensure that Poetry includes it in the `whl`, or does it happen automatically?

This is all that's needed, the file will be included in the `whl`. You can verify this by running

```
poetry build
unzip -l dist/flwr-1.1.0-py3-none-any.whl | grep py.typed
```

I tested that this works as expected by running `mypy` on something very wrong like

```python
import flwr as fl

strategy: dict[str, str] = {}

fl.server.start_server(strategy=strategy)
```

Without `py.typed`, running `mypy` would fail with

```
client.py:1: error: Skipping analyzing "flwr": module is installed, but missing library stubs or py.typed marker
client.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

with `py.typed` instead the error is as expected

```
client.py:5: error: Argument "strategy" to "start_server" has incompatible type "Dict[str, str]"; expected "Optional[Strategy]"
Found 1 error in 1 file (checked 1 source file)
```

PR #1381, opened 2022-08-22
Update FedAvg documentation default values

rob-luke:
#### Reference Issues/PRs

I did not open an issue in advance, is this something you would prefer I do in the future? Or are you ok with issues as PRs?

Modified the doc string for FedAvg to reflect the default values in the function definition.

#### Any other comments?

Thanks for the great package


danieljanes (approved):
Good catch, thanks for the PR @rob-luke !

Issue #1376, opened 2022-08-14
Memory leak when ray.remote are repeatedly called in simulation


mofanv:
### Describe the bug

`ray.remote` in [ray_client_proxy.py](https://github.com/adap/flower/blob/main/src/py/flwr/simulation/ray_transport/ray_client_proxy.py#L113) are repeatedly called when running simulation. Because on default, it will go `Ray::IDLE` mode and rest in GPUs, the occupied GPU memory keeps increasing until CUDA out-of-memory errors happen. 

My solution is to add the `max_call` parameter when calling like `ray.remote(max_calls=1)`, based on ray documents [here](https://docs.ray.io/en/latest/ray-core/tasks/using-ray-with-gpus.html#workers-not-releasing-gpu-resources).

### Steps/Code to Reproduce

Run the simulation_pytorch example with `python main.py`, and monitor memory usage with `watch -n 0.1 nividia-smi`

### Expected Results

Clients release their occupied memory every communication round, to let other clients use it later in the next round.

### Actual Results

All clients have been executed keep the memory, which will cause CUDA OOM.

PR #1384, opened 2022-08-23
simulation, add max_calls arg to ray.remote to avoid rayidle in gpus

### LINKED ISSUE #1152: Exhausting computational resources with multiple fl.simulation.start_simulation runnings (ray)
### Describe the bug

I have been working with the simulation package `fl.simulation.start_simulation`, and after tweaking some parameters and re-running the code it seems (by the error message) to exhaust the computational resources. Probably some kind of garbage collector required ?

During the try-outs, it seems that after a while it works again (probably some gargabe collection from the OS?)

### Steps/Code to Reproduce

I am using as reference the [sim.ipynb](https://github.com/adap/flower/blob/main/examples/quickstart_simulation/sim.ipynb) notebook with a different dataset (bigger than MNIST) and different DNN model.

```
# Create FedAvg strategy
strategy=fl.server.strategy.FedAvg(
        fraction_fit=0.1,  # Sample 10% of available clients for training
        fraction_eval=0.05,  # Sample 5% of available clients for evaluation
        min_fit_clients=5,  # Never sample less than 10 clients for training
        min_eval_clients=5,  # Never sample less than 5 clients for evaluation
        min_available_clients=int(NUM_CLIENTS * 0.75),  # Wait until at least 75% clients are available
)

# Start simulation
fl.simulation.start_simulation(
    client_fn=client_fn,
    num_clients=NUM_CLIENTS,
    clients_ids=clients_list,
    num_rounds=5,
    strategy=strategy,
)
```

### Expected Results

Output of INFO, DEBUG, launch_and_fit, launch_and_evaluate messages.

### Actual Results

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [78], in <cell line: 13>()
      4 strategy=fl.server.strategy.FedAvg(
      5         fraction_fit=0.1,  # Sample 10% of available clients for training
      6         fraction_eval=0.05,  # Sample 5% of available clients for evaluation
   (...)
      9         min_available_clients=int(NUM_CLIENTS * 0.75),  # Wait until at least 75% clients are available
     10 )
     12 # Start simulation
---> 13 fl.simulation.start_simulation(
     14     client_fn=client_fn,
     15     num_clients=NUM_CLIENTS,
     16     clients_ids=clients_list,
     17     num_rounds=5,
     18     strategy=strategy,
     19 )

File ~/fl/venv/lib/python3.9/site-packages/flwr/simulation/app.py:143, in start_simulation(client_fn, num_clients, clients_ids, client_resources, num_rounds, strategy, ray_init_args)
    140     ray.shutdown()
    142 # Initialize Ray
--> 143 ray.init(**ray_init_args)
    144 log(
    145     INFO,
    146     "Ray initialized with resources: %s",
    147     ray.cluster_resources(),
    148 )
    150 # Initialize server and server config

File ~/fl/venv/lib/python3.9/site-packages/ray/_private/client_mode_hook.py:105, in client_mode_hook.<locals>.wrapper(*args, **kwargs)
    103     if func.__name__ != "init" or is_client_mode_enabled_by_default:
    104         return getattr(ray, func.__name__)(*args, **kwargs)
--> 105 return func(*args, **kwargs)

File ~/fl/venv/lib/python3.9/site-packages/ray/worker.py:933, in init(address, num_cpus, num_gpus, resources, object_store_memory, local_mode, ignore_reinit_error, include_dashboard, dashboard_host, dashboard_port, job_config, configure_logging, logging_level, logging_format, log_to_driver, namespace, runtime_env, _enable_object_reconstruction, _redis_max_memory, _plasma_directory, _node_ip_address, _driver_object_store_memory, _memory, _redis_password, _temp_dir, _metrics_export_port, _system_config, _tracing_startup_hook, **kwargs)
    895     ray_params = ray._private.parameter.RayParams(
    896         node_ip_address=node_ip_address,
    897         raylet_ip_address=raylet_ip_address,
   (...)
    927         metrics_export_port=_metrics_export_port,
    928         tracing_startup_hook=_tracing_startup_hook)
    929     # Start the Ray processes. We set shutdown_at_exit=False because we
    930     # shutdown the node in the ray.shutdown call that happens in the atexit
    931     # handler. We still spawn a reaper process in case the atexit handler
    932     # isn't called.
--> 933     _global_node = ray.node.Node(
    934         head=True,
    935         shutdown_at_exit=False,
    936         spawn_reaper=True,
    937         ray_params=ray_params)
    938 else:
    939     # In this case, we are connecting to an existing cluster.
    940     if num_cpus is not None or num_gpus is not None:

File ~/fl/venv/lib/python3.9/site-packages/ray/node.py:274, in Node.__init__(self, ray_params, head, shutdown_at_exit, spawn_reaper, connect_only)
    268         self.get_gcs_client().internal_kv_put(
    269             b"tracing_startup_hook",
    270             ray_params.tracing_startup_hook.encode(), True,
    271             ray_constants.KV_NAMESPACE_TRACING)
    273 if not connect_only:
--> 274     self.start_ray_processes()
    275     # we should update the address info after the node has been started
    276     try:

File ~/fl/venv/lib/python3.9/site-packages/ray/node.py:1114, in Node.start_ray_processes(self)
   1110 # Make sure we don't call `determine_plasma_store_config` multiple
   1111 # times to avoid printing multiple warnings.
   1112 resource_spec = self.get_resource_spec()
   1113 plasma_directory, object_store_memory = \
-> 1114     ray._private.services.determine_plasma_store_config(
   1115         resource_spec.object_store_memory,
   1116         plasma_directory=self._ray_params.plasma_directory,
   1117         huge_pages=self._ray_params.huge_pages
   1118     )
   1119 self.start_raylet(plasma_directory, object_store_memory)
   1120 if self._ray_params.include_log_monitor:

File ~/fl/venv/lib/python3.9/site-packages/ray/_private/services.py:1899, in determine_plasma_store_config(object_store_memory, plasma_directory, huge_pages)
   1895     plasma_directory = "/dev/shm"
   1896 elif (not os.environ.get("RAY_OBJECT_STORE_ALLOW_SLOW_STORAGE")
   1897       and object_store_memory >
   1898       ray_constants.REQUIRE_SHM_SIZE_THRESHOLD):
-> 1899     raise ValueError(
   1900         "The configured object store size ({} GB) exceeds "
   1901         "/dev/shm size ({} GB). This will harm performance. "
   1902         "Consider deleting files in /dev/shm or increasing its "
   1903         "size with "
   1904         "--shm-size in Docker. To ignore this warning, "
   1905         "set RAY_OBJECT_STORE_ALLOW_SLOW_STORAGE=1.".format(
   1906             object_store_memory / 1e9, shm_avail / 1e9))
   1907 else:
   1908     plasma_directory = ray._private.utils.get_user_temp_dir()

ValueError: The configured object store size (33.928609792 GB) exceeds /dev/shm size (33.928609792 GB). This will harm performance. Consider deleting files in /dev/shm or increasing its size with --shm-size in Docker. To ignore this warning, set RAY_OBJECT_STORE_ALLOW_SLOW_STORAGE=1.
```

mofanv:
#### Reference Issues/PRs
Fixes #1152 and #1376 

Issue: `@ray.remote` in [ray_client_proxy.py](https://github.com/adap/flower/blob/main/src/py/flwr/simulation/ray_transport/ray_client_proxy.py#L113) is repeatedly called when running simulation. By default, after each client finishes its training, the ray worker still rests in GPUs as `Ray::IDLE`. This accumulates and causes CUDA memory to run out. 

Change: By adding one argument `@ray.remote(max_calls=1)`, each ray worker will be removed after every client finishes.

#### Any other comments?
None

### DIFF
--- a/src/py/flwr/simulation/ray_transport/ray_client_proxy.py
+++ b/src/py/flwr/simulation/ray_transport/ray_client_proxy.py
@@ -91,7 +91,7 @@ class RayClientProxy(ClientProxy):
         return common.DisconnectRes(reason="")  # Nothing to do here (yet)
 
 
-@ray.remote
+@ray.remote(max_calls=1)
 def launch_and_get_properties(
     client_fn: ClientFn, cid: str, get_properties_ins: common.GetPropertiesIns
 ) -> common.GetPropertiesRes:
@@ -100,7 +100,7 @@ def launch_and_get_properties(
     return client.get_properties(get_properties_ins)
 
 
-@ray.remote
+@ray.remote(max_calls=1)
 def launch_and_get_parameters(
     client_fn: ClientFn, cid: str, get_parameters_ins: common.GetParametersIns
 ) -> common.GetParametersRes:
@@ -109,7 +109,7 @@ def launch_and_get_parameters(
     return client.get_parameters(get_parameters_ins)
 
 
-@ray.remote
+@ray.remote(max_calls=1)
 def launch_and_fit(
     client_fn: ClientFn, cid: str, fit_ins: common.FitIns
 ) -> common.FitRes:
@@ -118,7 +118,7 @@ def launch_and_fit(
     return client.fit(fit_ins)
 
 
-@ray.remote
+@ray.remote(max_calls=1)
 def launch_and_evaluate(
     client_fn: ClientFn, cid: str, evaluate_ins: common.EvaluateIns
 ) -> common.EvaluateRes:


danieljanes:
Thanks for the PR @mofanv !

CC'ing @pedropgusmao @jafermarq @Ryan0v0

jafermarq:
Thanks for looking into this @mofanv. Have you noticed any slowdown in training when for example you use many workers? (i.e. more than what you can have running concurrently). I suspect, using `max_calls=1` will create some overheads when spawning a new worker in the same round.

mofanv:
Hi, I have not noticed any significant slowdown with `max_calls=1` to close ray workers. How many workers do you think I can test for such spawning overheads?
My setting now is using 10 out of 100 clients each round as in https://github.com/adap/flower/blob/aa9b9467fbb7b25d321ae0208434adc71e28bff2/examples/simulation_pytorch/main.py#L159. I changed the model to `mobilenetv3` and it started to have `CUDA out of memory` error.

jafermarq:
I'd imagine the slowdown will only happen in settings were you N clients per round but your system (e.g. GPUs and CPUs) can only accommodate M at a given time, M<N. For example. If you have 50 clients per round but your system can only run 4 concurrently, then the remaining 46 will be scheduled by ray to run later, once some resources have been freed. In this example, I would imagine that having `max_calls=1` hurts performance because resources need to be deallocated (when client ends -- i.e. when the `call` ends) and re-allocated again (when a new client in the round is spawned)

About the `CUDA out of memory` error, how are you setting the `num_gpus` resources for the clients? This would be a fractional number [0:1] representing the portion of the GPU memory that a single client needs. (this value is different depending on how much VRAM your GPU has)

mofanv:
Thanks for the clarification. I wonder will such re-allocation causes a significant overhead. As I can imagine that most of the execution time is due to model training on GPUs? Also, if I understand it correctly, resting clients, i.e., ray workers, for later use still holds the memory. But this may occupy a large amount of memory when clients are training a large model?

I tried 8 clients on 8 GPUs (i.e., `num_gpus=1`) and also 8 clients on 4GPUs (i.e., `num_gpus=0.5`), both have this GPU OOM problem. My machine has 8 V100 GPUs. Each GPU has 16GB memory. I monitored GPUs usage by `watch -n0.1 nvidia-smi`, and observed more `Ray::IDLE` processes resting in GPUs every round. Each round costs around 1.5G of memory for `mobilenetv3`. 

I have also tested on the original `simulation_pytorch` example, and the accumulated memory issue also happens, but just less severe because the model size is small.

jafermarq:
Which version of Ray do you use? Ray 1.12 resulted in an issue similar to what you describe "more `Ray::IDLE` processes resting in GPUs every round". Could you downgrade to 1.11.1 and try again? `pip install ray==1.11.1

mofanv:
I see. I was using Ray 1.13, and by downgrading it to 1.11.1, no `Ray::IDLE` anymore. I can run it on `simulation_pytorch` example now, but failed with my own project. It stops with 
```
The configured object store size (10.99511808 GB) exceeds /dev/shm size (10.99511808 GB). This will harm performance. Consider deleting files in /dev/shm or increasing its size with --shm-size in Docker. To ignore this warning, set RAY_OBJECT_STORE_ALLOW_SLOW_STORAGE=1.
```
With Ray 1.13, there is no such problem anyway.

So I wonder which do you suggest for now? i) use Ray 1.11.1 and wait for Ray 1.13 or later version to fix the problem, or ii) just add `max_calls=1` in order to use Ray 1.13 without OOM

mofanv:
I checked again. This OOM problem also exists in Ray 2.0.0

mofanv:
> @mofanv Where should the `ray.remote(max_calls=1)` parameter? When I add it to the clients `fit` function, the functionality is broken.

Not sure is it still a issue in the current version. I think the official suggestion is to downgrade Ray version.
But if you still want to do the trick, then changing `ray.remote` in this and all following lines here
https://github.com/adap/flower/blob/713c008ad5bcec53d5d954c20824c82322d1d3e4/src/py/flwr/simulation/ray_transport/ray_client_proxy.py#L118

vtsouval:
@mofanv Thank you. 

Another workaround that worked for me was to limit the number number of CPU cores that ray can utilize to be equal to the utilized CPU cores from client per round. This can be set in the initialization parameters of ray in `start_simulation` (i.e. ray_init_args("num_cpus": 10)). This way all IDLE clients are removed in each federated round.

pedropgusmao:
`max_calls=1` can be a problem if you increase the number of clients per round. I'm running experiments where, **without** `max_calls=1`, each round with 100 clients takes about 40 seconds. If I set `max_calls=1`, the required time goes up to 10 minutes!

kuihao:
Thank you very much, you really saved my life, I checked all night long and finally solved the problem, the method that changing all `@ray.remote` to `@ray.remote(max_calls=1)` really works. Not only does it save space, but it also reduces the running time from 15 minutes to 1 minute.

As for **downloading the ray version**, it **didn't work** for me. After I downloaded the version from `ray==2.2.0` to `ray==1.11.1`, all the processes died as soon as they were sent to cuda, and the flower server didn't know why it didn't wait for the cuda to return and end itself.

mxbi:
Potentially useful: https://docs.ray.io/en/latest/ray-core/tasks/using-ray-with-gpus.html#workers-not-releasing-gpu-resources
The issue seems to be that Ray assumes GPU resources are freed entirely after a task is finished, which is far from guaranteed.

In my case, a very simple FL on a tiny dataset ran out of memory on my 24GB RTX 3090 after about 15 clients/5 rounds, because new threads were being launched for every round and not being freed.

Additionally, this behaviour can vary from system-to-system because Ray seems to keep around `num_cpus` clients lazily before it starts re-using them. In my case, it worked on a Colab notebook with 2 threads, but not on my machine with 48 threads. Essentially, ray's default behaviour is to delay garbage collection until it runs out of _CPU threads_, rather than running out of GPU memory.

As @vtsouval suggested, pointing users to set `ray_init_args={"num_cpus": MAX_WORKERS}` could be a better option, where `MAX_WORKERS=n_clients` ensures maximum parallelisation (assuming you have the VRAM for it). This fixed the memory leak for me **while retaining the performance benefit of re-using workers between rounds**. Setting `max_calls=1` seems to destroy and recreate the process on every call, which as noted can harm performance greatly.

(Caveat: I'm a first-time Flower user, so I could be off-base here).

jafermarq:
@mxbi  if you use Ray 1.11 it will work just fine (as i indicated earlier in this thread https://github.com/adap/flower/pull/1384#issuecomment-1226118068) . Afaik this issue with Flower+Ray happens with Ray 2+

@pedropgusmao @danieljanes why aren't we pinning Ray 1.11 to Flower simulation?

pedropgusmao:
Tbh, I don't think keeping an older version as the default is good long-term as this might lead to broken dependencies later on. 
Ray already moved to`max_calls=1` in their [2.2 version](https://docs.ray.io/en/latest/ray-core/tasks/using-ray-with-gpus.html?highlight=max_calls#workers-not-releasing-gpu-resources), but I still think we might want to look for alternatives.

alexkyllo:
I'm running into this issue with flwr==1.3.0 and ray==2.2.0. I need to train 10 clients per round on 1 GPU, but only 5 clients will fit in my GPU memory. Ray keeps creating processes and leaving them "IDLE", consuming memory until I get a CUDA OOM error. I've tried changing the arguments like `start_simulation(client_resources={"num_cpus": 8,"num_gpus": 1})`. I can't downgrade to ray==1.11.1 because it only supports Python <=3.9 and I'm on Python 3.10. Does anyone know of another workaround?

kuihao:
@alexkyllo  Although I am using python 3.9 (ray 2.2.0, flwr1.2.0), I have chosen not to downgrade the ray version and still use ray 2.2.0. 

Firstly, open `ray_client_proxy.py` (<pip package installed dir>/lib/python3.9/site-packages/flwr/simulation/ray_transport/ray_client_proxy.py) and **select all `@ray.remote` and replace them with `@ray.remote(max_calls=1)`**. This will ensure that the ray actors (clients) are cleared when idle, rather than occupying v-ram on the GPU (Nvidia RTX 3090, GTX 1080 Ti, GTX 1070, A6000, A4000, P5000). At least, this method has been effective for me. You may want to give it a try.

alexkyllo:
> @alexkyllo  Although I am using python 3.9 (ray 2.2.0, flwr1.2.0), I have chosen not to downgrade the ray version and still use ray 2.2.0. 
> 
> Firstly, open `ray_client_proxy.py` (<pip package installed dir>/lib/python3.9/site-packages/flwr/simulation/ray_transport/ray_client_proxy.py) and **select all `@ray.remote` and replace them with `@ray.remote(max_calls=1)`**. This will ensure that the ray actors (clients) are cleared when idle, rather than occupying v-ram on the GPU (Nvidia RTX 3090, GTX 1080 Ti, GTX 1070, A6000, A4000, P5000). At least, this method has been effective for me. You may want to give it a try.

Thanks, this works, but I will need to fork the flwr package to make it work across my environments. 

I was a little confused because @pedropgusmao 's comment mentioned that `max_calls=1` was the default in Ray 2.2 but this doesn't appear to be the case after all. 

Would it be a good idea to enhance flwr to make this `max_calls` thing configurable via an argument to `start_simulation` or something?

PR #1401, opened 2022-09-01
Upgrading to Ray 2.0

pedropgusmao:
#### Reference Issues/PRs
NA

Upgrades Ray version to 2.0.0

#### Any other comments?

Fixed the Private Variable Access warning by `ray.worker.get()` to `ray.get()`  as suggested by the API.

### DIFF
--- a/examples/simulation_pytorch/main.py
+++ b/examples/simulation_pytorch/main.py
@@ -38,7 +38,7 @@ class FlowerClient(fl.client.NumPyClient):
         set_params(self.net, parameters)
 
         # Load data for this client and get trainloader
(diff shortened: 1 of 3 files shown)


danieljanes:
Thanks for the PR @pedropgusmao, do we need to change other instances or `ray.worker.get` as well?

pedropgusmao:
Sorry, I missed the example and two other calls to `ray.worker.` I checked with `grep` and couldn't find any other calls. The example also works.

danieljanes (approved):
Thanks for the update @pedropgusmao, lgtm!

@tanertopal thoughts?

tanertopal (approved):

PR #1413, opened 2022-09-06
Checkpointing Pytorch Example

Linardos:

This is a minimal implementation of PyTorch-based checkpointing. The logic is sound, but I must be missing something as performance is not maintained upon loading the stored state_dict.

#### Any other comments?

This PR presents a minimal reproducible example of a checkpointing issue I'm facing. The code runs normally and stores the checkpoint, but the checkpoint doesn't maintain the performance once I load it. I'm hoping we can figure out why by working on this together.

### DIFF
--- a/examples/checkpointing_pytorch/server_checkpointing.py
+++ b/examples/checkpointing_pytorch/server_checkpointing.py
@@ -1,0 +1,53 @@
+from typing import List, Optional, Tuple
+from collections import OrderedDict
+import glob
+import os
+
+import torch
+import flwr as fl
+
+import cifar
+
+DEVICE: str = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+# Load PyTorch model
+net = cifar.Net().to(DEVICE)
+continue_from_checkpoint = True 
+class SaveModelStrategy(fl.server.strategy.FedAvg):
+    def aggregate_fit(
+        self,
+        rnd: int,
+        results: List[Tuple[fl.server.client_proxy.ClientProxy, fl.common.FitRes]],
+        failures: List[BaseException],
+    ) -> Optional[fl.common.Weights]:
+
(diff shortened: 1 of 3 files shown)


examples/checkpointing_pytorch/server_checkpointing.py
```diff
@@ -0,0 +1,59 @@
...
+    def load_parameters_from_disk():
+        # import Net
+        list_of_files = [fname for fname in glob.glob("./model_round_*")]
+        latest_round_file = max(list_of_files, key=os.path.getctime)
+        print("Loading pre-trained model from: ", latest_round_file)
+        state_dict = torch.load(latest_round_file)
+        net.load_state_dict(state_dict)
+
+strategy = SaveModelStrategy(
+    initial_parameters=load_parameters_from_disk() if continue_from_checkpoint else None
```
danieljanes, line 52: `load_parameters_from_disk()` needs to return a `Parameters` object, but the function definition above does not even have a return statement.

You could something like this after `net.load_state_dict(state_dict)` on line 49 (no guarantees):

```
weights = [val.cpu().numpy() for _, val in self.net.state_dict().items()]
return fl.common.ndarrays_to_parameters(weights)
```

See also: https://github.com/adap/flower/blob/2728140a28514b3174fafee0c64d368e8b568645/src/py/flwr/server/strategy/fedavg.py#L72

danieljanes (changes requested):

WilliamLindskog:
Thanks for opening this! 

I'm closing this PR as it is now built in - see [ref](https://flower.ai/docs/framework/main/en/tutorial-series-build-a-strategy-from-scratch-pytorch.html).

This other [how-to page](https://flower.ai/docs/framework/main/en/how-to-save-and-load-model-checkpoints.html) also present the same idea, assuming one is a bit familiar with `serverapp`/`clientapp`.

PR #1446, opened 2022-10-05
Create experimental mt-pytorch code example

danieljanes:


examples/mt-pytorch/driver.py
```diff
@@ -0,0 +1,86 @@
...
+            req=get_results_req
+        )
+        # ------------------------------------------------------------------ Driver SDK
+
+        results: List[Result] = get_results_res.results
+        print(f"Got {len(get_results_res.results)} results")
+
+        all_results += results
+        if len(all_results) == len(task_ids):
+            break
```
tanertopal, line 76: @danieljanes We don't have an early stop here but as this is all experimental I am just noting my thoughts here :)

tanertopal: Oh, also would be cool to use `tqdm` here too.

tanertopal (approved):
As all of this is experimental I only left a note but apart from that lgtm.

examples/mt-pytorch/driver.py
danieljanes: I agree that these things would be cool, but my goal was more to create a minimal example here, which is why I'd suggest to keep is focussed :) regarding the early stopping, my assumption was that the server would enforce a timeout on all tasks anyways, so the driver can just wait until it gets either a full result or something that tells it that a particular task did not finish in time.

tanertopal (approved):

PR #1450, opened 2022-10-07
Support optional client methods in VCE

danieljanes:


tanertopal (approved):

tanertopal (approved):
@danieljanes Not sure about the `maybe_*` naming, but I also have no better idea.

Overall looks good to me!

PR #1462, opened 2022-10-23
Move client-side message handler into a separate package

danieljanes:


src/py/flwr/client/message_handler/message_handler.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
```
tanertopal, line 12: Would it maybe make sense to move this one up? So it is only `flwr.client import message_handler`? @danieljanes wdyt?

danieljanes: @tanertopal that'd be an option. Having it in a separate package makes it easier to break `message_handler` up into separate files if the complexity increases (which I think it will), which is nice. Wdyt?

tanertopal (approved):

PR #1466, opened 2022-10-26
Create initial FED-0002 outline

danieljanes:


enhancements/0002-20221026-experimental-rest-transport-layer.md
```diff
@@ -0,0 +1,156 @@
...
+- Server responds in one of two ways:
+  - Reconnect at a later time.
+  - Execute a task and send the result.
+- Server has a (mild) form of control over the conversation by suggesting to the client when it should connect again.
+
+Example:
+- Client connects to the server: "I'm available"
+- Server doesn't have any tasks: "reconnect in 5min"
+- Client sleeps for 5min, then connects again
+- Server say: here's a TASK and a TOKEN, work on it and send me your RESULT
```
LorenzoMinto, line 76: an idea for TOKEN implementation that supports anonymous redemption: https://privacypass.github.io/

LorenzoMinto, line 76: Would it be possible (not sure how this ties in with the multi-tenancy update) to send down a list of available tasks instead of a single one? We were thinking of having some locally enforced task-specific eligibility criteria (i.e. does client have more than X samples for task A?), so not all tasks might be _runnable_ by a client that has advertised its availability. The current version with the single TASK shipped down can be a special case of the list of TASKs

LorenzoMinto, line 76: What does a TASK object look like?

enhancements/0002-20221026-experimental-rest-transport-layer.md
```diff
@@ -0,0 +1,156 @@
...
+3. Return `Task` + `token` to the client in the `GetTasksResponse`
+
+**POST /results**
+
+Request body: `SubmitResultRequest` containing `Result` + `token`
+Response body: `None`
+
+Implementation:
+1. Check if the `token` is known
+2. Save `Result` if true, else return error
```
LorenzoMinto, line 101: If the token has expired and new task is in an early "distribution" phase we could distribute the update task with a new token since we indirectly know the client is available? This might be a way to recover the straggler, but not sure how much of an optimization this actually is

enhancements/0002-20221026-experimental-rest-transport-layer.md
```diff
@@ -0,0 +1,156 @@
...
+Messages types should stay identical to the existing ones, particularly the ones used in the Driver API:
+
+- `Task` wrapping a `ServerMessage` with its nested message types (`FitIns`, `EvaluateIns`, ...)
+- `Result` wrapping a `ClientMessage` with its nested message types (`FitRes`, `EvaluateRes`, ...) 
+
+### Security considerations
+
+**HTTPS-only:** Clients are expected to connect to the server only via HTTPS, never via plain HTTP.
+
+**Poisoning attacks:** [TODO, which mechanisms do we want to support/enforce here?]
```
LorenzoMinto, line 122: would be very interesting to explore https://privacypass.github.io/ as a solution

enhancements/0002-20221026-experimental-rest-transport-layer.md
LorenzoMinto: leaving the client to pick the task may add to the inefficiency of the protocol, but a simple priority/pacing system for the task might help with that

enhancements/0002-20221026-experimental-rest-transport-layer.md
```diff
@@ -0,0 +1,156 @@
...
+Task: User-defined specification of what a client has to do, with the expectation that the client returns a result.
+
+### Protocol
+
+General idea:
+- Client always initiates conversation
+- Server responds in one of two ways:
+  - Reconnect at a later time.
+  - Execute a task and send the result.
+- Server has a (mild) form of control over the conversation by suggesting to the client when it should connect again.
```
LorenzoMinto, line 70: is the goal here to have some sort of pace steering?

enhancements/0002-20221026-experimental-rest-transport-layer.md
pedropgusmao: We might need to understand this protocol better first.

pedropgusmao: Actually, the implementation will return a list of tasks instead of a single task. We need to update the document.

enhancements/0002-20221026-experimental-rest-transport-layer.md
pedropgusmao: That sounds interesting. We might introduce this check, and if (the token has expired and there are some tasks for that client) then send the list of tasks right away.

enhancements/0002-20221026-experimental-rest-transport-layer.md
pedropgusmao: Yes. There won't be anything "forcing" the client to follow this, but it would be good practice. This can be seen as both pace steering and limiting the client selection.

enhancements/0002-20221026-experimental-rest-transport-layer.md
LorenzoMinto: I think it's completely fine if we implement a simple token at first and then we implement this protocol at a later moment

enhancements/0002-20221026-experimental-rest-transport-layer.md
danieljanes: Confirmed, I've updated the doc to reflect this.

enhancements/0002-20221026-experimental-rest-transport-layer.md
danieljanes: True, the wording should have been more precise, I've updated the doc

enhancements/0002-20221026-experimental-rest-transport-layer.md
```diff
@@ -0,0 +1,163 @@
...
+  - Reconnect at a later time.
+  - Execute a task and send the result.
+- Server has a (mild) form of control over the conversation by suggesting to the client when it should connect again. This can be used, among other things, to implement pace steering.
+
+Example:
+- Client connects to the server: "I'm available"
+- Server doesn't have any tasks: "reconnect in 5min"
+- Client sleeps for 5min, then connects again
+- Server say: here's a set of TASK-TOKEN-PAIRS, work on them and send me the RESULTS
+- Client receives the set of TASK-TOKEN-PAIRS, executes them locally (concurrently or sequentially, up to the client), and returns the RESULT (one after another or multiple results together, up to the client)
```
LorenzoMinto, line 77: I would make it optional for the client to execute the tasks. Maybe it's already like this, but from here it sounds like the server's response is an imperative and all assigned tasks have to be executed. 

The client might be unable to run a task because it doesn't have enough data for it or some other criteria doesn't check. But also it's a privacy advantage if there is some plausible deniability at the client side (i.e. it is not certain that a client will contribute to this round--straggler or not).

Oversampling might be enough to account for this potential "local rejection", or else the client could send a message/receipt to the server with the tasks it's going to run (in a more privacy preserving version we would add some noise to this _task execution promise_)

enhancements/0002-20221026-experimental-rest-transport-layer.md
danieljanes: Interesting, we discussed something along these lines where the server would reply in a way that tells the client what to do next. The most generic version of this might be a design where we only have one method: the client tells the server "hey, I'm here" or "hey, I'm here and I have these result/token pairs for you" and the server replies "reconnect later" or "here are one or more tasks, please work on them".

enhancements/0002-20221026-experimental-rest-transport-layer.md
danieljanes: Thanks @LorenzoMinto, good points, I updated the doc. Feel free to resolve this thread if you feel that the doc fully reflects our understanding.

PR #1472, opened 2022-10-29
Fixed epoch_loss calculation in intro tutorial

birnbaum:

There seems to be a tiny bug in the tutorial notebook, the `train` function sums up the loss values for each value in the *train* dataset but then divides by the number of *test* samples.


danieljanes:
Good catch @birnbaum, thanks for the PR!

danieljanes (approved):

PR #1478, opened 2022-10-31
Create experimental REST stack

danieljanes:
How to use it:
1. Install wheel with extra `rest`: `pip install 'flwr-1.2.0-py3-none-any.whl[rest]'`
2. Start the server: ` flower-server --rest` or `uvicorn flwr.server.rest_server.rest_api:app --port 8000`
4. Start the client:
    1. `cd examples/quickstart_rest_api`
    2. `python3 client.py` (check if the port in `client.py` is the same as the port used in step 1)

### DIFF
--- a/examples/quickstart_rest_api/client.py
+++ b/examples/quickstart_rest_api/client.py
@@ -1,0 +1,16 @@
+from collections import OrderedDict
+
+import flwr as fl
+
+
+# Define Flower client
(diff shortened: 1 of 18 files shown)


pedropgusmao:
Runnig Server: `uvicorn flwr.server.rest_server.rest_api:app --reload --port 8080` check if the port is the same of the client

PR #1496, opened 2022-12-01
Change docs title to 'Configure Clients'

adam-narozniak:

For our documentation, we’ve started to use the Diàtaxis framework: [https://diataxis.fr/](https://diataxis.fr/)

Our “How to” guides should have titles that continue the sencence “How to …”, for example, “How to upgrade to Flower 1.0”.

Most of our guides do not follow this new format yet, and changing their title is (unfortunately) more involved than one might think.

This issue is about changing the title “Configuring Clients” to “****Configure Clients****”. Does this pass our check?

Before: ”How to c****onfiguring clients****” ❌

After: ”How to c****onfigure clients****” ✅


danieljanes (approved):
Lgtm!

PR #1504, opened 2022-12-09
Add Python version modularity in venv scripts

charlesbvll:

The `dev/venv-create.sh` and `baselines/dev/venv-create.sh` scripts didn't work with versions of Python that were not `3.7.12`.  Now the scripts will also install the right Python version through `pyenv install $version` before creating the virtual environment.

By default the Python version used is `3.7.12`, but an other version can be given as argument of the script, for example :

`./dev/create-venv.sh 3.8.10` will create the virtual environment `flower-3.8.10` using `Python 3.8.10`.

Note that the `dev/venv-delete.sh` and `baselines/dev/venv-delete.sh` will work in the same fashion. By default, it will delete the virtual environment that is using `Python 3.7.12` but an argument can be given to delete a virtual environment that uses another version. If we continue with our previous example, 

`./dev/delete-venv.sh 3.8.10` will delete the virtual environment `flower-3.8.10` that we created with the previous command.

#### Any other comments?

This [page](https://flower.dev/docs/getting-started-for-contributors.html) of the docs would need to be updated accordingly (even though the changes shouldn't break the explained steps) .

### DIFF
--- a/baselines/dev/venv-create.sh
+++ b/baselines/dev/venv-create.sh
@@ -4,7 +4,8 @@ cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/../
 
 version=${1:-3.7.12}
 
-# Check if the directory does not exist and if so, install python
+# Check if the directory for the Python version does not exist and if so, 
+# install the right Python version through pyenv
 [[ ! -d $PYENV_ROOT/versions/$version ]] && pyenv install $version
 
 pyenv virtualenv $version baselines-$version
--- a/baselines/dev/venv-create.sh
+++ b/baselines/dev/venv-create.sh
@@ -2,5 +2,9 @@
 set -e
 cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/../
 
-pyenv virtualenv 3.7.12 baselines-3.7.12
-echo baselines-3.7.12 > .python-version
+version=${1:-3.7.12}
+
+[[ ! -d $PYENV_ROOT/versions/$version ]] && pyenv install $version
+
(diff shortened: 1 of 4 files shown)


baselines/dev/venv-create.sh
```diff
@@ -2,5 +2,9 @@
 set -e
 cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/../
 
-pyenv virtualenv 3.7.12 baselines-3.7.12
-echo baselines-3.7.12 > .python-version
+version=${1:-3.7.12}
+
+[[ ! -d $PYENV_ROOT/versions/$version ]] && pyenv install $version
```
tanertopal, line 7:
```suggestion
# Check if the directory does not exist and if so, install python
[[ ! -d $PYENV_ROOT/versions/$version ]] && pyenv install $version
```

dev/venv-create.sh
```diff
@@ -2,5 +2,9 @@
 set -e
 cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/../
 
-pyenv virtualenv 3.7.12 flower-3.7.12
-echo flower-3.7.12 > .python-version
+version=${1:-3.7.12}
+
+[[ ! -d $PYENV_ROOT/versions/$version ]] && pyenv install $version
```
tanertopal, line 7:
```suggestion
# Check if the directory does not exist and if so, install python
[[ ! -d $PYENV_ROOT/versions/$version ]] && pyenv install $version
```

tanertopal (changes requested):
@charlesbvll I like it! I just suggested adding two comments so its easier to understand whats happening. We can merge it after you apply the suggested changes.

danieljanes (approved):

PR #1506, opened 2022-12-13
Setup script

charlesbvll:

Adds scripts in `dev/` and `baselines/dev/` to set everything up automatically with the default parameters, that is: 

* Install `pyenv` with `virtualenv`
* Add it to `$PATH`
* Create the virtual environment (`Python 3.7.12`)
* Install all the dependancies

#### Any other comments?

This might need to be added to the [contributing docs](https://flower.dev/docs/getting-started-for-contributors.html).


baselines/dev/setup-defaults.sh
```diff
@@ -0,0 +1,25 @@
...
+    # Init pyenv with the shell
+    echo 'eval "$(pyenv init -)"' >> $rcfile
+    echo 'eval "$(pyenv virtualenv-init -)"' >> $rcfile
+    source $rcfile
+fi
+
+# Create the virtual environment for Flower baselines
+$( dirname "${BASH_SOURCE[0]}" )/venv-create.sh
+
+# Install the dependancies inside the virtual environment
```
danieljanes, line 24:
```suggestion
# Install the dependencies inside the virtual environment
```

baselines/dev/setup-defaults.sh
```diff
@@ -0,0 +1,25 @@
...
+    rcfile=$HOME/.$(basename $SHELL)rc
+
+    # Add pyenv to $PATH
+    echo 'export PATH="$HOME/.pyenv/bin:$PATH"' >> $rcfile
+
+    # Init pyenv with the shell
+    echo 'eval "$(pyenv init -)"' >> $rcfile
+    echo 'eval "$(pyenv virtualenv-init -)"' >> $rcfile
+    source $rcfile
+fi
```
danieljanes, line 19: Should we add an `else` block that updates `pyenv` (and the virtualenv plugin) by doing a `git pull`?

dev/setup-defaults.sh
```diff
@@ -0,0 +1,25 @@
...
+    # Init pyenv with the shell
+    echo 'eval "$(pyenv init -)"' >> $rcfile
+    echo 'eval "$(pyenv virtualenv-init -)"' >> $rcfile
+    source $rcfile
+fi
+
+# Create the virtual environment for Flower
+$( dirname "${BASH_SOURCE[0]}" )/venv-create.sh
+
+# Install the dependancies inside the virtual environment
```
danieljanes, line 24:
```suggestion
# Install the dependencies inside the virtual environment
```

danieljanes (changes requested):

baselines/dev/setup-defaults.sh
charlesbvll: Yes good idea ! I actually used the `pyenv update` built-in command.

baselines/dev/setup-defaults.sh
```diff
@@ -0,0 +1,28 @@
...
+    # Add pyenv to $PATH
+    echo 'export PATH="$HOME/.pyenv/bin:$PATH"' >> $rcfile
+
+    # Init pyenv with the shell
+    echo 'eval "$(pyenv init -)"' >> $rcfile
+    echo 'eval "$(pyenv virtualenv-init -)"' >> $rcfile
+    source $rcfile
+else
+    # If pyenv is already installed, check for a newer version
+    pyenv update
```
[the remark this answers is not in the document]
charlesbvll: Oh yes, I will add a check, it is a plugin that also updates all the other plugins, I will use it if it exists otherwise I will manually perform the pulls.

danieljanes (approved):

danieljanes:
Looks great, thanks @charlesbvll!

PR #1510, opened 2022-12-15
Fix scripts

charlesbvll:

After thorough testing the install script now works completely, before the environment variables were not set which caused errors when trying to run the `venv-create.sh` and `bootstrap.sh` scripts.
The user now has to run the `bootstrap.sh` script themselves after `setup-defaults.sh`, this is because the shell needs to be changed between the execution of the two scripts. This was not taken into account before.  

#### Any other comments?

I will be adding info in the docs after this branch is merged into main.


danieljanes (approved):
Thanks for the update, looks great!

PR #1513, opened 2022-12-16
Flower Baseline: FedProx MNIST

charlesbvll:

Adds a new baseline for FedProx+MNIST, it is a Flower implementation of this [paper](https://arxiv.org/abs/1812.06127).


doc/source/changelog.md
```diff
@@ -26,6 +26,8 @@ We would like to give our special thanks to all the contributors who made the ne
 
   A new code example (`quickstart_fastai`) demonstrates federated learning with [fastai](https://www.fast.ai/) and Flower. You can find it here: [quickstart_fastai](https://github.com/adap/flower/tree/main/examples/quickstart_fastai).
 
+- **Introduce new Flower Baseline: FedProx MNIST** ([#1513](https://github.com/adap/flower/pull/1513))
```
danieljanes, line 29: Move this up to a new "Unreleased" section

baselines/flwr_baselines/publications/fedprox_mnist/tests.py
```diff
@@ -0,0 +1,35 @@
+"""Used to test the model and the data partitionning."""
```
danieljanes, line 1: The entire module `tests.py` should be renamed to belong to one of the other modules, for example, `model_test.py`. If functions from multiple modules are being tested (e.g., `model` and `dataset`), then we should split this up into multiple test modules, each belonging to one of the existing non-test modules

baselines/flwr_baselines/publications/fedprox_mnist/model.py
```diff
@@ -0,0 +1,166 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.nn.parameter import Parameter
+from torch.utils.data import DataLoader
+
+
+class Net(nn.Module):
+    """Convolutional Neural Network architecture as described in McMahan 2017
+    paper :
```
danieljanes, line 15: Split this onto two separate lines (with an empty line in between - the first sentence in a docstring should not overflow to the next line).

danieljanes (changes requested):
Looks good overall, I added some small comments

danieljanes (approved):

jafermarq:
This baseline needs to be revisited:
* It should use 1000 clients (instead of 100)
* Different hyperparameters compared to paper? (see Appendix C2: https://arxiv.org/pdf/1812.06127.pdf) . It should use 10 clients per round (instead of full participation. currently is `fraction_fit`=1.0); Default learning rate should be changed to 0.3

PR #1514, opened 2022-12-16
Align scripts from baselines to scripts on root

charlesbvll:

Before the `bootstrap.sh` in `baselines/` was resetting the virtual environment, which was not the case in the root `dev/bootstrap.sh` script. Resetting the virtual environment is now left to `venv-reset.sh` and the 2 bootstrapping scripts have the same behavior.

### DIFF
--- a/baselines/dev/bootstrap.sh
+++ b/baselines/dev/bootstrap.sh
@@ -11,5 +11,4 @@ python -m pip install -U setuptools==63.2.0
 python -m pip install -U poetry==1.1.14
 
 # Use `poetry` to install project dependencies
-python -m poetry install \
-  --extras "simulation"
+python -m poetry install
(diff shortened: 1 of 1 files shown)


baselines/dev/bootstrap.sh
```diff
@@ -2,22 +2,14 @@
...
-
 # Upgrade/install spcific versions of `pip`, `setuptools`, and `poetry`
 python -m pip install -U pip==22.2
 python -m pip install -U setuptools==63.2.0
 python -m pip install -U poetry==1.1.14
 
 # Use `poetry` to install project dependencies
-python -m poetry install
+python -m poetry install \
+  --extras "simulation"
```
danieljanes, line 15:
```suggestion
python -m poetry install
```

danieljanes: The `simulation` extra is only available in the `flwr` package, which is a direct dependency of `flwr_baselines` (`flwr_baselines` doesn't have extra dependencies).

danieljanes (changes requested):

baselines/dev/bootstrap.sh
charlesbvll: Oops, fixed it

danieljanes (approved):

PR #1522, opened 2022-12-29
Upgrade pip, setuptools and poetry

charlesbvll:

Currently the different tests from the workflows fail because of an install error related to `typing_extensions` not being installed, this PR aims at fixing this by upgrading pip, setuptools and poetry.


danieljanes (approved):
Lgtm!

PR #1525, opened 2022-12-29
Create DriverClientManager

danieljanes:


src/py/flwr/server/driver/driver_client_manager.py
```diff
@@ -0,0 +1,126 @@
...
+from ..criterion import Criterion
+
+
+class DriverClientManager(ClientManager):
+    """Provides a pool of available clients."""
+
+    def __init__(self) -> None:
+        self._cv = threading.Condition()
+        self.cids_to_ids: Dict[str, int] = {}
+        self.ids_to_cps: Dict[int, ClientProxy] = {}
```
tanertopal, line 33: How about using a single Dict here of type `Dict[str, Tuple[int, ClientProxy]]` to avoid anything getting out of sync?

src/py/flwr/server/driver/driver_client_manager.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def all_ids(self) -> Set[int]:
+        """Return all available client ids.
+
+        Returns
+        -------
+        ids : Set[int]
+            The IDs of all currently available clients.
+        """
+        return set(self.ids_to_cps)
```
tanertopal, line 107: Given the adjustment suggested above, this would have to be something along the lines of:

```suggestion
        return set([t[1] for t in self. cids_to_ids])
```

src/py/flwr/server/driver/driver_client_manager.py
danieljanes: Sure, we can do that

src/py/flwr/server/driver/driver_client_manager.py
danieljanes: Discarding this suggestion because the suggestion would not work (iterating over a dict returns the keys of that dict)

tanertopal (approved):

PR #1549, opened 2023-01-09
Upgrade to Poetry 1.3.2

charlesbvll:
#### Reference Issues/PRs

#1547 

Upgrading to `Poetry` 1.3.1, which also allows us to upgrade `Ray` to 2.2.0.

#### Any other comments?

None


charlesbvll:
@danieljanes I will probably open another PR afterwards for the `Ray` issue, it might be solved by updating `Mypy`, because when I played with it on my machine it worked, it's only the `Mypy` tests that failed.

PR #1559, opened 2023-01-11
Add fed-0003 for blinded token authentication

LorenzoMinto:


dalmouiee:
Hello @danieljanes, @tanertopal, just wondering where this sits in the priority list of things to do. Would be great to have some level of authentication between the server and clients, as it's absence prevents flower's usage in highly critical networks. Thanks

PR #1573, opened 2023-01-13
Fix Jax quickstart example

charlesbvll:

Rename function which name was changed after a Jax update.


danieljanes:
@charlesbvll should we also update the JAX version in `pyproject.toml`?

charlesbvll:
> @charlesbvll should we also update the JAX version in `pyproject.toml`?

Done !

danieljanes (approved):

PR #1579, opened 2023-01-13
Fix dp-sgd-mnist example

charlesbvll:

Replace obsolete nomenclature in dp-sdg-mnist example.


danieljanes (changes requested):
Looks good, but we have to update the `flwr` version as well

danieljanes (approved):

PR #1590, opened 2023-01-17
Change `setup-envs.sh` script to work with non-Bash shells

charlesbvll:

The `setup-envs.sh` script doesn't work with non-Bash shells by default.


tanertopal (approved):
lgtm!

PR #1594, opened 2023-01-18
Create new REST API for client nodes

danieljanes:


src/py/flwr/client/rest_client/connection.py
```diff
@@ -0,0 +1,202 @@
...
+from flwr.proto.node_pb2 import Node
+from flwr.proto.task_pb2 import Task, TaskIns, TaskRes
+from flwr.proto.transport_pb2 import ClientMessage, ServerMessage
+
+PATH_PULL_TASK_INS: str = "api/v0/fleet/pull-task-ins"
+PATH_PUSH_TASK_RES: str = "api/v0/fleet/push-task-res"
+
+
+@contextmanager
+def rest_not_a_connection(
```
tanertopal, line 42:
```suggestion
def http_request_response(
```

src/py/flwr/client/app.py
```diff
@@ -130,11 +135,15 @@ class `flwr.client.Client`.
     >>>     root_certificates=Path("/crts/root.pem").read_bytes(),
     >>> )
     """
+
     event(EventType.START_CLIENT_ENTER)
 
+    # Use either gRPC bidirectional streaming or REST request/response
+    connection = rest_not_a_connection if use_rest else grpc_connection
```
tanertopal, line 142:
```suggestion
    connection = http_request_response if use_rest else grpc_connection
```

src/py/flwr/client/app.py
```diff
@@ -49,6 +49,7 @@
 from .numpy_client import has_fit as numpyclient_has_fit
 from .numpy_client import has_get_parameters as numpyclient_has_get_parameters
 from .numpy_client import has_get_properties as numpyclient_has_get_properties
+from .rest_client.connection import rest_not_a_connection
```
tanertopal, line 52:
```suggestion
from .rest_client.connection import http_request_response
```

src/py/flwr/client/app.py
```diff
@@ -107,6 +109,9 @@ class `flwr.client.Client`.
             The PEM-encoded root certificates as a byte string. If provided, a secure
             connection using the certificates will be established to a
             SSL-enabled Flower server.
+        use_rest: bool (default: False)
+            Defines whether or not client is interacting with the server using the
+            REST API.
```
tanertopal, line 114:
```suggestion
        use_rest: bool (default: False)
            Defines whether or not the client is interacting with the server using the
            experimental REST API. This feature is experimental and might be renamed
            or changed in future versions.
```

src/py/flwr/client/app.py
```diff
@@ -191,6 +204,9 @@ def start_numpy_client(
         The PEM-encoded root certificates a byte string. If provided, a secure
         connection using the certificates will be established to a
         SSL-enabled Flower server.
+    use_rest: bool (default: False)
+        Defines whether or not client is interacting with the server using the
+        REST API.
```
tanertopal, line 209:
```suggestion
    use_rest: bool (default: False)
        Defines whether or not the client is interacting with the server using the
        experimental REST API. This feature is experimental and might be renamed
        or changed in future versions.
```

src/py/flwr/server/app.py
```diff
@@ -438,17 +504,45 @@ def _add_args_common(parser: argparse.ArgumentParser) -> None:
...
+    ex_group.add_argument(
+        "--grpc",
+        action="store_const",
+        dest="fleet_api_type",
+        const="grpc",
+        default="grpc",
+        help="Start a gRPC-based Fleet API server",
+    )
+    ex_group.add_argument(
+        "--rest",
```
tanertopal, line 527: Should we name this also `--use-rest`?

src/py/flwr/server/app.py
```diff
@@ -280,25 +304,40 @@ def run_server() -> None:
...
     )
 
     # Block
     driver_server.wait_for_termination()
-    fleet_server.wait_for_termination()
 
 
 def _register_exit_handlers(
     grpc_servers: List[grpc.Server],
+    bckg_threads: List[threading.Thread],
```
tanertopal, line 340: What does `bckg_threads` mean? Can we rename this?

examples/mt-pytorch/client.py
```diff
@@ -31,6 +31,7 @@ def evaluate(self, parameters, config):
 
 # Start Flower client
 fl.client.start_numpy_client(
-    server_address="[::]:9092",
+    server_address="0.0.0.0:9093",
```
tanertopal, line 34: Should we mention that when using "use_rest", one needs to use port 9093 by default, or should we always use the same port on the server side so users on the local host only need to change the boolean flag? Wdyt?

src/py/flwr/server/app.py
```diff
@@ -280,25 +304,40 @@ def run_server() -> None:
...
+            args=(args.rest_fleet_api_address, state_factory),
+        )
+        fleet_thread.start()
+        bckg_threads.append(fleet_thread)
+    elif args.fleet_api_type == "grpc":
+        fleet_server = _run_fleet_api_grpc_bidi(
+            address=args.grpc_fleet_api_address,
+            state_factory=state_factory,
+        )
+        grpc_servers.append(fleet_server)
```
tanertopal, line 323: Rest seems to have preceded before grpc. Should we fail if somewhere if both options are passed?

src/py/flwr/server/app.py
danieljanes: It's a list of running background threads. We can rename it in an update, the solution isn't ideal and might change anyways.

tanertopal: ✅

examples/mt-pytorch/client.py
danieljanes: Both approaches are valid. In the current version, different communication stacks run on different ports (by default) because we might have multiple stacks running together at some point. Happy to change it though.

src/py/flwr/server/app.py
danieljanes: It's not possible to pass both options because they are defined as mutually exclusive.

src/py/flwr/server/app.py
danieljanes: We could, but then we'd have to also rename `--grpc` to `--use-grpc`

danieljanes: Thinking about it, I'd prefer not to. We'd then also have to rename `--rest-fleet-api-address` to `--use-rest-fleet-api-address`

danieljanes: We could rename the `start_client` argument `use_rest` to `rest` though

examples/mt-pytorch/client.py
tanertopal: Fine for me either way. Let's keep it and see what comes back from users.

src/py/flwr/server/app.py
tanertopal: Either way is fine for me. I am going to approve you can change it if you want to. I can reapprove afterwards if you do.

src/py/flwr/server/app.py
tanertopal: 👍🏼

tanertopal (approved):

PR #1609, opened 2023-01-24
Fix `setup-envs.sh` script to remove crashes

charlesbvll:

It adds a default case to the case statement and only compare the basename of `$SHELL`.


tanertopal (approved):
Lgtm! Thank you!

PR #1619, opened 2023-01-31
Add FedProx strategy

charlesbvll:

Adds a strategy for `FedProx`, it is very similar to `FedAvg`, there is just one more class attribute, `proximal_mu` and an explanation in the docstring about how to implement the client for `FedProx`.


src/py/flwr/server/strategy/fedprox.py
```diff
@@ -0,0 +1,315 @@
...
+
+# flake8: noqa: E501
+class FedProx(Strategy):
+    """Configurable FedProx strategy implementation."""
+
+    # pylint: disable=too-many-arguments,too-many-instance-attributes,line-too-long
+    def __init__(
+        self,
+        *,
+        proximal_mu: float,
```
danieljanes, line 58: Let's make this the last parameter, so that the initial parameters are the same as in `FedAvg`

src/py/flwr/server/strategy/fedprox.py
```diff
@@ -0,0 +1,315 @@
...
+
+WARNING_MIN_AVAILABLE_CLIENTS_TOO_LOW = """
+Setting `min_available_clients` lower than `min_fit_clients` or
+`min_evaluate_clients` can cause the server to fail when there are too few clients
+connected to the server. `min_available_clients` must be set to a value larger
+than or equal to the values of `min_fit_clients` and `min_evaluate_clients`.
+"""
+
+# flake8: noqa: E501
+class FedProx(Strategy):
```
danieljanes, line 51: If we just extend `FedAvg`, we can remove most method implementations because they are inherited from `FedAvg` (review some of the other strategies to see how they still manage to customize, for example, `configure_fit`):
```suggestion
class FedProx(FedAvg):
```

doc/source/changelog.md
```diff
@@ -6,9 +6,15 @@
...
-- **Add support for ** `workload_id` **and** `group_id` **in Driver API** ([#1595](https://github.com/adap/flower/pull/1595))
+- **Add support for** `workload_id` **and** `group_id` **in Driver API** ([#1595](https://github.com/adap/flower/pull/1595))
 
 - **Make Android example compatible with** `flwr >= 1.0.0` **and the latest versions of Android** ([#1603](https://github.com/adap/flower/pull/1603))
+
+- **Add new `FedProx` strategy** ([#1619](https://github.com/adap/flower/pull/1619))
+  
+  This [strategy](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/fedprox.py) is almost identical to [`FedAvg`](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/fedavg.py), 
+ but helps users replicate what is described in this [paper](https://arxiv.org/abs/1812.06127). It essentially adds a parameter called `proximal_mu` to 
+ regularize the local models with respect to the global models. You might want to checkout this [blog post](https://flower.dev/blog/2023-02-02-fl-starter-pack-fedprox-mnist-cnn/) for more information.
```
danieljanes, line 17: This needs to be updated once the blog post is available

danieljanes (changes requested):

doc/source/changelog.md
charlesbvll: But the blog post will only be published once this is merge and can be used in the FedProx baseline, so I will create a snap issue on notion and resolve the issue here.

src/py/flwr/server/strategy/fedprox.py
```diff
@@ -0,0 +1,188 @@
...
+        # Sample clients
+        sample_size, min_num_clients = self.num_fit_clients(
+            client_manager.num_available()
+        )
+        clients = client_manager.sample(
+            num_clients=sample_size, min_num_clients=min_num_clients
+        )
+
+        # Return client/config pairs
+        return [(client, fit_ins) for client in clients]
```
danieljanes, line 188: The implementation of `configure_fit` should call `super().configure_fit`, and then just customize the return values

danieljanes (changes requested):

danieljanes (approved):
Lgtm!

PR #1621, opened 2023-02-01
Swift SDK & Sample

NumericalMax:
**Flower Swift SDK**

_About_
The branch implements the Flower SDK for Swift based on the initial draft `examples/ios`.

_Additional_
It includes an iOS sample application called FLiOS with benchmarking functionality for testing the SDK.

_Major Changes_
1. Moving initial implementation of the Swift SDK from `examples/ios` to `src/swift`
2. Extension of the existing SDK so that it fits to the proto of Flower 1.0
3. Fixing race conditions and threading bugs
4. Provide a standard implementation of the client protocol within the SDK such that loading data or wrapping layers does not need to be implemented in the client application - unless there is specific need for customization
5. Unit tests of the Swift SDK. It may require extension due to medium test coverage
6. Basic documentation of the Swift SDK code
8. Providing FLiOS - a sample app with benchmark functionalities for testing and benchmarking purposes
9. The file `examples/ios/server.py` has been streamlined for the sole purpose of running the FL process
10. Included Github Actions to unit test the Swift SDK

_How to test_
Move to the folder `examples/ios` and open the file `FLiOS.xcodeproj` with XCode. Run the application FLiOS on a simulator or a physical device with iOS version 14 or higher. Execute the server/python script `examples/ios/server.py`. Go through the sequential steps of FLiOS and adjust the _server hostname_ and _port_ accordingly before starting the federated learning process. For test purposes, the standard MNIST sample is used.

_Remark_
Github Actions for automatically running lint operations to look for possible code flaws, are not included in this PR.


.gitignore
```diff
@@ -160,3 +160,6 @@ dmypy.json
 # Poetry
 poetry.lock
 *.pkl
+
+# iOS
+xcuserdata/
```
charlesbvll, line 165: This could be moved to the example folder right?

examples/ios/FLiOS/Benchmark/BenchmarkSuite.swift
```diff
@@ -0,0 +1,98 @@
...
+        let url = self.getBenchmarkFileUrl()
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .iso8601
+        
+        do {
+            let completeBenchmark = try encoder.encode(BenchmarkSnaptshot(batteryHistory: self.batteryHistory, networkHistory: self.networkHistory, actionHistory: self.actionHistory, deviceID: self.deviceID))
+            let benchmarkString = String(data: completeBenchmark, encoding: .utf8)!
+            try benchmarkString.write(to: url, atomically: true, encoding: .utf8)
+            let test = try String(contentsOf: url)
+            print(test)
```
charlesbvll, line 91: These can be removed I think

examples/ios/FLiOS/ClientModel.swift
```diff
@@ -6,17 +6,22 @@
 //
```
charlesbvll: This file should be removed

examples/ios/FLiOS/CoreMLClient/MLModelInspect.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+import flwr
```
charlesbvll, line 9: I think the comment on line 14 can be removed  (I can't annotate it as from the previous version of the iOS example)

examples/ios/README.md
```diff
@@ -1,13 +1,16 @@
 # Flower iOS - A Flower SDK for iOS Devices with Example
-Flower iOS contains a Swift Package for Flower clients in iOS and an example that demonstrates a federated learning setup with iOS clients. The training is done on a MNIST dataset using CoreML with a updatable digit recognition model. 
+
+Flower iOS contains a Swift Package for Flower clients in iOS and an example that demonstrates a federated learning setup with iOS clients. The training is done on a MNIST dataset using CoreML with a updatable digit recognition model.
 
 ## Project Setup
 
 Requirements for setting up a server:
+
 - Conda for managing environment
```
charlesbvll, line 9: Is this a hard requirement? Because elsewhere we tend to use `pyenv` or `poetry`. `Poetry` might be the simplest one to use in this case. You might want to look at the setup we recommend in other examples, it's usually very basic. From the `quickstart_pytorch` example:
```bash
poetry install
poetry shell
```

src/swift/flwr/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+This package implements the Swift SDK for Flower - a friendly federated learning framework https://flower.dev/
+
+## Installation
+
+You can download the Flower project and integrate the package manually.
+
+## Usage
+
+A comprehensive example is available in: ```examples/iOS/```. To give information about the usage structurally:
```
charlesbvll, line 11:
```suggestion
A comprehensive example is available in: [```examples/ios/```](https://github.com/adap/flower/tree/main/examples/ios). To give information about the usage structurally: 
```

src/swift/flwr/Package.resolved
```diff
@@ -0,0 +1,122 @@
+{
```
charlesbvll, line 1: Is this file necessary?

charlesbvll (commented):
Just a few comments! It looks good otherwise

.gitignore
danielnugraha: I think structure wise it is better to have it included in global gitignore, but we can move it if you think it is more appropriate inside example folder.

examples/ios/FLiOS/Benchmark/BenchmarkSuite.swift
danielnugraha: removed.

examples/ios/FLiOS/CoreMLClient/MLModelInspect.swift
danielnugraha: removed.

examples/ios/README.md
danielnugraha: readme edited with poetry.

src/swift/flwr/README.md
danielnugraha: readme adapted.

src/swift/flwr/Package.resolved
danielnugraha: removed.

.gitignore
danieljanes: Both ok for me

.github/workflows/swift.yml
```diff
@@ -0,0 +1,28 @@
+name: Swift
+
+on: 
+    push:
+        branches:
+            - main
+    pull_request:
+        branches:
+            - main
```
tanertopal, line 9: Can we limit this to running only if we have changes to the Swift code? If I understand the macos runner pricing correctly, 13 minutes of CI runtime (this job) would cost us a dollar on each run of this job. We are running the CI a lot, and I want to avoid unnecessary costs.

Let me know if I am mistaken somewhere.

src/swift/flwr/README.md
```diff
@@ -0,0 +1,34 @@
+# flwr Swift SDK
+
+This package implements the Swift SDK for Flower - a friendly federated learning framework https://flower.dev/
```
tanertopal, line 3:
```suggestion
This package implements a Swift SDK for Flower.
```

src/swift/flwr/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+...
+let mlFlwrClient = MLFlwrClient(layerWrappers: layerWrappers, dataLoader: dataLoader, compiledModelUrl: compiledModelUrl)
+let flwrGRPC = FlwrGRPC(serverHost: hostname, serverPort: port)
+startFlwrGRPC(client: mlFlwrClient) {
+    // completion handler
+    print("Federated learning completed")
+}
+...
+```
```
tanertopal, line 24: I feel a bit lost here. I think it would be helpful to either just refer to the example or provide a bit more overview of how the Swift Client SDK is structured.

src/swift/flwr/Sources/Flower/Common/Parameter.swift
```diff
@@ -0,0 +1,214 @@
...
+    
+    public static let shared = ParameterConverter()
+    
+    private init() {
+        initGroup()
+    }
+    
+    private func initGroup() {
+        if group == nil {
+            print("Opening Python event loop group")
```
tanertopal, line 34: We generally avoid using print statements in Python, and I think this would be desirable in the Swift SDK too. Do you think it would be much effort to use, e.g. https://developer.apple.com/documentation/os/logger, with proper log levels so, e.g. a caught error is logged with `logger.error`?

src/swift/flwr/Sources/Flower/Common/Parameter.swift
```diff
@@ -0,0 +1,214 @@
...
+            group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
tanertopal, line 35: What is this group used for? Could you elaborate in a comment? Why do we use `numberOfThreads: 1` and not more? What would be the impact?

src/swift/flwr/Sources/Flower/Common/Parameter.swift
```diff
@@ -0,0 +1,214 @@
...
+            do {
+                try future?.wait()
+            } catch let error {
+                print(error)
+            }
+        }
+    }
+    
+    private func multiArrayToNumpy(multiArray: MLMultiArray) -> PythonObject? {
+        let pointer = try! UnsafeBufferPointer<Float>(multiArray)
```
tanertopal, line 52: Won't this crash the app if it does not work out? Would it be reasonable to rewrite it, so the user gets a proper error log?

src/swift/flwr/Sources/Flower/CoreML/MLFlwrClient.swift
```diff
@@ -0,0 +1,160 @@
+//
+//  File.swift
```
tanertopal, line 2:
```suggestion
//  MLFlwrClient.swift
```

src/swift/flwr/Sources/Flower/CoreML/MLFlwrClient.swift
```diff
@@ -0,0 +1,160 @@
...
+@available(iOS 14.0, *)
+public class MLFlwrClient: Client {
+    private var eventLoopGroup: EventLoopGroup?
+    
+    private var parameters: MLParameter
+    private var dataLoader: MLDataLoader
+    
+    private var compiledModelUrl: URL
+    private var tempModelUrl: URL
+
```
tanertopal, line 27: Can you add some documentation regarding the purpose of this class?

src/swift/flwr/Sources/Flower/GRPC/FlwrGRPC.swift
```diff
@@ -0,0 +1,130 @@
+//
+//  File.swift
```
tanertopal, line 2: copy/paste?

src/swift/flwr/Sources/Flower/CoreML/MLFlwrClient.swift
tanertopal: I am no Swift expert and I am wondering if this should be rather user-level code but I am not sure. Could you explain a bit about the reasoning this should go into the flwr core and not be part of the example?

src/swift/flwr/Sources/Flower/GRPC/FlwrGRPC.swift
danielnugraha: When creating a new file, File.swift is automatically created. Will be changed accordingly.

src/swift/flwr/Sources/Flower/Common/Parameter.swift
danielnugraha: We need Python to convert numpy to CoreML MultiArray, and since under the hood Python is still single threaded, using one thread will ensure the swift application to not violate PyStateGILCheck.

src/swift/flwr/Sources/Flower/Common/Parameter.swift
danielnugraha: It will not crash the app, instead it sends a nil (null) value which will be checked by the caller. Will add a log message.

src/swift/flwr/Sources/Flower/CoreML/MLFlwrClient.swift
danielnugraha: The idea is to provide one Client as a default, just like numpy Client in Python. Do you think it is more appropriate to move it to example?

.github/workflows/swift.yml
danielnugraha: I think it is already set to only run when swift code changes (line 15)

tanertopal: @danielnugraha, the line with `working-directory` just defines the from where the commands are run from. 

You can use this to limit the directory in which the change occurred:
* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

src/swift/flwr/Sources/Flower/Common/Parameter.swift
tanertopal: Sounds reasonable. Can you add that as a comment?

src/swift/flwr/Sources/Flower/Common/Parameter.swift
tanertopal: Understood! Thank you.

src/swift/flwr/Sources/Flower/CoreML/MLFlwrClient.swift
tanertopal: @danielnugraha, that depends on how well it generalizes. NumpyClient mostly takes care of the conversion to ensure no one has to be an expert in protobuf. Most researchers are pretty familiar with Numpy. Do you think this reasoning applies here, too? If so, we can keep it and improve on it later. I am looking forward to hearing your thoughts.

src/swift/flwr/Sources/Flower/GRPC/FlwrGRPC.swift
tanertopal: Thank you.

tanertopal (approved):
Huge PR. Thank you for the great work!

PR #1627, opened 2023-02-03
Fix poetry install for native M1 Mac by constraining python version

adam-narozniak:
The solution from: https://github.com/python-poetry/poetry/issues/5628. Thanks to Charles for suggesting it.


danieljanes (approved):
Lgtm!

PR #1632, opened 2023-02-04
Enable TaskIns for anonymous nodes

danieljanes:


src/py/flwr/server/driver/state.py
```diff
@@ -57,8 +59,16 @@ def get_task_ins(self, node_id: int, limit: Optional[int]) -> List[TaskIns]:
...
+                node_id is not None  # Not anonymous
+                and task_ins.task.consumer.anonymous is False
+                and task_ins.task.consumer.node_id == node_id
+                and task_ins.task.delivered_at == ""
+            ) or (
+                node_id is None  # Anonymous
+                and task_ins.task.consumer.anonymous is True
+                and task_ins.task.consumer.node_id == 0
                 and task_ins.task.delivered_at == ""
             ):
```
tanertopal, line 73: Doesn't this silently ignore all those Tasks with a non-None 'node_id' and `task_ins.task.consumer.anonymous` = True? It might be better to throw some exceptions here or handle it somehow.

danieljanes: Not quite, it would deliver those tasks (`task_ins.task.consumer.anonymous==True`) to an anonymous client node (`node_id=None`). It's a restrictive way of delivering `TaskIns` that doesn't send `TaskIns` that are scheduled for anonymous client nodes to client nodes that have identity. We could, at a later point, open this up and deliver `TaskIns` that were scheduled for anonymous client nodes to client nodes that have identity. The reverse (making it more restrictive), however, would not be easily possible without breaking user code.

tanertopal (approved):

PR #1640, opened 2023-02-06
Add new metrics to event

tanertopal:

Add new events to understand better how users are using the virtual client engine.


src/py/flwr/common/telemetry.py
```diff
@@ -58,6 +59,42 @@ def log(msg: Union[str, Exception]) -> None:
...
+
+    # If no source_file exists create one and write it
+    if not source_file.exists():
+        try:
+            source_file.touch(exist_ok=True)
+            source_file.write_text(str(uuid.uuid4()), encoding="utf-8")
+        except PermissionError:
+            return source_id
+
+    source_id = source_file.read_text(encoding="utf-8").strip()
```
danieljanes, line 93: Should we verify if the string we read is a valid UUID?

src/py/flwr/common/telemetry_test.py
```diff
@@ -68,3 +69,42 @@ def test_telemetry_disabled(self) -> None:
...
+        source_id = _get_source_id()
+
+        # Assert
+        # source_id should be len 36 as it's a uuid4 in the current
+        # implementation
+        self.assertIsNotNone(source_id)
+        self.assertEqual(len(source_id), 36)
+
+    def test_get_source_id_no_home(self) -> None:
+        """Test if _get_source_id returns None without a home dir."""
```
danieljanes, line 92: This description is not accurate

danieljanes (changes requested):

src/py/flwr/common/telemetry.py
tanertopal: I think it's better to send something like an invalid UUID rather than "unavailable". I would not validate it, therefore.

src/py/flwr/common/telemetry_test.py
tanertopal: Good catch. Fixed it.

danieljanes (approved):
Lgtm!

PR #1653, opened 2023-02-08
Add formatting to automatically clean notebooks

charlesbvll:

These formatting scripts would avoid people from committing notebooks containing metadata that can potentially harm the version control process.


pyproject.toml
```diff
@@ -94,6 +94,7 @@ sphinxcontrib-mermaid = "==0.7.1"
 furo = "==2022.12.7"
 sphinx-reredirects = "==0.1.1"
 nbsphinx = "==0.8.12"
+nbstripout = "^0.6.1"
```
danieljanes, line 97:
```suggestion
nbstripout = "==0.6.1"
```

dev/format.sh
```diff
@@ -14,3 +14,9 @@ find src/proto/flwr/proto -name *.proto | grep "\.proto" | xargs clang-format -i
 # Examples
 python -m black -q examples
 python -m docformatter -i -r examples
+
+
+# Notebooks
+KEYS="metadata.celltoolbar metadata.language_info metadata.toc metadata.notify_time metadata.varInspector metadata.accelerator metadata.vscode cell.metadata.id cell.metadata.heading_collapsed cell.metadata.hidden cell.metadata.code_folding cell.metadata.tags cell.metadata.init_cell"
+python -m nbstripout doc/source/tutorial/*.ipynb --extra-keys "$KEYS"
+python -m nbstripout examples/*/*.ipynb --extra-keys "$KEYS"
```
danieljanes, line 22: Can we add these in "check-only mode" to `test.sh`?

danieljanes (changes requested):
Lgtm, we can almost merge it

dev/format.sh
charlesbvll: @danieljanes Well I looked into it, there is an option to just check if the files would need formatting but I don't think there is an easy way to halt the execution of the script if so.

danieljanes (approved):

Issue #1604, opened 2023-01-23
Incorrect calculation of running loss in PyTorch example code


kuihao:
### Describe the bug

Dear maintainers,

I am a loyal user of **Flwr + Tensorflow**, and recently started using **Flwr + PyTorch**. However, I have noticed that the [Flwr example: quickstart_pytorch](https://github.com/adap/flower/blob/main/examples/quickstart_pytorch/client.py#L54) does not calculate the loss correctly.

According to the official [PyTorch CrossEntropyLoss documentation](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#torch.nn.CrossEntropyLoss), `torch.nn.CrossEntropyLoss` is calculated by default as `reduction='mean'`, meaning the function outputs the "mean of the minibatch" obtained from each iteration of the dataloader. So to calculate `loss per epoch` directly, it should be `running_loss / len(dataloader)`, because `len(dataloader)` gives "how many steps of dataloader (number of minibatches)".

The current algorithm in the [example code: quickstart_pytorch](https://github.com/adap/flower/blob/main/examples/quickstart_pytorch/client.py#L54):
```python 
def test(net, testloader):
    """Validate the model on the test set."""
    criterion = torch.nn.CrossEntropyLoss()
    correct, total, loss = 0, 0, 0.0
    with torch.no_grad():
        for images, labels in tqdm(testloader):
            outputs = net(images.to(DEVICE))
            labels = labels.to(DEVICE)
            loss += criterion(outputs, labels).item()
            total += labels.size(0)
            correct += (torch.max(outputs.data, 1)[1] == labels).sum().item()
    return loss / len(testloader.dataset), correct / total
```
To correct this, `running_loss / len(dataloader.dataset)` should be applied to `torch.nn.CrossEntropyLoss(reduction='sum')` to calculate the average of the entire dataset losses.

Addendum: 
- The term "**running_loss**" is a reference from the [PyTorch official tutorial](https://pytorch.org/tutorials/beginner/introyt/tensorboardyt_tutorial.html#graphing-scalars-to-visualize-training), which refers to the **accumulation of loss values during training**.
[...]

PR #1672, opened 2023-02-17
Correct testing loss in the quickstart_pytorch example

charlesbvll:
#### Reference Issues/PRs

#1604 

Makes the total testing loss in the `quickstart_pytorch` example more consistent by changing the reduction method of the criterion from `mean` to `sum`.


danieljanes (approved):
Lgtm!

PR #1676, opened 2023-02-20
Add SQLite state

tanertopal:

Add new SQLite state.


src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+    producer_node_id        INTEGER,
+    consumer_anonymous      BOOLEAN,
+    consumer_node_id        INTEGER,
+    created_at              TEXT,
+    delivered_at            TEXT,
+    ttl                     TEXT,
+    ancestry                TEXT,
+    legacy_server_message   BLOB,
+    legacy_client_message   BLOB
+    -- FOREIGN KEY(task_id) REFERENCES node(id)
```
danieljanes, line 52: This doesn't seem right, `task_id` is not a foreign key, but `producer_node_id` and `consumer_node_id` are (in cases where the node is not anonymous).

However, given that we can register and unregister nodes, should we perhaps just not add any foreign key constraint here?

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
```
danieljanes, line 72: Same as above

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        self,
+        database_path: str,
+    ) -> None:
+        """Initialize an SqliteState.
+
+        Parameters
+        ----------
+        database : (path-like object)
+            The path to the database file to be opened. Pass ":memory:" to open
+            a connectionto a database that is in RAM instead of on disk.
```
danieljanes, line 92:
```suggestion
            a connection to a database that is in RAM, instead of on disk.
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        """
+        self.database_path = database_path
+        self.conn: Optional[sqlite3.Connection] = None
+
+    def initialize(self) -> List[Tuple[str]]:
+        """Create tables if they don't exist yet."""
+        self.conn = sqlite3.connect(self.database_path)
+        self.conn.row_factory = dict_factory
+        self.conn.set_trace_callback(lambda query: log(1, query))
```
danieljanes, line 101: Do we need this in the production version, or is it only for debugging?

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        res = cur.execute("SELECT name FROM sqlite_schema;")
+
+        return res.fetchall()
+
+    def query(
+        self,
+        query: str,
+        data: Optional[Union[List[DictOrTuple], DictOrTuple]] = None,
+    ) -> List[Dict[str, Any]]:
+        """Run SQLQuery."""
```
danieljanes, line 118:
```suggestion
        """Execute a SQL query."""
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        if self.conn is None:
+            raise Exception("State is not initialized.")
+
+        if data is None:
+            data = []
+
+        # Cleanup whitespace to make the logs nicer
```
danieljanes, line 125:
```suggestion
        # Clean up whitespace to make the logs nicer
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+                    # pylint: disable-next=C0123
+                    and (type(data) == tuple or type(data) == list)
+                    # pylint: disable-next=C0123
+                    and (type(data[0]) == tuple or type(data[0]) == dict)
+                ):
+                    rows = self.conn.executemany(query, data)
+                else:
+                    rows = self.conn.execute(query, data)
+
+                # Extract results before commiting to support
```
danieljanes, line 141:
```suggestion
                # Extract results before committing to support
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+                #   INSERT/UPDATE ... RETURNING
+                # style queries
+                result = rows.fetchall()
```
danieljanes, line 144: The comment explains why this happens before the commit, but where does the actual commit happen?

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+        Stores the value of the task_ins in the state and, if successful, returns the
+        task_id (UUID) of the task_ins. If, for any reason, storing the task_ins fails,
+        `None` is returned.
+
+        Constraints
+        -----------
+        If `task_ins.task.consumer.anonymous` is `True`, then
+        `task_ins.task.consumer.node_id` MUST NOT be set (equal 0). Any implemenation
+        may just override it with zero instead of validating.
```
danieljanes, line 163: Are we overriding this anywhere? If not, we should probably not mention it in the docs here.

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        errors = validate_task_ins_or_res(task_ins)
+        if any(errors):
+            log(ERROR, errors)
+            return None
+
+        # Create and set task_id
+        task_id = uuid4()
+        task_ins.task_id = str(task_id)
+
+        # Set created_at
```
danieljanes, line 178:
```suggestion
        # Create and set created_at
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        created_at: datetime = _now()
+        task_ins.task.created_at = created_at.isoformat()
+
+        # Set ttl
```
danieljanes, line 182:
```suggestion
        # Create and set ttl
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        query = f"INSERT INTO task_ins VALUES({columns});"
+
+        self.query(query, data)
+
+        return task_id
+
+    def get_task_ins(
+        self, node_id: Optional[int], limit: Optional[int]
+    ) -> List[TaskIns]:
+        """Get TaskIns optionally filtered by node_id.
```
danieljanes, line 198:
```suggestion
        """Get undelivered TaskIns for one node (either anonymous or with ID).
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+            1. the `task_ins.task.consumer.node_id` equals `node_id` AND
+            2. the `task_ins.task.consumer.anonymous` equals `False` AND
+            3. the `task_ins.task.delivered_at` equals `""`.
+
+        If `node_id` is `None`, retrieve all TaskIns where the
+        `task_ins.task.consumer.node_id` equals `0` and
+        `task_ins.task.consumer.anonymous` is set to `True`.
+
+        If `delivered_at` MUST BE set (not `""`) otherwise the TaskIns MUST not be in
```
danieljanes, line 215:
```suggestion
        `delivered_at` MUST BE set (i.e., not `""`) otherwise the TaskIns MUST not be in
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+        If `limit` is not `None`, return, at most, `limit` number of `task_ins`. If
+        `limit` is set, it has to be greater zero.
+        """
+        if limit is not None and limit < 1:
+            raise AssertionError("`limit` must be >= 1")
+
+        # Retrieve all anonymous Tasks
+        if node_id == 0:
+            return []
```
danieljanes, line 226: Why do we immediately return here?

`node_id == 0` should never happen, right? It can only be `0` for anonymous nodes, in which case we'd expect the `node_id` to be `None`. I'd suggest raising an `Exception` here, `node_id==0` seems like an error. An empty list suggests that everything went well.

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+        data: Dict[str, Union[str, int]] = {}
+
+        # Retrieve all anonymous Tasks
+        if node_id is None:
+            query = """
```
danieljanes, line 232:
```suggestion
        if node_id is None:
            # Retrieve all anonymous Tasks
            query = """
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+                SELECT task_id
+                FROM task_ins
+                WHERE consumer_anonymous == 1
+                AND   consumer_node_id == 0
+                AND   delivered_at = ""
+            """
+
+        else:
+            query = """
```
danieljanes, line 241:
```suggestion
        else:
            # Retrieve all TaskIns for node_id
            query = """
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+            rows = self.query(query, data)
+
+        result = [dict_to_task_ins(row) for row in rows]
+
+        return result
+
+    def store_task_res(self, task_res: TaskRes) -> Optional[UUID]:
+        """Store one TaskRes.
+
+        Usually, the Fleet API calls this for Nodes returning results.
```
danieljanes, line 285:
```suggestion
        Usually, the Fleet API calls this when Nodes return their results.
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+        Stores the TaskRes and, if successful, returns the `task_id` (UUID) of
+        the `task_res`. If storing the `task_res` fails, `None` is returned.
+
+        Constraints
+        -----------
+        If `task_res.task.consumer.anonymous` is `True`, then
+        `task_res.task.consumer.node_id` MUST NOT be set (equal 0). Any implemenation
+        may just override it with zero instead of validating.
```
danieljanes, line 294: Same as above

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        errors = validate_task_ins_or_res(task_res)
+        if any(errors):
+            log(ERROR, errors)
+            return None
+
+        # Create and set task_id
+        task_id = uuid4()
+        task_res.task_id = str(task_id)
+
+        # Set created_at
```
danieljanes, line 309:
```suggestion
        # Create and set  created_at
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+        created_at: datetime = _now()
+        task_res.task.created_at = created_at.isoformat()
+
+        # Set ttl
```
danieljanes, line 313:
```suggestion
        # Create and set ttl
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+        self.query(query, data)
+
+        return task_id
+
+    def get_task_res(self, task_ids: Set[UUID], limit: Optional[int]) -> List[TaskRes]:
+        """Get TaskRes for task_ids.
+
+        Usually, the Driver API calls this for Nodes planning to work on one or more
+        TaskIns.
```
danieljanes, line 330: Not accurate

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+        Retrieves all TaskRes for the given `task_ids` and returns and empty list of
+        none could be found.
```
danieljanes, line 333:
```suggestion
        Retrieves all TaskRes for the given `task_ids` and returns and empty list if
        none could be found.
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+        Constraints
+        -----------
+        If `limit` is not `None`, return, at most, `limit` number of TaskRes. The limit
+        will only take effect if enough task_ids are in the set AND are currently
+        available. If `limit` is set, it has to be greater zero.
```
danieljanes, line 339:
```suggestion
        available. If `limit` is set, it has to be greater than zero.
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+    def get_nodes(self) -> Set[int]:
+        """Retrieve all currently stored node IDs as a set."""
+        query = "SELECT * FROM node;"
+        rows = self.query(query)
+        result: Set[int] = {row["id"] for row in rows}
+        return result
+
+
+def _now() -> datetime:
+    return datetime.now(tz=timezone.utc)
```
danieljanes, line 469: I believe we have this somewhere else, too. Should we put it somewhere in `flwr.common`?

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+from flwr.proto.node_pb2 import Node
+from flwr.proto.task_pb2 import Task, TaskIns, TaskRes
+from flwr.proto.transport_pb2 import ClientMessage, ServerMessage
+from flwr.server.utils.validator import validate_task_ins_or_res
+
+from .state import State
+
+SQL_CREATE_TABLE_NODE = """
+CREATE TABLE IF NOT EXISTS node(
+    id INTEGER UNIQUE
```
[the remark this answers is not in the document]
tanertopal: My thought process was that

* `task_res.task_id`
* `task_ins.task_id`

is each only 1-to-1 mapping of the proto structure while the node table with `node.id` is not. I prefer having just `id` in there. Do you have a strong preference for `node_id`?

src/py/flwr/server/state/sqlite_state.py
tanertopal: The `--` basically means its a comment. I think I had it in there to just test it, but than also thought right now we can't have an FK as we simply remove the node id from the table. I will remove the SQL comments.

src/py/flwr/server/state/sqlite_state.py
tanertopal: doing it

src/py/flwr/server/state/sqlite_state.py
tanertopal: See previous comment.

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -0,0 +1,582 @@
...
+
+    def __init__(
+        self,
+        database_path: str,
+    ) -> None:
+        """Initialize an SqliteState.
+
+        Parameters
+        ----------
+        database : (path-like object)
```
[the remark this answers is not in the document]
tanertopal: `path-like object` is actually correct, but mypy can't handle that correctly in Python 3.7/3.8 iirc. The type is taken from the [SQLite docs](https://docs.python.org/3/library/sqlite3.html#sqlite3.connect). Even "database_path" is not really correct, as you can pass, e.g. ":memory:". We can apply this for now, and I'll fix it in the subsequent clean-up when I fix the threading issue. wdyt?

src/py/flwr/server/state/sqlite_state_test.py
```diff
@@ -0,0 +1,58 @@
...
+# ==============================================================================
+"""Test for utility functions."""
+# pylint: disable=no-self-use, invalid-name, disable=R0904
+
+import unittest
+
+from flwr.server.state.sqlite_state import task_ins_to_dict
+from flwr.server.state.state_test import create_task_ins
+
+INS_KEYS = [
```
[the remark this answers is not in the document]
tanertopal: The only way I see would be to regex them out of `state_sqlite.SQL_CREATE_TABLE_TASK_INS`, which I think is rather complex. Wouldn't invest time into that right now.

src/py/flwr/server/state/sqlite_state.py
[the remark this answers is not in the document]
danieljanes: No strong preference. However, I thought of the `node` table to be an almost-1-to-1-mapping of the `Node` proto structure (just omitting the `anonymous` field), that's where the comment came from ^^

src/py/flwr/server/state/sqlite_state_test.py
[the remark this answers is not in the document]
tanertopal: Moving them into the one test in which they are used.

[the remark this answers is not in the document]
danieljanes: Ok for now. In the long run, we might want to move the table names into constants (in `sqlite_state`), there is quite a bite of repetition.

src/py/flwr/server/state/sqlite_state_test.py
```diff
@@ -0,0 +1,58 @@
...
+    "legacy_server_message",
+    "legacy_client_message",
+]
+
+
+class SqliteStateTest(unittest.TestCase):
+    """Test utilitiy functions."""
+
+    def test_ins_res_to_dict(self) -> None:
+        """Validate that a new state has no TaskIns."""
```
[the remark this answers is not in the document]
tanertopal: fixed

src/py/flwr/server/state/state_test.py
```diff
@@ -401,5 +401,53 @@ def state_factory(self) -> State:
...
+        state = self.state_factory()
+
+        # Execute
+        result = state.query("SELECT name FROM sqlite_schema;")
+
+        # Assert
+        assert len(result) == 6
+
+
+class SqliteFileBaseTest(StateTest, unittest.TestCase):
```
danieljanes, line 427:
```suggestion
class SqliteFileBasedTest(StateTest, unittest.TestCase):
```

src/py/flwr/common/__init__.py
```diff
@@ -15,7 +15,9 @@
 """Flower utilities shared between server and client."""
 
 
+from .date import now
```
danieljanes, line 18:
```suggestion
from .date import now as now
```

src/py/flwr/common/__init__.py
```diff
@@ -15,7 +15,9 @@
...
 from .grpc import GRPC_MAX_MESSAGE_LENGTH
+from .logger import log
```
danieljanes, line 20:
```suggestion
from .logger import log as log
```

src/py/flwr/server/state/sqlite_state.py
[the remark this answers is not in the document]
danieljanes: Either way is ok for me - up to you!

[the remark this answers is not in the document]
tanertopal: Ok, keeping it for now.

src/py/flwr/server/state/sqlite_state.py
tanertopal: Good catch. That's an outdated comment. Removing it.

src/py/flwr/server/state/sqlite_state.py
tanertopal: Making it consistent with `in_memory_state`

src/py/flwr/server/state/sqlite_state.py
tanertopal: So this was before I realized the connection could be used as a context manager, although it still makes sense. Basically, what's happing when `with self.conn:` is used is that a "BEGIN" query is executed, and at the exit of the context manager a "COMMIT" is executed. SQLite requires the "COMMIT" to write any change to disk.

src/py/flwr/server/state/sqlite_state.py
tanertopal: True, adjusting it.

tanertopal: Let me know what you think about the comment.

src/py/flwr/server/state/sqlite_state.py
danieljanes: Got it, thanks!

danieljanes (approved):

PR #1693, opened 2023-02-24
Create StateFactory

danieljanes:

### DIFF
--- a/src/py/flwr/server/state/state_factory.py
+++ b/src/py/flwr/server/state/state_factory.py
@@ -1,0 +1,47 @@
+# Copyright 2022 Adap GmbH. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Factory class that creates State instances."""
+
+
+from logging import DEBUG
+from typing import Optional
+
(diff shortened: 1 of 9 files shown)


src/py/flwr/server/state/state_factory.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, database: str) -> None:
+        self.database = database
+        self.state_instance: Optional[State] = None
+
+    def state(self) -> State:
+        """Return a State instance and create it, if necessary."""
+
+        # InMemoryState
+        if self.database == ":flwr-in-memory-state:":
```
tanertopal, line 39:
```suggestion
        if self.database == ":flwr-in-memory-state:":
```
This format is valid for SqliteState, too. Is that intentional?

src/py/flwr/server/state/state_factory.py
```diff
@@ -0,0 +1,49 @@
...
+            if self.state_instance is None:
+                self.state_instance = InMemoryState()
+            log(DEBUG, "Using InMemoryState")
+            return self.state_instance
+
+        # SqliteState
+        state = SqliteState(self.database)
+        state.initialize()
```
tanertopal, line 47:
```suggestion
        state = SqliteState(self.database)
        state.initialize()
```
Just a remark: I already had a draft of a threading-compatible version of the SqliteState. Let's not build too many workarounds around it. I assume that's the intention here, as it seems easily adjustable with the `self.state_instance` property.

src/py/flwr/server/state/state_factory.py
danieljanes: It's inspired by it, but we could use any magic string here.

src/py/flwr/server/state/state_factory.py
danieljanes: Exactly, a different kind of `State` implementation would only require changes in the `StateFactory.state` method.

src/py/flwr/server/state/state_factory.py
danieljanes: (btw, easy to change it later)

tanertopal (approved):

PR #1694, opened 2023-02-24
Federated XGBoost horizontal quickstart

dannymcy:


src/py/flwr/common/fedxgb.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2020 Adap GmbH. All Rights Reserved.
```
charlesbvll, line 1: This file should sit in `src/py/flwr/server/strategy` rather than `common`. `common` is for functions used both by `server` and `client`.

charlesbvll (changes requested):
I have only added one small comment about the location of a file but other than that there still is the question of the `xgboost` dependency. There is no perfect way to do it, but you probably want to add it to the global `pyproject.toml` under the extra dependencies, in a similar fashion to `ray`.

danieljanes (approved):

charlesbvll (approved):

PR #1702, opened 2023-03-06
Extend Ruff by E F W B checks

tanertopal:

Extend Ruff linter checks by "E", "F", "W", "B"
* E = pycodestyle Error
* W = pycodestyle Warning
* F = Pyflakes
* B = flake8-bugbear

Fix:
* B011 Do not assert False (python -O removes these calls), raise AssertionError()
Fixed by changing assert False to raise AssertionError()
* B024 {name} is an abstract base class, but it has no abstract methods
It regards NumPyClient and Client
Fixed by ignoring this error.
* B904 Within an except clause, raise exceptions with raise ... from err or raise ... from None to distinguish them from errors in exception handling
Fixed by adding as err
and then from err
* B905 zip() without an explicit strict= parameter
Fixed by changing the Python target version to the currently lowest supported by Flower = 3.7. The zip's strict mode was introduced in Python 3.10
* E501 Line too long
Fixed by adding (similar solution exists for flake8 in .flake8)
```
[tool.ruff.per-file-ignores]
"src/py/flwr/server/strategy/*.py" = ["E501"]
```


pyproject.toml
```diff
@@ -149,3 +150,35 @@ disallow_untyped_calls = false
...
 follow_imports = "skip"
 follow_imports_for_stubs = true
+
+[tool.ruff]
+select = ["E", "F", "W", "B"]
+fixable = ["E", "F", "W", "B"]
+ignore = ["B027"]
+# ruff is less lenient than pylint and does not make any exceptions
+# (for docstrings, strings and comments in particular).
+line-length = 115
```
danieljanes, line 160: Shouldn't this match the black setting (`88`)?

tanertopal: That would be correct. I copy that config from the pylint config. That is how they configure `ruff` for the pylint repo. We could merge it with 88 and think of generally increasing it separately.

tanertopal: Changed it to 88

adam-narozniak:
* B011  Do not `assert False` (`python -O` removes these calls), raise `AssertionError()`
Fixed by changing `assert False` to `raise AssertionError()`
* B024 `{name}` is an abstract base class, but it has no abstract methods
It regards NumPyClient and Client 
Fixed by ignoring this error.
* B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
Fixed by adding `as err` 
and then `from err`
* B905 `zip()` without an explicit `strict=` parameter
It got fixed by changing the Python target version to the currently lowest supported by Flower = 3.7. The zip's strict mode was introduced in Python 3.10

danieljanes (approved):

PR #1705, opened 2023-03-08
Remove "optional" parameter attribute from documentation where needed

nepiskopos:


danieljanes (approved):
Good catch, thanks for the PR!

PR #1755, opened 2023-03-28
Validate server and client IP address

charlesbvll:

The server and client IPs are given as strings but are not validated. This PR adds a utility function to parse IPv4 and IPv6 addresses.


src/py/flwr/client/app.py
```diff
@@ -150,35 +152,41 @@ class `flwr.client.Client`.
...
+                break
+            # Sleep and reconnect afterwards
+            log(
+                INFO,
+                "Disconnect, then re-establish connection after %s second(s)",
+                sleep_duration,
+            )
+            time.sleep(sleep_duration)
+        else:
+            sys.exit(f"Client IP address ({server_address}) cannot be parsed.")
```
danieljanes, line 189: `server_address` is the address (host+IP) that the server is running on:

```suggestion
            sys.exit(f"Server address ({server_address}) cannot be parsed.")
```

src/py/flwr/client/app.py
```diff
@@ -150,35 +152,41 @@ class `flwr.client.Client`.
...
-            log(INFO, "Disconnect and shut down")
-            break
-        # Sleep and reconnect afterwards
-        log(
-            INFO,
-            "Disconnect, then re-establish connection after %s second(s)",
-            sleep_duration,
-        )
-        time.sleep(sleep_duration)
+        parsed_address = parse_address(server_address)
```
danieljanes, line 155: Is there a reason for the address validation to happen inside the loop? If not, let's do it right after `event(...)`.

src/py/flwr/client/app.py
```diff
@@ -80,7 +82,7 @@
 ClientLike = Union[Client, NumPyClient]
 
 
-# pylint: disable=import-outside-toplevel
+# pylint: disable=import-outside-toplevel, too-many-locals
```
danieljanes, line 85:
```suggestion
# pylint: disable=import-outside-toplevel,too-many-locals
```

src/py/flwr/common/address_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+from .address import parse_address
+
+
+def test_ipv4_correct() -> None:
+    """Test if a correct IPv4 address is correctly parsed."""
+
+    # Prepare
+    addresses = [
```
danieljanes, line 25: @tanertopal, are there any other cases like `localhost` that we should consider?

src/py/flwr/common/address_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # Assert
+        assert actual is None
+
+
+def test_ipv6_correct() -> None:
+    """Test if a correct IPv6 address is correctly parsed."""
+
+    # Prepare
+    addresses = [
```
danieljanes, line 62: @tanertopal, same question as above: are there any other cases we need to consider?

src/py/flwr/server/app.py
```diff
@@ -63,7 +64,7 @@ class ServerConfig:
     round_timeout: Optional[float] = None
 
 
-def start_server(  # pylint: disable=too-many-arguments
+def start_server(  # pylint: disable=too-many-arguments, too-many-locals
```
danieljanes, line 67:
```suggestion
def start_server(  # pylint: disable=too-many-arguments,too-many-locals
```

src/py/flwr/server/app.py
```diff
@@ -226,21 +234,27 @@ def run_driver_api() -> None:
     # Initialize StateFactory
     state_factory = StateFactory(args.database)
 
-    # Start server
-    grpc_server: grpc.Server = _run_driver_api_grpc(
-        address=args.driver_api_address,
-        state_factory=state_factory,
-    )
+    parsed_address = parse_address(args.driver_api_address)
+    if parsed_address:
```
danieljanes, line 238: Can we reduce the "blast radius" of this change by doing the check + `sys.exit` here (and then leave the remainder of this function unchanged / unindented)?

src/py/flwr/server/app.py
```diff
@@ -258,18 +272,33 @@ def run_fleet_api() -> None:
 
     # Start Fleet API
     if args.fleet_api_type == "rest":
-        fleet_thread = threading.Thread(
-            target=_run_fleet_api_rest,
-            args=(args.rest_fleet_api_address, state_factory),
-        )
-        fleet_thread.start()
-        bckg_threads.append(fleet_thread)
+        parsed_address = parse_address(args.rest_fleet_api_address)
```
danieljanes, line 275: Can we do the validation in between `event(...)` and state initialization?

src/py/flwr/server/app.py
```diff
@@ -258,18 +272,33 @@ def run_fleet_api() -> None:
...
+                f"Fleet IP address ({args.rest_fleet_api_address}) cannot be parsed."
+            )
     elif args.fleet_api_type == "grpc":
-        fleet_server = _run_fleet_api_grpc_bidi(
-            address=args.grpc_fleet_api_address,
-            state_factory=state_factory,
-        )
-        grpc_servers.append(fleet_server)
+        parsed_address = parse_address(args.grpc_fleet_api_address)
+        if parsed_address:
```
danieljanes, line 290: Same as above

danieljanes (changes requested):

src/py/flwr/server/app.py
charlesbvll: @danieljanes I changed it, but it means that we know check 2 times if we are using rest or grpc.

src/py/flwr/server/app.py
```diff
@@ -308,18 +354,33 @@ def run_server() -> None:
...
     # Start Fleet API
     if args.fleet_api_type == "rest":
-        fleet_thread = threading.Thread(
-            target=_run_fleet_api_rest,
-            args=(args.rest_fleet_api_address, state_factory),
-        )
-        fleet_thread.start()
-        bckg_threads.append(fleet_thread)
+        parsed_address = parse_address(args.rest_fleet_api_address)
+        if parsed_address:
```
danieljanes, line 358: We can reduce the size of the diff here quite a bit by using `if not parsed_address: sys.exit(...)`

src/py/flwr/common/address_test.py
tanertopal: Things we could test for:
```
("flower.dev:123", ("flower.dev", 123, False)),
("localhost:123", ("localhost", 123, False)),
("[::]:123", ("::", 123, True)),
("[0000:0000:0000:0000:0000:0000:0000:0001]:443", ("0000:0000:0000:0000:0000:0000:0000:0001", 443, True)),
("[0:0:0:0:0:0:0:1]:80", ("0:0:0:0:0:0:0:1", 80, True)),
("[::1]:80", ("::1", 80, True)),
("[0000:0000:0000:0000:0000:0000:0000:0001]:90", ("0000:0000:0000:0000:0000:0000:0000:0001", 90, True)),
```

The square brackets on the IPv6 are only needed when passing the port with the string. It's to make sure one can separate the port from the rest.

src/py/flwr/server/app.py
danieljanes: @charlesbvll good point, I agree that the previous version was nicer

src/py/flwr/common/address_test.py
tanertopal: Just saw the next comment. Probably some of these are for the other test.

danieljanes (changes requested):

src/py/flwr/server/app.py
danieljanes: ... and we should use the same pattern consistently in all places where we have the address check

src/py/flwr/common/address_test.py
tanertopal: Also @charlesbvll the first two are not IPs but rather domains. They are still valid addresses which might resolve to either Ipv4 or Ipv6.

charlesbvll: @tanertopal Should I use something like `socket.gethostbyname(<DOMAIN_NAME>)` to get the actual IP before validating it? Otherwise, it doesn't really make sense to validate a domain. So, should I change the tests so that instead of returning the domain and the port, it returns the IP and the port?

src/py/flwr/server/app.py
charlesbvll: It looks way better indeed! Thanks for the suggestion

src/py/flwr/server/app.py
```diff
@@ -308,15 +344,28 @@ def run_server() -> None:
...
+            args=(host, port, state_factory),
         )
         fleet_thread.start()
         bckg_threads.append(fleet_thread)
     elif args.fleet_api_type == "grpc":
+        parsed_address = parse_address(args.grpc_fleet_api_address)
+        if not parsed_address:
+            sys.exit(
+                f"Fleet IP address ({args.grpc_fleet_api_address}) cannot be parsed."
+            )
```
danieljanes, line 364:
```suggestion
        address_arg = args.grpc_fleet_api_address
        parsed_address = parse_address(address_arg)
        if not parsed_address:
            sys.exit(f"Fleet IP address ({address_arg}) cannot be parsed.")
```

danieljanes: @charlesbvll we can use the same approach wherever `sys.exit` get formatted to span three lines.

src/py/flwr/common/address.py
```diff
@@ -0,0 +1,54 @@
...
+        a bool that is True if the address is IPv6 and False otherwise.
+    """
+    try:
+        raw_host, _, raw_port = address.rpartition(":")
+        if raw_host.count(".") in [1, 2]:
+            raw_host = socket.gethostbyname(raw_host)
+        elif raw_host == "localhost":
+            raw_host = "127.0.0.1"
+        host, port = raw_host.translate({ord(i): None for i in "[]"}), int(raw_port)
+        if port > 65535:
```
danieljanes, line 50: Do we need to check for negative values here as well?

src/py/flwr/common/address.py
```diff
@@ -0,0 +1,54 @@
...
+    Union[Tuple[str, int, bool], None]
+        If the string provided is not a correct IPv6 or IPv4,
+        the function will return None, otherwise it will return the host,
+        as a string, the port number, as an int, and
+        a bool that is True if the address is IPv6 and False otherwise.
+    """
+    try:
+        raw_host, _, raw_port = address.rpartition(":")
+        if raw_host.count(".") in [1, 2]:
+            raw_host = socket.gethostbyname(raw_host)
```
danieljanes, line 46: @charlesbvll @tanertopal I'd say we should not do this ourselves but leave it up to the actual libraries (gRPC, uvicorn, requests) to do this (and rather just forward the original address as-is)

danieljanes: This would also mean removing the `localhost` conversion below.

danieljanes (changes requested):

src/py/flwr/client/app.py
```diff
@@ -137,6 +139,14 @@ class `flwr.client.Client`.
...
     event(EventType.START_CLIENT_ENTER)
 
+    # Parse IP address
+    parsed_address = parse_address(server_address)
+    if parsed_address:
+        host, port, is_v6 = parsed_address
+        address = f"[{host}]:{port}" if is_v6 else f"{host}:{port}"
+    else:
+        sys.exit(f"Server address ({server_address}) cannot be parsed.")
+
```
danieljanes, line 149:
```suggestion
    parsed_address = parse_address(server_address)
    if not parsed_address:
        sys.exit(f"Server address ({server_address}) cannot be parsed.")
    host, port, is_v6 = parsed_address
    address = f"[{host}]:{port}" if is_v6 else f"{host}:{port}"

```

src/py/flwr/common/address.py
```diff
@@ -0,0 +1,60 @@
...
+"""Flower IP address utils."""
+
+
+from ipaddress import ip_address
+from typing import Tuple, Union
+
+IPV6: int = 6
+
+
+def parse_address(address: str) -> Union[Tuple[str, int, bool], None]:
```
danieljanes, line 24:
```suggestion
def parse_address(address: str) -> Optional[Tuple[str, int, bool]:
```

src/py/flwr/common/address.py
```diff
@@ -0,0 +1,60 @@
...
+    """Parses a string representing an IP address into a host, port and
+    version.
```
danieljanes, line 26: Shorten to one line

src/py/flwr/common/address.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Parameters
+    ----------
+    address : str
+        The string representation of an IPv4 or IPV6 address with the port number.
+        For example, '127.0.0.1:8080', or [::1]:8080.
+
+    Returns
+    -------
+    Union[Tuple[str, int, bool], None]
```
danieljanes, line 36: Same as above

src/py/flwr/common/address_test.py
danieljanes: I'd propose that the parse function should return the host in it's original form (either and IP address or a host string like `flower.dev` or `localhost`). Wdyt @charlesbvll @tanertopal ?

danieljanes (changes requested):

src/py/flwr/common/address_test.py
charlesbvll: Yes, you're right, that makes more sense, I changed it already.

src/py/flwr/common/address.py
charlesbvll: @danieljanes This actually breaks the `docformatter` test, should I try to reduce the text so it fits in one line? Like `Parses an IP address into a host, port, and version.`?

src/py/flwr/common/address_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+def test_ipv4_correct() -> None:
+    """Test if a correct IPv4 address is correctly parsed."""
+
+    # Prepare
+    addresses = [
+        ("127.0.0.1:8080", ("127.0.0.1", 8080, False)),
+        ("0.0.0.0:12", ("0.0.0.0", 12, False)),
+        ("flower.dev:123", ("flower.dev", 123, False)),
+        ("localhost:123", ("localhost", 123, False)),
```
danieljanes, line 29:
```suggestion
        ("flower.dev:123", ("flower.dev", 123, False)),
        ("flower.dev:123", ("flower.dev", 123, False)),
        ("sub.flower.dev:123", ("sub.flower.dev", 123, False)),
        ("sub2.sub1.flower.dev:123", ("sub2.sub1.flower.dev", 123, False)),
        ("sub5.sub4.sub3.sub2.sub1.flower.dev:123", ("sub5.sub4.sub3.sub2.sub1.flower.dev", 123, False)),
        ("localhost:123", ("localhost", 123, False)),
```

src/py/flwr/common/address_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # Prepare
+    addresses = [
+        ("127.0.0.1:8080", ("127.0.0.1", 8080, False)),
+        ("0.0.0.0:12", ("0.0.0.0", 12, False)),
+        ("flower.dev:123", ("flower.dev", 123, False)),
+        ("flower.dev:123", ("flower.dev", 123, False)),
+        ("sub.flower.dev:123", ("sub.flower.dev", 123, False)),
+        ("sub2.sub1.flower.dev:123", ("sub2.sub1.flower.dev", 123, False)),
+        ("sub5.sub4.sub3.sub2.sub1.flower.dev:123", ("sub5.sub4.sub3.sub2.sub1.flower.dev", 123, False)),
```
danieljanes, line 32:
```suggestion
        ("s5.s4.s3.s2.s1.flower.dev:123", ("s5.s4.s3.s2.s1.flower.dev", 123, False)),
```

charlesbvll:
@danieljanes Should we use an external library like this [one](https://validators.readthedocs.io/en/latest/#) to check if a string is a valid domain?

tanertopal (commented):
@danieljanes as far as I can see this looks good. Let's merge it.

danieljanes (approved):

PR #1761, opened 2023-03-30
Advanced PyTorch example simplified

charlesbvll:

This advanced PyTorch example needed a high end machine to run in a reasonable time, now, by default, it will run with less data and less clients.


examples/advanced_pytorch/README.md
```diff
@@ -2,7 +2,6 @@
 
 This example demonstrates an advanced federated learning setup using Flower with PyTorch. It differs from the quickstart example in the following ways:
 
-- 10 clients (instead of just 2)
 - Each client holds a local dataset of 5000 training examples and 1000 test examples
```
tanertopal, line 5: @charlesbvll is this true when passing the `--toy` option? Can you make a remark about it?

charlesbvll: Technically it is still true that each client holds that much data, but indeed when running the script, it will by default use the `--toy` option which will only use a very small subset of this. I will change this accordingly.

examples/advanced_pytorch/server.py
```diff
@@ -99,7 +99,7 @@ def main():
         fraction_evaluate=0.2,
         min_fit_clients=2,
         min_evaluate_clients=2,
-        min_available_clients=10,
+        min_available_clients=2,
```
danieljanes, line 102: If we default to `--toy True`, can we continue to run it with 10 clients?

danieljanes (changes requested):

examples/advanced_pytorch/server.py
charlesbvll: Ok so it turns out this happens even with 2 clients, so there must be another issue at hand.

adam-narozniak (approved):
LGTM

examples/advanced_pytorch/server.py
danieljanes: @charlesbvll why are we not waiting until all 10 clients are available?

charlesbvll: Oh right, this is something I forgot to revert, good catch!

Issue #1610, opened 2023-01-24
Implement a new strategy: TrimmedMean


edogab33:
### Describe the type of feature and its functionality.

I propose to add TrimmedMean as a new strategy robust against poisoning attacks.

It works by removing the largest and smallest `beta` parameters from the local models received by the server and finally average their values to make the new global model.

Trimmed Mean: https://arxiv.org/pdf/1803.01498.pdf

### Describe step by step what files and adjustments are you planning to include.

I am planning to add a new strategy `trimmedmean.py` and its tests in `trimmedmean_test.py`.

danieljanes:
@edogab33 good idea

edogab33:
Can we close this issue?

@danieljanes @adam-narozniak

PR #1769, opened 2023-03-31
Add FedTrimmedMean strategy

m-danilow:
#### Reference Issues/PRs

#1610

A strategy that removes the indicated percentage of the largest and smallest weights from the local models received by the server and finally averages their values to create a new global model

#### Any other comments?

Please check TODO comments and I would also need some guidance in fixing CI ;)

### DIFF
--- a/src/py/flwr/server/strategy/aggregate.py
+++ b/src/py/flwr/server/strategy/aggregate.py
@@ -134,13 +134,17 @@ def _compute_distances(weights: List[NDArrays]) -> NDArray:
             distance_matrix[i, j] = norm**2
     return distance_matrix
 
-def aggregate_trimmed_avg(results: List[Tuple[NDArray, int]], proportiontocut: float) -> NDArray:
+
+def aggregate_trimmed_avg(
+    results: List[Tuple[NDArray, int]], proportiontocut: float
+) -> NDArray:
     """Compute trimmed average."""
     # Create a list of weights and ignore the number of examples
     weights = [weights for weights, _ in results]
 
     trimmed_w: NDArrays = [
-        stats.trim_mean(np.asarray(layer), axis=0, proportiontocut=proportiontocut) for layer in zip(*weights)
+        stats.trim_mean(np.asarray(layer), axis=0, proportiontocut=proportiontocut)
+        for layer in zip(*weights)
     ]
 
     return trimmed_w
--- a/src/py/flwr/server/strategy/aggregate.py
+++ b/src/py/flwr/server/strategy/aggregate.py
@@ -19,6 +19,7 @@ from functools import reduce
 from typing import List, Tuple
 
 import numpy as np
+from scipy import stats
 
 from flwr.common import NDArray, NDArrays
 
@@ -132,3 +133,14 @@ def _compute_distances(weights: List[NDArrays]) -> NDArray:
             norm = np.linalg.norm(delta)  # type: ignore
             distance_matrix[i, j] = norm**2
     return distance_matrix
+
+def aggregate_trimmed_avg(results: List[Tuple[NDArray, int]], proportiontocut: float) -> NDArray:
+    """Compute trimmed average."""
+    # Create a list of weights and ignore the number of examples
+    weights = [weights for weights, _ in results]
+
+    trimmed_w: NDArrays = [
+        stats.trim_mean(np.asarray(layer), axis=0, proportiontocut=proportiontocut) for layer in zip(*weights)
+    ]
+
+    return trimmed_w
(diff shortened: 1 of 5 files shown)


src/py/flwr/server/strategy/fedtrimmedavg.py
```diff
@@ -0,0 +1,136 @@
...
+                Optional[Tuple[float, Dict[str, Scalar]]],
+            ]
+        ] = None,
+        on_fit_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
+        on_evaluate_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
+        accept_failures: bool = True,
+        initial_parameters: Optional[Parameters] = None,
+        fit_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        evaluate_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        proportiontocut: float = 0.2,
```
edogab33, line 50: I propose to call it beta, so that we stick to the naming convention of the original paper

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -132,3 +133,18 @@ def _compute_distances(weights: List[NDArrays]) -> NDArray:
...
+
+def aggregate_trimmed_avg(
+    results: List[Tuple[NDArrays, int]], proportiontocut: float
+) -> NDArrays:
+    """Compute trimmed average."""
+    # Create a list of weights and ignore the number of examples
+    weights = [weights for weights, _ in results]
+
+    trimmed_w: NDArrays = [
+        stats.trim_mean(np.asarray(layer), axis=0, proportiontocut=proportiontocut)
```
edogab33, line 146: Maybe it would be better to not import an entire external library to do a simple operation like trimmed mean

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -19,6 +19,7 @@
 from typing import List, Tuple
 
 import numpy as np
+from scipy import stats
```
edogab33, line 22: see: https://github.com/adap/flower/pull/1769/files#r1161094950

src/py/flwr/server/strategy/fedmedian.py
```diff
@@ -118,6 +118,7 @@ def __init__(
             fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
             evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
         )
+        # TODO Aren't these two lines below redundant?
```
edogab33, line 121: One function is used to aggregate metrics about the fitting process, the other one is used to aggregate metrics about the evaluation. The metrics I am talking about are the ones returned by those methods [here](https://github.com/adap/flower/blob/main/src/py/flwr/client/numpy_client.py) (`fit()` and `evaluate()`) that are sent back to the server.

Also note that this file (`fedmedian.py`) is not part of this pull request. To avoid conflicts with other PRs I suggest to not modify files that are not meant to be touched by your PRs, otherwise we loose the meaning of opening them for each issue (that is, tracking every change to the main branch) :)
However, if you have any question, a comment in the code is not the right place to ask it. You can use the issue or discussion sections.

edogab33:
Hi @m-danilow, I'm not a team of Flower (just a simple contributor) but I've posted some comments to help you with the PR. Hope you find them useful.

src/py/flwr/server/strategy/fedmedian.py
edogab33: I see it now, yes it is redundant

src/py/flwr/server/strategy/fedtrimmedavg.py
edogab33: Definitely more descriptive, but it is common to use scientific notation in such python libraries, Flower included.

See for example: 
- https://github.com/adap/flower/blob/60a47015219913600880940d6a2402fc30e186f4/src/py/flwr/server/strategy/fedadagrad.py
- https://github.com/adap/flower/blob/60a47015219913600880940d6a2402fc30e186f4/src/py/flwr/server/strategy/fedadam.py

P.S. As for the other strategy implementations, the original paper should be cited and it is better to keep the original name or the common one used in literature. So, the class name becomes TrimmedMean along with the file name. 

However these are just details that improve the PR, you are doing a good job overall IMHO :)

danieljanes:
Thanks for the PR @m-danilow & thanks for the early review @edogab33 !

At first glance, the PR seems to be in good shape. Let me know if you need help with getting the CI to pass.

adam-narozniak:
Hi @m-danilow! I will merge your PR into a new branch, so I can make modifications to fix the tests before merging it into main. Thanks for your contribution! 🌼

PR #1773, opened 2023-04-06
Export unexported modules for better LSP hint; fix wrong `fl` alias usage

SichangHe:

I noticed that my LSP (`pyright`) complains that `fl.server.strategy` is not a valid field. After some investigation, I found out that `strategy` is not exported in `server`. Adding the export resolves the error.

I went ahead and checked through the whole codebase and found another unexported module. I also found a misuse of `flwr` when `import flwr as fl` is used. So, I applied the fix.

Formatted with black and isort.

#### Any other comments?

In `/examples/opacus/dp_cifar_simulation.py`, on line 54, `get_evaluate_fn` has `fl.common.Weights` in its type annotation. `fl.common.Weights` does not actually exist, so this must be a mistake instead of a missed export.


danieljanes:
Thanks for the PR and the detailed description @SichangHe !

It's true that `Weights` does not exist any more. Or, to be more precise, it has been renamed to `NDArrays`. Do you want to update this PR to include the fix for this as well?

SichangHe:
Thanks for the feedback, @danieljanes.

I have renamed `Weights` in type annotations to `NDArrays`.

I also noticed that `common.ParameterRes` is not picked up by `pyright` nor `pylint`, but that class seems to be defined in Cpp so I don't know how to deal with it.

danieljanes:
Thanks for the big update @SichangHe!

`ParameterRes` has been renamed to `GetParametersRes` since Flower 1.0, so it's right that it does not get picked up (it doesn't exist anymore) - the code should be updated to use `GetParametersRes` instead.

One last thing that needs to be updated before we can merge: `flwr.common.logger` should not be exported from `flwr.common`. `log` already gets re-exported on the level of `flwr.common`, the right way to use it in applications is, for example, `from flwr.common import log`. Not exporting `logger` enables us to move the `log` implementation to a different module in `flwr.common` without breaking user code.

SichangHe:
> One last thing that needs to be updated before we can merge: `flwr.common.logger` should not be exported from `flwr.common`. `log` already gets re-exported on the level of `flwr.common`, the right way to use it in applications is, for example, `from flwr.common import log`. Not exporting `logger` enables us to move the `log` implementation to a different module in `flwr.common` without breaking user code.

`flwr.common.logger.configure` is still used around the codebase. What is up with that?

danieljanes:
Good catch @SichangHe - @tanertopal , should we export `configure` from `flwr.common`?

SichangHe:
@danieljanes, I have removed the export for `flwr.common.logger`. I think I shall leave `flwr.common.logger.configure` to you to deal with.

danieljanes:
Thanks @SichangHe, and thanks again for the PR! We can merge now 🎉

danieljanes (approved):
🌼

SichangHe:
Thank you, @danieljanes! 🎉 🌼

PR #1788, opened 2023-04-17
Export logger.configure from flwr.common

danieljanes:


tanertopal (dismissed):
lgtm

src/py/flwr/common/__init__.py
```diff
@@ -12,12 +12,13 @@
...
 # limitations under the License.
 # ==============================================================================
-"""Flower utilities shared between server and client."""
+"""Common components shared between server and client."""
 
 
 from .date import now as now
 from .grpc import GRPC_MAX_MESSAGE_LENGTH
 from .logger import log as log
+from .logger import configure as configure
```
danieljanes, line 21:
```suggestion
from .logger import configure as configure
from .logger import log as log
```

tanertopal (approved):

PR #1790, opened 2023-04-17
feat(ci): add `poetry` cache to CI

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Set the `cache` and `cache-dependency-path` argument to enable caching to speed up CI times.

### DIFF
--- a/.github/workflows/baselines.yml
+++ b/.github/workflows/baselines.yml
@@ -24,6 +24,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: 3.8.15
+          cache: "poetry"
+          cache-dependency-path: "**/pyproject.toml"
       - name: Install build tools
         run: |
           python -m pip install -U pip==22.3.1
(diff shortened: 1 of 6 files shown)


tanertopal:
@SauravMaheshkar The PR is much appreciated! Unfortunately, it does not seem to have an effect. If you can figure out how to make this work we'd love to have it.

danieljanes:
Thanks for the PR @SauravMaheshkar!

`setup-python` supports Poetry: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

SauravMaheshkar:
@tanertopal It is working isn't it ? 

Screenshot 2023-04-18 at 1 30 36 PM

It seems to use cached versions of packages and successfully restores cache.

@danieljanes I think the `poetry` argument requires a `poetry.lock` file instead of a `pyproject.toml`

danieljanes:
Thanks @SauravMaheshkar , the logs you posted look good. Is it the same for dependencies that are installed via Poetry? (`pip` is only used to install a specific version of our build tools)

SauravMaheshkar:
@tanertopal should I revert to [32cbbe3](https://github.com/adap/flower/pull/1790/commits/32cbbe37d023a69e563c330eb2fcab73fac04bee) ? At that point the CI was green

PR #1798, opened 2023-04-18
Make spacing before imports consistent

charlesbvll:

Makes the spacing at the beginning of Python files consistent, by always adding 2 line breaks between the module doc and the import section (this PR only applies to `flwr`, `flwr_tools`, and `examples` but not to `baselines`, `flwr_experimental`, and `flwr_examples`).


danieljanes (approved):
Thanks for the PR! I wish `black` was doing this automatically

PR #1800, opened 2023-04-18
Add http and https to address validation regex

charlesbvll:

Allows `http://` or `https://` addresses to be passed to the client or the server.

### DIFF
--- a/src/py/flwr/common/address.py
+++ b/src/py/flwr/common/address.py
@@ -22,14 +22,14 @@ from typing import Optional, Pattern, Tuple
 IPV6: int = 6
 
 DOMAIN_PATTERN: Pattern[str] = re.compile(
-    r"^(localhost)|(([a-zA-Z]{1})|([a-zA-Z]{1}[a-zA-Z]{1})|"
+    r"^(https?:\/\/)?(localhost)|(([a-zA-Z]{1})|([a-zA-Z]{1}[a-zA-Z]{1})|"
     r"([a-zA-Z]{1}[0-9]{1})|([0-9]{1}[a-zA-Z]{1})|"
     r"([a-zA-Z0-9][-_.a-zA-Z0-9]{0,61}[a-zA-Z0-9]))\."
     r"([a-zA-Z]{2,13}|[a-zA-Z0-9-]{2,30}.[a-zA-Z]{2,3})$"
 )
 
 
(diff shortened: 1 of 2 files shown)


src/py/flwr/common/address.py
```diff
@@ -22,14 +22,14 @@
 IPV6: int = 6
 
 DOMAIN_PATTERN: Pattern[str] = re.compile(
-    r"^(localhost)|(([a-zA-Z]{1})|([a-zA-Z]{1}[a-zA-Z]{1})|"
+    r"^(https?:\/\/)?(localhost)|(([a-zA-Z]{1})|([a-zA-Z]{1}[a-zA-Z]{1})|"
```
tanertopal, line 25:
```suggestion
    r"^(https?:\/\/)?(localhost)|(([a-zA-Z]{1})|([a-zA-Z]{1}[a-zA-Z]{1}))"
```

 think there is a closing `)` missing.

src/py/flwr/common/address.py
```diff
@@ -22,14 +22,14 @@
...
     r"([a-zA-Z]{1}[0-9]{1})|([0-9]{1}[a-zA-Z]{1})|"
```
tanertopal, line 26: Whats this going to match? Do you have an example? Are you sure about the final `|`

src/py/flwr/common/address.py
charlesbvll: @tanertopal Nope, the closing bracket is on the 3rd line

src/py/flwr/common/address.py
charlesbvll: Most of this regex is from https://validators.readthedocs.io/en/latest/_modules/validators/domain.html#domain, we just added `localhost`, `http://`, and `https://` to it.

charlesbvll: To answer your question more specifically, I think the requirements it will match are the following:

- A domain name consists of minimum two and maximum 63 characters.
- All letters from a to z, all numbers from 0 to 9 and a hyphen (-) are possible.
- A domain name mustn't consist of a hyphen (-) on the third and fourth position at the same time

You can see some basic examples of this in the `address_test.py`.

src/py/flwr/common/address.py
tanertopal: my bad. Did not realize this is a multi line string.

tanertopal (approved):

PR #1811, opened 2023-04-20
Make node_id in driver state manager sqlite compatible

tanertopal:

Make node_id generation in driver state manager sqlite compatible by limiting the maxint size.

### DIFF
--- a/src/py/flwr/server/grpc_server/driver_client_manager.py
+++ b/src/py/flwr/server/grpc_server/driver_client_manager.py
@@ -16,7 +16,7 @@
 
 
 import threading
-import uuid
+import random
 from typing import Dict, List, Optional, Set, Tuple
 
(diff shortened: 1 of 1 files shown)


src/py/flwr/server/grpc_server/driver_client_manager.py
```diff
@@ -72,7 +72,7 @@ def register(self, client: ClientProxy) -> bool:
             return False
 
         # Generate random integer ID
-        random_node_id: int = uuid.uuid1().int >> 64
+        random_node_id: int = random.randrange(9223372036854775806)
```
danieljanes, line 75: The max value is 9223372036854775807, so shouldn't this be:
```suggestion
        random_node_id: int = random.randrange(9223372036854775808)
```

PR #1842, opened 2023-05-04
Add E2E tests for PyTorch and TensorFlow examples

charlesbvll:

Improve current tests for PyTorch and TensorFlow examples by validating them from end to end.


.github/workflows/examples.yml
```diff
@@ -28,33 +28,34 @@ jobs:
...
+          path: "./tests/pytorch/data"
           key: pytorch-datasets
       - name: Download Datasets
         run: |
-          cd examples/quickstart_pytorch
-          python -c "from torchvision.datasets import CIFAR10; CIFAR10(\"./dataset\", download=True)"
-      - name: Test
+          cd tests/pytorch
+          python -c "from torchvision.datasets import CIFAR10; CIFAR10('./data', download=True)"
+      - name: Run FL
```
tanertopal, line 42:
```suggestion
      - name: Run test
```

.github/workflows/examples.yml
```diff
@@ -95,22 +101,23 @@ jobs:
       - name: Download Datasets
         run: |
           python -c "import tensorflow as tf; tf.keras.datasets.cifar10.load_data()"
-      - name: Test
+      - name: Run FL
```
tanertopal, line 104:
```suggestion
      - name: Run tests
```

.github/workflows/examples.yml
```diff
@@ -141,44 +153,24 @@ jobs:
...
-          cd examples/pytorch_from_centralized_to_federated
+          cd tests/bare
           python -m poetry install
-      - name: Cache Datasets
-        uses: actions/cache@v2
-        with:
-          path: "./examples/pytorch_from_centralized_to_federated/dataset"
-          key: pytorch-datasets
-      - name: Download Datasets
+      - name: Run FL
```
tanertopal, line 158:
```suggestion
      - name: Run tests
```

tests/bare/result
```diff
@@ -0,0 +1 @@
+SUCCESS
```
tanertopal, line 1: Whats the purpose of this file?

tanertopal:
@danieljanes @charlesbvll My thoughts are:
* Should we instead name the top-level directory just `e2e` to avoid anyone wanting to put their unit or integration tests? I don't see any other than e2e tests at the top level.
* other comments in the code

tests/bare/result
charlesbvll: It is generated in case of success of the training, it should not be committed, I will remove it

tests/bare/server.py
```diff
@@ -0,0 +1,19 @@
...
+)
+
+if hist.losses_distributed[-1][1] != 0:
+    with open("result", "w") as file:
+        file.write("FAIL")
+    sys.exit("Loss did not decrease.")
+else:
+    with open("result", "w") as file:
+        file.write("SUCCESS")
+    sys.exit()
```
danieljanes, line 19: Can we just set a 0 or 1 exit code here and use that as the exit code of `test.sh`? That would prevent us from having to write a file etc.

tests/bare/client.py
```diff
@@ -0,0 +1,28 @@
...
+import numpy as np
+
+SUBSET_SIZE = 1000
+
+
+model_params = np.array([1])
+objective = 5
+
+# Define Flower client
+class DummyClient(fl.client.NumPyClient):
```
danieljanes, line 11:
```suggestion
class FlowerClient(fl.client.NumPyClient):
```

tests/bare/client.py
```diff
@@ -0,0 +1,28 @@
...
+    def get_parameters(self, config):
+        return model_params
+
+    def fit(self, parameters, config):
+        model_params = parameters
```
danieljanes, line 16: This line is not necessary if you use `parameters` on the next line

tests/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This directory contains folders for different scenarios that need to be tested and validated before a change can be added to Flower.
+
+Currently the following end-to-end tests have been implemented:
+
+* [bare](bare): testing Flower in a bare minimum scenario, that is, with a dummy model and dummy operations. This is mainly to test the core functionnality of Flower independently from any framework. It can easily be extendended to test more complex communication set-ups
+
+* [pytorch](pytorch): testing Flower with PyTorch by using the CIFAR10 dataset and a CNN.
+
+* [tensorflow](tensorflow): testing Flower with Tensorflow by using the CIFAR10 dataset and a MobileNetV2 model.
```
danieljanes, line 11: Let's remove the list here and add the details in the README of each test case. Otherwise, we have to modify this top-level README every time we add a new test case.

tests/bare/server.py
```diff
@@ -0,0 +1,19 @@
...
+import flwr as fl
+
+
+# Start Flower server
+hist = fl.server.start_server(
+    server_address="0.0.0.0:8080",
+    config=fl.server.ServerConfig(num_rounds=3),
+)
+
+if hist.losses_distributed[-1][1] != 0:
```
danieljanes, line 12: Should we check for an exact value here?

danieljanes (changes requested):

tests/bare/client.py
charlesbvll: This in order to store the parameters in between rounds

tanertopal:
@danieljanes @charlesbvll I would like to either rename the top-level directory as mentioned here: https://github.com/adap/flower/pull/1842#issuecomment-1538615380 into `e2e` or move all `e2e` tests into `tests/e2e`. IMO it's important to clearly indicate that the other tests are in the src directory as I do not want to create impressions that's all the tests we have or confuse contributors wrt to where to put a test.

tanertopal:
One more thought: We have quite some frameworks specific code in the tests, and I wonder if that's the best approach. If we test the examples, that would be covered, so we do not need to duplicate the effort here. I believe we should make the E2E minimal while testing our APIs. It would be more valuable to, e.g. test failure cases such as providing an invalid loss or, e.g. NumPy arrays with different shapes. @danieljanes @charlesbvll wdyt?

danieljanes:
> One more thought: We have quite some frameworks specific code in the tests, and I wonder if that's the best approach. If we test the examples, that would be covered, so we do not need to duplicate the effort here. I believe we should make the E2E minimal while testing our APIs. It would be more valuable to, e.g. test failure cases such as providing an invalid loss or, e.g. NumPy arrays with different shapes. @danieljanes @charlesbvll wdyt?

My perspective is that we should make a clear distinction between testing examples and testing the framework. Examples are written with readability in mind, and that will often lead to situations where the example is not a good E2E test case. If we use examples as E2E tests, we might face difficulties in deciding whether to write them in a way that's good for the reader (our main objective) or good for our testing. I'm in favor of not using examples as E2E tests (and removing the ones we have from the testing). How to test the examples themselves (so that they don't break over time) is an open question, and it seems like we don't have a good answer yet.

That being said, in the E2E testing of the framework, we need to cover each framework at least once. We need to ensure that the typical workflow works with each framework: sending parameters to the client node, deserializing them, "using" them in the ML framework (PyTorch, TensorFlow, ...), getting the out of the framework again, serializing them, and sending them back to the server. Once each framework has at least one E2E test, we can cover the rest of the E2E scenarios with more "low-level" way (for example, by just using NumPy NDArrays).

tanertopal:
> > One more thought: We have quite some frameworks specific code in the tests, and I wonder if that's the best approach. If we test the examples, that would be covered, so we do not need to duplicate the effort here. I believe we should make the E2E minimal while testing our APIs. It would be more valuable to, e.g. test failure cases such as providing an invalid loss or, e.g. NumPy arrays with different shapes. @danieljanes @charlesbvll wdyt?
> 
> My perspective is that we should make a clear distinction between testing examples and testing the framework. Examples are written with readability in mind, and that will often lead to situations where the example is not a good E2E test case. If we use examples as E2E tests, we might face difficulties in deciding whether to write them in a way that's good for the reader (our main objective) or good for our testing. I'm in favor of not using examples as E2E tests (and removing the ones we have from the testing). How to test the examples themselves (so that they don't break over time) is an open question, and it seems like we don't have a good answer yet.
> 
> That being said, in the E2E testing of the framework, we need to cover each framework at least once. We need to ensure that the typical workflow works with each framework: sending parameters to the client node, deserializing them, "using" them in the ML framework (PyTorch, TensorFlow, ...), getting the out of the framework again, serializing them, and sending them back to the server. Once each framework has at least one E2E test, we can cover the rest of the E2E scenarios with more "low-level" way (for example, by just using NumPy NDArrays).

If we are not considering the examples as E2E tests, I agree with testing including framework-specific code. I'll review the code more thoroughly.

tanertopal:
@charlesbvll When I run the test for TensorFlow I get this one: 

image

Can you check if you can reproduce it?

charlesbvll:
> @charlesbvll When I run the test for TensorFlow I get this one:
> 
> ```shell
> INFO flwr 2023-05-24 21:15:19,210 | app.py:165 | Starting Flower server, config: ServerConfig(num_rounds=3, round_timeout=None)
> INFO flwr 2023-05-24 21:15:19,222 | app.py:179 | Flower ECE: gRPC server running (3 rounds), SSL is disabled
> INFO flwr 2023-05-24 21:15:19,222 | server.py:89 | Initializing global parameters
> INFO flwr 2023-05-24 21:15:19,222 | server.py:279 | Requesting initial parameters from one random client
> Traceback (most recent call last):
>   File "...me/.pyenv/versions/3.7.15/lib/python3.7/threading.py", line 300, in wait
>     gotit = waiter.acquire(True, timeout)
>   File "...me/development/adap/flower/.venv/lib/python3.7/site-packages/ray/_private/worker.py", line 1727, in sigterm_handler
>     sys.exit(signum)
> SystemExit: 15
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "...me/development/adap/flower/src/py/flwr/server/client_manager.py", line 126, in wait_for
>     lambda: len(self.clients) >= num_clients, timeout=timeout
>   File "...me/.pyenv/versions/3.7.15/lib/python3.7/threading.py", line 331, in wait_for
>     self.wait(waittime)
>   File "...me/.pyenv/versions/3.7.15/lib/python3.7/threading.py", line 300, in wait
>     gotit = waiter.acquire(True, timeout)
>   File "...me/development/adap/flower/.venv/lib/python3.7/site-packages/ray/_private/worker.py", line 1727, in sigterm_handler
>     sys.exit(signum)
> SystemExit: 15
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "server.py", line 7, in <module>
>     config=fl.server.ServerConfig(num_rounds=3),
>   File "...me/development/adap/flower/src/py/flwr/server/app.py", line 185, in start_server
>     config=initialized_config,
>   File "...me/development/adap/flower/src/py/flwr/server/app.py", line 224, in run_fl
>     hist = server.fit(num_rounds=config.num_rounds, timeout=config.round_timeout)
>   File "...me/development/adap/flower/src/py/flwr/server/server.py", line 90, in fit
>     self.parameters = self._get_initial_parameters(timeout=timeout)
>   File "...me/development/adap/flower/src/py/flwr/server/server.py", line 280, in _get_initial_parameters
>     random_client = self._client_manager.sample(1)[0]
>   File "...me/development/adap/flower/src/py/flwr/server/client_manager.py", line 180, in sample
>     self.wait_for(min_num_clients)
>   File "...me/development/adap/flower/src/py/flwr/server/client_manager.py", line 126, in wait_for
>     lambda: len(self.clients) >= num_clients, timeout=timeout
>   File "...me/.pyenv/versions/3.7.15/lib/python3.7/threading.py", line 244, in __exit__
>     return self._lock.__exit__(*args)
> RuntimeError: cannot release un-acquired lock
> Training had an issue
> ```
> 
> Can you check if you can reproduce it?

Well I am unable to reproduce it but this is very weird for a few reasons:
- You shouldn't be able to call the script from outside the `e2e/tensorflow` folder, it should error out because it won't find `../test.sh`
- There shouldn't be any interactions with Ray as this is not using the simulation engine at all and it is not part of the dependencies
- I have had some issues before when other Flower processes where running and I forgot to kill them, that might be a possible cause of your issue (even though I don't really understand the first 2 errors)

The most plausible explanation might be that you have a `test.sh` script sitting at the root of the Flower dir which is being called by the symlink.

e2e/bare/test.sh
```diff
@@ -0,0 +1 @@
+../test.sh
```
tanertopal, line 1:
```suggestion
../test.sh

```

e2e/pytorch/test.sh
```diff
@@ -0,0 +1 @@
+../test.sh
```
tanertopal, line 1:
```suggestion
../test.sh

```

e2e/tensorflow/test.sh
```diff
@@ -0,0 +1 @@
+../test.sh
```
tanertopal, line 1:
```suggestion
../test.sh

```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,101 @@
...
+          python -m poetry install
+      - name: Cache Datasets
+        uses: actions/cache@v2
+        with:
+          path: "./e2e/pytorch/data"
+          key: pytorch-datasets
+      - name: Download Datasets
+        run: |
+          cd e2e/pytorch
+          python -c "from torchvision.datasets import CIFAR10; CIFAR10('./data', download=True)"
```
tanertopal, line 41: @charlesbvll, how about running the cache@v3 and checking for a cache hit before triggering the download once more?

=> https://github.com/actions/cache/tree/main#restoring-and-saving-cache-using-a-single-action

It might reduce flaky jobs in case the remote servers are not reachable.

e2e/bare/test.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/
+
+bash ../test.sh
```
tanertopal, line 5:
```suggestion
bash ../test.sh

```

.github/workflows/e2e.yml
tanertopal: @charlesbvll I am merging it. Feel free to do this in a follow-up PR if you think its worth it.

tanertopal (approved):
lgtm! Great work!

PR #1843, opened 2023-05-05
Support many uvicorn workers

adam-narozniak:
* Add DATABASE environment variable
* Remove starting _run_fleet_api_rest from a thread (reason: Signal only works in main thread)
* Add on_startup  event for StateFactory creation with environment variable to the fastAPI app (rest_api.py file)


WilliamLindskog:
Closing this for now.

PR #1849, opened 2023-05-07
Fix behavior with null loss

charlesbvll:

Currently when a loss is `0.0` it won't be stored in the `History` object. 

The PR also makes the check for `None` objects consistent.


danieljanes (approved):
Good catch!

PR #1855, opened 2023-05-08
Remove redundant warning check for strategies subclassing FedAvg

adam-narozniak:
Some strategies subclassing FedAvg still define the warning check that is already implemented in FedAvg.

```
        if (
            min_fit_clients > min_available_clients
            or min_evaluate_clients > min_available_clients
        ):
            log(WARNING, WARNING_MIN_AVAILABLE_CLIENTS_TOO_LOW)
```
They also define the string in `WARNING_MIN_AVAILABLE_CLIENTS_TOO_LOW`. This is redundant and results in double warnings.


danieljanes (approved):

adam-narozniak:
@danieljanes The error I got after removing the warnings meant that the __init__ in fedmedian and fedxgb_nn_avg were redundant. I removed them. Alternatively, the __init__ can be kept, and I can suppress "W0235: Useless super delegation in method '__init__' (useless-super-delegation)" so we can keep the docs and consistent structure. Let me know what you think

danieljanes:
@adam-narozniak I'm in favor of removing the redundant `__init__` to reduce the amount of code duplication

PR #1858, opened 2023-05-09
Push Swift Code to Downstream Repository to Enable Publishing in Swift Package Manager

danielnugraha:
Add a new CI job script that automatically pushes Swift code changes in src/swift to downstream repository flower-swift (https://github.com/adap/flower-swift).


.github/workflows/flower-swift_sync.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Pushes src/swift to flower-swift repository
+        uses: cpina/github-action-push-to-another-repository@main
+        env:
+          SSH_DEPLOY_KEY: ${{ secrets.FLOWER_SWIFT_SSH }}
+        with:
+          source-directory: 'src/swift/'
+          destination-github-username: 'adap'
+          destination-repository-name: 'flower-swift'
```
tanertopal, line 23:
```suggestion
          destination-repository-name: 'flower-swift'

```

.github/workflows/flower-swift_sync.yml
```diff
@@ -0,0 +1,23 @@
+name: Sync flower-swift
+
+on:
+  push:
+    branches: ['main']
+    paths: ['src/swift/**']
+  pull_request:
+    branches: ['main']
+    paths: ['src/swift/**']
```
tanertopal, line 9: Run this only when the code merges with the main branch instead of always running it.

tanertopal, line 9:
```suggestion
on:
  push:
    branches: ['main']
    paths: ['src/swift/**']
```

Run this only when the code merges with the main branch instead of always running it.

.github/workflows/flower-swift_sync.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Pushes src/swift to flower-swift repository
+        uses: cpina/github-action-push-to-another-repository@main
```
tanertopal, line 17: Do not use @main but instead a specific git commit, e.g. "cbe757f", which is release v1.7. We do not want to enable a dependency which might get compromised to get full access to all our environment variables.

tanertopal (approved):

PR #1859, opened 2023-05-09
Add tensorflow-macos support 

adam-narozniak:
Currently, TensorFlow for macOS is not supported out of the box. To fix this I propose adding markers to dependencies specification for both `tensorflow-cpu` and `tensorflow-macos`.


danieljanes (approved):
Lgtm!

PR #1879, opened 2023-05-17
Make tensorflow optional when using tensorboard

charlesbvll:
### Description

Currently when using tensorboard, the app will crash in an unhelpful way if `tensorflow` is not installed.

### Related issues/PRs

N/A

### Explanation

If `tensorflow` is not installed we display a warning to the user, and instead use the regular strategy that was passed to the wrapper so the execution can continue.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


jafermarq:
afaik tenforflow is not needed to be installed to use Tensorboard. On a clean env i can do just `pip install tensorboard` and the following will work just fine:
```python
import tensorboard as tb
w = tb.summary.Writer('test') # creates dir `test` in current path
w.add_scalar(data=0.99, tag='meh', step=0)
 w.add_scalar(data=1.99, tag='meh', step=1)
w.flush()
w.close()
quit()
```
Then i can launch the dashboard and see the two datapoints: `tensorboard --logdir test`

charlesbvll:
> 

@jafermarq We actually don't use Tensorboard directly in the code, we only use `tensorflow.summary` , hence why I was doing the check on Tensorflow. Maybe the code could be rewritten without `tensorflow`, but I'm not sure...

danieljanes (approved):

PR #1887, opened 2023-05-21
Fix code documentation in aggregate_krum()

edogab33:
### Description

Just realized I wrote a wrong comment here. So, `aggregate_krum` does not return the index of the best model, but the best model itself.

### DIFF
--- a/src/py/flwr/server/strategy/aggregate.py
+++ b/src/py/flwr/server/strategy/aggregate.py
@@ -87,7 +87,7 @@ def aggregate_krum(
         best_results = [results[i] for i in best_indices]
         return aggregate(best_results)
 
-    # Return the model which minimizes the score (Krum)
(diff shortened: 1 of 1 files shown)


src/py/flwr/server/strategy/aggregate.py
```diff
@@ -87,7 +87,7 @@ def aggregate_krum(
         best_results = [results[i] for i in best_indices]
         return aggregate(best_results)
 
-    # Return the index of the client which minimizes the score (Krum)
+    # Return the model which minimizes the score (Krum)
```
danieljanes, line 90:
```suggestion
    # Return the model parameters that minimize the score (Krum)
```

danieljanes (approved):

danieljanes:
Thanks for the PR @edogab33! Made a tiny tweak & approved

PR #1817, opened 2023-04-23
Implement Bulyan

edogab33:
Implements #1612

PR #1891, opened 2023-05-23
Implement Bulyan (#1817)

adam-narozniak:


src/py/flwr/server/strategy/bulyan_test.py
```diff
@@ -0,0 +1,241 @@
...
+    """Test num_fit_clients function."""
+    # Prepare
+    strategy = Bulyan()
+    expected = 20
+
+    # Execute
+    actual, _ = strategy.num_fit_clients(num_available_clients=20)
+
+    # Assert
+    assert expected == actual
```
danieljanes, line 48: Let's remove test like this if they only test functionality not modified in the `Bulyan` strategy.

src/py/flwr/server/strategy/bulyan_test.py
```diff
@@ -0,0 +1,241 @@
...
+    param_5: Parameters = ndarrays_to_parameters(
+        [array([0.1, 0.1, 0.1, 0.1], dtype=float32)]
+    )
+    bridge = MagicMock()
+    client_0 = GrpcClientProxy(cid="0", bridge=bridge)
+    client_1 = GrpcClientProxy(cid="1", bridge=bridge)
+    client_2 = GrpcClientProxy(cid="2", bridge=bridge)
+    client_3 = GrpcClientProxy(cid="3", bridge=bridge)
+    client_4 = GrpcClientProxy(cid="4", bridge=bridge)
+    client_5 = GrpcClientProxy(cid="5", bridge=bridge)
```
danieljanes, line 175: There's no need to make this test depend on `GrpcClientProxy`. It seems sufficient to use `MagicMock` for the `ClientProxy` instances in the `results` list.

danieljanes (changes requested):
Lgtm! I added two comments to improve the tests a bit.

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -91,6 +91,90 @@ def aggregate_krum(
...
+BYZANTINE_RESILIENT_MANY_RETURN_MODELS_AGGREGATION = (
+    []
+)  # Brute, Medoid (but not implemented yet)
+
+
+# pylint: disable=too-many-locals
+def aggregate_bulyan(
+    results: List[Tuple[NDArrays, int]],
+    num_malicious: int,
+    aggregation_rule: Callable,
```
adam-narozniak, line 106: @danieljanes Could you give me some feedback regarding the proposed solution, mainly regarding the aggregation_rule parameter? 

The Bulyan uses the Byzantine resilient aggregation internally, and the specific aggregation choice should be up to the user. This functionality comes with complex hint typing for callable that might not cover all choices of aggregation rule (introduced in the future). I'm not sure if I should follow by trying to capture the possibilities or maybe you have any suggestions for a neater solution for this problem. I am not satisfied with the list in the middle of the file that captures the different types of Byzantine resilient aggregation (with the division implied by the Bulyan aggregation method).

danieljanes: @adam-narozniak I agree that enumerating all the possible typing options would be too complex, let's stick to just `Callable`.

Regarding the two list constants, since they are used in only one place, should we just inline the possible values there?

danieljanes (changes requested):

src/py/flwr/server/strategy/bulyan.py
```diff
@@ -0,0 +1,161 @@
...
+        **aggregation_rule_kwargs: Any,
+    ) -> None:
+        """Configurable Bulyan strategy.
+
+        Implementation based on https://arxiv.org/pdf/1803.01498v1.pdf
+
+        Parameters
+        ----------
+        fraction_fit : float, optional
+            Fraction of clients used during training. Defaults to 0.1.
```
jafermarq, line 76: doesn't match the actual default set in `__init__()`, same goes for `fraction_evaluate`. I presume we want `0.1` in both as is the default in FedAvg for instance? But, if this strategy is more tailored to the `cross-silo` setting then I guess 1.0 makes sense.

src/py/flwr/server/strategy/bulyan.py
```diff
@@ -0,0 +1,161 @@
...
+        accept_failures: bool = True,
+        initial_parameters: Optional[Parameters] = None,
+        fit_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        evaluate_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        first_aggregation_rule: Callable = aggregate_krum,  # type: ignore
+        **aggregation_rule_kwargs: Any,
+    ) -> None:
+        """Configurable Bulyan strategy.
+
+        Implementation based on https://arxiv.org/pdf/1803.01498v1.pdf
```
jafermarq, line 71: This points to an old (2018) version of the paper. do we want this? I'd be inclined in pointing to the abstract page instead of the pdf (just like at the top of the file): `https://arxiv.org/abs/1803.01498`

src/py/flwr/server/strategy/bulyan.py
```diff
@@ -0,0 +1,161 @@
...
+            ]
+        ] = None,
+        on_fit_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
+        on_evaluate_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
+        accept_failures: bool = True,
+        initial_parameters: Optional[Parameters] = None,
+        fit_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        evaluate_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        first_aggregation_rule: Callable = aggregate_krum,  # type: ignore
+        **aggregation_rule_kwargs: Any,
```
jafermarq, line 67: Wouldn't it be better to have `aggregation_rule_kwargs` as a `Dict ` type? So later `aggregate_bulyan` will pass them as `**aggregate_bulyan` to `aggregation_rule`?

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -91,6 +91,88 @@ def aggregate_krum(
...
+    aggregated_parameters: NDArrays
+        Aggregated parameters according to the Bulyan strategy.
+    """
+    byzantine_resilient_single_ret_model_aggregation = [aggregate_krum]
+    # also GeoMed (but not implemented yet)
+    byzantine_resilient_many_return_models_aggregation = []  # type: ignore
+    # Brute, Medoid (but not implemented yet)
+
+    num_clients = len(results)
+    if num_clients < 4 * num_malicious + 3:
```
jafermarq, line 125: shouldn't this be changed to `num_clients >= 4 * num_malicious + 3`? as per the paper and the message in ValueError?

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -170,3 +252,86 @@ def aggregate_trimmed_avg(
...
+    for idx, weights in enumerate(list_of_weights):
+        if _check_weights_equality(reference_weights, weights):
+            return idx
+    raise ValueError("The reference weights not found in list_of_weights.")
+
+
+def _aggregate_n_closest_weights(
+    reference_weights: NDArrays, results: List[Tuple[NDArrays, int]], beta_closest: int
+) -> NDArrays:
+    """It calculates element-wise mean of the `N` the closest values.
```
jafermarq, line 299: remove `the` after `N`

src/py/flwr/server/strategy/bulyan.py
adam-narozniak: I'd stick to `**aggregation_rule_kwargs`. Later it can be used as `aggregation_rule(aggregation_rule_kwargs)`

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -91,6 +91,88 @@ def aggregate_krum(
...
+            "equal to the 4 * num_malicious + 3. This is the assumption of this method."
+            "It is needed to ensure that the method reduces the attacker's leeway to "
+            "the one proved in the paper."
+        )
+    selected_models_set: List[Tuple[NDArrays, int]] = []
+
+    # Create a list of weights and ignore the number of examples
+    weights = [weights for weights, _ in results]
+
+    theta = len(weights) - 2 * num_malicious
```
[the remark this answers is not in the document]
adam-narozniak: Done

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -91,6 +91,88 @@ def aggregate_krum(
...
+        list_of_weights = [weights for weights, num_samples in results]
+        # This group gives exact result
+        if aggregation_rule in byzantine_resilient_single_ret_model_aggregation:
+            best_idx = _find_reference_weights(best_model, list_of_weights)
+        # This group requires finding the closest model to the returned one
+        # (weights distance wise)
+        elif aggregation_rule in byzantine_resilient_many_return_models_aggregation:
+            # when different aggregation strategies available
+            # write a function to find the closest model
+            best_idx = 0
```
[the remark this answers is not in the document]
adam-narozniak: Yes, you're right. I've added the the `NotImplementedError`

src/py/flwr/server/strategy/aggregate.py
adam-narozniak: Done

jafermarq (approved):
looks ready.

danieljanes (approved):
Great work @edogab33 @adam-narozniak @jafermarq !!

PR #1917, opened 2023-06-13
Add Drop Clients Based on Performance tutorial

adam-narozniak:
This tutorial uses custom Strategy and Criterion to demonstrate how to control sampling. Each fit client is evaluated on the centralized dataset and dropped if its performance is below a computed threshold.


WilliamLindskog:

PR #1923, opened 2023-06-14
Fix D105 Missing docstring in magic method

adam-narozniak:
* Add docstring to __repr__
* Fix this problem in all the files besides strategies (the linter didn't detect all the strategies that had this problem)

CAUTION: Please review this PR in the order mentioned https://github.com/adap/flower/pull/1920 for your easy review process.

### DIFF
--- a/src/py/flwr/server/history.py
+++ b/src/py/flwr/server/history.py
@@ -73,6 +73,20 @@ class History:
             self.metrics_centralized[key].append((server_round, metrics[key]))
 
     def __repr__(self) -> str:
+        """Create a representation of History.
+
+        The representation consists of the following data (for each round) if present:
+        * distributed loss,
+        * centralized loss,
+        * distributed training metrics,
+        * distributed evaluation metrics,
+        * centralized metrics.
+
+        Returns
+        -------
+        representation : str
+            The string representation of the history object.
+        """
         rep = ""
         if self.losses_distributed:
             rep += "History (loss, distributed):\n" + reduce(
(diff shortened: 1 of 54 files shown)


src/py/flwr/server/history.py
```diff
@@ -73,6 +73,20 @@ def add_metrics_centralized(
...
 
     def __repr__(self) -> str:
+        """Create a representation of History.
+
+        The representation consists of the following data (for each round) if present:
+        * distributed loss,
+        * centralized loss,
+        * distributed training metrics,
+        * distributed evaluation metrics,
+        * centralized metrics.
```
tanertopal, line 83:
```suggestion
        * distributed loss
        * centralized loss
        * distributed training metrics
        * distributed evaluation metrics
        * centralized metrics
```

Did we use this style somewhere else in bullet lists? Otherwise, I suggest removing the "," and ".".

Wdyt @adam-narozniak ?

adam-narozniak: Yes, but I haven't checked it while adding. We used dots at the end of each bullet point
https://github.com/adap/flower/blob/34e9a4e889c95e6607b11b0ba0dcaa48f2735e24/src/py/flwr/driver/app.py#L77-L82
And here
https://github.com/adap/flower/blob/34e9a4e889c95e6607b11b0ba0dcaa48f2735e24/src/py/flwr/server/fleet/grpc_bidi/grpc_server.py#L106-L112
Should I move to the dots?

tanertopal: Yes, additionally, add the new line I would suggest. Let's keep it consistent.

adam-narozniak: Done - "." at the end of each bullet point and the new line between the ":" and the first bullet point.

tanertopal (approved):

PR #1934, opened 2023-06-15
Fix flake8 noqa

adam-narozniak:
CAUTION: Please review this PR in the order mentioned https://github.com/adap/flower/pull/1920 for your easy review process.
### Description

At the top of many strategies, there was a `# flake8: noqa: E501` which was supposed to be used for ignoring E501 for the whole file. However, it's not how it worked - it was interpreted as disabling all the errors for the whole file. 

Flake8 doesn't allow (without any plugins) to disable checks for the whole files. It works in two modes: 
1. Per line disabling of a specific error
eg. 
```
a = 5 # noqa: code
```
2. Per file disabling of ALL errors
```
# flake8: noqa

code here
```

`# flake8: noqa: E501` is undesired behavior because it disables all the errors detected in the whole fine (not as expected only E501 - line too long). It affects the following tools:
1. Flake
2. Ruff 

### Related issues/PRs

N/A

### Explanation

Remove the `# flake8: noqa: E501` and add new lines to keep the width of the lines below 88.
It also enabled the removal of the `pylint: disable=line-too-long` used for the whole function (it was applied on a per-line basis in only a single file).

Note that E501 was probably introduced to keep the hint types in documentation in a single line:
```
evaluate_fn : Optional[Callable[[int, NDArrays, Dict[str, Scalar]], Optional[Tuple[float, Dict[str, Scalar]]]]]
```
However, the new line to keep the line width below 88 doesn't break anything. The documentation is created as expected.


src/py/flwr/server/strategy/fedprox.py
```diff
@@ -58,14 +57,15 @@ def __init__(
...
+        be adjusted.
         A proximal term needs to be added to the loss function during the training:
 
         .. math::
             \\frac{\\mu}{2} || w - w^t ||^2
 
-        Where $w^t$ are the global parameters and $w$ are the local weights the function will
-        be optimized with.
+        Where $w^t$ are the global parameters and $w$ are the local weights the function
+         will be optimized with.
```
tanertopal, line 68:
```suggestion
        will be optimized with.
```

tanertopal (approved):

PR #1936, opened 2023-06-15
Return `Node` instances from `GetNodes` 

danieljanes:
### Description

The Driver API returns individual `int` node IDs, not full `Node` instances. When scheduling a `Task`, one has to construct a full `Node` instance (setting the `node_id` and the `anonymous` flag). Changes to the `Node` message are not reflected in `GetNodes` and it is unclear how additional `Node` properties would be communicated to the driver script.

### Related issues/PRs

None

### Explanation

 If `GetNodes` returned full `Node` instances, one could just sample from that list and set one of the `Node` instances as the `consumer` of the new `Task`. Changes to the `Node` message (e.g., additional properties) would automatically be reflected in `GetNodes`.

This PR returns full `Node` instances instead of `int` node IDs from `GetNodes`.

### Checklist

- [ ] Write tests
- [x] Make CI checks pass


tanertopal:
@danieljanes, my thoughts on this:

* Changing from int64 to a sub-message will increase the response size, making it less scalable.
* The Driver potentially has to pull all node ids, e.g. 10M+, which was our initial thinking for choosing int64.
* If we need a more verbose response, how about splitting it into two methods?

tanertopal (approved):

tanertopal (approved):

PR #1946, opened 2023-06-19
Add Ruff UP check

adam-narozniak:
### Description

The current code is not checked if the updates recommended for new Python versions were made. 

### Related issues/PRs
NA
## Proposal
Add Ruff check implemented from `pyupgrade`.
Fix the UP008 Use super() instead of super(__class__, self)


adam-narozniak:
@tanertopal By setting the target version to `py37` the roof tool checks for possible code updates up to Python 3.7 that come as a result of Python updates. It checks for the changes that were introduced in Python recommendations up to Python 3.7. It doesn't upgrade any code itself.
We can't use Python 3.11 as the target because we use features that are compatible with 3.7, and by having the higher target version, we would have to use features introduced later than in 3.7 e.g.
```
UP006 [*] Use `tuple` instead of `Tuple` for type annotation
B905 `zip()` without an explicit `strict=` parameter
UP007 [*] Use `X | Y` for type annotations
```

danieljanes (approved):

PR #1949, opened 2023-06-19
Add twine wheel check

adam-narozniak:
## Issue
Currently, the wheel is not tested; therefore, any potential problems can be detected only manually.

## Proposal
Add `twine check`.

### Explanation
 It checks the distribution's long description rendering on PyPI.


dev/test-wheel.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/../
+
+echo "=== test-wheel.sh ==="
+
+# Build the wheel first
+echo "Building the wheel: start"
+./dev/build.sh
+echo "Building the wheel: done"
```
danieljanes, line 10: @adam-narozniak this should not be part of the test script (otherwise re re-build the wheel every time we run the wheel tests)

adam-narozniak: @danieljanes I thought it should be done like it. 
But I can change it, and require the build by a user. I can check if the directory is empty, yet it won't be resilient to testing an old wheel accidentally.

PR #1957, opened 2023-06-22
dynamically decorate function with ray.remote to fix GPU OOM

noah822:
### Description
According to [ray doc](https://docs.ray.io/en/latest/ray-core/api/doc/ray.remote.html), without setting ``max_calls=1``, using GPUs with ray may lead to memory leakage issue. Because ray actor process might be reused multiple times and the resources allocated to it won't be reclaimed until the process terminates.
There are two simple fixes I could think of

- Statically decorate all relevant functions with ``@ray.remote(max_calls=1)``. But this fix might be suboptimal, because in other settings you might need ``max_calls`` to be large for better performance.
- Use ``{max_calls : 1}`` as one of the key-value pair of ``client_resources`` passed to ``flwr.simulation.start_simulation``. In the [source code](https://github.com/adap/flower/blob/main/src/py/flwr/simulation/ray_transport/ray_client_proxy.py), ``client_resources`` is then passed to ray actor through ``.options()``. But due to constraints imposed by ray, ``max_calls`` is the only exception, which can not be overwritten is ``.options()``. Check the response by ray team [here](https://discuss.ray.io/t/behavior-of-max-calls-of-ray-remote-by-default/11088)

Therefore, I am using a function factory kind of thing to dynamically decorate ray actors. User is then able to pass ``{max_calls : 1}`` as one of the configuration for ray.remote through ``client_resources``. If this PR gets merged, the doc probably needs to be updated to reflect this.

### Related issues/PRs

- GPU OOM problem related to this fix has been elaborated in my previous [issue post](https://github.com/adap/flower/issues/1942).
- There is [another PR](https://github.com/adap/flower/pull/1384) related to this.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/src/py/flwr/simulation/ray_transport/ray_client_proxy.py
+++ b/src/py/flwr/simulation/ray_transport/ray_client_proxy.py
@@ -46,8 +46,8 @@ class RayClientProxy(ClientProxy):
         self, ins: common.GetPropertiesIns, timeout: Optional[float]
     ) -> common.GetPropertiesRes:
         """Return client's properties."""
-        future_get_properties_res = launch_and_get_properties.options(
-            **self.resources,
+        future_get_properties_res = _ray_actor_factory(
+            launch_and_get_properties, self.resources
         ).remote(self.client_fn, self.cid, ins)
         try:
             res = ray.get(future_get_properties_res, timeout=timeout)  # type: ignore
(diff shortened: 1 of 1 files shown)


jafermarq:
hi @noah822 , thank you for looking into this issue and for proposing a solution that users can very easily make use of.

We are aware that some users have been experiencing problems when training on GPU. As you point out the direct fix to this issue is to use `max_calls=1` in the `@ray.remote` decorator of Ray Tasks. However, we have been advising users to, instead of doing this, to downgrade to a previous version of Ray that didn't result in this problem. After Ray `1.11.1` some (I presume low level) behaviour in how Ray Tasks are handled once they are completed changed. Up until that version (and at least since `1.4.1`) when a Ray Task was completed, another will be executed by "re-using" some of the GPU context (here i'im using this term very loosely) resulting in consecutive tasks (not concurrent) to inmediately do work (i.e. without apparent delay due to `nvidia runtime` initialisation etc) -- at least this is what I can deduce from my observations.

For example, if you run `examples/simulation_pytorch/main.py` with Ray `1.11.1` you'll notice it runs perfectly fine when you set clients to use gpu. With the fix you propose, newer versions of Ray (including the latest `2.5.1`) also runs without OOM but you'll notice it takes much much longer. This is due to the overheads of initialising Tasks making use of the GPU instead of "sharing some parts" of such initialisation (as it was the case before `1.12.0`). 

I have just added a draft PR #1969 switching from Tasks to Actors. I'm very curious to hear what you think. It has some limitations has i mention towards the end of my (long) PR message.

Issue #1942, opened 2023-06-17
GPU OOM problem. GPU memory not cleared after each of the client finishes training even after one simulation round ends


noah822:
### What is your question?

I am running my flower instance with 10 clients, models of which are written in PyTorch and one CUDA card with round 24GB RAM.
When monitoring my dashboard,  I found that the memory used by each of client can not be properly released, double-checked by ``nvidia-smi``. The framework will terminate after running for a while because GPU hits its memory budget. There are two major problems

- In one simulation round, the memory used by each of the client cannot be released after they finish their local training. For my setting, each client roughly occupies 2G RAM. I've tried to tune the ``client_resources`` passed to ``start_simulation`` function to see what happens if I fit each of the client sequentially into my GPU, specifically I set ``{"num_gpus : 1"}``. It turns out that when the second client starts training, RAM used by the first will not be released. Therefore, the memory consumption scales linearly as the number of clients involved increases.  I notice that there is an intentional design enforced by the ``ray`` backend, at least in ``ray==2.5.0``, which set ``max_calls=1`` by default when using ``@ray.remote`` with GPU. It intentionally disables actor reuse for GPU memory release purpose, so that GPU resources can be released after the task process finishes. For more detail, please refer to [ray's document](https://docs.ray.io/en/latest/ray-core/tasks/using-ray-with-gpus.html). It should be the case that when ``launch_and_fit`` finishes, GPU memory gets cleared. But I have no idea why it does not work on my machine.
 
- Another even more weird problem is, after one simulation round ends, the RAM used can not be collected. Therefore, my framework will be forced to terminate because GPU OOM after running a couple of rounds.

noah822:
I am using ``python==3.8.10`` ,``flwr==1.4.0``, ``ray==2.5.0``
<img width="1299" alt="launch_and_fit" src="https://github.com/adap/flower/assets/103807711/970bcd1c-8932-4499-be65-0df6f9a83801">
On my ray dashboard, I noticed that there is no GPU resource/memory allocated to ``ray::launch_and_fit`` actor, which is launched by flwr, though I have passed ``client_resources={num_gpus:1}`` to ``start_simulation``. I am kind of new to ray. I would expect that ``ray::launch_and_fit`` holds the resources specified in ``client_resources`` parameter, then after client finishes its local training, this process can properly release the resources it previously occupies.

noah822:
[...]

Issue #1383, opened 2022-08-23
How can I use one GPU to simulate over 100 concurrent clients and total 1000 clients in the pool? 


wizard1203:
### What is your question?

The flower paper says that the Virtual Client Engine can 

> enables large-scale single-machine or multi- machine experiments by executing workloads in a resource-aware fashion.

> It creates a ClientProxy for each client, but defers instantiation of the actual client object (including local model and data) until the resources to execute the client-side task (training, evaluation) become available.

Then I guess that flower can automatically schedule the simulation of different clients based on the resources. That means we can increase the number of clients largely, even infinite number of clients.

I tried 10 clients on one GPU, using the example https://github.com/adap/flower/blob/main/examples/advanced_pytorch/run.sh . But it cannot successfully run, with a CUDA out-of-memory error (16 GB memory). Could you please give example codes to implement this?  And even more clients than 10?

mofanv:
I had the same problem `CUDA OOM`, but was running another script `simulation_pytorch`.
Can you try monitor how the OOM happened with a command like,
```
watch -n0.1 nvidia-smi
```
 to see if there are many `Ray::IDLE` processes occupying your GPU memory?

wizard1203:
> I had the same problem `CUDA OOM`, but was running another script `simulation_pytorch`. Can you try monitor how the OOM happened with a command like,
> 
> ```
> watch -n0.1 nvidia-smi
> ```
> 
> to see if there are many `Ray::IDLE` processes occupying your GPU memory?

@mofanv Hi, I tried the script `simulation_pytorch`, it can successfully simulate 100 clients out of 1000 clients via one GPU, without OOM error. I'm trying to use more GPUs and more machines, do you know how to implement this? Like 2 machines with 8 GPUs, and 1000 concurrent clients out of 10000 clients (pool size).

wizard1203:
It seems that indicating the num_gpus for each client can accelerate simulating, making fully utility of multiple GPUs?https://github.com/adap/flower/blob/aa9b9467fbb7b25d321ae0208434adc71e28bff2/baselines/flwr_baselines/publications/adaptive_federated_optimization/main.py#L32

wizard1203:
@mofanv It seems that when I define
```
  client_resources = {
      "num_cpus": args.num_client_cpus,
      "num_gpus": 8,
  } 
```
The GPU memory will continuously increase. And there are many ray::IDLE process. 
If when I define 
```
  client_resources = {
[...]

PR #1969, opened 2023-06-25
Introduce new VCE with ActorPool

jafermarq:
This is a draft implementation of the VCE making use of ActorPools as a replacement of using Tasks. 

## Context
The reasons why we want to move away from using Ray Tasks is because _something_ changed after ray `1.11.1` that made the _context_ (here i'm using this term very loosely) not being shared between tasks, which affects negatively those that make use of GPU resources quite evidently. As observed by several users #1942 #1376 #1383  this leads to GPU OOM.

A working solution to the above is using `max_calls=1` in the `@ray.remote` decorator throughout the `RayClientProxy`, however, this comes with big overheads when the Ray Tasks (in our case executing a method in a flower client) can be completed in a few seconds. Currently this is the most typical situation in FL: clients don't not hold large volumes of data that requires several minutes of training. In this scenario, using `max_calls=1` results in several seconds of _task initialisation_ that represents a significant portion of the total FL simulation time. This gets worse the more Tasks are spawned, i.e., when there are more clients being simulated. 

To mitigate this, the advice has been to downgrade to ray `1.11.1`. However, such version of Ray is **directly** incompatible with the latest version of Flower due each requiring conflicting `grpcio` versions. This can be fixed by building the wheels for Ray `1.11.1` after bumping the version of `grpcio`. You can find this in my [Ray fork](https://github.com/jafermarq/ray/tree/fl) (just execute the docker command indicated in `python/README-building-wheels.md` and you'll get the wheels). This makes that older version of Ray compatible with the current Flower version and there is no OOM issue (and yes, no need to set up `max_calls`)

## Using ActorPool

The approach implemented in this PR to solve the above issue, instantiates an [ActorPool](https://docs.ray.io/en/latest/ray-core/api/doc/ray.util.ActorPool.html#ray-util-actorpool) upon `start_simulation()` is executed. The pool contains as many actors (see [Actor](https://docs.ray.io/en/latest/ray-core/actors.html)) as can be fit in the system given the `client_resources` that can be set by the user (the same behaviour as with the previous implementation using Tasks is preserved). Now, each virtual `ClientProxy` (for which i have drafted a new one I call `RayClientProxyForActorPool`) instead of launching a Task, they submit a job to the actor pool. Each job is executed by an idle actor. 

### How can I test this?

```bash
git checkout VCE_with_actorpool
git pull
pip install -e .
# set `num_gpus` in the `client_resources` dictionary
# ensure you pass `is_simulation=True` to the ServerConfig when calling `start_simulation`
python examples/simulation_pytorch/main.py
```

How does this perform compared to the other two alternatives when we set `num_gpus=0.2` (i.e. 5 clients can run concurrently on GPU) ?

*    Using Tasks w/ `ray==1.11.1` --> ~22s
*    Using ActorPool w/ `ray==2.5.1` (i.e. the code in this PR) --> ~22s
*    Using Tasks and `max_calls=1` w/ `ray==2.5.1` --> ~110s

The above is tested on a machine with AMD 5900X and NVIDIA 3090Ti. 

The difference was even more dramatic is setting `num_gpus=0.5`: 33s for ActorPool  vs 197s for Task with `max_calls=1`.

As expected, using `max_calls` for these workloads that take just a couple of seconds is harmful in terms of wall clock time. Given the change in the (low level?) handling of Ray Tasks post `1.11.1`, I think the best alternative is to use entities that persist throughout the FL experiment (therefore not needing to release GPU/CPU resources): Ray Actors.

### Limitations
*   It is fine to submit one job from each `ClientProxy` but if we also want each proxy to return a result upon method completion (i.e. when calling `fit()`), things get a bit tricky if we want these to be in order. The current implementation gets *a* result (which might not be the one corresponding to the client). This can be mitigated easily by using `cid` as identifiers of job submission/retrieval. This is the reason why i'm passing the `is_simulation` flag around.
* I'm inclined to say that although the current implementation works, it would be better to submit all jobs in one go, using `map`, and then have a single blocking statement until all jobs are completed. Something like this in the `fit_clients()` and `evaluate_clients()` functions in `src/py/flwr/server/server.py`:

  ```python
      round_tasks = [client_fn for _ in range(N)] # let's add N tasks (N = clients)
      # let's submit the tasks to the pool
      gen = pool.map(lambda a, v : a.run_client.remote(v), round_tasks)
      round_results = list(gen)
  ```


pedropgusmao:
Nice. 
Do we know how Ray performs resource allocation for Actors? Is it done once per actor using `CUDA_VISIBLE_DEVICES` or does it change every time an actor is called?

I ask because I wonder if using actors will eventually allow specific GPUs to be used by specific actors, and prevent multiple actors from keeping memory in GPUs they are not using. This would probably solve the `max_calls=1` requirement, right?

jafermarq:
I'm unsure if `max_calls` can be used with actors... but actors can be removed from the pool if idle ([ray.util.ActorPoo.pop_idle](https://docs.ray.io/en/latest/ray-core/api/doc/ray.util.ActorPool.pop_idle.html#ray.util.ActorPool.pop_idle)). But again, we don't want anything related to `max_calls=1` if using it results in big penalties to run the workloads in the clients (which more often than not are lightweight enough that the cost of "re-init" the process is harmful in terms of overall wall-clock). If the GPU resources allocated for each actor correctly budget for the max VRAM a client would use, an actor would become IDLE only when no more work is supposed to be done in the round. 

If specific actors could be mapped to specific GPUs (i guess here you are thinking of the setting where not all GPUs are the same, specially in terms of VRAM) that would be nice. Maybe something [Placement Groups](https://docs.ray.io/en/latest/ray-core/scheduling/placement-group.html) can achieve?

noah822:
I personally think this implementation makes a lot of sense. Some comments on the **limitation**: the second note there might be a better practice as you've said. ``.map()`` api of actor will (probably) return an ordered collection of result, so that there is no need to track ``cid`` to associate result with the corresponding client.
I am not quite familiar with the mechanism of ray actor and how it schedules its resource usage. Can it eliminate the OOM problem, i.e release memory after a client finishes its local training? The performance in terms of the wall clock time compared with ray task is pretty impressive tho.

pedropgusmao:
Regarding the limitations:
- Could `fit()` be changed to return either its single result (for backward compatibility) or multiple results `Union[Tuple[Paramenters, int, Metrics], List[Tuple[Paramenters, int, Metrics]]`. This would allow Actors to receive a list of `cid`s and process them in sequence. I think this would also require changing `flwr.client.Client.fit() -> Union[FitRes,ListFitRes]`  the creation of a gRPC message:
```
message ListFitRes {
    repeated FitRes fit_res = 1;
  }
```
I used `List` in the example above to aggregate results, but maybe `Dict` would be even better, with the gRPC message being 
```
message ListFitRes {
    map<string, FitRes>fit_res = 1;
  }
```
Removing the requirement of `cid` being an integer. 

- Regarding the mapping, maybe this could be made available for the user to **choose** a specific mapping `cid` -> `Actor`. This would allow the experimentation of novel virtual client allocation strategies.

jafermarq:
@pedropgusmao , yeah ideally we keep the ClientProxy fully backwards compatible and under the same API as the standard client proxy dealing with gRPC devices. What I have done is use a "buffer" where results are cached if it doesn't belong to a particular `ClientProxy`. When this happens, the client proxy waits until its results is added to the cache. All clients will retrieve one result from the actor pool so eventually all `ClientProxies` will be paired with their respective `FitRes` (or other type depending on the task)

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -115,6 +117,133 @@ def reconnect(
         return common.DisconnectRes(reason="")  # Nothing to do here (yet)
 
 
+class RayClientProxyForActorPool(ClientProxy):
```
danieljanes, line 120: How about:
```suggestion
class RayActorClientProxy(ClientProxy):
```
We could also rename the legacy one to `RayTaskClientProxy`.

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -115,6 +117,133 @@ def reconnect(
...
+    """Flower client proxy which delegates work using Ray."""
+
+    def __init__(self, client_fn: ClientFn, cid: str, actor_pool: VirtualClientEngineActor, cache: Dict):
```
danieljanes, line 123:
```suggestion
    def __init__(self, client_fn: ClientFn, cid: str, actor_pool: ActorPool, cache: Dict):
```

danieljanes (changes requested):
A few initial comments

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -115,6 +116,117 @@ def reconnect(
...
+            )
+            res = self.actor_pool.get_client_result(self.cid, timeout)
+
+        except Exception as ex:
+            if self.actor_pool.num_actors == 0:
+                # At this point we want to stop the simulation.
+                # since no more client workloads will be executed
+                log(ERROR, "ActorPool is empty!!! Disconnecting VirtualClient")
+                # TODO: the below does nothing?
+                self.reconnect()
```
danieljanes, line 142: Is it possible that more actors are available at a later point?

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -115,6 +116,117 @@ def reconnect(
...
+    """Flower client proxy which delegates work using Ray."""
+
+    def __init__(
+        self, client_fn: ClientFn, cid: str, actor_pool: VirtualClientEngineActorPool
+    ):
+        super().__init__(cid)
+        self.client_fn = client_fn
+        self.actor_pool = actor_pool
+
+    def _submit_job(self, job: Callable, timeout: Optional[float]):
```
danieljanes, line 129: How about this naming for consistency?
```suggestion
    def _submit_job(self, job_fn: Callable, timeout: Optional[float]):
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+        if self._check_and_remove_actor_from_pool(actor):
+            future = fn(actor, value)
+            future_key = tuple(future) if isinstance(future, List) else future
+            self._future_to_actor[future_key] = (self._next_task_index, actor, cid)
+            self._next_task_index += 1
+
+            # update with future
+            self._cid_to_future[cid]["future"] = future_key
+
+    def submit_client_job(self, fn: Any, value: Callable, cid: str) -> None:
```
danieljanes, line 137: Should this be the following?
```suggestion
    def submit_client_job(self, fn: Any, job_fn: Callable, cid: str) -> None:
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+        # --> add more actors to the pool if so
+
+    def __reduce__(self):
+        """Make this class serialisable (needed due to lock)."""
+        return VirtualClientEngineActorPool, (
+            self.client_resources,
+            self.actor_max_restarts,
+        )
+
+    def submit(self, fn: Any, value: Callable, cid: str) -> None:
```
danieljanes, line 125:
```suggestion
    def submit(self, fn: Any, job_fn: Callable, cid: str) -> None:
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+@ray.remote
+class VirtualClientEngineActor:
+    """A Ray Actor class that runs client workloads."""
+
+    def terminate(self):
+        """Terminate Actor."""
+        log(WARNING, f"Manually terminating {self.__class__.__name__}")
+        ray.actor.exit_actor()
+
+    def run(self, client_fn: Callable, client_id):
```
danieljanes, line 45: `client_fn` might be confused with the `client_fn` that returns a `Client`/`NumPyClient`. I'd propose:
```suggestion
    def run(self, job_fn: Callable, cid: str):
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+        """Run a client workload."""
+        # execute tasks and return result
+        # return also cid which is needed to ensure results
+        # from the pool are correctly assigned to each ClientProxy
+        try:
+            client_results = client_fn()
```
danieljanes, line 51:
```suggestion
            job_results = job_fn()
```

danieljanes (changes requested):
Adding a few more comments

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+            )
+            raise ClientException(message) from ex
+
+        return client_id, client_results
+
+
+def pool_size_from_resources(client_resources: Dict):
+    """Calculate number of Actors that fit in pool given the resources in the.
+
+    cluster and those required per client.
```
danieljanes, line 68: TODO

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+        else:
+            num_actors = 0
+
+    if num_actors == 0:
+        log(
+            WARNING,
+            f"Your ActorPool is empty. Your system ({num_cpus = }, {num_gpus = }) "
+            "does not meet the criteria to host at least one client with resources:"
+            f" {client_resources}. Consider lowering your `client_resources`",
+        )
```
danieljanes, line 88: Should this be an error that stops the simulation?

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+        actors = [
+            VirtualClientEngineActor.options(
+                **client_resources, max_restarts=max_restarts
+            ).remote()
+            for _ in range(num_actors)
+        ]
+
+        super().__init__(actors)
+
+        self._cid_to_future = {}  # a dict
```
danieljanes, line 109: Should we add a type here?

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+        self.actor_to_remove: Set[str] = set()  # a set
+        self.num_actors = len(actors)
+
+        self.lock = threading.RLock()
+
+        # TODO: asyncio check every N seconds if cluster has grown
+        # --> add more actors to the pool if so
```
danieljanes, line 116: We could use such a mechanism also to replace the `max_restarts` setting. The cluster could also shrink, so monitoring the current size could be useful to either add or remove actors from the pool.

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,280 @@
...
+        """Make this class serialisable (needed due to lock)."""
+        return VirtualClientEngineActorPool, (
+            self.client_resources,
+            self.actor_max_restarts,
+        )
+
+    def submit(self, fn: Any, value: Callable, cid: str) -> None:
+        """Take idle actor and assign it a client workload."""
+        actor = self._idle_actors.pop()
+        if self._check_and_remove_actor_from_pool(actor):
```
danieljanes, line 128: What happens in the `else` case?

danieljanes (changes requested):
Adding another batch of comments

src/py/flwr/simulation/app.py
```diff
@@ -176,27 +178,75 @@ def start_simulation(  # pylint: disable=too-many-arguments
...
+    log(
+        INFO,
+        "Flower VCE: Resources for each Virtual Client: %s",
+        client_resources,
+    )
+
+    # instantiate ActorPool
+    # TODO: maybe we want `max_restarts` to be user-defined ?
+    max_restarts = 1  # how many times an actor that crashes should be restarted
+    # after these many restarts, it will be removed from the pool
```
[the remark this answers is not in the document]
jafermarq: for example in a multi-node simulation, if one node goes offline, actors need 1 restart to be assigned to a different (still running) node. But i'm also fine using `max_restarts=0` as default.

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
jafermarq: could be if the user adds a new node to Ray. this is a TODO (mentioned at the end of `VirtualClientEngineActorPool` constructor.

src/py/flwr/simulation/app.py
```diff
@@ -176,27 +178,75 @@ def start_simulation(  # pylint: disable=too-many-arguments
...
+    pool = VirtualClientEngineActorPool(client_resources, max_restarts)
+
+    log(
+        INFO,
+        "Flower VCE: Creating %s with %s actors",
+        pool.__class__.__name__,
+        pool.num_actors,
     )
 
-    # Register one RayClientProxy object for each client with the ClientManager
```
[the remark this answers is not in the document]
jafermarq: you are totally right. I'll add it back

src/py/flwr/simulation/app.py
```diff
@@ -142,6 +143,7 @@ def start_simulation(  # pylint: disable=too-many-arguments
         strategy=strategy,
         client_manager=client_manager,
     )
+    # Setting simulation ON for server
```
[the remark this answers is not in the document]
jafermarq: i don't know why i added that. removing...

src/py/flwr/simulation/ray_transport/ray_actor.py
jafermarq: ok

src/py/flwr/simulation/ray_transport/ray_actor.py
jafermarq: yes

src/py/flwr/simulation/ray_transport/ray_actor.py
jafermarq: yeah, the shrinking is already supported. An actor is added back to the pool after completing a job iif it fits back into the pool (see `process_unordered_future()` and in particular the logic surounding `_check_actor_fits_in_pool()`)

src/py/flwr/simulation/ray_transport/ray_actor.py
jafermarq: the actor is removed from the pool

src/py/flwr/simulation/app.py
```diff
@@ -176,27 +178,75 @@ def start_simulation(  # pylint: disable=too-many-arguments
...
+            "recover from it. Try launching your simulation with more generous "
+            f"`client_resources` setting (i.e. it seems {client_resources} is "
+            "not enough for your workload). Use fewer concurrent actors. "
+            "\n\t\t - You were running a multi-node simulation and all worker nodes "
+            "disconnected. The head node might still be alive but cannot accommodate "
+            f"any actor with resources: {client_resources}."
+            "\n\t\t - Your Actors crashed because of an unknown reason and all their "
+            f"restarts attempts ({max_restarts=}) have been exhausted.",
+        )
+        hist = None
```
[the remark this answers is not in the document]
jafermarq: which one is less relevant to the actors? (maybe i'm not understanding your comment)

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
jafermarq: yes. this has been added using a `threading.Timer` outside the actorpool. I think it's better to expand the pool from outside the object.

jafermarq: yes. this is now supported.

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+                "\n\tException triggered on the client side: " + client_trace,
+            )
+            raise ClientException(str(message)) from ex
+
+        return cid, job_results
+
+
+@ray.remote
+class DefaultActor(VirtualClientEngineActor):
+    """A Ray Actor class that runs client workloads."""
```
danieljanes, line 79: Do we need this, or could we just annotate `VirtualClientEngineActor` with `@ray.remote`, and not make it an `ABC`?

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def pool_size_from_resources(client_resources: Dict[str, Union[int, float]]) -> int:
+    """Calculate number of Actors that fit in the cluster.
+
+    For this we consider the resources available on each node and those required per
+    client.
+    """
+    total_num_actors = 0
+    # We calculate the number of actors that fit in a node per node basis. This is
```
danieljanes, line 89:
```suggestion
    total_num_actors = 0

    # We calculate the number of actors that fit in a node per node basis. This is
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # If a node has detached, it is still in the list of nodes
+        # however, its resources will be empty.
+        if not node_resources:
+            continue
+
+        num_cpus = node_resources["CPU"]
+        num_gpus = node_resources.get("GPU", 0)  # There might not be GPU
+        num_actors = int(num_cpus / client_resources["num_cpus"])
+        # If a GPU is present and client resources do require one
```
danieljanes, line 107:
```suggestion
        num_actors = int(num_cpus / client_resources["num_cpus"])

        # If a GPU is present and client resources do require one
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+            else:
+                num_actors = 0
+        total_num_actors += num_actors
+
+    if total_num_actors == 0:
+        log(
+            WARNING,
+            "Your ActorPool is empty. Your system (%s, %s) "
+            "does not meet the criteria to host at least one client with resources:"
+            " %s. Consider lowering your `client_resources`",
```
danieljanes, line 123:
```suggestion
            "The ActorPool is empty. The system (%s, %s) "
            "does not meet the criteria to host at least one client with resources:"
            " %s. Lowering the `client_resources` could help.",
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+        """
+        job_fn, cid = value
+        actor = self._idle_actors.pop()
+        if self._check_and_remove_actor_from_pool(actor):
+            future = fn(actor, job_fn)
+            future_key = tuple(future) if isinstance(future, List) else future
+            self._future_to_actor[future_key] = (self._next_task_index, actor, cid)
+            self._next_task_index += 1
+
+            # update with future
```
danieljanes, line 222:
```suggestion
            # Update with future
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+        create_actor_fn: Callable[[], Type[VirtualClientEngineActor]],
+        client_resources: Dict[str, Union[int, float]],
+        actor_list: Optional[List[Type[VirtualClientEngineActor]]] = None,
+    ):
+        self.client_resources = client_resources
+        self.create_actor_fn = create_actor_fn
+
+        if actor_list is None:
+            # Figure out how many actors can be created given the cluster resources
+            # and the resources the user indicates each VirtualClient will need.
```
danieljanes, line 169:
```suggestion
            # Figure out how many actors can be created given the cluster resources
            # and the resources the user indicates each VirtualClient will need
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def submit_client_job(
+        self, actor_fn: Any, job: Tuple[Callable[[], ClientRes], str]
+    ) -> None:
+        """Submit a job while tracking client ids."""
+        # We need to put this behind a lock since .submit() involves
+        # removing and adding elements from a dictionary. Which creates
+        # issues in multi-threaded settings
+
+        _, cid = job
```
danieljanes, line 233:
```suggestion
        _, cid = job

        # We need to put this behind a lock since .submit() involves
        # removing and adding elements from a dictionary. Which creates
        # issues in multi-threaded settings
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+        with self.lock:
+            # TODO: w/ timestamp check, call ray.resources()  # pylint: disable=fixme
```
danieljanes, line 235: Is this still relevant?

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+            # Creating cid to future mapping
```
danieljanes, line 236:
```suggestion
            # Create cid to future mapping
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+                WARNING,
+                "Cluster resources have changed. Number of actors in the pool should be"
+                " reduced from %s down to %s. This"
+                " might take several intermediate steps",
+                self.num_actors,
+                num_actors_updated,
+            )
+            # We are preventing one actor to be added back in the queue, so we just
+            # decrease the number of actors by one. Eventually `self.num_actors`
+            # should be equal what pool_size_from_resources(self.resources) returns
```
danieljanes, line 341:
```suggestion
            # We are preventing one actor from being added back to the queue, so we just
            # decrease the number of actors by one. Eventually `self.num_actors`
            # should be equal to what `pool_size_from_resources(self.resources)` returns
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+            self.num_actors -= 1
+            return False
+
+        return True
+
+    def process_unordered_future(self, timeout: Optional[float] = None) -> None:
+        """Similar to parent's get_next_unordered() but without final ray.get()."""
+        if not self.has_next():  # type: ignore
+            raise StopIteration("No more results to get")
+        # Block until one result is ready
```
danieljanes, line 351:
```suggestion

        # Block until one result is ready
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+        if res:
+            [future] = res
+        else:
+            raise TimeoutError("Timed out waiting for result")
+
+        with self.lock:
+            # Get actor that completed a job
+            _, actor, cid = self._future_to_actor.pop(future, (None, None, -1))
+            if actor is not None:
+                # Still space in queue ? (no if a node in the cluster died)
```
danieljanes, line 363:
```suggestion
                # Still space in queue? (no if a node in the cluster died)
```

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -0,0 +1,388 @@
...
+                    # and fetch its result
+                    self._flag_future_as_ready(cid)
+                else:
+                    # The actor doesn't fit in the pool anymore.
+                    # Manually terminate the actor
+                    actor.terminate.remote()
+
+    def get_client_result(self, cid: str, timeout: Optional[float]) -> ClientRes:
+        """Get result from VirtualClient with specific cid."""
+        # Loop until all jobs submitted to the pool are completed. Break early
```
danieljanes, line 378:
```suggestion

        # Loop until all jobs submitted to the pool are completed. Break early
```

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -30,8 +31,12 @@
 )
 from flwr.common.logger import log
 from flwr.server.client_proxy import ClientProxy
+from flwr.simulation.ray_transport.ray_actor import VirtualClientEngineActorPool
 
 ClientFn = Callable[[str], ClientLike]
+ClientRes = Union[
+    common.GetPropertiesRes, common.GetParametersRes, common.FitRes, common.EvaluateRes
+]
```
danieljanes, line 39: This is redundant with the definition in `ray_actor`. It should be defined in only one module, and then imported in the other.

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -115,6 +120,119 @@ def reconnect(
...
+            )
+            res = self.actor_pool.get_client_result(self.cid, timeout)
+
+        except Exception as ex:
+            if self.actor_pool.num_actors == 0:
+                # At this point we want to stop the simulation.
+                # since no more client workloads will be executed
+                log(ERROR, "ActorPool is empty!!!")
+                # TODO: Implement ClientProxy disconnect  # pylint: disable=fixme
+                # self.reconnect()
```
danieljanes, line 148: Is this still relevant?

danieljanes (changes requested):

src/py/flwr/simulation/ray_transport/ray_actor.py
jafermarq: It's not possible to inherit from classes with @ray.remote decorator.

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
jafermarq: good point. now fixed

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
jafermarq: removed

danieljanes (approved):
Lgtm, merging it now 🚀

PR #1982, opened 2023-06-30
Update mxnet examples dependency versions

danielnugraha:
### Description

There are some inconsistencies with the examples that uses mxnet as its dependencies, resulting in error when poetry tries to resolve the dependencies. Using mxnet with version 1.7.1 results in error.

### Related issues/PRs

See also #1980 and #1981.

### Explanation

Following changes in pyproject.toml and requirements.txt are made to the examples that uses mxnet as its dependencies:
1. Fix mxnet version to 1.6.0
2. Fix python version to >=3.8, <3.11.
3. ^ on other dependencies are removed.
4. Flwr local dependency path is changed to the latest flwr-nightly version.
5. Flwr version is fixed to >=1.0, <2.0.

### Checklist

- [x] Implement proposed change
- [x] Make CI checks pass

PR #1984, opened 2023-06-30
Update tensorflow examples dependency versions

danielnugraha:
### Description

There are some inconsistencies with the examples that uses tensorflow as its dependencies, resulting in error when poetry tries to resolve the dependencies, for example resolving tensorflow-privacy will result in error.

### Related issues/PRs

See also #1980, #1981, #1982.

### Explanation

Following changes in pyproject.toml and requirements.txt are made to the examples that uses tensorflow as its dependencies:
1. Fix tensorflow-cpu and -macos version to >=2.9.1, != 2.11.1
2. Fix python version to >=3.8, <3.11.
3. ^ on other dependencies are removed.
4. Flwr local dependency path is changed to the latest flwr-nightly version.
5. Flwr version is fixed to >=1.0, <2.0.

### Checklist

- [x] Make CI checks pass


examples/quickstart_tensorflow/pyproject.toml
```diff
@@ -1,4 +1,4 @@
-[build-system]
+p[build-system]
```
tanertopal, line 1: Is that a `p` there? I think that might be a typo

tanertopal: > Is that a `p` there? I think that might be a typo

examples/quickstart_tabnet/pyproject.toml
```diff
@@ -10,8 +10,8 @@ authors = ["The Flower Authors <hello@flower.dev>"]
...
-flwr = "^1.0.0"
-tensorflow-cpu = {version = "^2.9.1, !=2.11.1", markers="platform_machine == 'x86_64'"}
-tensorflow-macos = {version = "^2.9.1, !=2.11.1", markers="sys_platform == 'darwin' and platform_machine == 'arm64'"}
-tensorflow_datasets = "^4.8.3"
-tabnet = "^0.1.6"
+flwr = ">=1.0,<2.0"
+tensorflow-cpu = {version = ">=2.9.1,<2.11.1 || >2.11.1", markers = "platform_machine == \"x86_64\""}
+tensorflow-macos = {version = ">=2.9.1,<2.11.1 || >2.11.1", markers = "sys_platform == \"darwin\" and platform_machine == \"arm64\""}
+tensorflow_datasets = "4.8.3"
+tabnet = "0.1.6"
```
tanertopal, line 17:
```suggestion
tabnet = "0.1.6"

```

PR #1987, opened 2023-06-30
Add format and test scripts

adam-narozniak:
## Issue
The baseline created using the updated way doesn't have the format and test scripts. Additionally, they differ slightly - each has its own pyproject.toml
### Description
The test scripts include:
* structure (whether they follow the baseline_template structure)
* formatting (similar to the main flwr code)

### Related issues/PRs

NA

## Proposal
Create a script that checks a single baseline.

### Explanation

Baselines have to be checked individually since they have different pyproject.tomls. Additionally, check performed on-baseline-basis will help to run only the CI test for the baseline that was modified (all baseline checks, according to me is not feasible to singe the env setup for a single baseline is ~3 minutes, and we plan on having 50 baselines).

The test and format work as follows:
./dev/test-baseline.sh <baseline-name>

* Structure tests might be relaxed (in rare cases) by adding the directory to the `structure_exceptions=()` in test-baseline-structure.sh
* The unit tests are executed only when they are present.
```python -m pytest . || ([ $? -eq 5 ] || [ $? -eq 0 ])```
The 5 return status means: no test detected. The 0 return status means correctly executed. We accept both.

### CODE REVIEW

**Comment by jafermarq** [comment]
@adam-narozniak @danieljanes should we include this before we launch SoR? at the very least the formatting script?

**Comment by adam-narozniak** [comment]
@jafermarq I think we should add it. There will be a next PR that will automatically test only the changed baseline in the CI, but ppl can use that locally for testing and formatting.

**Comment by jafermarq** [comment]
@adam-narozniak, i feel like we should provide instructions for people to know how to run the `format-baseline.sh`? For example something like.

```bash
# the top level directory for baselines run:
./dev/format_baseline.sh <name-of-you-baseline>
```

Also, i think we need to add `ruff` to the pyproject toml in the baseline?

jafermarq (approved):

PR #1893, opened 2023-05-24
Add SecureAggregation field to Task message

panh99:
### Checklist

- [x] Implement proposed change
- [ ] Write tests
- [ ] Update [documentation](https://flower.dev/docs/writing-documentation.html)
- [ ] Update [changelog](https://github.com/adap/flower/blob/main/doc/source/changelog.rst)
- [ ] Make CI checks pass
- [ ] Ping maintainers on [Slack](https://flower.dev/join-slack/) (channel `#contributions`)

PR #2007, opened 2023-07-05
Adopt TaskIns/TaskRes-based communications on the client side

panh99:
### Description

Current ServerMessage/ClientMessage-based communication on the client side cannot support Secure Aggregation,  as SA-related info is stored in the Task message.

### Related issues/PRs
Add SecureAggregation field to Task message #1893
adopt TaskIns/TaskRes-based communications on the client side #2007
Update serde.py and validator.py for Secure Aggregation #2008

### Explanation
Change the `send()` and `receive()` in `connection.py` of `grpc_client`, `grpc_rere_client`, and `rest_client` to send TaskRes and receive TaskIns.

In `message_handler.py`, change the `handle()` function to handle TaskIns and return TaskRes, and rename original `handle()` function to `handle_legacy_message()`

Add utility functions `get_task_ins_from_pull_task_ins_response()`, `validate_task_ins()`, `get_server_message_from_task_ins()` and `wrap_client_message_in_task_res()` in `task_handler.py`.

Modify `message_handler_test.py` and `connection_test.py` correspondingly.

PS: I squashed _**14 commits**_ into one, cuz I was struggling with pylint false positives, saying that `no task in TaskRes`.

### Possibly Important Details
As far as I know, `ReconnectIns` is filtered out when using `grpc_rere_client` or `rest_client`. Hence, in this implementation, the `receive()` functions for `grpc_rere_client` and `rest_client` perform checks on `TaskIns` and discard it if it is `ReconnectIns`.

### Checklist

- [x] Write tests
- [x] Make CI checks pass


src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -172,17 +168,22 @@ def send(client_message_proto: ClientMessage) -> None:
...
+        task_res.MergeFrom(
+            TaskRes(
+                task_id="",  # This will be generated by the server
+                group_id=task_ins.group_id,
+                workload_id=task_ins.workload_id,
+            )
+        )
+
+        # fill `producer`, `consumer`, and `ancestry` in Task
+        task_res.task.MergeFrom(
```
tanertopal, line 181: The behaviour of `MergeFrom` is non-intuitive. I think we should add a comment here. I am also not sure, but maybe add a link to the docs.

Afaik fields will be overwritten, lists will be concatenated, and sub-messages will be recursively merged. I can imagine this to result in subtle and hard-to-debug bugs.

panh99: Hi @tanertopal ! Thanks so much for pointing that out. I've checked the API documentation and it is true that certain fields may not be overwritten when calling `MergeFrom`. 
I've pushed a new commit, adding `ClearField` before `MergeFrom` and some comments.

src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -172,17 +173,34 @@ def send(client_message_proto: ClientMessage) -> None:
...
-            group_id=task_ins.group_id,
-            workload_id=task_ins.workload_id,
-            task=Task(
+        # Clear fields to be set
+        task_res.ClearField("task_id")
+        task_res.ClearField("group_id")
+        task_res.ClearField("workload_id")
+        task_res.task.ClearField("producer")
+        task_res.task.ClearField("consumer")
+        task_res.task.ClearField("ancestry")
```
danieljanes, line 182: Why do these fields need to be cleared? `wrap_client_message_in_task_res` does not seem to set any one of these fields.

I assume that if any of these fields are set at this point in the code, it'd be due to an error/bug in a different part of the code. Is this understanding correct? If so, we could test if any of these fields are set here, and abort with an error if so.

src/py/flwr/client/message_handler/message_handler.py
```diff
@@ -24,15 +24,51 @@
...
+        reconnect later (False).
+    """
+    server_msg = get_server_message_from_task_ins(task_ins, exclude_reconnect_ins=False)
+    if server_msg is not None:
+        client_msg, sleep_duration, keep_going = handle_legacy_message(
+            client, server_msg
+        )
+        task_res = wrap_client_message_in_task_res(client_msg)
+        return task_res, sleep_duration, keep_going
+    raise NotImplementedError()
```
danieljanes, line 68: A small idea to decrease the amount of indentation:
```suggestion
    if server_msg is None:
        raise NotImplementedError()
    client_msg, sleep_duration, keep_going = handle_legacy_message(
        client, server_msg
    )
    task_res = wrap_client_message_in_task_res(client_msg)
    return task_res, sleep_duration, keep_going
```

src/py/flwr/client/message_handler/message_handler_test.py
```diff
@@ -14,6 +14,7 @@
 # ==============================================================================
 """Client-side message handler tests."""
 
+import uuid
```
danieljanes, line 17: We recently started to leave two empty lines between the module docstring and the first import:
```suggestion

import uuid
```

danieljanes (changes requested):

src/py/flwr/client/grpc_rere_client/connection.py
panh99: I agree with you, but it is a bit complicated. If we adopt TaskIns/TaskRes-based communication, `send()` will need to take `TaskIns` as input and then set some unset fields, e.g., `task_id`, `group_id`. 

`MergeFrom` is very convenient to fill unset fields after a message has been created. But it may cause some unforeseeable problems and make it hard to debug (as Taner pointed out), because `MergeFrom` treat different types of fields differently.
FYI,

> MergeFrom(other_msg)
> Merges the contents of the specified message into current message.
> 
> This method merges the contents of the specified message into the current message. Singular fields that are set in the specified message overwrite the corresponding fields in the current message. Repeated fields are appended. Singular sub-messages and groups are recursively merged.

So I clear the fields to be set before `MergeFrom()`. But I agree with you that we should throw an error if those fields have been set somewhere else cuz they should not be.

danieljanes (approved):
Lgtm!

PR #2063, opened 2023-07-07
Fix baseline creation on linux-base systems

adam-narozniak:
### Description
The create-baseline script doesn't fully work on Linux. It leads to `sed: can't read : No such file or directory` on linux - it doesn't change the name in pyproject.toml. 

### Related issues/PRs

NA

### Explanation

The sed works slightly differently on Mac and Linux. The fix causes the script to work differently on Linux and Mac.


jafermarq (approved):
amazing!

PR #1985, opened 2023-06-30
Update other examples dependency versions

danielnugraha:
### Description

There are some inconsistencies with the examples dependencies, resulting in error when poetry tries to resolve it. For example, quickstart_jax uses a yanked version of jax as its dependencies.

### Related issues/PRs

See also #1980, #1981, #1982, #1984.

### Explanation

Following changes in pyproject.toml and requirements.txt are made to the examples that uses mxnet as its dependencies:
1. Fix python version to >=3.8, <3.11.
2. ^ on other dependencies are removed.
3. Flwr local dependency path is changed to the latest flwr-nightly version.
4. Flwr version is fixed to >=1.0, <2.0.

### Checklist

- [x] Implement proposed change
- [x] Make CI checks pass

PR #2064, opened 2023-07-07
Check if requirements.txt is synced with pyproject.toml

danielnugraha:
### Description

The existing solution to sync requirements.txt and pyproject.toml is not stable so it needs to be changed. Also the dependencies are incompatible with each other.

### Related issues/PRs

See also #1985 #1981 #1982 #1984.

### Explanation

Following changes are made:
1. Make dependencies compatible with each other.
2. Replace version ^ with ==.
3. Use poeareq as a tool to generate pyproject.toml from requirements.txt.
4. Rewrite generate-requirements-txt.sh.
5. Remove pipreqs from the pyproject.toml
6. Add action to flower.yml (only when python version == 3.10.9)

### Checklist

- [x] Write tests
- [x] Make CI checks pass


.github/workflows/flower.yml
```diff
@@ -35,3 +35,9 @@ jobs:
         run: ./dev/check-protos.sh
       - name: Lint + Test (isort/black/docformatter/mypy/pylint/flake8/pytest)
         run: ./dev/test.sh
+      - name: Check if pyproject.toml and requirements.txt are in sync
+        run: |
+          python -m pip install poetry-add-requirements.txt
+          ./dev/check-requirements-txt.sh
+        if: ${{ matrix.python == '3.10.9' }}
```
tanertopal, line 42: We never do this, so I'd prefer not to start now. Just let it run in each version in the matrix.

danielnugraha: The tool requires python version >=3.10

tanertopal: Got it! Thank you.

tanertopal (dismissed):

PR #2065, opened 2023-07-07
Fix flake8 error E266 in baseline template

adam-narozniak:
### Description

Currently, the baseline template fails flake8 check:  `E266 too many leading '#' for block comment`

## Proposal
Remove double ## in main.py.
## Explanation
In order to introduce the flake8 tests the current code should meet its standards.


jafermarq (approved):
looks optimal.

PR #2074, opened 2023-07-10
Add simulation to E2E tests

charlesbvll:
### Description

Only edge client is tested during E2E tests.

### Related issues/PRs

First these PRs need to be merged: 

#2073 
#2072 
#2071 
#2070 
#2069 
#2068 
#2067 

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


tanertopal (approved):
lgtm! Merging it, but one thing to note - which we are testing at other places already is that we are now installing the simulation extras even in the non-simulation tests, and therefore the non-simulation tests will have theoretically access to Ray.

PR #2080, opened 2023-07-11
Exclude baselines/doc from tests in CI

adam-narozniak:
### Description

The new directory `doc` in `baselines` makes the pipeline fail.
### Related PRs
https://github.com/adap/flower/pull/2060 (it makes this change fail the tests)

## Proposal
Exclude the `baseline/doc` from checks.

### Explanation
The check didn't take into account the possibility of new directories, therefore it's detected as a change and then the tests fail. This solves it by excluding this directory therefore no changes should be detected.

### DIFF
--- a/.github/workflows/baselines.yml
+++ b/.github/workflows/baselines.yml
@@ -54,7 +54,13 @@ jobs:
             # Pull request event triggered for any commit to a pull request
             change_references="main..HEAD"
           fi
-          dirs=$(git diff --dirstat=files,0 ${change_references} . | awk '{print $2}' | grep -E '^baselines/[^/]*/$' | grep -v -e '^baselines/dev' -e '^baselines/baseline_template' -e '^baselines/flwr_baselines' -e '^baselines/doc' | sed 's/^baselines\///')
+          dirs=$(git diff --dirstat=files,0 ${change_references} . | awk '{print $2}' | grep -E '^baselines/[^/]*/$' | \
+          grep -v \
+            -e '^baselines/dev' \
+            -e '^baselines/baseline_template' \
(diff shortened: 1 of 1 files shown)


.github/workflows/baselines.yml
```diff
@@ -54,7 +54,7 @@ jobs:
             # Pull request event triggered for any commit to a pull request
             change_references="main..HEAD"
           fi
-          dirs=$(git diff --dirstat=files,0 ${change_references} . | awk '{print $2}' | grep -E '^baselines/[^/]*/$' | grep -v -e '^baselines/dev' -e '^baselines/baseline_template' -e '^baselines/flwr_baselines' | sed 's/^baselines\///')
+          dirs=$(git diff --dirstat=files,0 ${change_references} . | awk '{print $2}' | grep -E '^baselines/[^/]*/$' | grep -v -e '^baselines/dev' -e '^baselines/baseline_template' -e '^baselines/flwr_baselines' -e '^baselines/doc' | sed 's/^baselines\///')
```
tanertopal, line 57:
```suggestion
          dirs=$(git diff --dirstat=files,0 ${change_references} . | awk '{print $2}' | grep -E '^baselines/[^/]*/$' | \
          grep -v \
            -e '^baselines/dev' \
            -e '^baselines/baseline_template' \
            -e '^baselines/flwr_baselines' \
            -e '^baselines/doc' \
          | sed 's/^baselines\///')
```

@adam-narozniak would this also work to make it more readable?

adam-narozniak: @tanertopal yes, I check it also on my repo for tests too

tanertopal (approved):

PR #2083, opened 2023-07-11
Optimize CI file structure by moving redundant steps into a common action

tanertopal:
### Description

The CI workflows have a lot of redundant definitions, which make updating error prone and hard to debug in case something goes wrong. This PR adds an action which merges the repetitive preparation steps into one and will make adding caching actions down the line easier.

### Checklist

- [x] Make CI checks pass


.github/actions/bootstrap/action.yml
```diff
@@ -0,0 +1,29 @@
+name: "bootstrap"
+description: "Bootstrap Python environment including Python dependencies and configure optimizations."
+inputs:
+  python-version:
+    description: "Version range or exact version of Python or PyPy to use, using SemVer's version range syntax."
```
danieljanes, line 5: Why PyPy?

.github/actions/bootstrap/action.yml
```diff
@@ -0,0 +1,29 @@
+name: "bootstrap"
+description: "Bootstrap Python environment including Python dependencies and configure optimizations."
```
danieljanes, line 2:
```suggestion
description: "Bootstrap Python environment (install and configure Python version and build tools)"
```

.github/workflows/e2e.yml
```diff
@@ -17,16 +17,8 @@ jobs:
...
-        with:
-          python-version: 3.7.15
-      - name: Install
-        run: |
-          python -m pip install -U pip==22.3.1
-          python -m pip install -U setuptools==65.6.3
-          python -m pip install poetry==1.3.2
-          poetry config virtualenvs.create false
+      - name: Bootstrap
+        uses: ./.github/actions/bootstrap
```
danieljanes, line 21: This implicitly changes Python from 3.7 to 3.8

.github/workflows/e2e.yml
```diff
@@ -258,16 +202,8 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.8.17
```
danieljanes, line 264: Why is the default for the new bootstrap action 3.8.16 if .17 is available?

danieljanes (changes requested):

.github/actions/bootstrap/action.yml
tanertopal: @danieljanes this is just passed through to the office action and the description is a copy.

.github/workflows/e2e.yml
tanertopal: I'll update it to .17. 

As a sidenote in the improve_ci_speed PR there will be some more changes and explanations on that.

.github/workflows/e2e.yml
tanertopal: I'll update it to .17. 

As a sidenote in the improve_ci_speed PR there will be some more changes and explanations on that.

.github/workflows/e2e.yml
danieljanes: It's often good to separate refactoring changes ("moving stuff around") and actual changes (changing how something behaves) into separate PR. A refactoring should not change any behaviour, which also makes it easier to verify that the refactoring is correct. The actual change (in this case: changing the version of those CI jobs) is trivial later on and should also be done in isolation. If we find an issue later on, this makes it easier to identify whether the issue has been introduced by the refactoring or by the version change.

TL;DR: let's stick with 3.7 in this PR

tanertopal: Not sure I agree in this case as we are modifying the CI jobs but I'll revert the changes so each uses its own previous Python version and open PR's afterwards for each job.

danieljanes: If we don't change the Python version, the DRY part of this PR is a typical refactoring. It just "moves code around", but the CI behaviour should not change. In fact, if it changes, it's a bug. The only small exception is the Pandoc installation because there the installation order changes.

tanertopal: @danieljanes, all functional changes are reverted now.

danieljanes (approved):

PR #2130, opened 2023-07-15
Adding Logging Page to Docs

jafermarq:
This PR introduces a new page in the documentation presenting a basic overview of the Flower logger. It demonstrates:
- how to output the log messages to a file (in addition to the terminal)
- Shows how to add new messages to the logger.
- It also hints at the possibility of using the HTTPHandler if there is a remote logging listener.


tanertopal (commented):
@danieljanes this lgtm. wdyt?

tanertopal:
@jafermarq @danieljanes merging it. We can make improvements if Daniels sees something later.

PR #2138, opened 2023-07-18
Rename examples using dash instead of underscore

charlesbvll:
### Description

We use underscore for the naming of the examples which is bad for URLs.

### Related issues/PRs

#2133 depends on this.

### Explanation

Use dashes instead of underscores.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


danieljanes (changes requested):
The renaming itself looks good. We also need to change the references to those examples in READMEs and the docs.

tanertopal (approved):

PR #2144, opened 2023-07-19
Fix PyTorch Quickstart Example

jafermarq:
The version of `torchvision` in pytorch quickstart is not compatible with the Pytorch version set in `pyproject.toml`. Now it's fixed. The change is also applied to `requirements.txt`.


jafermarq:
I presume this is needed for all other examples after #1993

PR #2176, opened 2023-07-30
Documentation part 1 of DP in FL with Flower

ElnathanTiokou:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/doc/source/_templates/sidebar/versioning.html
+++ b/doc/source/_templates/sidebar/versioning.html
@@ -1,0 +1,76 @@
+{% if versions %}
+
+<style>
+input[type='checkbox'] {
+  display: none;
+}
+
+.lbl-toggle {
+  display: block;
+
+  padding: 1rem;
+
+  cursor: pointer;
+
+  border-radius: 7px;
+  transition: all 0.25s ease-out;
+}
+
+.lbl-toggle::before {
+  content: ' ';
+  display: inline-block;
+
+  border-top: 5px solid transparent;
+  border-bottom: 5px solid transparent;
+  border-left: 5px solid currentColor;
+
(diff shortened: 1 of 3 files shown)


panh99:
Hi @ElnathanTiokou ! Thanks for your great work. The tutorial is well-written and well-organized. 
Just some minor concerns. First, I would suggest that the DP tutorial should focus on the DP topic and elaborate on how to use what's already there in Flower. Discussing Secure Aggregation (SA) and the combination of two techniques may might go beyond the intended scope. Besides, I believe that the GIF used to illustrate SA + DP in this tutorial originates from [Google Research](https://ai.googleblog.com/2023/03/distributed-differential-privacy-for.html). I don't think we can use that GIF for legal reasons.

ElnathanTiokou:
Thank you for the feedback @panh99 I actually notified @danieljanes about the visuals, all of them need to be customized for Flower with our Color. 
Thank you I take note on elaborating on how to use what's already there in Flower. ( This is actually what I'm working on right now. The step by steps guide with the Flower DP functions. 
This first document was more of an explanatory file around the whole concept of DP in FL and a little bit beyond (An overview). Thank you for the remarks.

panh99:
> 

Hi @ElnathanTiokou ! We would recommend avoiding discussing the combination of SA and DP within this tutorial. It would be better to separate the SA + DP thing into a different tutorial. Besides, just to clarify, you mean that the GIF from Google Research will be replaced by other usable images, right?

ElnathanTiokou:
Well noted. I'll stick to DP. And yes the GIF was supposed to be replaced. But since SA shouldn't figure in, I'll just take it off.

PR #2251, opened 2023-08-24
Add CreateWorkload

panh99:
Add `CreateWorkload` Driver API that requests and returns a `workload_id`

Update `in_memory_state.py` and `sqlite3_state.py`

Add validation processes to check the validity of `workload_id`


examples/mt-pytorch/driver.py
```diff
@@ -54,8 +54,14 @@ def weighted_average(metrics: List[Tuple[int, Metrics]]) -> Metrics:
 
 # -------------------------------------------------------------------------- Driver SDK
 driver.connect()
+create_workload_res: driver_pb2.CreateWorkloadResponse = driver.create_workload(
+    req=driver_pb2.CreateWorkloadRequest()
+)
 # -------------------------------------------------------------------------- Driver SDK
 
+workload_id = create_workload_res.workload_id
+print(f"Get workload id {workload_id}")
```
danieljanes, line 63:
```suggestion
print(f"Created workload id {workload_id}")
```

src/py/flwr/server/state/in_memory_state.py
```diff
@@ -186,3 +193,10 @@ def unregister_node(self, node_id: int) -> None:
     def get_nodes(self) -> Set[int]:
         """Return all available client nodes."""
         return self.node_ids
+
+    def create_workload(self) -> str:
+        """Create one workload."""
+        # Create, store, and return workload ID
+        workload_id = str(random.randrange(9223372036854775808))
+        self.workload_ids.add(workload_id)
```
danieljanes, line 201: Maybe validate that the workload ID is not in the set?

src/py/flwr/server/state/state_test.py
```diff
@@ -341,7 +385,10 @@ def test_num_task_res(self) -> None:
 
 
 def create_task_ins(
-    consumer_node_id: int, anonymous: bool, delivered_at: str = ""
+    consumer_node_id: int,
+    anonymous: bool,
+    delivered_at: str = "",
+    workload_id: str = "",
```
danieljanes, line 391: It might be safer to remove the defaults here and force callers that relied on the default to manually pass `""`.

danieljanes (changes requested):
The PR is in great shape, I only added two minor comments.

src/py/flwr/server/state/in_memory_state.py
panh99: sure.

src/py/flwr/driver/driver_client_manager.py
```diff
@@ -35,6 +35,9 @@ class DriverClientManager(ClientManager):
 
     def __init__(self, driver: Driver) -> None:
         self.driver = driver
+        self.workload_id = driver.create_workload(
+            driver_pb2.CreateWorkloadRequest()
+        ).workload_id
```
danieljanes, line 40: Let's do this lazily when one of the other methods is called

src/py/flwr/server/state/in_memory_state.py
```diff
@@ -183,6 +190,21 @@ def unregister_node(self, node_id: int) -> None:
...
         """Return all available client nodes."""
+        if workload_id not in self.workload_ids:
+            log(ERROR, "`workload_id` is invalid")
+            return set()
         return self.node_ids
+
+    def create_workload(self) -> str:
+        """Create one workload."""
+        # Create, store, and return workload ID
+        for _ in range(100):
```
danieljanes, line 203: Let's remove this. We can just try it once, and raise an error if the workload_id already exists.

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -457,13 +471,35 @@ def unregister_node(self, node_id: int) -> None:
...
         return result
 
+    def create_workload(self) -> str:
+        """Create one workload and store it in state."""
+        for _ in range(100):
+            # String representation of random integer from 0 to 9223372036854775807
+            workload_id = str(int.from_bytes(os.urandom(8), "little") >> 1)
+            # Check conflicts
+            query = "SELECT COUNT(*) FROM workload WHERE workload_id = ?;"
+            # If workload_is does not exist
```
danieljanes, line 495:
```suggestion
            # If workload_id does not exist
```

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -457,13 +471,35 @@ def unregister_node(self, node_id: int) -> None:
...
         self.query(query, {"node_id": node_id})
 
-    def get_nodes(self) -> Set[int]:
+    def get_nodes(self, workload_id: str) -> Set[int]:
         """Retrieve all currently stored node IDs as a set."""
+        # Validate workload ID
+        query = "SELECT COUNT(*) FROM workload WHERE workload_id = ?;"
+        if self.query(query, (workload_id,))[0]["COUNT(*)"] == 0:
+            log(ERROR, "`workload_id` is invalid")
+            return set()
```
tanertopal, line 480: What's the purpose of passing the `workload_id` into the function?

Additionally, if the goal of the function is to return all existing `node_ids` it should not expose a method to probe for valid workload_ids. That would be a security risk.

Was the goal of this change to get all `node_ids` which are related to a certain `workload_id`?

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -36,6 +37,12 @@
 );
 """
 
+SQL_CREATE_TABLE_WORKLOAD = """
+CREATE TABLE IF NOT EXISTS workload(
+    workload_id TEXT UNIQUE
```
tanertopal, line 42: `SQL_CREATE_TABLE_TASK_INS` should be modified to make workload_id a foreign key to check the constraint automatically.

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -177,6 +185,12 @@ def store_task_ins(self, task_ins: TaskIns) -> Optional[UUID]:
             log(ERROR, errors)
             return None
 
+        # Validate workload ID
+        query = "SELECT COUNT(*) FROM workload WHERE workload_id = ?;"
+        if self.query(query, (task_ins.workload_id,))[0]["COUNT(*)"] == 0:
+            log(ERROR, "`workload_id` is invalid")
+            return None
```
tanertopal, line 192: This will not be required if `workload_id` is a foreign key in `task_ins`

src/py/flwr/server/state/sqlite_state.py
```diff
@@ -457,13 +471,35 @@ def unregister_node(self, node_id: int) -> None:
...
+            log(ERROR, "`workload_id` is invalid")
+            return set()
+
+        # Get nodes
         query = "SELECT * FROM node;"
         rows = self.query(query)
         result: Set[int] = {row["node_id"] for row in rows}
         return result
 
+    def create_workload(self) -> str:
```
tanertopal, line 488: We don't let the state create the `node_id` and handle that responsibility outside of the state module. Should we do the same here? I feel like it should be unified.

src/py/flwr/server/state/sqlite_state.py
panh99: Yes. Including `workload_id` in the `GetNodesRequest` and validating it is suggested by @danieljanes to prevent access from invalid workload and also to return different lists of nodes based on `workload_id` in the future.

Btw, I am a bit confused about the security risk you just mentioned. Would you mind elaborating on why workload validation can possibly be a security risk when the aim of the function `get_nodes()` is to return all existing nodes?

src/py/flwr/server/state/sqlite_state.py
panh99: sure, makes a lot of sense. will change this.

src/py/flwr/server/state/sqlite_state.py
panh99: I am thinking of the same issue. We can also pick a random integer as workload_id in the DriverAPI and include that in the `CreateWorkloadRequest`. @danieljanes which option would you prefer, sample `workload_id` on the driver or on the server?

src/py/flwr/driver/driver_client_manager.py
panh99: sure

src/py/flwr/server/state/state.py
```diff
@@ -134,5 +134,9 @@ def unregister_node(self, node_id: int) -> None:
         """Remove `node_id` from state."""
 
     @abc.abstractmethod
-    def get_nodes(self) -> Set[int]:
+    def get_nodes(self, workload_id: str) -> Set[int]:
         """Retrieve all currently stored node IDs as a set."""
```
tanertopal, line 138: @panh99 can you similar to how it's done in `get_task_res` and some others write a spec here? For example, we could mention as a constraint:

```
If the provided `workload_id` does not exist or has no matching nodes, an empty `Set` MUST be returned.
```

src/py/flwr/server/state/sqlite_state.py
tanertopal: I am unsure if I misread yesterday, but I thought we should not allow probing through the call to `get_nodes` if a `workload_id` exists. The return value in both of the following cases should be an empty `Set`:

Case 1: `workload_id` does not exist
Case 2: `workload_id` has no nodes

src/py/flwr/server/state/state.py
tanertopal: Additionally, we can define in the spec if the `workload_id` should be generated by the state or will be passed into the state.

src/py/flwr/server/state/state.py
```diff
@@ -134,5 +134,9 @@ def unregister_node(self, node_id: int) -> None:
...
+
+    @abc.abstractmethod
+    def create_workload(self) -> str:
+        """Create one workload."""
```
tanertopal, line 142: Similar to what I mentioned above, we should write a spec here that clearly states who is responsible for `workload_id` generation and what type the `workload_id` should have.

tanertopal: I would prefer IDs to be generated by the state and not outside of the state. I think for legacy reasons, in the case of the `node_id`, we generate it outside of the state.

src/py/flwr/server/state/sqlite_state.py
danieljanes: @panh99 @tanertopal let's use the same approach that we use for `node_id`

src/py/flwr/server/state/state.py
danieljanes: As discussed offline, we will move ID generation into the `State` in a subsequent PR (and so so for all IDs, not just the `workload_id`)

danieljanes (approved):

PR #2270, opened 2023-08-31
Add Spanish translation to docs

jafermarq:
My attempt at getting docs translated.


charlesbvll:
@jafermarq I added the Spanish flag to the website with the correct link and made a few necessary changes, I think this PR is ready now! (I will let you mark it as ready if you think so as well)

jafermarq:
Many thanks! @charlesbvll , I'll review the translations and update them before going ahead.

charlesbvll (approved):

PR #2291, opened 2023-09-05
Add new wheel building job and use it in E2E tests

charlesbvll:
### Description

Currently we have to test the wheels manually, which is slow and inefficient.

### Related issues/PRs

N/A

### Explanation

We could build the wheel and use it for the E2E tests.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


.github/workflows/e2e.yml
```diff
@@ -12,9 +12,36 @@ env:
...
+      - name: Upload wheel
+        id: upload
+        env:
+          AWS_DEFAULT_REGION: ${{ secrets. AWS_DEFAULT_REGION }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets. AWS_SECRET_ACCESS_KEY }}
+        run: |
+          cd ./dist
+          echo "WHL_PATH=$(ls *.whl)" >> "$GITHUB_OUTPUT"
+          aws s3 cp --content-disposition "attachment" --cache-control "no-cache" ./ s3://artifact.flower.dev/py/${{ github.event.number }}/${{ github.sha }} --recursive
```
tanertopal, line 37: Does `github.event.number` correspond to the GitHub PR number? What is the value when merging a PR into the `main` branch?

I am thinking about life cycle policies on S3, and it would be easier if we could just give it a top-level path like

* `s3://artifact.flower.dev/py/pr/....` to clean up after a while.

Instead of `github.event.number` we could also use `github.ref_name`, which would be just the branch name and should work the same in PR's and the `main` branch.

That being that it would be nice to have everything in main under:

* `s3://artifact.flower.dev/py/main/...`

So that each wheel built on the main branch can be found as:

* `s3://artifact.flower.dev/py/main/SHORT_SHA/flwr....whl`

Wdyt?

charlesbvll: I do agree that it's better to use the branch name instead of the PR number, it was the initial approach, but I thought it would make it more complicated for cleaning. I will make the change! 
I will also try to use the short SHA

tanertopal (approved):

PR #2293, opened 2023-09-05
Add in-place FedAvg

jafermarq:
A faster and more memory-efficient version of `FedAvg`.


src/py/flwr/server/strategy/fedavginplace.py
```diff
@@ -0,0 +1,92 @@
...
+            for x in parameters_to_ndarrays(results[0][1].parameters)
+        ]
+        for i, (_, fit_res) in enumerate(results[1:]):
+            res = (
+                scaling_factors[i + 1] * x
+                for x in parameters_to_ndarrays(fit_res.parameters)
+            )
+            params = [
+                reduce(np.add, layer_updates) for layer_updates in zip(params, res)
+            ]
```
charlesbvll, line 79: Should we move this to an `aggregate_inplace(results: List[Tuple[ClientProxy, FitRes]])` function sitting in `aggregate.py` ?

jafermarq: Yeah. But how we make it easy to use? adding a new `inplace` flag to the existing `FedAvg` that would call this `aggregate_inplace` instead of doing what currently `FedAvg` does in the first half of its `aggregate_fit()`?

charlesbvll: I personally like the idea of the `inplace` flag on the existing `FedAvg` that would call the `aggregate_inplace` function. 
One benefit would also be that we make the aggregation step a bit more explicit. I noticed quite a few people that where going through `FedAvg` and struggled to figure out that the aggregation happens in `aggregate.py`. I attribute it to the fact that `aggregate` is called inside `ndarrays_to_parameters` which doesn't jump to the eye. Whereas if we have a check for the aggregation method, it will be more visible.

jafermarq: I made the change. I also moved the post-aggregation conversion `ndarrays_to_parameters` in the existing implementation.

danieljanes (changes requested):

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -41,6 +42,31 @@ def aggregate(results: List[Tuple[NDArrays, int]]) -> NDArrays:
     return weights_prime
 
 
+def aggregate_inplace(results: List[Tuple[ClientProxy, FitRes]]) -> NDArrays:
```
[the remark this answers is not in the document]
jafermarq: I'll have to change then that in `aggregate`, since otherwise, passing `results` as `List[Tuple[NDArrays, int]]` to `aggregate_inplace` would require a copy of all updates sent by clients.

src/py/flwr/server/strategy/fedavg_test.py
```diff
@@ -120,3 +130,51 @@ def test_fedavg_num_evaluation_clients_minimum() -> None:
 
     # Assert
     assert expected == actual
+
+
+def test_inplace_aggregate_fit_equivalence() -> None:
+    """Test aggregate_fit equivalence between fedavg and its inplace version."""
+    # Prepare
```
danieljanes, line 137:
```suggestion
    """Test aggregate_fit equivalence between FedAvg and its inplace version."""
    # Prepare
```

src/py/flwr/server/strategy/fedavg_test.py
```diff
@@ -120,3 +130,51 @@ def test_fedavg_num_evaluation_clients_minimum() -> None:
...
+    fedavg_reference = FedAvg(inplace=False)
+    fedavg_inplace = FedAvg()
+
+    # Execute
+    reference, _ = fedavg_reference.aggregate_fit(1, results, failures)
+    assert reference
+    inplace, _ = fedavg_inplace.aggregate_fit(1, results, failures)
+    assert inplace
+
+    # convert to numpy to check similarity
```
danieljanes, line 174:
```suggestion
    # Convert to NumPy to check similarity
```

src/py/flwr/server/strategy/aggregate.py
```diff
@@ -41,6 +42,31 @@ def aggregate(results: List[Tuple[NDArrays, int]]) -> NDArrays:
...
+    # Count total examples
+    num_examples_total = sum([fit_res.num_examples for _, fit_res in results])
+
+    # Compute scaling factors for each result
+    scaling_factors = [
+        fit_res.num_examples / num_examples_total for _, fit_res in results
+    ]
+
+    # Let's do in-place aggregation
+    # get first result, then add up each other
```
danieljanes, line 56:
```suggestion
    # Get first result, then add up each other
```

danieljanes (approved):

PR #2314, opened 2023-09-07
Migrate PyTorch quickstart to Flower Datasets

adam-narozniak:
## Issue
Quickstart PyTroch example does not use partitioning (each client gets the same full dataset).

## Proposal
Use `flwr-datasets` to partition the data into 3 iid partitions.

### Explanation
* Keep the same PyTorch transforms
* Keep the federated evaluation.


adam-narozniak:
I can also use `click` for the command line args instead of parsearg (it will shorten the example; which might be crucial for that). WDYT?
It'd be then
```python
@click.command()
@click.option('--node_id', ....)
def load_dataset(node_id):
```

examples/quickstart-pytorch/client.py
```diff
@@ -56,30 +59,48 @@ def test(net, testloader):
     criterion = torch.nn.CrossEntropyLoss()
     correct, loss = 0, 0.0
     with torch.no_grad():
-        for images, labels in tqdm(testloader):
-            outputs = net(images.to(DEVICE))
-            labels = labels.to(DEVICE)
+        for batch in tqdm(testloader, "Testing"):
+            images = batch["img"]
+            labels = batch["label"]
+            outputs = net(images)
```
jafermarq, line 65:
```suggestion
            outputs = net(images.to(DEVICE))
```

examples/quickstart-pytorch/client.py
```diff
@@ -56,30 +59,48 @@ def test(net, testloader):
...
 
+# Get node id
+parser = argparse.ArgumentParser(description="Flower")
+parser.add_argument("--node-id", choices=[0, 1, 2], type=int, help="Partition of the "
+"dataset, which is divided into 3 iid partitions created artificially.")
+node_id = parser.parse_args().node_id
 # Load model and data (simple CNN, CIFAR-10)
 net = Net().to(DEVICE)
-trainloader, testloader = load_data()
+trainloader, testloader = load_data(node_id)
```
jafermarq, line 103: I just wonder if we should redesign slightly the example so things are not "floating" around in the global scope but instead are inside a `def main()` or something like that... just a thought

jafermarq (commented):
@adam-narozniak, I think `click` is cool but I wonder if it might confuse some people that are only familiar with `arparse`?

Some comments below.

examples/quickstart-pytorch/client.py
adam-narozniak: Actually, that was my first thought too. I even wrote the code that way and reverted but in the end, I think it is the best format that can also be added to the flower.dev e.g. as it is now on the main page.

adam-narozniak:
@jafermarq I'll try `click` to shorten the example. If we get any bad feedback or decide on the `def main()` structure let's revert to `argparse`.

examples/quickstart-pytorch/client.py
```diff
@@ -79,8 +96,7 @@ def load_data():
 
 # Load model and data (simple CNN, CIFAR-10)
 net = Net().to(DEVICE)
-trainloader, testloader = load_data()
-
+trainloader, testloader = load_data(standalone_mode=False)
```
adam-narozniak, line 99: the `standalone_mode=False` is needed to return anything from the `click` function

examples/quickstart-pytorch/pyproject.toml
```diff
@@ -11,6 +11,8 @@ authors = ["The Flower Authors <hello@flower.dev>"]
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
 flwr = ">=1.0,<2.0"
+flwr-datasets = ">=0.0.2,<1.0.0"
 torch = "1.13.1"
 torchvision = "0.14.1"
```
jafermarq, line 16:
```suggestion
torch = "2.1.1"
torchvision = "0.16.1"
```

examples/quickstart-pytorch/requirements.txt
```diff
@@ -1,4 +1,6 @@
 flwr>=1.0, <2.0
+flwr-datasets>=0.0.2, <1.0.0
 torch==1.13.1
 torchvision==0.14.1
```
jafermarq, line 4:
```suggestion
torch==2.1.1
torchvision==0.16.1
```

jafermarq (commented):
maybe we can use this opportunity to bump up to torch 2+. works fine

jafermarq (dismissed):

examples/quickstart-pytorch/pyproject.toml
```diff
@@ -11,6 +11,7 @@ authors = ["The Flower Authors <hello@flower.dev>"]
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
 flwr = ">=1.0,<2.0"
-torch = "1.13.1"
-torchvision = "0.14.1"
+flwr-datasets = {extras = ["vision"], version = ">=0.0.2,<1.0.0" }
```
danieljanes, line 14:
```suggestion
flwr-datasets = { extras = ["vision"], version = ">=0.0.2,<1.0.0" }
```

examples/quickstart-pytorch/requirements.txt
```diff
@@ -1,4 +1,5 @@
 flwr>=1.0, <2.0
-torch==1.13.1
-torchvision==0.14.1
+flwr-datasets>[vision]=0.0.2, <1.0.0
```
danieljanes, line 2: Shouldn't it be `[vision]>`?

examples/quickstart-pytorch/run.sh
```diff
@@ -2,16 +2,13 @@
 set -e
 cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/
 
-# Download the CIFAR-10 dataset
-python -c "from torchvision.datasets import CIFAR10; CIFAR10('./data', download=True)"
```
danieljanes, line 6: How does HF DS handle concurrent requests when the dataset isn't cached on disk yet? We experienced issues with `torchvision.datasets` where concurrent clients trying to download the dataset corrupted each other.

danieljanes (changes requested):

examples/quickstart-pytorch/requirements.txt
adam-narozniak: Yes, fixed now

examples/quickstart-pytorch/run.sh
adam-narozniak: In empirical testing, I found it works ok but haven't found that stated explicitly in any docs. I would need to investigate the code or ask a question in the HF forum to make it 100% certain.

danieljanes (approved):

PR #2321, opened 2023-09-07
Unify task_res generation

panh99:
Unified task_res generation of gRPC, gRPC-rere, and REST in one place.

Added `configure_task_res()` function in `task_handler.py`. The function accepts task_res, corresponding task_res, and producer node as input and returns a new task_res with `group_id`, `workload_id`, `producer`, `consumer`, `ancestry` fields.


src/py/flwr/server/fleet/grpc_bidi/ins_scheduler.py
```diff
@@ -100,16 +101,10 @@ def _worker(
...
-            task=Task(
-                producer=Node(node_id=client_proxy.node_id, anonymous=False),
-                consumer=Node(node_id=0, anonymous=True),
-                legacy_client_message=client_message_proto,
-                ancestry=[task_ins.task_id],
-            ),
+        task_res = configure_task_res(
+            TaskRes(task=Task(legacy_client_message=client_message_proto)),
+            task_ins,
+            Node(node_id=0, anonymous=True),
```
danieljanes, line 107: Don't we need `Node(node_id=client_proxy.node_id, anonymous=False),` here? The `consumer` field needs to be `Node(node_id=0, anonymous=True),`

src/py/flwr/client/message_handler/task_handler.py
```diff
@@ -131,3 +132,35 @@ def wrap_client_message_in_task_res(client_message: ClientMessage) -> TaskRes:
...
+    task_res = TaskRes(
+        task_id="",  # This will be generated by the server
+        group_id=ref_task_ins.group_id,
+        workload_id=ref_task_ins.workload_id,
+        task=task_res.task,
+    )
+    # pylint: disable-next=no-member
+    task_res.task.MergeFrom(
+        Task(
+            producer=producer,
```
danieljanes, line 161: Can we get this value from `task_ins.consumer`, or do we need to provide it manually?

src/py/flwr/server/fleet/grpc_bidi/ins_scheduler.py
panh99: yes. changed to `Node(node_id=client_proxy.node_id, anonymous=False)`

src/py/flwr/client/message_handler/task_handler.py
panh99: I think we can actually get producer from `task_ins`. The producer arg here is bc, conventionally, we obtain `node`  from `node: Node = cast(Node, node_store[KEY_NODE])` and set it to be the `producer`. I am okay with both options.

danieljanes: Got it, let's leave it as-is

danieljanes (approved):

PR #2330, opened 2023-09-11
Add Flower Datasets docs setup

adam-narozniak:
Add the missing element needed for the documentation to be created.


datasets/doc/source/conf.py
```diff
@@ -26,13 +26,15 @@
 
 
 # Fixing path issue for autodoc
-sys.path.insert(0, os.path.abspath("../../src/py"))
+sys.path.insert(0, os.path.abspath("../../"))
+sys.path.insert(0, os.path.abspath("../../flwr_datasets"))
+sys.path.insert(0, os.path.abspath("../../flwr_datasets/federated_dataset"))
```
danieljanes, line 31: Are these two necessary?

datasets/doc/source/conf.py
```diff
@@ -47,6 +49,8 @@
 extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.autodoc",
+    "sphinx.ext.autosummary",
+    "sphinx_autodoc_typehints",
```
danieljanes, line 53: Should we add these to other docs projects as well?

adam-narozniak: I use the auto-summary in the project to generate links to classes and methods like here image

adam-narozniak: I will delete the `sphinx_autodoc_typehints`. I didn't finally use it.

datasets/doc/source/conf.py
adam-narozniak: Now, I removed the bottom one, and everything works just fine.

danieljanes: And how the one ending in `flwr_datasets`? The previous version seems to only point at the directory, not the package itself

adam-narozniak: When the `sys.path.insert(0, os.path.abspath("../../flwr_datasets"))` is removed then I have as shorter API Reference
image

The full reference:
image

adam-narozniak: In any cases when I keep just one of them doesn't work (to produce the full API ref). The only setting that works is the 
```
sys.path.insert(0, os.path.abspath("../../"))
sys.path.insert(0, os.path.abspath("../../flwr_datasets"))
```

danieljanes (approved):

PR #2333, opened 2023-09-11
Remove old versions from switcher of v1.5.0 docs

charlesbvll:
### Description

When switching to the v1.5.0 version of the docs, the older versions appeared in the switcher.

### Related issues/PRs

N/A

### Explanation

For v1.5.0 we need to use the snippet from the new `source/conf.py` file.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


doc/build-versioned-docs.sh
```diff
@@ -49,18 +49,36 @@ for current_version in ${versions}; do
       # Copy updated version of locales
       cp -r ${tmp_dir}/locales/$current_language locales/
       changed=true
-
     fi
 
+    # Only for v1.5.0, update the versions listed in the switcher
```
tanertopal, line 54: @charlesbvll does this mean this code segment will not be applied in 1.5.1 and subsequent releases?

charlesbvll: Yes, because for versions above 1.5.0, the correct code is already written in `conf.py`.

tanertopal: Sounds good! merging!

tanertopal (approved):

tanertopal (approved):

PR #2335, opened 2023-09-11
Add FDS get-started docs

adam-narozniak:
Add FDS get-started section of the documentation:
* Installation
* quickstart (with a short version of conversions).


datasets/doc/source/get-started-installation.rst
```diff
@@ -0,0 +1,33 @@
+Installation
+============
+
+Python Version
+--------------
+
+Federated Datasets requires `Python 3.8 <https://docs.python.org/3.8/>`_ or above.
```
danieljanes, line 7:
```suggestion
Flower Datasets requires `Python 3.8 <https://docs.python.org/3.8/>`_ or above.
```

danieljanes (changes requested):
There's no "get started" category in the Diataxis documentation framework, so I'd propose to make the installation page a how-to guide, and the quickstart a tutorial.

adam-narozniak:
Let me push back on that a little bit.
Neither there's a division into tutorial-series and tutorial (that corresponds to tutorial and quickstart-tutorial on the website). Notice that the tutorial-series serves as something very similar to get-started (a clearly defined learning path - therefore, provides a good user experience, especially for new users). While none of the mentioned categories are stated directly in the Diataxis documentation framework, we broke it for Flower. I would suggest treating it as a guiding principle here too, because we should prioritize user experience rather than keeping the strict structure according to some reference. Having something like a tutorial-series or get-started provides a clear learning flow to learn the essential parts of the library.
WDYT?

PR #2336, opened 2023-09-11
Add FDS API reference documentation

adam-narozniak:
The API reference is structured such that the navigation between modules is easier:
* single file per class,
* structure that creates toggle buttons on the side bar.

### DIFF
--- a/datasets/doc/source/ref-federated_dataset.rst
+++ b/datasets/doc/source/ref-federated_dataset.rst
@@ -1,0 +1,34 @@
+﻿federated\_dataset
+==================
+
+.. toctree::
+   :hidden:
+
+   ref-federated_dataset.FederatedDataset
+
+.. automodule:: federated_dataset
+
+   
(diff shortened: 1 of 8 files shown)


datasets/doc/source/ref-federated_dataset.rst
```diff
@@ -0,0 +1,34 @@
+﻿federated\_dataset
```
danieljanes, line 1: We probably don't need this module documents. We expect users to use `from flwr_datasets import FederatedDataset`, and never use `flwr_datasets.federated_dataset`. The same goes for the `partitioner` package: we only want users to use whatever gets exported from that package (i.e., mentioned in `__init__.__all__`), but we don't want users to depend on the modules in that package (we might want to rename/restructure them later, the module name should be treated as an implementation detail).

danieljanes (changes requested):

PR #2331, opened 2023-09-11
Decouple client instantiation and task execution for simulation

jafermarq:
Decoupling `client_fn` (callable that creates the client instance) and `job_fn` (callable that instantiates client and runs a particular workload, e.g. `fit()`) allows for the VCE to add support to, among other things, stateful clients. This is a necessary first step to achieve that goal.

PR #2303, opened 2023-09-06
Add client_fn argument to start_server

danieljanes:
### Checklist

- [ ] Implement proposed change
- [ ] Write tests
- [ ] Update [documentation](https://flower.dev/docs/writing-documentation.html)
- [ ] Update [changelog](https://github.com/adap/flower/blob/main/doc/source/changelog.rst)
- [ ] Make CI checks pass
- [ ] Ping maintainers on [Slack](https://flower.dev/join-slack/) (channel `#contributions`)

PR #2341, opened 2023-09-11
Add support for stateful clients

jafermarq:
This PR introduces simple (but sufficient for most cases) support for stateful clients. All clients are stateful now (with state being a simple dataclass with nothing but the client-id). The state is transparently injected each time a client is instantiated and the result is retrieved before the client object is destroyed (recall in simulation clients are ephemeral). 

This PR assumes #2331 is merged. it is also compatible with the ideas in #2303.

This PR replaces the contents of #2148 and #2160 so they should be closed/deleted


jafermarq:
Let's first work on and merge #2389, where the definition of `ClientState` and `WorkloadState` is introduced.

phdcybersec:
Hello there, and thanks for the work on supporting stateful clients. I see this PR is closed because the feature is supposed to be supported via the other mentioned issues. However, I'm not sure to understand how. Are those released in Flower 1.6? Is there any documentation on using stateful clients that I missed? Thanks again.

PR #2345, opened 2023-09-12
Add Flower Datasets tests as GitHub workflow

adam-narozniak:
Run standard tests for Flower Dataset based on GitHub workflow.


.github/workflows/datasets.yml
```diff
@@ -0,0 +1,42 @@
...
+    branches:
+      - main
+    paths:
+      - "datasets/flwr_datasets/**"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "datasets/flwr_datasets/**"
+
```
tanertopal, line 14:
```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.event.pull_request.number || github.ref }}
  cancel-in-progress: true

```

.github/workflows/datasets.yml
```diff
@@ -0,0 +1,46 @@
...
+        # https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#python
+        # In case of a mismatch, the job has to download Python to install it.
+        # Note: Due to a bug in actions/setup-python we have to put 3.10 in
+        # qoutes as it will otherwise will assume 3.1
+        python: [3.8, 3.9, '3.10']
+
+    name: Python ${{ matrix.python }}
+
+    steps:
+      - uses: actions/checkout@v3
```
tanertopal, line 38:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/datasets.yml
```diff
@@ -0,0 +1,46 @@
...
+        # Note: Due to a bug in actions/setup-python we have to put 3.10 in
+        # qoutes as it will otherwise will assume 3.1
+        python: [3.8, 3.9, '3.10']
+
+    name: Python ${{ matrix.python }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Bootstrap
+        uses: ./../.github/actions/bootstrap
```
tanertopal, line 40:
```suggestion
        uses: ./.github/actions/bootstrap
```

.github/workflows/datasets.yml
```diff
@@ -0,0 +1,46 @@
...
+  push:
+    branches:
+      - main
+    paths:
+      - "datasets/flwr_datasets/**"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "datasets/flwr_datasets/**"
```
tanertopal, line 13:
```suggestion
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

.github/workflows/datasets.yml
```diff
@@ -0,0 +1,42 @@
+name: Flower Baselines
```
tanertopal, line 1:
```suggestion
name: Flower Datasets
```

.github/workflows/datasets.yml
adam-narozniak: @tanertopal Why did you remove the `datasets/flwr_datasets/**`. I thought to run the tests only when there was any change in the library itself

tanertopal: Is their no cross depend code? Meaning if anything changes in Flower does that have any influence on datasets?

adam-narozniak: On the Flower's side, there'll be examples that install the Flower Datasets library (the dependency is added to the pyproject.toml). They are already implemented - it's just a modification to the existing partition code of our current examples. It can be merged when the library is released. 
On the Flower Datasets's side, I don't think we should add such tests (there will be tests that for direct ML frameworks integration).
To sum up, there should not be any case that the change of the code in one library (without a release) affects the other library.

tanertopal (approved):

tanertopal (approved):

PR #2349, opened 2023-09-12
Refactor Flower client

jafermarq:
Refactored Flower client:
* created a `typing.py` in client where types used throughout flower (ClientLike and ClientFn) are defined. It makes easier to access them like this instead that from client app where many things are imported.
* Moved all the numpy wrapping logic to `numpy_client_wrapper.py`, again so we can access important functions such as `to_client()` w/o having to worry about circular imports due to it being previously defined in client app

### DIFF
--- a/src/py/flwr/client/__init__.py
+++ b/src/py/flwr/client/__init__.py
@@ -20,11 +20,17 @@ from .app import start_client as start_client
 from .app import start_numpy_client as start_numpy_client
 from .client import Client as Client
 from .numpy_client import NumPyClient as NumPyClient
+from .numpy_client_wrapper import to_client as to_client
+from .typing import ClientFn as ClientFn
+from .typing import ClientLike as ClientLike
 
 __all__ = [
     "Client",
     "NumPyClient",
+    "ClientLike",
+    "ClientFn",
     "run_client",
     "start_client",
     "start_numpy_client",
+    "to_client",
 ]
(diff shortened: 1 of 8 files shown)


src/py/flwr/client/__init__.py
```diff
@@ -15,20 +15,16 @@
...
 from .numpy_client import NumPyClient as NumPyClient
 
 __all__ = [
     "Client",
-    "ClientLike",
     "NumPyClient",
     "run_client",
     "start_client",
     "start_numpy_client",
-    "to_client",
```
danieljanes, line 33: Can we continue to export these from `flwr.client`, and also include `ClientFn`, just like we do in `flwr.common` and `flwr.common.typing`?

jafermarq: done

src/py/flwr/client/__init__.py
```diff
@@ -15,18 +15,20 @@
...
+from .numpy_client_wrapper import to_client as to_client
+from .typing import ClientFn as ClientFn
+from .typing import ClientLike as ClientLike
 
 __all__ = [
     "Client",
-    "ClientLike",
     "NumPyClient",
+    "ClientLike",
+    "ClientFn",
```
danieljanes, line 31:
```suggestion
    "ClientFn",
    "ClientLike",
    "NumPyClient",
```

danieljanes (approved):

PR #2383, opened 2023-09-17
Shell changes

achiverram28:
## Issue

Modifying some of the things in the shell

### Description

1. In `test-baseline-structure.sh` , `baseline_path=$(pwd)/$1` would take the dev path , but for 
testing we need the baseline path

2. While testing even ruff cache would be there , so adding a way out to rm even that

## Proposal

1. Modified the current usage of `baseline_path`
2. Added a new feature in flwr/dev/rm-cache.sh to remove ruff_cache

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


dev/test.sh
```diff
@@ -19,7 +19,7 @@ python -m black --exclude "src\/py\/flwr\/proto" --check src/py/flwr
 echo "- black: done"
 
 echo "- init_py_check: start"
-python -m flwr_tool.init_py_check src/py/flwr src/py/flwr_tool
+python -m src.py.flwr_tool.init_py_check src/py/flwr src/py/flwr_tool
```
tanertopal, line 22: Can you elaborate on this line? It should work as is if `bootstrap.sh` runs correctly. It might be I am missing something.

baselines/dev/test-baseline-structure.sh
```diff
@@ -6,7 +6,7 @@ if [ "$#" -ne 1 ]; then
     exit 1
 fi
 baseline_dir=$1
-baseline_path=$(pwd)/$1
+baseline_path=$(cd .. && pwd)/$1
```
tanertopal, line 9: @jafermarq can you review this?

tanertopal (commented):
@achiverram28 Thank you for opening up the PR! Left one question for you.

baselines/dev/test-baseline-structure.sh
jafermarq: @achiverram28 we discussed about this earlier today. I think we concluded this is not needed, right?

dev/test.sh
jafermarq: yes, the problem was that the poetry env was not sourced before running the tests. @achiverram28 , should we revisit the instructions about environment setup we provide for contributors? Is there something we should make more clear or even update?

baselines/dev/test-baseline-structure.sh
achiverram28: yes

PR #2390, opened 2023-09-18
Unify client types

jafermarq:
This PR removes `ClientLike` usage and therefore everywhere a client object of type `Client` is expected. For this, the idea is that users will call `<their_client>.to_client()` when returning inside their `ClientFn`. `to_client()` is a new method that both `Client` and `NumPyClient` have. 

For backwards compatibility:
- the entry point via `start_numpy_client` when users supply a client object (that should be of type `ClientNumPy`). In this case, `.to_client()` is called internally and transparently to the user. However a Deprecation warning is triggered encouraging users to use `start_client` instead.
- Similarly in spirit, for simulation, we trigger a warning (once) if the type of client returned in `client_fn` is not of type `Client`. Then, the client is converted to `Client` type. 

In addition this PR:
- The content of `numpy_client_wrapper.py` has been moved to `client.py` (where it used to be not long ago)

Part of the content above was done in #2493.


src/py/flwr/client/app.py
```diff
@@ -268,8 +269,16 @@ def start_numpy_client(
     >>>     root_certificates=Path("/crts/root.pem").read_bytes(),
     >>> )
     """
-    # Start
-    _check_actionable_client(client, client_fn)
+    warnings.warn(
+        "flwr.client.start_numpy_client() is deprecated and will"
+        "be removed in a future version of Flower.",
+        DeprecationWarning,
```
danieljanes, line 275: Shouldn't this be the following?
```suggestion
        warnings.DeprecationWarning,
```

src/py/flwr/client/app.py
```diff
@@ -268,8 +269,16 @@ def start_numpy_client(
     >>>     root_certificates=Path("/crts/root.pem").read_bytes(),
     >>> )
     """
-    # Start
-    _check_actionable_client(client, client_fn)
+    warnings.warn(
+        "flwr.client.start_numpy_client() is deprecated and will "
+        "be removed in a future version of Flower.",
```
danieljanes, line 274: Let's be more helpful here and add guidance on how to migrate from `start_numpy_client` to `start_client` using a multi-line warning that shows before/after.

src/py/flwr/client/app.py
```diff
@@ -224,7 +225,7 @@ def start_numpy_client(
         the same machine on port 8080, then `server_address` would be
         `"[::]:8080"`.
     client_fn : Optional[Callable[[str], NumPyClient]]
-        A callable that instantiates a NumPyClient. (default: None)
+        A callable that returns a client of type NumPyClient. (default: None)
```
danieljanes, line 228: Should we just remove the support for `client_fn` from `start_numpy_client`? We added the capability after the Flower 1.5 stable release, which means it was never available to Flower stable users.

src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Ray-based Flower Actor and ActorPool implementation."""
 
-
 import threading
```
danieljanes, line 17:
```suggestion

import threading
```

danieljanes: We usually leave two empty lines between docstring and first import

src/py/flwr/simulation/ray_transport/utils.py
```diff
@@ -55,3 +60,25 @@ def enable_tf_gpu_growth() -> None:
             log(ERROR, traceback.format_exc())
             log(ERROR, ex)
             raise ex
+
+
+def check_clientfn_returns_client(client: Client) -> Client:
+    """Warn once that clients returned in `clinet_fn` should be of type Client.
```
danieljanes, line 66: Are we sure this will be displayed only once? It looks like it will be run every time `VirtualClientEngineActor.run` runs.

danieljanes (changes requested):

src/py/flwr/simulation/ray_transport/utils.py
jafermarq: yes. once each actor runs a client_fn for the first time.

src/py/flwr/client/app.py
jafermarq: yes, removed

danieljanes (approved):

PR #2393, opened 2023-09-19
Decouple client callable

danieljanes:
- [x] Merge #2559 first


src/py/flwr/client/app.py
```diff
@@ -146,20 +178,31 @@ class `flwr.client.Client` (default: None)
...
+        def _load_app() -> Flower:
+            return Flower(client_fn=client_fn)
 
-        client_fn = single_client_factory
+        load_callable_fn = _load_app
+    else:
+        warn_experimental_feature("`load_callable_fn`")
+
+    # At this point, only `load_callable_fn` should be used
+    # Both `client` and `client_fn` must not be used directly
```
tanertopal, line 206: Are we planning to deprecate `client_fn`? Maybe you can explain the thinking in the comment.

src/py/flwr/client/flower.py
```diff
@@ -0,0 +1,109 @@
...
+
+    task_res: TaskRes
+    state: WorkloadState
+
+
+FlowerCallable = Callable[[Fwd], Bwd]
+
+
+class Flower:
+    """Flower callable."""
```
tanertopal, line 48: Could you elaborate a bit on the general thinking behind the callable here? Maybe even give a usage example.

src/py/flwr/client/flower.py
```diff
@@ -0,0 +1,109 @@
...
+            state=WorkloadState(state={}),
+        )
+
+
+class LoadCallableError(Exception):
+    """."""
+
+
+def load_callable(module_attribute_str: str) -> Flower:
+    """."""
```
tanertopal, line 74: Even though this might change in the near term it would be helpful to document a bit the behaviour and thinking.

tanertopal (commented):
I left a few questions / comments on things I find hard to understand.

examples/mt-pytorch-callable/client.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def evaluate(self, parameters, config):
+        self.set_parameters(parameters)
+        loss, accuracy = test(net, testloader)
+        return loss, len(testloader.dataset), {"accuracy": accuracy}
+
+
+def client_fn(cid: str):
+    """."""
+    return FlowerClient().to_client()
```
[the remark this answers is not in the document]
danieljanes: `to_client()` needs to be called on client implementations that derive from `NumPyClient`. This is related to a recent change that requires users to return a subclass of `Client`. Subclasses of `NumPyClient` are not supported. In the case of `start_client`, this is not a breaking change b/c we never supported `client_fn` before.

src/py/flwr/client/app.py
danieljanes: This is unrelated to deprecation. The comment merely indicates that, after this line, `client` and `client_fn` should be considered out of scope. Not using those two directly, but only through `load_callable_fn`, ensures that all three ways of starting a client behave (almost) exactly the same. If also frees us from having if/else statements in many places because we handle all cases via `load_callable_fn`.

src/py/flwr/client/flower.py
danieljanes: Done

src/py/flwr/client/flower.py
danieljanes: Done

src/py/flwr/client/app.py
tanertopal: Thx for explaining!

tanertopal (approved):

PR #2401, opened 2023-09-20
Move `node_id` generation into `State` (#2401)

panh99:
### Description

> This PR does two things: 1) generate `node_id` and `workload_id` as a random 64-bit unsigned integer instead of a positive int64; 2) move the generation of `node_id` code into `state`.
> However, it may not be super necessary to use "real" uint64, given that `sqlite3` does not natively support uint64. We can always use 63 bits instead of all 64 bits if the information loss due to 1 bit less is trivial. 

Refactoring `State.register_node(self, node_id: int) -> None` and `State.unregister_node(self, node_id: int) -> None` to `State.create_node(self) -> int` and `State.delete_node(self, node_id: int) -> None`. The two methods are renamed to fit their new functionalities.

Move the `node_id` generation within the `State.create_node()`.

Make `node_id` true uint64, via `random.getrandbits(64)`.

Update `InMemoryState` and `SqliteState`.

Store uint64 values as int64 in `SqliteState` and convert them back to uint64 on retrieval.

Update corresponding tests, comments, names, and protocols.


danieljanes (approved):
Lgtm!

PR #2403, opened 2023-09-21
Enable more partitioners specification in more types

adam-narozniak:
Enable specifying partitioners as type `Dict[str, Union[Paritioner, int]]`

How the library is currently used to **create 10 partitions**:
```python
fds = FederatedDataset(dataset="mnist", partitioners={"train": 10})
```

Here are the new ways of specifying partitioners. All of the examples accomplish the same goal:

Dictionary-based object creation with Partitioner(s)
```
fds = FederatedDataset(dataset="mnist", partitioners={"train": IidPartitioner(num_partitions=10)})
```

Dictionary-based object creation with ints
```
fds = FederatedDataset(dataset="mnist", partitioners={"train":10})
```

Another use case is.
Dictionary-based object creation with mixed types
```
fds = FederatedDataset(
  dataset="mnist", 
  partitioners={
    "train":10,
    "valid": IidPartitioner(num_partitions=100)
  }
)
```


datasets/flwr_datasets/utils.py
```diff
@@ -29,23 +29,40 @@
...
+def _instantiate_partitioners(
+    partitioners: Dict[str, Union[Partitioner, int]]
+) -> Dict[str, Partitioner]:
     """Transform the partitioners from the initial format to instantiated objects.
 
     Parameters
     ----------
-    partitioners: Dict[str, int]
-        Partitioners specified as split to the number of partitions format.
+    partitioners:  Dict[str, Union[Partitioner, int]]
```
danieljanes, line 39:
```suggestion
    partitioners: Dict[str, Union[Partitioner, int]]
```

danieljanes (changes requested):

danieljanes (approved):
Lgtm!

PR #2409, opened 2023-09-21
Driver exception handling

panh99:
Add simple exception handling in Driver.

The driver will retry to send the message via `DriverStub` when encountering `grpc.StatusCode.UNAVAILABLE`. Specifically, it will retry for at most N times and double the sleep time after each retry.


src/py/flwr/driver/driver.py
```diff
@@ -73,52 +104,68 @@ def disconnect(self) -> None:
...
+                            retry_count + 1,
+                            retry_interval,
+                        )
+
+                    time.sleep(retry_interval)
+                    # Double the retry interval for exponential backoff
+                    retry_interval *= 2
+                else:
+                    raise
+        raise RuntimeError()
```
charlesbvll, line 171: Should we add a message to this error?

src/py/flwr/driver/driver.py
```diff
@@ -15,25 +15,56 @@
...
+    PullTaskResResponse,
+    PushTaskInsRequest,
+    PushTaskInsResponse,
+)
+from flwr.proto.driver_pb2_grpc import DriverStub
 
 DEFAULT_SERVER_ADDRESS_DRIVER = "[::]:9091"
 
+INITIAL_RETRY_INTERVAL = 1.0
+RETRY_TIMES = 3
```
charlesbvll, line 42: Shouldn't we rather make those optional arguments to the `Driver` constructor?

Issue #2369, opened 2023-09-13
FedAvgM expecting initial_parameters as Optional


gubertoli:
### Describe the bug

The current implementation of FedAvgM strategy considers the `initial_parameters` as Optional in the class init method ([here](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/fedavgm.py#L62C1-L62C57)):

`initial_parameters: Optional[Parameters] = None`

However, the FedAvgM strategy requires the `initial_parameters` ([here](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/fedavgm.py#L155C1-L158C78)):

```
            assert (
                self.initial_parameters is not None
            ), "When using server-side optimization, model needs to be initialized."
            initial_weights = parameters_to_ndarrays(self.initial_parameters)
```

### Steps/Code to Reproduce

```
strategy = fl.server.strategy.FedAvgM()

fl.simulation.start_simulation(
    client_fn=client_fn,
    num_clients=NUM_CLIENTS,
    config=fl.server.ServerConfig(num_rounds=5),
    strategy=strategy,
)
```

### Expected Results

To run the a simulation with vanilla FedAvgM strategy without initial_parameters.

### Actual Results

Assert exception:  **"When using server-side optimization, model needs to be initialized."**

DanielCardeal:
As far as I can tell, `initial_parameters` are required when using server side optimization in the FedAvgM strategy, as indicated by [this](https://github.com/adap/flower/blob/1cef0d47096220f00ccb004d0cc4fe24ef0bf5f0/src/py/flwr/server/strategy/fedavgm.py#L153) if statement above the assertion clause.

Server side optimization is enabled whenever `server_learning_rate` or `server_momentum` constructor parameters are set to values different from the defaults ([here](https://github.com/adap/flower/blob/1cef0d47096220f00ccb004d0cc4fe24ef0bf5f0/src/py/flwr/server/strategy/fedavgm.py#L116C1-L118C10)). As such, calling the `FedAvgM` constructor without any parameters works fine, while calling it with `server_learning_rate = 0.9`, for example, don't.

Considering all of that, it could be nice to add this information to the constructors' documentation, as to avoid confusion in the future.

achiverram28:
I guess the assert is an explanatory point and agree with @DanielCardeal  where the user can understand that if there is a server side optimisation , the `initial_parameters` shouldnt be done , but , for more clarity we can add it in the docstrings of the strategy class

[...]

PR #2416, opened 2023-09-22
Set `initial_parameters` as non-optional for FedAvgM (#2369)

DanielCardeal:
## Issue

#2369 

### Description

The constructor parameter `initial_parameters` of the FedAvgM strategy is typed as `Optional`, but is required whenever server-side optimization is enabled. This can be misleading, since server-side optimization is enabled "automatically" and in a non-transparent way.

## Proposal

Change type of `intial_parameters` from `Optional[Parameters]` to `Parameters` in FedAvgM.  Also, documents which conditions must be met for server-side optimization to be enabled.

### Explanation

This change aims to better communicate to users the necessity of `initial_parameters` in case of server-side optimization, as well as expose which conditions must be met to enable server-side optimization.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


jafermarq:
Hey @DanielCardeal, thanks for helping with the issue! The proposed change sounds reasonable. Would it make sense to you keeping it as optional and, in the first round, override the `configure_fit()` to initialise a local copy of the global parameters? (for context, when the `initial_parameters` are not set, the sever will sample one random client for it's model parameters and those will become the initial state of the global model.) 

Any thoughts on this @DanielCardeal @charlesbvll @adam-narozniak @panh99

gubertoli:
@jafermarq, I aggree that from a user perspective retrieving the parameters from a random client would be much easier. 

If possible (probably as a new issue) doing the same for FedOpt would also be great. Initializing these parameters is a recurrent question in Slack.

DanielCardeal:
@jafermarq @gubertoli Great idea!

I will try to add this changes in the next week as a new PR :)

adam-narozniak:
Hi, great catch. I'm in favor of @jafermarq 's idea. 

I think we should also add a warning saying if at least one of the default values for `momentum_vector` or `server_learning_rate` is not changed, then it is just a vanilla FedAvg running. I think this will help the users to understand that aspect too.

DanielCardeal:
Hi

I added the following changes to the PR:

- Automatic initialization of `intial_parameters` using the global model 
- Warn users when either `server_momentum` and `server_learning_rate` are set to default values and server side optimization is disabled.

Are there any further changes that need to be implemented?

src/py/flwr/server/strategy/fedavgm.py
```diff
@@ -59,7 +67,7 @@ def __init__(
         on_fit_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
         on_evaluate_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
         accept_failures: bool = True,
-        initial_parameters: Optional[Parameters] = None,
+        initial_parameters: Optional[Parameters],
```
jafermarq, line 70:
```suggestion
        initial_parameters: Optional[Parameters] = None,
```

jafermarq: It is preferable to leave as `None` as default. In this way, users can still use this strategy without having to worry about passing a (often randomly initialised) model.

jafermarq (commented):
Hi @DanielCardeal,

just a minor comment. I think we still want to not force users to pass `initial_parameters`.

I like the `WARNING_NO_SERVER_OPT` message you wrote to alert users. Do we want it to be displayed right upon strategy instantiation? Maybe yes? thoughts @adam-narozniak @danieljanes
It looks something like the below:
```bash
python run.py
WARNING flwr 2023-11-16 16:11:34,348 | fedavgm.py:132 | 
Setting both `server_momentum` and `server_learning_rate` to default values
cause FedAvgM to work as a vanilla FedAvg strategy. Server optimization with
momentum is enabled if either `server_momentum` is set to a value greater than
0.0 or `server_learning_rate` is set to a value lower than 1.0.

INFO flwr 2023-11-16 16:11:34,349 | app.py:178 | Starting Flower simulation, config: ...
....
```

WilliamLindskog:
@jafermarq something we can merge? Went through the comments and seems like most things have been cleared.

WilliamLindskog:
Hi everyone, after revisiting this internally and reviewing our latest examples, we think there’s an argument for making `initial_parameters` required again.

- Most of the examples already initialize and pass the global model manually when starting the server.
- Requiring `initial_parameters` explicitly would remove the hidden logic (sampling from random client) and make behavior more predictable and transparent.

Thus, @DanielCardeal's original idea (making `initial_parameters` non-optional) is the right direction for `FedAvgM`.

@DanielCardeal would you be interested in bringing this PR up-to-date with current codebase e.g. location of file?

PR #2422, opened 2023-09-26
Fix Flower Datasets api docs

adam-narozniak:
## Issue

Currently, the Flower Datasets API reference docs are empty (just the headers - names of the modelues)
image

### Explanation

The problem occurs due to the workflow logic (that I missed). The library that is installed is `flwr` yet for the automatic docs creation (using the sphinx by `make docs`), the library is needed to exist. That brings the need to install the `flwr-datasets` library prior to running the `make docs` command.

## Proposal
Create a separate script for docs building and workflow for uploading the HTML.

### Explanation

In order not to complicate the original workflow, a new workflow is created that specifically creates just the Flower Datasets docs.


tanertopal:
@adam-narozniak I think it's acceptable to just deploy the docs in the same job. No need for a separate workflow file.

I don't see any particular benefit in separating these as it's near zero additional cost right now to just deploy all in one workflow.

tanertopal:
I just reread the description and understand now the need. I'd still avoid an extra workflow file. Can you move the job into the other workflow file?

danieljanes (changes requested):
I'd prefer to have a consistent process across different projects (framework/baselines/datasets), if possible. I understand the need to separate out the datasets doc building process. To keep things consistent, I would therefore suggest to also separate out the baselines docs building process.

adam-narozniak:
@danieljanes @tanertopal 
I think your answers differ slightly in terms of what I should do.

so I should create a new workflow and a build script for the baselines? Or, should I try to move the current build docs for the dataset into another workflow?

I would propose to keep the workflow for the datasets docs to have the responsibility separated for each workflow (otherwise I think I could add it to the datasets after the tests, but I'm not in favor of that). And then I can create a script for the baselines, and I don't know what exactly @danieljanes meant but I can create a separate workflow (no need for flwr install there because there are no autodocs), or if we want to keep the number of them down, I can just use the script in the current docs workflow.

danieljanes:
@adam-narozniak if possible, I'm in favor of having everything in a single workflow via `docs.yml` (for now)

adam-narozniak:
@danieljanes In order to keep everything in the docs workflow, I'll have to add the `flwr-datasets` installing there, which I wanted to avoid to avoid potential dependency conflicts. Should I do it anyway?

What about the scripts? Should I keep `build-flwr-dataset-docs.sh` script and create one for the baselines?

danieljanes:
> @danieljanes In order to keep everything in the docs workflow, I'll have to add the `flwr-datasets` installing there, which I wanted to avoid to avoid potential dependency conflicts. Should I do it anyway?
> 
> What about the scripts? Should I keep `build-flwr-dataset-docs.sh` script and create one for the baselines?

Yes, let's install `flwr-datasets` there. Finding a dependency conflict would actually be very helpful since we want to make sure that users can always install all our libraries together (we don't have a test for that yet).

For scripts: up to you, as long as we follow the same pattern for all sub-projects. So, either one script that does everything, or one script that orchestrates sub-scripts for each project.

adam-narozniak:
Ok, got it. I went back to the single script that orchestrates all builds.

danieljanes (approved):
Lgtm!

PR #2423, opened 2023-09-26
Make the split keyword optional for load_partition

adam-narozniak:
## Issue

Currently, when you create the `FederatedDataset` object and want to load a partition, you need to do the following.
```
fds = FederatedDataset(dataset="mnist", partitioners={"train": 10})
partition = fds.load_partition(idx=0, split="train")
```
Yet it is clear that there is only one instance of `Partitioner`. Therefore, the call can be simplified. To say it differently, if when specyfing the split keyword in the `load_partitition(idx=idx, split="train")` `train` is the only possibility since you denote which partitioner will be used and only a single partitioner was used.

## Proposal

Make the split keyword optional for `load_partition` in case of having just one partitioner:
```
fds = FederatedDataset(dataset="mnist", partitioners={"train": 10})
partition = fds.load_partition(idx=0) # no need to give the split here, it will be taken from "train"
```

### Explanation
This PR does the following:
* make the split optional - code modification of `FederatedDataset`,
* check if it works - add tests
* (create a utility function to compare the `Datasets` objects).


datasets/flwr_datasets/federated_dataset.py
```diff
@@ -97,8 +97,12 @@ def load_partition(self, idx: int, split: str) -> Dataset:
...
         idx: int
             Partition index for the selected split, idx in {0, ..., num_partitions - 1}.
-        split: str
-            Name of the (partitioned) split (e.g. "train", "test").
+        split: Optional[str]
+            Name of the (partitioned) split (e.g. "train", "test"). You can skip this
+            parameter if there is only one partitioner for the dataset. The name will be
+             inferred automatically. (e.g. if partitioners={"train": 10} you do not need
+              to give this parameter, but if partitioners={"train": 10, "test": 100} you
+              need to give it, to differentiate which partitioner should be used.
```
danieljanes, line 105: These lines should not be indented

danieljanes:
```suggestion
            inferred automatically. For example, if `partitioners={"train": 10}`, you do not need
            to provide this argument, but if `partitioners={"train": 10, "test": 100}`, you
            need to set it to differentiate which partitioner should be used.
```

datasets/flwr_datasets/federated_dataset.py
```diff
@@ -188,3 +195,10 @@ def _resplit_dataset_if_needed(self) -> None:
         if self._resplitter:
             self._dataset = self._resplitter(self._dataset)
         self._resplit = True
+
+    def _check_if_no_split_keyword_possible(self) -> None:
+        if len(self._partitioners) != 1:
+            raise ValueError(
+                "Please give the split argument. You can omit the split keyword only if"
+                " there is a single partitioner specified."
```
danieljanes, line 203:
```suggestion
                "Please set the `split` argument. You can only omit the split keyword if"
                " there is exactly one partitioner specified."
```

danieljanes (changes requested):

danieljanes (dismissed):

danieljanes (approved):

PR #2432, opened 2023-09-29
Fix not matching return types in Strategy docs

adam-narozniak:
## Issue
We keep the types parameters as type hints in code and in docs. In `Strategy` the docs return types do not reflect the type hints return types.

### Description

`aggregate_fit` and `aggregate_evaluate` docs return types do not match the type hints in the code

## Proposal
Fix the docs return types.

### Explanation
The type hints are correct based on the code functionality therefore, the docs types need to be changed.


doc/source/ref-changelog.md
```diff
@@ -8,6 +8,10 @@
 
 ### What's new?
 
+- **Fix the incorrect return types of Strategy** ([#2432](https://github.com/adap/flower/pull/2432/files))
+
+  The types of the return values in the docstrings in two methods: aggregate_fit and aggregate_evaluate now match the hint types in the code.
```
danieljanes, line 13:
```suggestion
  The types of the return values in the docstrings in two methods (`aggregate_fit` and `aggregate_evaluate`) now match the hint types in the code.
```

danieljanes (changes requested):

danieljanes (approved):

PR #2435, opened 2023-09-30
Update baseline creation instruction

jafermarq:
Small change, to indicate the baseline name should be a single  **lowercase** word. 

Edited baselines docs:
* existing "tutorial" pages are moved to "how-to" pages.
* minor changes here and there

- [ ] do we need the Tutorials/References section in the `index.rst`?


baselines/doc/source/index.rst
```diff
@@ -19,30 +19,32 @@ The Flower Community is growing quickly - we're a friendly group of researchers,
...
 .. BASELINES_TABLE_ANCHOR
 
-Tutorials
-~~~~~~~~~
 
-A learning-oriented series of tutorials, the best place to start.
+How-to guides
+~~~~~~~~~~~~~
+
+Problem-oriented how-to guides show step-by-step how to achieve a specific goal.
```
danieljanes, line 30: It's ok to move our two docs pages to the `How-To` section, but let's keep the order of Tutorial / How-To / Explanation / Reference the same as it is for other projects

danieljanes (changes requested):
Looks great, just one comment

baselines/doc/source/index.rst
jafermarq: ok

danieljanes (approved):

PR #2447, opened 2023-10-02
Let `start_simulation` crash if flag is set.

jafermarq:
In some settings we might want to let simulation's `app.py` to throw an exception. This is useful for instance if some action needs to be taken when this happens.


danieljanes:
@jafermarq converted to draft as there are merge conflicts

danieljanes (approved):

PR #2452, opened 2023-10-03
Add script to generate contributor log

charlesbvll:
### Description

For each release, we use the same way to generate the list of contributors, but we have no script to automate it.

### Related issues/PRs

N/A

### Explanation

Add a new script that adds the contributor list along side the message to the changelog file.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


dev/add-shortlog.sh
```diff
@@ -0,0 +1,22 @@
...
+set -e
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/../
+
+tags=$(git tag --sort=-creatordate)
+new_version=$(echo "$tags" | sed -n '1p')
+old_version=$(echo "$tags" | sed -n '2p')
+
+shortlog=$(git shortlog "$old_version".."$new_version" -s | grep -vEi '(\(|\[)bot(\)|\])' | awk '{name = substr($0, index($0, $2)); printf "%s`%s`", sep, name; sep=", "} END {print ""}')
+
+token="<!---TOKEN_$new_version-->"
```
tanertopal, line 11: Why do we need the token and when do we remove it?

dev/add-shortlog.sh
```diff
@@ -0,0 +1,22 @@
...
+thanks="\n### Thanks to our contributors\n\nWe would like to give our special thanks to all the contributors who made the new version of Flower possible (in \`git shortlog\` order):\n\n$shortlog $token"
+
+# Check if the token exists in the markdown file
+if ! grep -q "$token" doc/source/ref-changelog.md; then
+    # If the token does not exist in the markdown file, append the new content after the version
+    awk -v version="$new_version" -v text="$thanks" \
+        '{print} $0 ~ "## " version {print text}' doc/source/ref-changelog.md > temp.md && mv temp.md doc/source/ref-changelog.md
+else
+    # If the token exists, replace the line containing the token with the new shortlog
```
tanertopal, line 20: @charlesbvll  Why do we need the token, and when do we remove it?

charlesbvll: @tanertopal The token is needed to update the shortlog if it already exists (and to check if it already exists), I tried multiple approaches without any tokens but they were unsuccessful. As the token is a markdown comment it is never displayed (so we never remove it).

tanertopal (approved):

PR #2465, opened 2023-10-05
Update quickstart-huggingface requirements

danielnugraha:
### Description

The requirements in quickstart-huggingface can not be resolved, since datasets and evaluate has incompatible versions, also the pyproject.toml and requirements.txt is not synced.

### Explanation

Remove dataset from dependencies and sync pyproject.toml and requirements.txt in quickstart-huggingface example.

### Checklist

- [x] Write tests
- [x] Make CI checks pass


tanertopal:
@danielnugraha isn't the example using `datasets`? Will it work if we remove it?

danielnugraha:
> @danielnugraha isn't the example using `datasets`? Will it work if we remove it?

@tanertopal evaluate depends on datasets so it will install datasets automatically. Previously it can't be resolved since we use evaluate 0.4.0 that depends on datasets >2.0 but we specify datasets v1.12 as dependency. I already tested it and it still works.

examples/quickstart-huggingface/pyproject.toml
```diff
@@ -12,10 +12,9 @@ authors = [
...
-python = "^3.8"
-flwr = "^1.0.0"
-torch = "^1.13.1"
+python = ">=3.8,<3.11"
+flwr = ">=1.0,<2.0"
+torch = "1.13.1"
 transformers = "4.30.0"
-datasets = "1.12.1"
 evaluate = "0.4.0"
 scikit-learn = "1.0"
```
jafermarq, line 20:
```suggestion
scikit-learn = "1.3.1"
```

jafermarq: w/ `1.0` installation fails.

examples/quickstart-huggingface/requirements.txt
```diff
@@ -1,6 +1,5 @@
...
-evaluate~=0.4.0
-flwr~=1.4.0
-numpy~=1.21.1
-torch~=2.0.1
-transformers~=4.30.2
+flwr>=1.0, <2.0
+torch==1.13.1
+transformers==4.30.0
+evaluate==0.4.0
+scikit-learn==1.0
```
jafermarq, line 5:
```suggestion
scikit-learn==1.3.1
```

jafermarq: w/ `1.0` installation fails.

jafermarq (commented):
just a minor comment. I wasn't able to do `poetry install` (tested with py3.8 and py3.9)

examples/quickstart-huggingface/requirements.txt
danielnugraha: I used python 3.10.9.

jafermarq: I didn't try the `requirements.txt` install

danielnugraha: I updated the requirements.txt as well and run pip install and it works in both 3.8 and 3.10.

jafermarq (approved):
works!

tanertopal:
> > @danielnugraha isn't the example using `datasets`? Will it work if we remove it?
> 
> @tanertopal evaluate depends on datasets so it will install datasets automatically. Previously it can't be resolved since we use evaluate 0.4.0 that depends on datasets >2.0 but we specify datasets v1.12 as dependency. I already tested it and it still works.

@danielnugraha, any dependency we explicitly import in the code must also be a dependency in the requirements.txt as otherwise, removing one dependency might break modules that do not import that one.

tanertopal (approved):

PR #2480, opened 2023-10-06
Migrate tutorial-series-get-started-with-flower-pytorch to FDS

adam-narozniak:
## Issue 
The tutorial series uses the custom dataset partitioning which might be implied.

## Proposal 
Migrate the 1st tutorial to use the Flower Datasets.

## Discussion 
I've updated the example so the structure remains the same - there is a list of dataloaders. There might be an alternative approach to just create a FederatedDataset abstraction and use it in the client function to fetch a partition, split train test and transform to DataLoader.


jafermarq (dismissed):
Looks good to me. I just added some small changes.

doc/source/tutorial-series-get-started-with-flower-pytorch.ipynb
```diff
@@ -640,9 +620,21 @@
    "toc_visible": true
   },
   "kernelspec": {
-   "display_name": "flower-3.7.12",
+   "display_name": "flwr-rtwXnbAq-py3.10",
```
tanertopal, line 623: @adam-narozniak just to be sure, is this intended?

doc/source/tutorial-series-get-started-with-flower-pytorch.ipynb
```diff
@@ -640,9 +620,21 @@
...
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.10.6"
```
tanertopal, line 637: @adam-narozniak do we need this change?

doc/source/tutorial-series-get-started-with-flower-pytorch.ipynb
jafermarq: that change probably comes from my side when re-running the notebook. i think this is not important. Can be discarded or kept.

doc/source/tutorial-series-get-started-with-flower-pytorch.ipynb
adam-narozniak: No, we don't need it. Our current `nbstripout` removed it. That was added in Javier's run and I didn't notice it.

doc/source/tutorial-series-get-started-with-flower-pytorch.ipynb
adam-narozniak: @tanertopal both "display_name" and "name" are required. I changed it to "flwr" now.
When they are not present I get "[E 09:57:23.746 NotebookApp] Notebook JSON is invalid: 'display_name' is a required property". And in UI, when starting the notebook I get the error (if not present). Similarly when "name" was not present in google colab, it complained and defaulted to "python3"

danieljanes (approved):

Issue #2042, opened 2023-07-06
FedSTAR


jafermarq:
# FedSTAR

* **Title:** Federated Self-training for Semi-supervised Audio Recognition
* **Venue:** ACM TEC 2022
* **Link to paper:** https://dl.acm.org/doi/10.1145/3520128

## Do you want to work on this baseline? 

> 🌻 Check everything about the `Summer of Reproducibility` on [`flower.dev/summer`](https://flower.dev/summer/)

> All available baselines are listed in the [Summer of Reproducibility Dashboard](https://github.com/orgs/adap/projects/6/) and also in the [GitHub Issues with the `summer-of-reproducibility` label](https://github.com/adap/flower/labels/summer-of-reproducibility). The content is the same.

> 📝 It is advised to complete these steps before your start working on your code. But if you can't wait to implement your baseline with Flower (we totally understand it 😄), please ensure you follow the steps on [how to contribute a new baseline](https://github.com/adap/flower/tree/main/baselines#contributing-a-new-baseline).

What follows are the steps 1 & 2 in the Summer of Reproducibility instructions.

### 1. Join the Summer of Reproducibility program
- [x] Join the [Flower Slack](https://flower.dev/join-slack/) and say _"hi! 👋"_ in the channel `#summer-of-reproducibility`.
- [x] Pick a baseline from our curated list **<---------------------------------------- [you are doing this now]**
### 2. Define the scope of your contribution
- [x] **What are you going to reproduce?** Add a comment to your issue and tell us about your plan regarding this baseline: what experiments from the paper are you reproducing?, for which datasets ? the more details you provide us with the better !
- [x] **Check if you are eligible for a reward**. 
    > As we have to comply with US/EU regulations, we have checked that individual contributors based on these countries or territories are eligible: Australia, Austria, Belgium, Bulgaria, Canada, Croatia, Cyprus, Czech Republic, Denmark, Estonia, Finland, France, Germany, Gibraltar, Greece, Hong Kong SAR China, Hungary, India, Ireland, Italy, Japan, Latvia, Liechtenstein, Lithuania, Luxembourg, Malta, Mexico, Netherlands, New Zealand, Norway, Poland, Portugal, Romania, Singapore, Slovakia, Slovenia, South Korea, Spain, Sweden, Switzerland, Thailand, United Arab emirates, United Kingdom United States.

[...]

PR #2482, opened 2023-10-06
FedStar

Raj-Parekh24:
## Issue
#2042 
### Description

Implementation of FedStar in flower.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


Raj-Parekh24:
Hi @jafermarq ,
I have replicated all codes and suggestions in this PR and also checked the working of the code. Please review the code.

Thanks for the support!

baselines/fedstar/README.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Federated Self-training for Semi-supervised Audio Recognition
+url: [dl.acm.org/doi/10.1145/3520128](https://dl.acm.org/doi/10.1145/3520128)
```
jafermarq, line 3:
```suggestion
url: https://dl.acm.org/doi/10.1145/3520128
```

baselines/fedstar/README.md
```diff
@@ -0,0 +1,59 @@
...
+labels: [Audio Classification, Semi Supervised learning]
+dataset: [Ambient Context, Speech Commands]
+---
+
+# FedStar: Federated Self-training for Semi-supervised Audio Recognition
+
+> Note: If you use this baseline in your work, please remember to cite the original authors of the paper as well as the Flower paper.
+
+****Paper:**** :[dl.acm.org/doi/10.1145/3520128](https://dl.acm.org/doi/10.1145/3520128)
```
jafermarq, line 12:
```suggestion
****Paper:****  [dl.acm.org/doi/10.1145/3520128](https://dl.acm.org/doi/10.1145/3520128)
```

jafermarq: In many places in the `README.md` there are two `:` followed, please remove them.

baselines/fedstar/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Environment Setup
+# Set python version
+pyenv local 3.10.6
+# Tell poetry to use python 3.10
+poetry env use 3.10.6
+# Now install the environment
+poetry install
+# Start the shell to activate your environment.
+poetry shell
```
jafermarq, line 38: please put this inside a block of code, not just like plain text. You can see in the readme it just doesn't look so good. (see how `FedProx` (or any other baseline currently under review) does it: https://github.com/adap/flower/tree/main/baselines/fedprox#environment-setup)

baselines/fedstar/README.md
```diff
@@ -0,0 +1,59 @@
...
+# Run dataset shell script
+. setup_datasets.sh
```
jafermarq, line 40: Similar to my comment above, this needs to be inside a block of code. Also, the script needs to be made executable. Maybe via `chmod +x setup_stasets.sh`.

Also it would be good if you put a paragraph explaining what the scrip is going to do.

baselines/fedstar/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Running the Experiments
+###### python -m fedstar.server
+###### python -m fedstar.client
```
jafermarq, line 44: Again, this should be inside a block of code, not plain text.

baselines/fedstar/fedstar/model.py
```diff
@@ -0,0 +1,239 @@
+import time
+import itertools
+import numpy as np
+import tensorflow as tf
+import tensorflow_addons as tfa
```
jafermarq, line 5: `tensorflow-addons` is not recommended anymore since it will stop being maintained in May'24. It seems you use it to define fairly standard layers so there must be another way that does not require this package.

baselines/fedstar/README.md
jafermarq: Also, i think you wanted to specify `fedstar.clients` (instead of `.client`)

jafermarq (commented):
Hi @Raj-Parekh24,

I have now completed a first pass through your code. I have highlighted a few formatting issues in the `README.md`. In general the main points are:
* It's better not to rely on `tensorflow_addons` since TF warns to not use. 
* I ran your code for 100 rounds with the default config (i.e. `ambient_context` dataset) but I only saw ~40% accuracy on the server side. 
* I'm not entirely sure if the logic around your GPU selection for clients is the best. On my system (with just one big GPU) only the server made use of it even though less than 5% of the VRAM was used. Wouldn't it be better to use Flower's simulation engine (i.e. use `start_simulation` as oppose to `start_server`+`start_client`?

Ping me again over here or Slack once you had a chance to look into the suggestions mentioned above and also those highlighted below. It would be good to start seeing some results on the `README.md` (even if still aren't fully reproducing those in the `FedStar` paper).

baselines/fedstar/fedstar/model.py
Raj-Parekh24: Thanks for the suggestion I have removed tfa completely.

baselines/fedstar/README.md
Raj-Parekh24: Thanks for the suggestion, I have transferred it to block of code.

baselines/fedstar/README.md
Raj-Parekh24: Thanks for poinitng out this error, I have remove all double colons.

baselines/fedstar/README.md
Raj-Parekh24: Thanks for the suggestion, updated the readme accordingly.

Raj-Parekh24:
Sorry for the delay in response.

I will try to convert the execution part to flower simulation, and simultaneously I will start creating the result section as well.

Raj-Parekh24:
Hi @jafermarq ,

Exactly how should I add the reproduced tables in the result. Like should I add the as of execution result or combined them altogether and show it as shown in paper.

Raj-Parekh24:
Hi @jafermarq ,

I have one more query, should I add configs for all the runs? For each cell in table a separate config is executed so should I add all of it?

baselines/fedstar/EXTENDED_README.md
jafermarq, line 1: please remove this file

baselines/fedstar/README.md
```diff
@@ -0,0 +1,64 @@
...
+# Set python version
+pyenv local 3.10.6
+# Tell poetry to use python 3.10
+poetry env use 3.10.6
+# Now install the environment
+poetry install
+# Start the shell to activate your environment.
+poetry shell
+# making the shell script executable
+chmod +x setup_stasets.sh
```
jafermarq, line 41:
```suggestion
chmod +x setup_datasets.sh
```

baselines/fedstar/README.md
```diff
@@ -0,0 +1,64 @@
...
+# The below script will download the datasets and create a directory structure requir to run this experiment.
+. setup_datasets.sh
```
jafermarq, line 43:
```suggestion
./setup_datasets.sh
```

jafermarq (commented):
Hi @Raj-Parekh24,

Thanks for applying the previously suggested changes. Please find below some other suggestions to improve your `README.md`.

To answer your questions:

* **Exactly how should I add the reproduced tables in the result?**

The best would be to present them similarly formatted as how it is done in the paper. For instance, since you are reproducing Table3  for _SpeechCommands_ and _AmbientContext_, I'd expect a table with that looks like Table 3 (excluding _VoxForge_). For Table 4 you can do the same.

* **should I add configs for all the runs?**

Yes, you should include all the configs and commands needed to reproduce the experiments. We prefer if you present these in the `README.md` directly instead of having them all in a big bash script. I can point you to a recent baseline that got merged that shows very clearly how to reproduce each table of results: [Check the `Expected Results` section for `DepthFL`](https://github.com/adap/flower/tree/main/baselines/depthfl#expected-results)

baselines/fedstar/README.md
Raj-Parekh24: Sorry for the typo error!

baselines/fedstar/README.md
Raj-Parekh24: Incorporated the suggested changes.

baselines/fedstar/EXTENDED_README.md
Raj-Parekh24: Incorporated in the next commits.

Raj-Parekh24:
Thanks for the suggestion I am adding configs and updating the readme with result tables.

Raj-Parekh24:
Hi @jafermarq ,

I have added table 3 output in the readme. To generate tables 3 and 4 around 30 config files are used and to invoke config files the path is required to be changed in the server.py file. Where should I mention the steps regrading it?

Raj-Parekh24:
Currently I am running the experiments of table 4 as well it will get completed by starting of the next week.

jafermarq:
> Hi @jafermarq ,
> 
> I have added table 3 output in the readme. To generate tables 3 and 4 around 30 config files are used and to invoke config files the path is required to be changed in the server.py file. Where should I mention the steps regrading it?

Hi @Raj-Parekh24, thanks for the update. 
I see configs for experiments in the same row are identical only differing on `server.num_clients`, `server.min_sample_size`, and `client.num_clients`. Wouldn't it be better to have a single config per row and then override via the command line the settings i just mentioned?
For example you can have a single config for Table 3 and `ambient_context` in `conf/table_3/ambient_context.yaml` that you can pass to your server and clients like this:

```bash
# clients_10
python -m fedstar.server --config-name table_3/ambient_context num_clients=10

# then run the exact same config but with 30 clients
python -m fedstar.server --config-name table_3/ambient_context num_clients=30
```

The above way of doing it is sufficient if you change your configs so they define a "global" `num_clients` and reference it to the settings i mentioned. Like this:

```yaml
# @package _global_ # <--------- this is important!!
---
num_clients: 5

server:
  num_clients: ${num_clients}
  min_sample_size: ${num_clients}
  ...

client:
  num_clients: ${num_clients}
  ...

```
That's how I'd approach this. Having so many config files is also prone to make small mistakes. 

For Table 4 you could follow a similar logic for rows; potentially incorporating the same ideas so you fuse the configs that iterate over both `num_clients` and `l_per | u_per`. The ideal setting is such that (1) it's easier to run the experiments; (2) it's easy for others to understand how to customise the hyperparameters; (3) is not error prone.

As for to where to add it. Please add it to your `Expected Results` section. It would be good if you can show exactly what commands need to be run to generate both table 3 and 4. If you think table 4 needs too many commands to execute to generate it fully, you can add those to a bash script and commit it.

Raj-Parekh24:
Hi @jafermarq ,

I have added the results of table 4 and also updated configs for table_3 as you suggested. Please check it if it looks good I will similarly add the configs command for table 4 as well.

jafermarq:
> Hi @jafermarq ,
> 
> I have added the results of table 4 and also updated configs for table_3 as you suggested. Please check it if it looks good I will similarly add the configs command for table 4 as well.

Hi @Raj-Parekh2, your tables look ok! I pushed some changes automatically applied after running the formatting scripts. I think the remaining TODO would be to ensure SpeechCommands is correctly pre-processed (that should make your accuracies match those from the paper).

jafermarq:
Hi @Raj-Parekh24, in my last commit i don't know what went wrong but your big table4 was erased by mistake. I've put it back. Sorry for this. Let me know if we should sync regarding the SpeechCommands pre-processing

Raj-Parekh24:
Hi @jafermarq ,

No worries,

I have added the preprocessing scripts for the silence category in speech commands. Let me check it on a small number of parameters. After that can we organize the one to one call.

Raj-Parekh24:
Hi @jafermarq ,

I have run few experiments with updated pre processing part. But there is no improvement compared to existing results.

jafermarq:
> Hi @jafermarq ,
> 
> I have run few experiments with updated pre processing part. But there is no improvement compared to existing results.

Hi @Raj-Parekh24, were you able to check if batches are balanced? (i.e. roughly an equal number or training examples from each class). Note this is important because the "other" category has many more training examples than the 10 keywords (i.e. "yes", "no", "up", "down", etc) of interest.

Raj-Parekh24:
The distribution of the data is the same among all clients. The data distributed among them follows the same distribution as that of source data. The experiment is carried out on IID data. So the distribution is the same, but the sample count for each is different.

jafermarq:
Hi @Raj-Parekh24, (following our chat on Slack) i have just pushed the changes. Main changes:
* Introduction of another [dataset pipeline stage](https://github.com/Raj-Parekh24/flower/blob/79029cb65aec8c52f61f45232d48f4f076480042/baselines/fedstar/fedstar/data.py#L308) applied to both `ds_train_L` and `ds_train_U` . It's purpose is to ensure batches used by clients are somewhat balanced. Since speechcommands with 12 classes has many more training examples of the _other_ class, this weighted sampling (via TF's [rejection_sample](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#rejection_resample)) seems necessary. With this I was able to get ~5% extra accuracy for SpeechCommands. Note this is also applied to `Ambient Context` (you might want to disable that)
* I have edited the `README.md` to indicate that the silence processing needs to be run. It also mentions some caveats of running it for a second time. 
* I updated the silence processing function and now it generates 1k 1-second audio clips from every background noise. Likely this can be done on-the-fly (i.e. without creating files) but I'm not sure how to do this (i'm usually work with PyTorch 😅 )
* In my previous couple of commits I made some edits to `server.py` and `clients.py`. They work fine assuming you run the Table 3 config. I haven't tested the Table 4 configs. In `clients.py` I had to edit [this line](https://github.com/Raj-Parekh24/flower/blob/79029cb65aec8c52f61f45232d48f4f076480042/baselines/fedstar/fedstar/clients.py#L36) else they will fail to find the path where the data is. Not sure how you were able to run the experiments before. I run the code from the top-level directory of `fedstar` baseline.

jafermarq:
Hi @Raj-Parekh24, I introduced some changes to the configs for Table 4. Now everything is based on a single config (`conf/table4.yaml`) and sub-configs in `conf/L/`. I left the old configs there too just in case you want to double check my changes. With the new changes, you can for instance run an experiment in Table 4 like this:

```
python -m fedstar.server --config-name table4 dataset_name=speech_commands num_clients=10 L=L5 # runs 10 clients with L=5%
python -m fedstar.clients --config-name table4 dataset_name=speech_commands num_clients=10 L=L5 
```
The above can be run in "supervised" mode if you additionally pass `fedstar=false`.

baselines/fedstar/fedstar/models.py
```diff
@@ -0,0 +1,426 @@
...
+        - num_batches (int, optional): Number of batches per epoch.
+        - T (float, optional): Temperature parameter for softmax in pseudo-labeling.
+        - verbose (int, optional): Verbosity mode.
+
+        Returns
+        -------
+        - History: An object containing training history information.
+        """
+        l_dataset, u_dataset = train_datasets
+        l_num_batch = len(l_dataset)
```
jafermarq, line 289: this fails when using `table4` experiments with `speech_commands`. I suspect it's due to the balancing. It should be possible to achieve the same functionality since you already know what the `l_per` and `u_per` values are and what's the total dataset length of the client right?

baselines/fedstar/fedstar/client.py
```diff
@@ -0,0 +1,154 @@
...
+            finished {self.local_train_round}th round of training with
+            loss {history.history['loss'][0]:.4f} and
+            accuracy {history.history['accuracy'][0]:.4f}"""
+        )
+        # Clear Memory
+        # tf.keras.backend.clear_session()
+        gc.collect()
+        return (
+            self.model.get_weights(),
+            self.num_examples_train,
```
jafermarq, line 138: is it correct reporting only the number of labelled examples?

jafermarq (commented):
Hi @Raj-Parekh24 , just a couple of points based on our last sync on slack.

baselines/fedstar/fedstar/client.py
Raj-Parekh24: I thought of putting on labeled samples because our semi-supervised architecture is run on the client side and only the unlabelled samples are available there. The server only knows about labeled samples.

jafermarq (approved):
Amazing job @Raj-Parekh24 reproducing `FedStar`, a FL method for semi-supervised audio recognition! It's great including a new TensorFlow baseline too.

Raj-Parekh24:
Thanks a lot @jafermarq,

For your guidance and support throughout the project!

PR #2488, opened 2023-10-08
Upgrade Pylint to 3.0.3

danieljanes:


src/py/flwr/server/server_test.py
```diff
@@ -88,24 +88,29 @@ def get_properties(
         self, ins: GetPropertiesIns, timeout: Optional[float]
     ) -> GetPropertiesRes:
         """Raise an Exception to simulate failure in the client."""
+        # pylint: disable-next=broad-exception-raised
         raise Exception()
```
danieljanes, line 92: Should we raise `NotImplementedError` instead and remove the `# pylint: disable` comments?

charlesbvll: I used the `# pylint: disable` because the function docstrings specifically indicate that an `Exception` is raised, but I can update both if you prefer

pyproject.toml
```diff
@@ -88,7 +88,7 @@ isort = "==5.12.0"
 black = { version = "==23.10.1", extras = ["jupyter"] }
 docformatter = "==1.7.5"
 mypy = "==1.6.1"
-pylint = "==2.13.9"
+pylint = "==3.0.2"
```
danieljanes, line 91:
```suggestion
pylint = "==3.0.3"
```

src/py/flwr/server/server_test.py
danieljanes: yes, let's update both to keep it consistent with the `SuccessClient` above (and also change the docstrings there)

tanertopal (dismissed):

tanertopal (approved):

PR #2493, opened 2023-10-09
Add `to_client` method to client classes

jafermarq:
Add `to_client()` method to both `Client` and `NumPyClient` classes. Moves `NumPyClient`->`Client` conversion logic. 

A follow up PR will remove `numpy_client_wrapper.py` entirely, currently not possible due to `ClientState` definition.


danieljanes (approved):
Lgtm!

PR #2496, opened 2023-10-10
Set `workload_id` fields in `Driver` class

panh99:
This PR updates `Driver` class. It can now set the `workload_id` fields automatically when `get_nodes()` or `push_task_ins()` is called.


src/py/flwr/driver/driver.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         self.certificates = certificates
         self.channel: Optional[grpc.Channel] = None
         self.stub: Optional[DriverStub] = None
+        self.workload_id: Optional[int] = None
```
danieljanes, line 60: Should we enable setting this through the constructor? Users might want to re-start the driver script after a crash, and continue to submit tasks for the same workload.

panh99: Sound reasonable. Will do

PR #2499, opened 2023-10-11
Add new auto release workflow

charlesbvll:
### Description

Currently the release of a new version must be made manually.

### Related issues/PRs

N/A

### Explanation

This workflow will run every time a new tag is pushed to the repo and will create a new draft release, containing the changelog of the release (obtained from `doc/source/ref-changelog.md`), and the wheel and the tar generated from the commit associated with the tag.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


.github/workflows/framework-release.yml
```diff
@@ -0,0 +1,47 @@
...
+          echo "TAG_NAME=$tag_name" >> "$GITHUB_ENV"
+          
+          wheel_name="flwr-${tag_name}-py3-none-any.whl"
+          echo "WHEEL_NAME=$wheel_name" >> "$GITHUB_ENV"
+          
+          tar_name="flwr-${tag_name}.tar.gz"
+          echo "TAR_NAME=$tar_name" >> "$GITHUB_ENV"
+
+          wheel_url="https://artifact.flower.dev/py//${GITHUB_SHA::7}/${wheel_name}"
+          tar_url="https://artifact.flower.dev/py//${GITHUB_SHA::7}/${tar_name}"
```
tanertopal, line 29:
```suggestion
          wheel_url="https://artifact.flower.dev/py/${GITHUB_SHA::7}/${wheel_name}"
          tar_url="https://artifact.flower.dev/py/${GITHUB_SHA::7}/${tar_name}"
```

.github/workflows/framework-release.yml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  publish:
+    name: Publish draft
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Download wheel
```
tanertopal, line 17: How can we ensure the job which actually creates and uploads the wheel is already finished? If I understand this  job right it would not create the wheel so the assumption is that the pipeline which run on the same commit is already finished right?

.github/workflows/framework-release.yml
```diff
@@ -0,0 +1,47 @@
...
+          curl $wheel_url --output $wheel_name
+          curl $tar_url --output $tar_name
+
+      - name: Generate body
+        run: |
+          ./dev/get-latest-changelog.sh > body.md
+          cat body.md
+          
+      - name: Release
+        uses: softprops/action-gh-release@v1
```
tanertopal, line 40: Should we use the SHA of the action here? Ensuring that when `softprops` reuses the v1 tag, we don't get something we don't want?

.github/workflows/framework-release.yml
charlesbvll: @tanertopal I completely overlooked that indeed, thanks for the catch, I added a new step to wait for the upload to be finished

.github/workflows/framework-release.yml
charlesbvll: @tanertopal Done!

tanertopal (approved):

PR #2513, opened 2023-10-13
Explain on the fly transforms for PyTorch

adam-narozniak:
## Issue
So far, immediate transforms have been used to create PyTorch `DataLoader` from `datasets.Dataset`. It is not how the typically downloaded Dataset from PyTorch `torchvision` behaves. The transforms applied to it execute at the access time. The difference might bring confusion to users or introduce unexpected behavior. 

## Proposal

Explain how to do the transforms on-the-fly that strictly resembles the PyTorch way of doing it (which users are probably familiar with and expect). 
The instant transform to all the dataset still exist in the documentation but the difference in the behavior is explained. Also, it's explained after the on-the-fly transforms.


datasets/doc/source/how-to-use-with-pytorch.rst
```diff
@@ -15,17 +15,37 @@ vary e.g. "img" or "image", "label" or "labels"::
...
+    batch["img"] = [transforms(img) for img in batch["img"]]
+    return batch
+
+  partition_torch = partition.with_transform(apply_transforms)
+  # At this point, you can check if you didn't make any mistakes by calling partition_torch[0]
+  dataloader = DataLoader(partition_torch, batch_size=64)
+
+
+Alternatively, you can use the `map() <https://huggingface.co/docs/datasets/v2.14.5/en/package_reference/main_classes#datasets.Dataset.map>`_
+function. Note that the operation is instant (contrary to the set_transform and with_transform). Remember that the map
```
danieljanes, line 46:
```suggestion
function. Note that the operation is instant (contrary to the ``set_transform`` and ``with_transform``). Remember that the ``map``
```

datasets/doc/source/how-to-use-with-pytorch.rst
```diff
@@ -15,17 +15,37 @@ vary e.g. "img" or "image", "label" or "labels"::
...
+will modify the existing dataset if the key in the dictionary you return is already present and append a new feature if
+it did not exist before. Below, we modify the "img" feature of our dataset.::
```
danieljanes, line 48:
```suggestion
it did not exist before. Below, we modify the ``"img"`` feature of our dataset.::
```

danieljanes (approved):

PR #2535, opened 2023-10-25
Add `driver_service_address` and `certificates` args to `Driver` class constructor

panh99:
Based on #2534


src/py/flwr/driver/driver.py
```diff
@@ -31,7 +31,13 @@
 class Driver:
     """`Driver` class provides an interface to the Driver API."""
 
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        driver_service_address: str = DEFAULT_SERVER_ADDRESS_DRIVER,
+        certificates: Optional[bytes] = None,
```
danieljanes, line 37: This should be documented in the docstring

danieljanes (changes requested):

src/py/flwr/driver/driver.py
panh99: sure

danieljanes (approved):

PR #2549, opened 2023-10-30
Replace `GrpcDriver` with `Driver` in the `start_driver()` function.

panh99:
- Replaced `GrpcDriver` with `Driver` in `start_driver()`
- Replaced `GrpcDriver` with `Driver` in `DriverClientProxy`
- Removed sync lock
- Added `ref_exit_flag = [False]`, which is used to "kill" the thread updating the client manager.
- Updated unit tests in `driver/app_test.py` accordingly (also accelerated tests by patching `time.sleep`).
- Changed name `certificates` to `root_certificates`


src/py/flwr/driver/app.py
```diff
@@ -113,7 +113,6 @@ def start_driver(  # pylint: disable=too-many-arguments, too-many-locals
         root_certificates = Path(root_certificates).read_bytes()
     driver = GrpcDriver(driver_service_address=address, certificates=root_certificates)
     driver.connect()
```
danieljanes, line 115: Given that we use `Driver` below, why do we have `GrpcDriver` here?

src/py/flwr/driver/app.py
```diff
@@ -208,3 +197,5 @@ def update_client_manager(
 
         # Sleep for 3 seconds
         time.sleep(3)
+    # Exit
+    del driver
```
danieljanes, line 201: Why is this necessary? Doesn't this happen automatically when we exit the function?

src/py/flwr/driver/app.py
```diff
@@ -197,9 +186,9 @@ def update_client_manager(
         for node_id in new_nodes:
             client_proxy = DriverClientProxy(
                 node_id=node_id,
-                driver=driver,
+                driver=cast(GrpcDriver, driver.grpc_driver),
                 anonymous=False,
-                workload_id=workload_id,
+                workload_id=cast(int, driver.workload_id),
```
danieljanes, line 191: Shouldn't we introduce a getter method on `Driver` for this? This would prevent us from doing the `cast` above

danieljanes (changes requested):

src/py/flwr/driver/app.py
panh99: Removed. Thanks for spotting it.

src/py/flwr/driver/app.py
panh99: Double checked. It is redundant. Removed.

src/py/flwr/driver/app.py
panh99: Make sense. I've updated the `DriverClientProxy` to use `Driver` instead of `GrpcDriver`.

jafermarq:
@panh99 , should we close this?

PR #2553, opened 2023-10-31
Fix `RetryInvoker`

panh99:
### Quick fix
- Change `wait_strategy` to `wait_factory`. This is to avoid continuing with the used `wait_strategy` generator when `invoke` method is called for multiple times.
- Updated the example in the class docstring accordingly.


src/py/flwr/common/retry_invoker.py
```diff
@@ -145,20 +145,20 @@ class RetryInvoker:
...
-    Initialize a `RetryInvoker` with exponential backoff and call a function:
+    Initialize a `RetryInvoker` with exponential backoff and invoke a function:
 
     >>> invoker = RetryInvoker(
-    >>>     exponential(),
-    >>>     grpc.RpcError,
-    >>>     max_tries=3,
-    >>>     max_time=None,
-    >>> )
+    ...     exponential,  # # or use lambda: exponential(3, 2) to pass arguments
```
danieljanes, line 151:
```suggestion
    ...     exponential,  # Or use `lambda: exponential(3, 2)` to pass arguments
```

danieljanes (approved):

danieljanes (approved):

PR #2588, opened 2023-11-10
Add shuffle seed parameters to FederatedDataset

adam-narozniak:
## Issue
The FDS doesn't allow shuffling datasets after the download, which might pose a problem when a dataset comes with a non-random order, e.g., ordered by label (firstly only 0, then only 1). Then partitioning in such cases might have unexpected behavior.

## Proposal
Introduce `shuffle` and `seed` keywords that enable shuffling of the dataset after the download.
### Explanation
The `shuffle = True` shuffles all the splits (each one separately). The self._dataset is DatasetDict so e.g. {"train": train_data, "test": test_data}. All splits get shuffled (with the same seed).

The default seed is set (let's make it reproducible out of the box).

The default of shuffling is True. Let's make only the informed user (that checked that the ds doesn't come with a non-random order change it).

This PR also merges the downloading, shuffling and resplitting into a single function. This had to be done in this PR; otherwise, a complex logic determining shuffling time should be added. It will also enable in the future easy choice whether to download the dataset easily or not.

## Breaking changes
Note: this change makes the load_partition and load_full behave differently than they did with no shuffling. The order in load_full is not kept. The sets of samples assigned to the node_id will differ due to the shuffling (unless the data is not divided, accomplished by just a single partition. In that case only the order will differ)


datasets/flwr_datasets/federated_dataset.py
```diff
@@ -50,6 +50,13 @@ class FederatedDataset:
         A dictionary mapping the Dataset split (a `str`) to a `Partitioner` or an `int`
         (representing the number of IID partitions that this split should be partitioned
         into).
+    shuffle: bool
+        Whether to randomize the order of samples. Applied prior to resplitting,
+        speratelly to each of the present splits in the dataset. It uses the `seed`
+        argument. Defaults to True.
+    seed: Optional[int]
+        Seed used for dataset shuffling. It has no effect if `shuffle` is False. The
+        seed con not be set in the later stages.
```
danieljanes, line 59:
```suggestion
        seed cannot be set in the later stages.
```

datasets/flwr_datasets/federated_dataset.py
```diff
@@ -50,6 +50,13 @@ class FederatedDataset:
         A dictionary mapping the Dataset split (a `str`) to a `Partitioner` or an `int`
         (representing the number of IID partitions that this split should be partitioned
         into).
+    shuffle: bool
```
danieljanes, line 53: We're following the NumPy docstring style, which has a space before the `:`:
```suggestion
    shuffle : bool
```

datasets/flwr_datasets/federated_dataset.py
```diff
@@ -176,15 +183,34 @@ def _assign_dataset_to_partitioner(self, split: str) -> None:
...
+        Notes
+        -----
+            The shuffling should happen before the resplitting. Here is the explanation.
+            If the dataset has a non-random order of samples e.g. each split has first
+            only label 0, then only label 1. Then in case of resplitting e.g.
+            someone creates: "train" train[:int(0.75 * len(train))], test: concat(
+            train[int(0.75 * len(train)):], test). The new test took the 0.25 of e.g.
+            the train that is only label 0 (assuming the equal count of labels).
+            Therefore, for such edge cases (for which we have split) the split should
+            happen before the resplitting.
```
danieljanes, line 205: This section should be unindented.

danieljanes (changes requested):

datasets/flwr_datasets/federated_dataset.py
adam-narozniak: Fixed also for other parameters

danieljanes (approved):

PR #2591, opened 2023-11-11
Make `flower-server` start server with HTTPS by default

panh99:
- Added `--insecure` argument to start server without HTTPS.
- Added `--certificates` argument to accept paths to (CA certificate, server certificate, and server private key)
- Added `certificates` argument to some functions to facilitate the above changes.
- Added E2E test `bare-https` to test starting server with SSL-enabled.
- Added necessary certificates under `e2e/certificates/` for testing purposes.
- Amended `e2e/test.sh` and `e2e/test_driver.sh`.

### Note
- `flower-server` command will fail without either `--insecure` or `--certificates a b c`.
- When `--insecure` is enabled, `--certificates` will be ignored. (This is mentioned in the help message.)
- `e2e/bare-https/simulation.py` is empty because simulation does not require any certificate.
- REST server part remains untouched b/c it's different from `grpc-rere` and `grpc-bidi` and I am not sure how to unify the three.


danieljanes (commented):
PR looks good! 

`run_driver_api` and `run_fleet_api` are the targets for two additional commands that allow users to start the Driver API and the Fleet API independent of each other (see `pyproject.toml`:

image

Those two commands will have to support `--insecure` and `--certificates`, too. The easiest way to do it is probably through `_add_args_common`.

danieljanes (dismissed):

e2e/bare-https/client.py
```diff
@@ -0,0 +1,35 @@
+import flwr as fl
+import numpy as np
+from pathlib import Path
+
+
```
tanertopal, line 5: What's the benefit of having objective and model_params here for the outcome of the test? Maybe we can just simplify fit/evaluate? I might be missing something though.

e2e/bare-https/simulation.py
```diff
@@ -0,0 +1 @@
+# No simulation test for bare-https
```
tanertopal, line 1:
```suggestion
# No simulation test for bare-https

```

src/py/flwr/server/app.py
```diff
@@ -441,6 +457,29 @@ def run_server() -> None:
         driver_server.wait_for_termination(timeout=1)
 
 
+def _try_obtain_certificates(
+    args: argparse.Namespace,
+) -> Optional[Tuple[bytes, bytes, bytes]]:
+    # Obtain certificates
+    if args.insecure:
+        log(WARN, "The server will be started without HTTPS.")
```
tanertopal, line 465: I would word this differently.

```suggestion
        log(WARN, "Option `--insecure` was found. Starting insecure HTTP server.")
```

e2e/bare-https/client.py
panh99: the `bare-https` test is copied from `bare` test, which is the simplest E2E test we have. Having a `model_params` can also make sure that flower can aggregate models successfully with HTTPS-enabled. I am open to change if you think we can simplify the fit/evaluate methods here.

e2e/certificates/ca.crt
```diff
@@ -0,0 +1,30 @@
+-----BEGIN CERTIFICATE-----
```
tanertopal, line 1: I would not commit these here but rather generate them each time the test runs and delete them afterwards. It's often easier to commit them, but at the same time, I would prefer to avoid having any secret keys or certificates even for testing committed to prevent any possibility someone might accidentally use them somewhere.

danieljanes: @tanertopal wdyt, should we also move the `certificates` directory into the `https-bare` dir? They are only required for that single test

e2e/bare-https/client.py
danieljanes: @tanertopal @panh99 I think it makes sense to keep `bare` and `bare-https` as similar as possible

e2e/certificates/ca.crt
tanertopal: If they are only required here, then let's move it. The directory could also contain the script to generate the certificates.

danieljanes (approved):

Issue #2595, opened 2023-11-13
Add Flower Baseline: FLANDERS


edogab33:
### Paper

Gabriele Tolomei, Edoardo Gabrielli, Dimitri Belli, Vittorio Miori. (2023). A Byzantine- Resilient Aggregation Scheme for Federated Learning via Matrix Autoregression on Client Updates.

### Link

https://arxiv.org/abs/2303.16668

### Maybe give motivations about why the paper should be implemented as a baseline.

I'm one of the authors. To the best of my knowledge, this is the first work to offer a new approach to the problem of model poisoning attacks by analyzing the history of model updates sent by the clients through an anomaly detection algorithm.

Our results show that FLANDERS can mitigate extremely severe attacks (i.e. where up to 95% of clients are byzantine).

### For first time contributors

- [x]  Read the [`first contribution` doc](https://flower.dev/docs/first-time-contributors.html)
- [x]  Complete the Flower tutorial
- [x]  Read the Flower Baselines docs to get an overview:
    - [x]  [https://flower.dev/docs/using-baselines.html](https://flower.dev/docs/using-baselines.html)
    - [x]  [https://flower.dev/docs/contributing-baselines.html](https://flower.dev/docs/contributing-baselines.html)

### Prepare - understand the scope

- [X] Read the paper linked above
- [x] Create the directory structure in Flower Baselines (just the `__init__.py` files and a `README.md`)
- [x] Before starting to write code, write down all of the specs of this experiment in a README (dataset, partitioning, model, number of clients, all hyperparameters, …)
- [x] Open a draft PR

### Implement - make it work

- [x] Implement some form of dataset loading and partitioning in a separate `dataset.py` (doesn’t have to match the paper exactly)
- [x] Implement the model in PyTorch
- [x] Write a test that shows that the model has the number of parameters mentioned in the paper
- [x] Implement the federated learning setup outlined in the paper, maybe starting with fewer clients
- [x] Plot accuracy and loss
- [x] Run it and check if the model starts to converge

### Align - make it converge

- [x] Implement the exact data partitioning outlined in the paper
- [x] Use the exact hyperparameters outlined in the paper
- [x] Make it converge to roughly the same accuracy that the paper states
- [x] Commit the final hyperparameters and plots
- [ ] Mark the PR as ready

jafermarq:
[...]

PR #2620, opened 2023-11-20
feat(baselines) Add `Flanders` baseline

edogab33:
## Issue
Implements #2595

### Description
Implementing FLANDERS baseline into Flower.


baselines/flanders/EXTENDED_README.md
jafermarq, line 1: you can remove this file once you are done going through it.

baselines/flanders/pyproject.toml
```diff
@@ -0,0 +1,137 @@
+[build-system]
+requires = ["poetry-core>=1.4.0"]
+build-backend = "poetry.masonry.api"
+
+[tool.poetry]
+name = "flanders" # <----- Ensure it matches the name of your baseline directory containing all the source code
+version = "1.0.0"
+description = "Flower Baselines"
```
jafermarq, line 8:
```suggestion
description = "FLANDERS"
```

jafermarq: or feel free to replace it with the whole title of your paper.

baselines/flanders/pyproject.toml
```diff
@@ -0,0 +1,137 @@
...
+black = "==23.1.0"
+docformatter = "==1.5.1"
+mypy = "==1.4.1"
+pylint = "==2.8.2"
+flake8 = "==3.9.2"
+pytest = "==6.2.4"
+pytest-watch = "==4.2.0"
+ruff = "==0.0.272"
+types-requests = "==2.27.7"
+
```
jafermarq, line 55:
```suggestion
virtualenv = "20.21.0"
```

jafermarq (commented):
Hi @edogab33,

Thanks for opening the PR for `FLANDERS`. I started reviewing it but I noted some parts of the `README` are not completed yet (which is fine!). Still, I added a couple of comments for the `pyproject.toml` (one of which will fix the current CI issue)

In addition to that:
* If you'd like to add some plots to your readme, please do so in a new directory named `_static` placed in the same directory as your README.
* We think it's better not to include the logs from the experiments (I see currently you are including the `outputs/` directory.
* Ideally, all the dependencies needed are added to the `pyproject.toml` so the environment can be built easily by simply doing `poetry install`. Depending on what packages you use, this might get tricky to achieve actually. If that's the case, it's fine provide additional instructions on how to install those packages **after** doing `poetry install`. Then you can install additional ones via the standard `pip install <>` method.

Ping me once there are some minimal instructions on how to setup the python environment and how to run an experiment. From that I can take a closer look at the code. Happy to discuss about your baseline anytime.

edogab33:
Hi @jafermarq, the baseline should be ready. Let me know if there are any issue or suggestions, I'll take care of them.

baselines/flanders/flanders/main.py
```diff
@@ -0,0 +1,168 @@
...
+    clients = {
+        "mnist": (MnistClient, mnist_evaluate),
+        "cifar": (CifarClient, cifar_evaluate),
+        "house": (HouseClient, house_evaluate),
+        "income": (IncomeClient, income_evaluate)
+    }
+
+    for dataset_name in cfg.dataset.name:
+        for attack_fn in cfg.server.attack_fn:
+            for num_malicious in cfg.server.num_malicious:
```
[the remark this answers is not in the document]
jafermarq: The idea is to keep the baseline more modular. Naturally, you could provide (as other baselines do) a single `run.sh` script to run all the experiments. But that script should contain two or more commands.

baselines/flanders/flanders/strategy.py
```diff
@@ -0,0 +1,234 @@
...
+        if server_round > 1:
+            win = self.window
+            if server_round < self.window:
+                win = server_round
+            M = load_all_time_series(dir="clients_params", window=win)
+            M = np.transpose(M, (0, 2, 1))  # (clients, params, time)
+
+            M_hat = M[:,:,-1].copy()
+            pred_step = 1
+            print(f"aggregate_fit - Computing MAR on M {M.shape}")
```
jafermarq, line 155: please replace all `print` statements in your baseline with `log(INFO, ...)` if you want them to be captured in the log. to do that you just need to import the below:
```python
from logging import INFO
```

baselines/flanders/README.md
```diff
@@ -0,0 +1,102 @@
+---
+title: A Byzantine-Resilient Aggregation Scheme for Federated Learning via Matrix Autoregression on Client Updates
+url: https://arxiv.org/abs/2303.16668
+labels: [robustness, model poisoning, anomaly detection, autoregressive model] # please add between 4 and 10 single-word (maybe two-words) labels (e.g. "system heterogeneity", "image classification", "asynchronous", "weight sharing", "cross-silo")
+dataset: [MNIST, Income, California Housing] # list of datasets you include in your baseline
+---
+
+****Paper:**** https://arxiv.org/abs/2303.16668
```
jafermarq, line 8:
```suggestion
****Paper:**** [arxiv.org/abs/2303.16668](https://arxiv.org/abs/2303.16668)
```

baselines/flanders/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## About this baseline
+
+****What’s implemented:**** The code in this directory replicates the results on MNIST and Income datasets under all attack settings (Gaussian, LIE, OPT and AGR-MM), but I've also implemented the code for California Housing and CIFAR-10.
+
+****Datasets:**** MNIST, Income
+
+****Hardware Setup:**** Apple M2 Pro, 16gb RAM
+
+****Contributors:**** Edoardo Gabrielli, University of Rome "La Sapienza"
```
jafermarq, line 23: if you want to add a link to any of your online profiles you can.

baselines/flanders/pyproject.toml
```diff
@@ -0,0 +1,143 @@
+[build-system]
+requires = ["poetry-core>=1.4.0"]
+build-backend = "poetry.masonry.api"
+
+[tool.poetry]
+name = "flanders" # <----- Ensure it matches the name of your baseline directory containing all the source code
+version = "1.0.0"
+description = "FLANDERS"
+license = "Apache-2.0"
+authors = ["The Flower Authors <hello@flower.dev>"]
```
jafermarq, line 10: maybe you want to add your "name <email>" here too.

baselines/flanders/pyproject.toml
```diff
@@ -0,0 +1,143 @@
...
+    "Topic :: Software Development :: Libraries",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+    "Typing :: Typed",
+]
+
+[tool.poetry.dependencies]
+python = ">=3.9, <3.12.0"  # pandas requires >=3.9
+hydra-core = "1.3.2" # don't change this
+flwr = {extras = ["simulation"], version = "1.6.0" }
+torch = "^2.1.1"
```
jafermarq, line 43:
```suggestion
torch = "2.1.1"
```

baselines/flanders/pyproject.toml
```diff
@@ -0,0 +1,143 @@
...
+pandas = "^2.1.3"
+scikit-learn = "1.3.2"
+torchvision = "^0.16.1"
```
jafermarq, line 46:
```suggestion
torchvision = "0.16.1"
```

baselines/flanders/README.md
```diff
@@ -0,0 +1,102 @@
...
+****Abstract:**** In this work, we propose FLANDERS, a novel federated learning (FL) aggregation scheme robust to Byzantine attacks. FLANDERS considers the local model updates sent by clients at each FL round as a matrix-valued time series. Then, it identifies malicious clients as outliers of this time series by comparing actual observations with those estimated by a matrix autoregressive forecasting model. Experiments conducted on several datasets under different FL settings demonstrate that FLANDERS matches the robustness of the most powerful baselines against Byzantine clients. Furthermore, FLANDERS remains highly effective even under extremely severe attack scenarios, as opposed to existing defense strategies. 
+
+
+## About this baseline
+
+****What’s implemented:**** The code in this directory replicates the results on MNIST and Income datasets under all attack settings (Gaussian, LIE, OPT and AGR-MM), but I've also implemented the code for California Housing and CIFAR-10.
+
+****Datasets:**** MNIST, Income
+
+****Hardware Setup:**** Apple M2 Pro, 16gb RAM
```
jafermarq, line 21: Could you provide an estimation of how long each experiment takes?

baselines/flanders/README.md
```diff
@@ -0,0 +1,102 @@
...
+****Paper:**** https://arxiv.org/abs/2303.16668
+
+****Authors:**** Gabriele Tolomei, Edoardo Gabrielli, Dimitri Belli, Vittorio Miori
+
+****Abstract:**** In this work, we propose FLANDERS, a novel federated learning (FL) aggregation scheme robust to Byzantine attacks. FLANDERS considers the local model updates sent by clients at each FL round as a matrix-valued time series. Then, it identifies malicious clients as outliers of this time series by comparing actual observations with those estimated by a matrix autoregressive forecasting model. Experiments conducted on several datasets under different FL settings demonstrate that FLANDERS matches the robustness of the most powerful baselines against Byzantine clients. Furthermore, FLANDERS remains highly effective even under extremely severe attack scenarios, as opposed to existing defense strategies. 
+
+
+## About this baseline
+
+****What’s implemented:**** The code in this directory replicates the results on MNIST and Income datasets under all attack settings (Gaussian, LIE, OPT and AGR-MM), but I've also implemented the code for California Housing and CIFAR-10.
```
jafermarq, line 17: Could you extend this description to point which Figure in the paper we can reproduce by running the code?

baselines/flanders/flanders/conf/base.yaml
```diff
@@ -0,0 +1,52 @@
...
+# a similar configuration structure and hence be easy to customise)
+#defaults:
+#  - _self_
+#  - override hydra/hydra_logging: disabled
+#  - override hydra/job_logging: disabled
+
+hydra:
+  output_subdir: null
+  run:
+    dir: .
```
jafermarq, line 13: it's preferable if outputs go to the directory automatically generated by hydra (that's how all baselines do it). You are welcome to rename that directory (defaults to `outputs/`) if you don't like it.

jafermarq (commented):
Hi @edogab33,

Thanks for the updates. I have done a preliminar review of your baseline. See some comments below this message. In addition to those:
* when i ran your code I see quite a few messages of the type: `ConvergenceWarning: Liblinear failed to converge, increase the number of iterations.` is this expected?
* Don't forget to run the formatting script by: (with your environment activated) `cd ..`, then `./dev/format-baseline.sh flanders`. This will automatically correct quite a few small issue, and highlight many more. Generally these are easy to fix.
* Once the above is completed, you'll have to run the ./dev/test-baseline.sh flanders` command. Likely other issues will be highlighted (again, easy to fix for the most part). But do ping me if some of them are not so easy to fix.

I think the main TODO is to change a bit the logic in your `main.py` so only one experiment runs each time `main.py` is executed. This helps to keep the code modular (and in line with how other baselines work).  Also please use the direcotry hydra creates automatically to save the experiment results (that's where the log will go too). This might require you to change the code you have to generate the plots. 

Maybe you want to take a look at how other baselines have done the above. For example: `FedProx`: https://github.com/adap/flower/tree/main/baselines/fedprox

Let me know if you'd like to discuss how to achieve the above. It's not hard but can't be done in 30mins.

baselines/flanders/flanders/conf/base.yaml
edogab33: Now I save it here `save_path = HydraConfig.get().runtime.output_dir`, however I don't use the history that the strategy outputs as FedProx does; I still delegate this to the strategy by means of the evaluation function. Do you think it is better to use the history?

edogab33:
Thanks for the review @jafermarq!

>     * when i ran your code I see quite a few messages of the type: `ConvergenceWarning: Liblinear failed to converge, increase the number of iterations.` is this expected?

We use only one local epoch and it fails to converge. That's normal.

>     * Don't forget to run the formatting script by: (with your environment activated) `cd ..`, then `./dev/format-baseline.sh flanders`. This will automatically correct quite a few small issue, and highlight many more. Generally these are easy to fix.
>
>     * Once the above is completed, you'll have to run the ./dev/test-baseline.sh flanders` command. Likely other issues will be highlighted (again, easy to fix for the most part). But do ping me if some of them are not so easy to fix.

I'm on the process of correcting these issues and I'm ready for another review.

edogab33:
I had to suppress lot of mypy errors on the code that I copied from FedAvg.

edogab33:
@jafermarq I see that here pylint doesn't pass (`flanders/utils.py:11:0: E0401: Unable to import 'natsort' (import-error)`), however locally it runs without warnings. Any clue?

baselines/flanders/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+It will generate the results in `results/all_results.csv`. To generate the plots, use the notebook in `plotting/plots.ipynb`.
+
+Expected maximum accuracy achieved across different number of malicious clients and different attacks:
+![](_static/max_acc.jpg)
+
+Expected distribution of accuracy from round $L$ to the final round across all the experiments:
+![](_static/boxplot_Income.jpg)
+
+![](_static/boxplot_MNIST.jpg)
```
edogab33, line 119: @jafermarq here the MNIST boxplot is not coming up on GitHub. Any idea about why this happens?

jafermarq: It seems adding a new line between the sentence and the line inserting `boxplot_Income.jpg` fixed it.

jafermarq:
> @jafermarq I see that here pylint doesn't pass (`flanders/utils.py:11:0: E0401: Unable to import 'natsort' (import-error)`), however locally it runs without warnings. Any clue?

I added `natsort` to `pyproject.toml` and now the test pass.

baselines/flanders/flanders/strategy.py
```diff
@@ -0,0 +1,419 @@
...
+
+WARNING_MIN_AVAILABLE_CLIENTS_TOO_LOW = """
+Setting `min_available_clients` lower than `min_fit_clients` or
+`min_evaluate_clients` can cause the server to fail when there are too few clients
+connected to the server. `min_available_clients` must be set to a value larger
+than or equal to the values of `min_fit_clients` and `min_evaluate_clients`.
+"""
+
+
+class Flanders:
```
jafermarq, line 35: Wouldn't it be better to make `Flanders` be a strategy class inheriting from Flower's `FedAvg`? This would make your baseline easier to maintain. Also, you wouldn't need to re-implement `num_fit_clients`, `num_evaluation_clients`, `initialize_parameters`, `evaluate` and `aggregate_evaluate` evaluate methods...

baselines/flanders/flanders/strategy.py
```diff
@@ -0,0 +1,419 @@
...
+    @typing.no_type_check
+    def initialize_parameters(
+        self, client_manager: ClientManager
+    ) -> Optional[Parameters]:
+        """Initialize global model parameters."""
+        initial_parameters = self.initial_parameters
+        self.initial_parameters = None  # Don't keep initial parameters in memory
+        return initial_parameters
+
+    @typing.no_type_check
```
jafermarq, line 168: I'm inclined to say It's better to try not to use this 😅. Are there some typing issues hard to solve in your implementation?

baselines/flanders/flanders/main.py
```diff
@@ -0,0 +1,172 @@
...
+    # 4. Define your strategy
+    strategy = instantiate(
+        cfg.strategy,
+        evaluate_fn=clients[dataset_name][1],
+        on_fit_config_fn=fit_config,
+        fraction_fit=1,
+        fraction_evaluate=0,
+        min_fit_clients=cfg.server.pool_size,
+        min_evaluate_clients=0,
+        warmup_rounds=cfg.server.warmup_rounds,
```
jafermarq, line 129: I believe this shouldn't be passed.

baselines/flanders/flanders/conf/base.yaml
```diff
@@ -0,0 +1,33 @@
...
+  num_rounds: 50
+  pool_size: 10
+  warmup_rounds: 30
+  sampling: 500
+  history_dir: clients_params
+  magnitude: 100
+  threshold: 1e-05
+  attack_fn: gaussian
+  num_malicious: 0
+  omniscent: True
```
jafermarq, line 33:
```suggestion
  omniscent: True
 
 server_device: cpu
 
 client_resources:
   num_cpus: 10
   num_gpus: 0.0
  
```

jafermarq: It's desirable to keep the `client_resources` as part of the config, so it's easier to modify from the command line. There is a comment related to this in `main.py`

baselines/flanders/flanders/main.py
```diff
@@ -0,0 +1,172 @@
...
+        min_available_clients=cfg.server.pool_size,
+        window=cfg.server.warmup_rounds,
+        distance_function=l2_norm,
+        maxiter=cfg.strategy.maxiter,
+    )
+
+    # 5. Start Simulation
+    fl.simulation.start_simulation(
+        client_fn=client_fn,
+        client_resources={"num_cpus": 10},
```
jafermarq, line 140:
```suggestion
        client_resources=cfg.client_resources,
```

jafermarq: See my comment in `conf.py`

baselines/flanders/README.md
```diff
@@ -0,0 +1,121 @@
...
+| MNIST | 10 | 100 | 50 | 30 | 32 | $10^{-3}$ | \ | \ | Adam | 0.2 | 0.0 | 0.0 | 1 | \ |
+
+Where $\lambda_1$ and $\lambda_2$ are Lasso and Ridge regularization terms, $L$ is the number of warmup rounds (rounds without any attack) and the number of malicious clients ranges from 0 to 9.
+
+
+## Environment Setup
+
+```bash  
+# Use a version of Python >=3.9 and <3.12.0.
+poetry env use 3.9.18
```
jafermarq, line 69:
```suggestion
pyenv local 3.10.12
poetry env use 3.10.12
```

jafermarq: just a bit more future proof (this is also what i'm using to test `FLANDERS`.

baselines/flanders/pyproject.toml
```diff
@@ -0,0 +1,144 @@
...
+[tool.poetry.dependencies]
+python = ">=3.9, <3.12.0"  # pandas requires >=3.9
+hydra-core = "1.3.2" # don't change this
+flwr = {extras = ["simulation"], version = "1.6.0" }
+torch = "2.1.1"
+pandas = "^2.1.3"
+scikit-learn = "1.3.2"
+torchvision = "0.16.1"
+ipykernel = "^6.27.1"
+natsort = "^8.4.0"
```
jafermarq, line 48:
```suggestion
natsort = "^8.4.0"
seaborn = "^0.13.0"
```

jafermarq: for plotting.

baselines/flanders/plotting/plots.ipynb
jafermarq, line 1: Paths to `.csv` need to be updated from:

```
'../results/all_results.csv'
```

to:

```
'../outputs/all_results_original.csv'
```

jafermarq (commented):
Hey @edogab33,

Thanks for updating `main.py` and how results are saved. I have added a few comments below. You'll see also I fixed the plotting in the `README.md` and now all tests pass. 

I still think there is a lot of value that can be added to this baseline if you could add at least one other strategy. All Flower baselines compare to at least one other strategy or method. Both [Bulyan](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/bulyan.py) and [Krum](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/krum.py) are implemented in Flower, so adding them as part of the `--multirun` shouldn't be too hard. What often works best is to use the `defaults:` structure in Hydra configs and have the strategy-specific part of the config in its own directory under `conf/strategy`. You can see how [FedNova](https://github.com/adap/flower/tree/main/baselines/fednova) did this. If these experiments become more computationally expensive to run. 

Finally, the target platform for all baselines is Ubuntu 22.04 x86 and ideally it would run both on CPU and GPU. We can leave this until the end. 

Let me know if you need some help with the above.

Merry Christmas!

baselines/flanders/flanders/conf/base.yaml
jafermarq: I think this is fine as long as all outputs that need to do plotting later are stored. I see you generate a `outputs/all_results.csv`. I guess this is sufficient?

baselines/flanders/flanders/strategy.py
edogab33: That what I used to do, but mypy gets angry because some methods do not have the same signature of FedAvg (if I recall correctly, the error was ` error: Signature of [method] incompatible with supertype "FedAvg".`). I think that it thinks that I am overriding, but I'm actually overloading and it should be a safe operation AFAIK

baselines/flanders/flanders/strategy.py
edogab33: It gives incompatible type errors, although it is code copied from FedAvg.

baselines/flanders/flanders/main.py
edogab33: You're right. Forgot to delete

edogab33:
> Hey @edogab33,
> 
> I still think there is a lot of value that can be added to this baseline if you could add at least one other strategy. All Flower baselines compare to at least one other strategy or method. Both [Bulyan](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/bulyan.py) and [Krum](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/krum.py) are implemented in Flower, so adding them as part of the `--multirun` shouldn't be too hard. What often works best is to use the `defaults:` structure in Hydra configs and have the strategy-specific part of the config in its own directory under `conf/strategy`. You can see how [FedNova](https://github.com/adap/flower/tree/main/baselines/fednova) did this. If these experiments become more computationally expensive to run.

Thanks for the insight! Let me see how I can do what you suggest.

> Another thing, is it posible to not have as part of the baseline the contents of `flanders/dataset_files` (i.e. can these be downloaded/preprocessed by the person running the baseline)? Those files add up to ~8MB which is not so little.

Unfortunately I don't remember where I've downloaded them and the House dataset was the result of a notebook found somewhere... so there were no automatic way to download it. The best I can do is to compress them and unzip locally in the main, so they should be way less heavy in the repo.

> Merry Christmas!

Merry Christmas to you and thanks again for your review!

baselines/flanders/flanders/strategy.py
jafermarq: ok. from a quick look at this, it seems the only source of incompatibility is passing `client_state` to `aggregate_fit`. You could pass this from your server using a new method for example (that can set a `self.client_state` inside the strategy and then access that when needed in `aggregate_fit`)

baselines/flanders/flanders/strategy.py
jafermarq: ok. I'll work out something and propose the change.

jafermarq:
> > Hey @edogab33,
> > I still think there is a lot of value that can be added to this baseline if you could add at least one other strategy. All Flower baselines compare to at least one other strategy or method. Both [Bulyan](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/bulyan.py) and [Krum](https://github.com/adap/flower/blob/main/src/py/flwr/server/strategy/krum.py) are implemented in Flower, so adding them as part of the `--multirun` shouldn't be too hard. What often works best is to use the `defaults:` structure in Hydra configs and have the strategy-specific part of the config in its own directory under `conf/strategy`. You can see how [FedNova](https://github.com/adap/flower/tree/main/baselines/fednova) did this. If these experiments become more computationally expensive to run.
> 
> Thanks for the insight! Let me see how I can do what you suggest.
> 
> > Another thing, is it posible to not have as part of the baseline the contents of `flanders/dataset_files` (i.e. can these be downloaded/preprocessed by the person running the baseline)? Those files add up to ~8MB which is not so little.
> 
> Unfortunately I don't remember where I've downloaded them and the House dataset was the result of a notebook found somewhere... so there were no automatic way to download it. The best I can do is to compress them and unzip locally in the main, so they should be way less heavy in the repo.
> 
> > Merry Christmas!
> 
> Merry Christmas to you and thanks again for your review!

ok! Let's not worry about the datasets for now. About using another strategy as point of comparison with `FLANDERS`. Maybe let's wait a bit until we have `FLANDERS` strategy API to match that of other strategies (like `FedAvg`). This is important else your custom server won't be compatible with other startegies.

baselines/flanders/flanders/strategy.py
edogab33: I realised that `client_state` in `aggregate_fit` was a typo. Now only the `evaluate` method is an overloading. I managed to solve the issue with mypy by using keyword arguments with default values in the signature, but pylint still complains: `W0221: Parameters differ from overridden 'evaluate' method (arguments-differ)`. At this point either I ignore the warning or I implement a new way of saving the results (the way the evaluate methods are designed are to first evaluate the model and then save the results in the csv file, that's why I need two more arguments). Maybe I can get this working by using the history returning from `start simulation()`... but I'm not sure.

baselines/flanders/flanders/conf/strategy/flanders.yaml
```diff
@@ -0,0 +1,7 @@
+---
+name: flanders
+
+strategy:
+  _target_: flanders.strategy.Flanders
+  to_keep: 1  # number of benign local models to keep for the aggregation (FedAvg)
```
edogab33, line 6: @jafermarq Sorry to bother you, but I can't wrap my head around this: how can I access a variable of this file from `main.py`?

jafermarq: It should be something like:
```bash
python -m flanders.main strategy.strategy.to_keep=2
```
The first `strategy` is because that's the name of the "node" in `defaults`. The second one is to access the `strategy` node in `flanders.yaml`. 

Does it work?

edogab33: Works. Thanks.

baselines/flanders/flanders/server.py
```diff
@@ -0,0 +1,360 @@
...
+
+        # Sort clients states
+        clients_state = {k: clients_state[k] for k in sorted(clients_state)}
+        log(INFO, "Clients state: %s", clients_state)
+
+        # Aggregate training results
+        log(INFO, "fit_round - Aggregating training results")
+        aggregated_result = self.strategy.aggregate_fit(server_round, results, failures)
+
+        if isinstance(self.strategy, Flanders):
```
edogab33, line 319: This is needed because `aggregate_fit()` of `FLANDERS` returns extra stuff that other strategies don't. 
I don't know if you need also the return variables to be the same, but I think this is the best solution. The alternative is to remove the extra return values `good_clients_idx` and `malicious_clients_idx` from `FLANDERS` and delegate part of the params saving mechanism to the strategy itself; however I think that splitting this task into two different classes is confusing.

edogab33:
Hi @jafermarq, happy new year!

I think that `FLANDERS` APIs are ok now. I left a comment for the only thing that changes with respect to other strategies.

I also splitted the config files to fit other strategies as requested. I will make experiments with Krum and include them in the plots as soon as I have the time to do so, but we can go ahead with other issues in the meanwhile.

Let me know what you think. Thank you.

baselines/flanders/plotting/plots.ipynb
```diff
@@ -0,0 +1,636 @@
...
+    "def trim_attack_fn(df):\n",
+    "    \"\"\"In column \"attack_fn\" of dataframe df, remove all\n",
+    "    superfluous characters from the attack name, leaving only\n",
+    "    the attack name itself.\n",
+    "    \"\"\"\n",
+    "    names = [\"gaussian_attack\", \"lie_attack\", \"fang_attack\", \"minmax_attack\"]\n",
+    "    # remove superfluous characters from attack names in df\n",
+    "    for name in names:\n",
+    "        df.loc[df[\"attack_fn\"].str.contains(name), \"attack_fn\"] = name\n",
+    "    names = {\"gaussian_attack\": \"GAUSS\", \"lie_attack\": \"LIE\", \"fang_attack\": \"OPT\", \"minmax_attack\": \"AGR-MM\"}\n",
```
jafermarq, line 32: The generated `outputrs/all_results.csv` doesn't contain those keys with the `_attack` suffix. I believe that line should be replaced with:
```
    names = {"gaussian": "GAUSS", "lie": "LIE", "fang": "OPT", "minmax": "AGR-MM"}
```

baselines/flanders/plotting/plots.ipynb
```diff
@@ -0,0 +1,636 @@
...
+    "    \"\"\"To save time, I didn't run experiments with num_malicious = 0\n",
+    "    for all datasets. This function adds those missing rows to the\n",
+    "    dataframe.\n",
+    "    \"\"\"\n",
+    "    df_lie = df.loc[df[\"num_malicious\"] == 0].assign(attack_fn=\"LIE\", ignore_index=True)\n",
+    "    df_opt = df.loc[df[\"num_malicious\"] == 0].assign(attack_fn=\"OPT\", ignore_index=True)\n",
+    "    df_agr = df.loc[df[\"num_malicious\"] == 0].assign(attack_fn=\"AGR-MM\", ignore_index=True)\n",
+    "    df = pd.concat([df, df_lie, df_opt, df_agr], ignore_index=True)\n",
+    "    return df\n",
+    "\n",
```
jafermarq, line 61: I think we can remove this function and run the experiment with 0 attackers.

jafermarq (commented):
Hi @edogab33,

I made some edits to how the results are saved: now a csv is stored in the subdirectory of each experiment but also they are appended to the file `outputs/all_results.csv` as you originally had. I've run the big _multirun_ but encountered that experiments with `fang` crashed sooner or later with both `MNIST` and `income` dataset and all number of malicious (0-9). This is the error I see:

```bash
[2024-01-27 22:31:04,285][flwr][ERROR] - arrays used as indices must be of integer (or boolean) type
[2024-01-27 22:31:04,286][flwr][ERROR] - Traceback (most recent call last):
  File ".../flwr/simulation/app.py", line 308, in start_simulation
    hist = run_fl(
  File ".../flwr/server/app.py", line 225, in run_fl
    hist = server.fit(num_rounds=config.num_rounds, timeout=config.round_timeout)
  File ".../flanders/baselines/flanders/flanders/server.py", line 122, in fit
    res_fit = self.fit_round(
  File ".../flanders/baselines/flanders/flanders/server.py", line 271, in fit_round
    results, others = self.attack_fn(
  File ".../flanders/baselines/flanders/flanders/attacks.py", line 187, in fang_attack
    _, _, _, distances = _krum(all_params, num_corrupted, 1)
  File ".../flanders/baselines/flanders/flanders/attacks.py", line 417, in _krum
    closest_indices = _get_closest_indices(
  File ".../flanders/baselines/flanders/flanders/attacks.py", line 463, in _get_closest_indices
    np.argsort(distance_matrix[idx])[1 : num_closest + 1].tolist()
IndexError: arrays used as indices must be of integer (or boolean) type
```
For all the other experiments, I did some small modifications to the notebook (see comments below) and this is what i obtained. They are a bit different compared to those in your readme...



Finally, would it be possible to remove the `dataset_files` directory and instead add a link to your google drive (or similar) ? Then include some minimal instructions in the `README.md` telling people how to download the dataset, how to decompress, and where to place it. Currently it's 8MB which is a bit big

Ping me when i can review this again!

edogab33:
Hey @jafermarq, sorry for the wait. We've been busy running new experiments and reworking most of the paper, which means we've also changed the code. We'll be updating the arXiv soon so you can see our latest results. In the meantime, I'll get started on updating the code in this pull request

baselines/flanders/flanders/server.py
edogab33: Removed in the new version.

edogab33:
@jafermarq I've updated the code, now it's alligned with the most recent version that I've written. Let me update the README and test everything before starting a code review.

jafermarq:
Hi @edogab33 , i fixed a couple of small formatting errors. All tests pass. Let me know if there is something pending you'd like to add to your baseline!

edogab33:
> Hi @edogab33 , i fixed a couple of small formatting errors. All tests pass. Let me know if there is something pending you'd like to add to your baseline!

Thanks a lot @jafermarq. I'll add a new notebook and perform the experiments again.

jafermarq:
> > Hi @edogab33 , i fixed a couple of small formatting errors. All tests pass. Let me know if there is something pending you'd like to add to your baseline!
> 
> Thanks a lot @jafermarq. I'll add a new notebook and perform the experiments again.

Great! Please ping me when ready or if you'd like me to look into something. 🙌

jafermarq:
What about this one @edogab33 ?

edogab33:
@jafermarq I'm working on it now. Experiments are almost done, I just have to update the notebook for the plots and the README. Things about this project changed a lot, here's the new version: https://arxiv.org/pdf/2303.16668

jafermarq:
> @jafermarq should I rename the title of the PR in: `feat(baselines) Add FLANDERS`?

yes please. maybe "`feat(baselines) Add FLANDERS baseline`"?

edogab33:
Hi @jafermarq, everything is ready. However, a test is not passing. Can't get why

edogab33:
@jafermarq do you know what's the problem with the test?

jafermarq:
> @jafermarq do you know what's the problem with the test?

All seems good now. I'll review soon.

edogab33:
> > @jafermarq do you know what's the problem with the test?
> 
> All seems good now. I'll review soon.

Thanks! :)

jafermarq (commented):
Hi @edogab33,

Thank you for for pushing your baseline forward. Sorry for the delay in my review after your updates. Below i leave a couple of comments. Let me know what you think.

I'm now re-running all experiments as you indicate in your `Expected Results` section. I'll approve the PR shortly after if everything looks alright!

baselines/flanders/run.sh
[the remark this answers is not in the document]
edogab33: > @edogab33 Should we mention this `run.sh` in the `README.md` ? how does it relate to the commands in the `Expected Results` section?

Definitely, that's the file to run all the experiments of the paper. I'll add it to the README

edogab33:
@jafermarq Thanks for the review! Hope the results are pretty much the same. 

I've addressed your issues and realized that if the .sh script is run instead of launching batches experiments individually, you'll get all results into one file. At the same time, the notebook expects that each experiment has a different one (i.e. `all_results_[dataset]_[flanders/no_flanders].csv`). 
Now the notebook should automatically organize results into the files it expects.

Hope to hear from you soon

jafermarq:
> @jafermarq Thanks for the review! Hope the results are pretty much the same as mine.
> 
> I've addressed your issues and realized that if the .sh script is run instead of launching batches of experiments individually, you'll get all results into one file. At the same time, the notebook expects that each experiment has a different one (i.e. `all_results_[dataset]_[flanders/no_flanders].csv`). Now the notebook should automatically organize results into the files it expects.
> 
> Hope to hear from you soon

Thanks! but to generate the results you show in the read me should I run the two commands in the `Expected Results` section or should I run those in `run.sh` instead? I'm doing the former but please let me know.

baselines/flanders/flanders/main.py
```diff
@@ -0,0 +1,277 @@
...
+
+    rounds, test_loss = zip(*history.losses_centralized)
+    _, test_accuracy = zip(*history.metrics_centralized["accuracy"])
+    _, test_auc = zip(*history.metrics_centralized["auc"])
+    _, truep = zip(*history.metrics_centralized["TP"])
+    _, truen = zip(*history.metrics_centralized["TN"])
+    _, falsep = zip(*history.metrics_centralized["FP"])
+    _, falsen = zip(*history.metrics_centralized["FN"])
+
+    path_to_save = [os.path.join(save_path, "results.csv"), "outputs/all_results.csv"]
```
jafermarq, line 238: I think the save in `outputs/all_results.csv` is going to fail if no `outputs/` dir exist. This is the case actually if a person runs your `--multirun` commands directly (which don't create the `outputs/` dir).

edogab33: Right! Good point

edogab33:
> Thanks! but to generate the results you show in the read me should I run the two commands in the `Expected Results` section or should I run those in `run.sh` instead? I'm doing the former but please let me know.

Good point, that's an oversight from my side. 

The best way --now that I've updated the notebook to divide the results automatically-- is to run the bash script. For this reason, I removed the former way to run the experiments to avoid mistakes. However, if you've already launched the experiments with that method it's perfectly fine, note that you should run them also for MNIST though.

By the way, as you suggested, I updated the `main.py` to create the output dir if not present.

jafermarq (approved):
It's great to have `FLANDERS` join Flower baselines! Thanks for the hard work @edogab33!

PR #2629, opened 2023-11-23
Update the dataset creation docs

adam-narozniak:
## Issue

The docs show how to divide the dataset into train and test sets. The indexing it proposes does not keep the dataset but is creates a dictionary which might come as a surprise to users.

## Proposal

Change the displayed way such that the Dataset is kept after the indexing by using `.select`


datasets/doc/source/how-to-use-with-pytorch.rst
```diff
@@ -70,8 +70,8 @@ If you want to divide the dataset, you can use (at any point before passing the
 Or you can simply calculate the indices yourself::
 
   partition_len = len(partition)
-  partition_train = partition[:int(0.8 * partition_len)]
-  partition_test = partition[int(0.8 * partition_len):]
+  partition_train = partition.select(range(int(0.8 * partition_len)))
+  partition_test = partition.select(range(int(0.8 * partition_len), partition_len))
```
jafermarq, line 74:
```suggestion
  # Split `partition` 80:20
  num_train_examples = int(0.8 * partition_len)
  partition_train = partition.select(range(num_train_examples)) ) # use first 80% 
  partition_test = partition.select(range(num_train_examples, partition_len)) ) # use last 20%
```

jafermarq: Let's not do it if you think is redundant.

adam-narozniak: Works for, let's have it explicit

jafermarq (approved):

danieljanes (approved):

PR #2636, opened 2023-11-25
Make `flower-client` HTTPS by default

panh99:
- Added `--insecure` and `--root-certificates` arguments for `flower-client`
- Modified `run_client` function to run with HTTPS-enabled by default. 
- Added to all `client.py` in E2E tests a `Flower` callable.
- Updated `e2e/test_driver.sh` to run clients via `flower-client`.
- Added `insecure` arg to `start_client`, `start_numpy_client`, `connection`, and `create_channel`.

### Note
- `flower-client` will start with HTTPS-enabled by default.
- `start_client` and `start_numpy_client` will still start in the insecure mode by default for backward compatibility.
- `flower-client` will fail when both `--insecure` and `--root-certificates` are set.


src/py/flwr/client/app.py
```diff
@@ -18,9 +18,12 @@
 import argparse
 import sys
 import time
-from logging import INFO
+from logging import INFO, WARN
+from pathlib import Path
 from typing import Callable, ContextManager, Optional, Tuple, Union
 
+import certifi
```
tanertopal, line 25: @danieljanes @panh99 Maybe I am missing something, but this is not a direct dependency (`requests` for example installs it). We should add it as such if we want to use it (which seems reasonable). It seems like a small utility library so I would be fine with it although

panh99: @tanertopal @danieljanes I just wanted to know your opinion on using the `certifi` library. Using `certifi` allows us to make minimal modifications to existing code. Alternatively, we could load system certificates bundled with gRPC, but that would require us to add an additional argument, such as enable_https, to specify whether to load system certificates when no certificates are provided. We would need to add enable_https to start_client, connection, and create_channel.

panh99: Added use_grpc_certificates arg to start_client, start_numpy_client, connection, and create_channel.

panh99: Renamed `use_grpc_certificates` to `insecure` and updated the logic.

src/py/flwr/client/app.py
```diff
@@ -75,6 +91,20 @@ def _parse_args_client() -> argparse.ArgumentParser:
...
 
+    parser.add_argument(
+        "--insecure",
+        action="store_true",
+        help="Run the client without HTTPS, regardless of whether certificate "
+        "paths are provided. By default, the client runs with HTTPS enabled. "
+        "Use this flag only if you understand the risks.",
+    )
+    parser.add_argument(
+        "--root_certificates",
```
danieljanes, line 102:
```suggestion
        "--root-certificates",
```

src/py/flwr/client/app.py
```diff
@@ -50,6 +51,19 @@ def run_client() -> None:
 
     args = _parse_args_client().parse_args()
 
+    # Obtain certificates
```
danieljanes, line 54: @panh99 can we add a check here to fail if both `--insecure` and `--root-certificates` are set?

src/py/flwr/common/grpc.py
```diff
@@ -38,14 +39,14 @@ def create_channel(
...
-    if root_certificates is not None:
+    if insecure:
+        channel = grpc.insecure_channel(server_address, options=channel_options)
+        log(INFO, "Opened insecure gRPC connection (no certificates were passed)")
+    else:
         ssl_channel_credentials = grpc.ssl_channel_credentials(root_certificates)
         channel = grpc.secure_channel(
             server_address, ssl_channel_credentials, options=channel_options
         )
         log(INFO, "Opened secure gRPC connection using certificates")
```
tanertopal, line 50: Should we throw an exception if `insecure` is `True` and `root_certificates` are passed along?

panh99: Updated.

src/py/flwr/client/app.py
panh99: Added.

danieljanes (approved):

tanertopal (approved):

PR #2632, opened 2023-11-23
Inject `WorkloadState` into instantiated client

jafermarq:
Passing `WorkloadState` so it's reachable by an instantiated client.

PR #2643, opened 2023-11-28
Add `WorkloadState` to virtual clients

jafermarq:
feature parity, as far as `WorkloadState` is concerned, between normal and virtual clients.

Assumes PR #2632 is merged.


src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -411,4 +424,5 @@ def get_client_result(self, cid: str, timeout: Optional[float]) -> ClientRes:
                 break
 
         # Fetch result belonging to the VirtualClient calling this method
+        # Return both result from taks and (potentially) updated workload state
```
danieljanes, line 427:
```suggestion
        # Return both result from tasks and (potentially) updated workload state
```

danieljanes (approved):

PR #2644, opened 2023-11-28
Add automatic docs

adam-narozniak:
## Issue
A single-page API documentation (according to me) is too long and hard to navigate.

## Proposal 
Introduce automatic docs creation using `autosummary` extension from `sphinx` where each class and function has a separate page.

### Description
* Update index.rst for the new docs generation
* Remove old docs `ref-api-flwr` (single docs file)
* Add autosummary templates (they dictate how each file looks like)
* (Keep the CLI docs; note that the functions are double - the reference is also present in the newly created docs but there are no argaparse arguments)
* Configure autosummary in `conf.py`

### Questions
Do we need to store the .rst files with documentation (I see some git diffs in the scripts for building the docs but I'm unsure if that affect this part)

### Diff between flwr and flwr-datasets
1) __init __ problem
In `flwr`, proper docs of SOME classes are present in the `__init__` and right after the class ClassName (I find it especially true in `Strategies`). In `flwr-datasets` added docs just after the class statement (no docs in __init __).
That implies that the parameters to the constructor are doubled in documentation (that was also the case in old docs); see the image below.
To account for that, I changed slightly the class template (to have the __init __ docs added).

PEP 257 says that the class Docstrings should be (immediately following the class definition) - the most common option in our docs. The constructor (__init __) Docstrings is optional. I think it should be removed in our case form all the classes subclassing `Strategy` and potentially others (if present).

image

2) new base template 
the base template will need to be added to `flwr-datasets`. The need for it was not apparent because there was none standalone function documentation. It is required for consistent naming conventions.


danieljanes:
@adam-narozniak thanks for the description.

Re Q1: I agree with your proposal. Let's use the class docstring as the main docstring going forward (instead of the __init__ docstring). This includes fixing it for classes where this isn't the case yet.

Re Q2: Also yes. Let's keep the approach and templates exactly the same across `flwr` and `flwr-datasets`.

danieljanes (approved):

PR #2658, opened 2023-11-29
Rename draft release workflow and create release

charlesbvll:
### Description

The current `release` workflow doesn't actually release the package to PyPi, it just creates a draft.

### Related issues/PRs

N/A

### Explanation

This new script (named `framework-release.yml`, while the old one is renamed `framework-draft-release.yml`) downloads the artifacts from the artifact store and publishes the package to PyPi once the draft is set to `published`.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


.github/workflows/framework-release.yml
```diff
@@ -1,63 +1,43 @@
-name: Release Framework
+name: Publish release
```
danieljanes, line 1:
```suggestion
name: Publish `flwr` release on PyPI
```

.github/workflows/framework-release.yml
```diff
@@ -1,63 +1,43 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+  
 jobs:
   publish:
-    if: ${{ github.repository == 'adap/flower' }}
-    name: Publish draft
+    name: Publish release
     runs-on: ubuntu-latest
```
danieljanes, line 14: Wouldn't it be better to target a specific Ubuntu version here?

.github/workflows/framework-release.yml
```diff
@@ -1,63 +1,43 @@
...
-      - name: Release
-        uses: softprops/action-gh-release@v1
-        with:
-          body_path: ./body.md
-          draft: true
-          name: Flower ${{ env.TAG_NAME }}
-          files: |
-            ${{ env.WHEEL_NAME }}
-            ${{ env.TAR_NAME }}
+        python -m poetry publish -u __token__ -p $PYPI_TOKEN
```
danieljanes, line 43: Should we update `dev/publish.sh` and use it here, or should we remove it altogether?

charlesbvll: In my opinion, such a simple line doesn't necessarily require a standalone file. Especially because people won't have to use the `/dev/publish.sh` script, so there is no advantage of providing it (and if they do, it's easier to have the `publish` command without providing arguments). But I guess @tanertopal will have the final say on this.

.github/workflows/framework-release.yml
```diff
@@ -1,63 +1,44 @@
...
+
+    - name: Bootstrap
+      uses: ./.github/actions/bootstrap
+
+    - name: Get artifacts and publish
+      env:
+        GITHUB_REF: ${{ github.ref }}
+        PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+      run: |
+        TAG_NAME=$(echo "${GITHUB_REF#refs/tags/}" | cut -c2-)
```
Robert-Steiner, line 30: can we use `$GITHUB_REF_NAME` as we do in the framework-draft-release?

.github/workflows/framework-release.yml
```diff
@@ -1,63 +1,44 @@
...
-      - name: Release
-        uses: softprops/action-gh-release@v1
-        with:
-          body_path: ./body.md
-          draft: true
-          name: Flower ${{ env.TAG_NAME }}
-          files: |
-            ${{ env.WHEEL_NAME }}
-            ${{ env.TAR_NAME }}
+        python -m poetry publish -u __token__ -p $PYPI_TOKEN
```
Robert-Steiner, line 44: I would prefer to not put the token into `env` (line 28) because all commands in the bash script will have access to it and it is only `poetry` that needs it.
```suggestion
        python -m poetry publish -u __token__ -p ${{ secrets.PYPI_TOKEN }}
```

Robert-Steiner (commented):
Look good to me! I added two small suggestions.

.github/workflows/framework-release.yml
tanertopal: @charlesbvll I am fine removing it. We don't want manual releases in the future anyway, so avoiding having that might be even better. It might be good to document how to do a manual release.

.github/workflows/framework-release.yml
```diff
@@ -1,63 +1,44 @@
...
+        TAG_NAME=$(echo "${GITHUB_REF#refs/tags/}" | cut -c2-)
+
+        wheel_name="flwr-${TAG_NAME}-py3-none-any.whl"        
+        tar_name="flwr-${TAG_NAME}.tar.gz"
+
+        wheel_url="https://artifact.flower.dev/py/release/v${TAG_NAME}/${wheel_name}"
+        tar_url="https://artifact.flower.dev/py/release/v${TAG_NAME}/${tar_name}"
+        
+        echo $wheel_url
+        echo $tar_url
```
tanertopal, line 39:
```suggestion
```

.github/workflows/framework-release.yml
charlesbvll: Good catch!

danieljanes (approved):

PR #2664, opened 2023-11-30
Add background task

danielnugraha:
### Description

In Android, FL happens in foreground, which is not ideal since it should be happening in the background.

### Explanation

Add periodic background task that schedules flwr workloads periodically based on a given interval.

### Checklist

- [x] Write tests
- [x] Make CI checks pass


tanertopal (approved):

src/kotlin/flwr/src/main/java/dev/flower/android/Worker.kt
```diff
@@ -0,0 +1,51 @@
...
+            inputData.getString("serverAddress")?.let { startFlowerRere(it, false, flwrClient) }
+        } else {
+            inputData.getString("serverAddress")?.let { startClient(it, false, flwrClient) }
+        }
+        val workManager: WorkManager = WorkManager.getInstance(context)
+        workManager.cancelWorkById(id)
+        return Result.success()
+    }
+
+    fun startFlwrPeriodicWorker(interval: Long, serverAddress: String) {
```
tanertopal, line 30: @danielnugraha can you add documentation which clarifies the format of `serverAddress` and TimeUnit for `interval` this function expects?

tanertopal (changes requested):

src/kotlin/flwr/src/main/java/dev/flower/android/Worker.kt
danielnugraha: done.

src/kotlin/flwr/src/main/java/dev/flower/android/Worker.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.work.PeriodicWorkRequest
+import androidx.work.PeriodicWorkRequestBuilder
+import androidx.work.WorkManager
+import androidx.work.WorkerParameters
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+
+class FlwrWorker(private val context: Context,
+                 workerParams: WorkerParameters,
+                 private val flwrClient: Client,
```
danieljanes, line 17: In Python, we call it just `client`:
```suggestion
                 private val client: Client,
```

PR #2667, opened 2023-11-30
Non-stateful client for XGBoost 

yan-gao-GY:
### Description
To use flower simulation with XGBoost, the clients should be non-stateful. However, the current client implementation is stateful.

### Related issues/PRs
This PR is created on the top of this [PR](https://github.com/adap/flower/pull/2666) (cyclic-xgboost).

## Proposal
Basically, we remove `self.bst` in previous implementation that was used to store the XGBoost object. Instead, we currently initialise client model from `bst = xgb.Booster(params=params)`, and then load the global model to this empty XGBoost object.


yan-gao-GY:
This PR has been combined with a later PR(https://github.com/adap/flower/pull/2668/)

PR #2668, opened 2023-11-30
XGBoost Flower simulation

yan-gao-GY:
### Description
Using Flower simulation is an easy way to simulate large-scale XGBoost experiments. However, the current XGBoost implementation does not support this.

### Related issues/PRs
This PR is created on the top of this [PR](https://github.com/adap/flower/pull/2667) (non-stateful client).

## Proposal
- We create `sim.py` to do Flower simulation. 
- Some shared functions are removed to `utils.py`.
- Create a new arguments parser for simulation in `utils.py`.
- Update readme with new experimental results for bagging and cyclic training.
- Add more explanation in readme.


danieljanes (approved):

examples/xgboost-comprehensive/sim.py
```diff
@@ -0,0 +1,156 @@
...
+            num_val = valid_data.shape[0]
+        else:
+            # Train/test splitting
+            train_data, valid_data, num_train, num_val = train_test_split(
+                partition, test_fraction=args.test_fraction, seed=args.seed
+            )
+
+        # Reformat data to DMatrix for xgboost
+        train_dmatrix = transform_dataset_to_dmatrix(train_data)
+        valid_dmatrix = transform_dataset_to_dmatrix(valid_data)
```
jafermarq, line 62: @yan-gao-GY , can we do this offline? (i.e. before calling `start_simulation`). I see this takes ~20s for each partition, slowing down the simulation by a lot.

yan-gao-GY: if we want to do the data re-formatting before partitioning, we need to first load the full dataset. I'm not sure how to do partitioning after loading the full dataset. I think it changes the logic of Flower Datasets. So I think we need to do something insides of Flower Datasets to achieve this. What do you think? Any better way? @jafermarq

jafermarq (approved):

danieljanes (approved):

PR #2673, opened 2023-12-01
Make the optional arg "--callable" in `flower-client` a required positional arg.

panh99:
- Made `client:flower` a positional arg. A client can be started via `flower-client client:flower`.
- Updated `test_driver.sh` and `mt-pytorch-callable` example accordingly.


src/py/flwr/client/app.py
```diff
@@ -98,6 +98,12 @@ def _parse_args_client() -> argparse.ArgumentParser:
         description="Start a long-running Flower client",
     )
 
+    parser.add_argument(
+        "target",
+        help="The target Flower object to run as a client. For example: "
+        "'client:flower' or 'project.package.module:wrapper.flower'. "
+        "This is a required positional argument.",
```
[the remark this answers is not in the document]
panh99: Updated.

jafermarq (commented):
I believe now `flower-client` comes with SSL by default so we need to specify `--insecure` unless we pass certificates. Also, maybe we can remove `run.sh`?

e2e/test_driver.sh
```diff
@@ -16,10 +16,10 @@ esac
 timeout 2m flower-server $server_arg &
 sleep 3
 
-timeout 2m flower-client $client_arg --callable client:flower --server 127.0.0.1:9092 &
+timeout 2m flower-client --insecure client:flower $client_arg --server 127.0.0.1:9092 &
```
danieljanes, line 19: What does the `$client_arg` do?

panh99: `$client_arg` is either `--insecure` or `--root-certificates xxx`. The `--insecure` in the command is redundant. I will remove it.

PR #2688, opened 2023-12-06
Create docker base images

Robert-Steiner:
## Issue

Building the server images for non-native platforms (e.g. arm) can take a long time because they run in an emulator. Most of the time is spent installing/compiling python. We want to keep our release CI fast therefore we need to find a way to speed it up.

A python version change will occur less frequently than releasing a new Flower version. The idea is to move this installation to a base image and use that image later when building the server image. This should reduce the build time of the server images. The base image can also be used later for the client images.

### Description

This PR moves the installation of system dependencies, python and  general python dependencies into its own Dockerfile. This change significantly reduces the build time of the server images. Building the a server images takes now [2 minutes](https://github.com/adap/flower/actions/runs/7116010432). I restructured the reusable workflow code so that it can be used to build the base and server image.

### DIFF
--- a/.github/workflows/docker-server.yml
+++ b/.github/workflows/docker-server.yml
@@ -38,7 +38,7 @@ jobs:
     name: Summary
     runs-on: ubuntu-22.04
     needs: build-server-images
-    timeout-minutes: 5
+    timeout-minutes: 10
     steps:
       - run: |
           echo "### Images" >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/docker-server.yml
+++ b/.github/workflows/docker-server.yml
@@ -8,20 +8,21 @@ on:
         required: true
         type: string
       base-image-version:
-        description: "Version of Flower base image."
+        description: "Version of the Flower base image."
         required: false
         type: string
-        default: "22.04"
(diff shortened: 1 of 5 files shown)


.github/workflows/docker-server.yml
```diff
@@ -7,42 +7,36 @@ on:
...
+        ${{ github.event.inputs.flwr-version }}
+        latest
     secrets:
       dockerhub-user: ${{ secrets.DOCKERHUB_USERNAME }}
       dockerhub-token: ${{ secrets.DOCKERHUB_TOKEN }}
 
   summary:
     runs-on: ubuntu-22.04
     needs: build-server-images
+    timeout-minutes: 5
```
tanertopal, line 39: @Robert-Steiner Should we increase this to 10min to give a bit more time in case the CI is slow due to e.g. bad network speed? I believe 10min is still acceptable for these jobs. Feel free to keep it at 5 if you think it's reasonable.

Robert-Steiner: I think we can increase it to 10 to be on the safe side.

tanertopal (approved):

PR #2691, opened 2023-12-06
Add `group_id` to `ClientProxy`

charlesbvll:
### Description

The `group_id` attribute of a task, supposed to describe the communication round or the current step of the task (both in sync and async FL), is never used and there is no way of using it without a custom driver script.

### Explanation

Add a `group_id` argument to the `fit` and `evaluate` of the `ClientProxy` class, this way, in the server's FL loop, we can send the current round to the clients proxies which will update the task's `group_id` (when using the `DriverClientProxy`).

### Checklist

- [x] Write tests
- [x] Make CI checks pass

### Any other comments?

N/A


src/py/flwr/server/fleet/grpc_bidi/ins_scheduler.py
```diff
@@ -162,6 +163,7 @@ def _call_client_proxy(
         evaluate_res: EvaluateRes = client_proxy.evaluate(
             ins=evaluate_ins,
             timeout=timeout,
+            group_id=None,
```
jafermarq, line 166: I'm not so sure about this part but, shouldn't there be some `group_id` passed in the `evaluate_ins` before that can be used to set `group_id` in the result? the same applies to `fit_ins` a couple of lines above.

charlesbvll: That would be ideal, but, from what I was able to see, there is no way from the `ins_scheduler` to get the round number. Also, from what I understood, this piece of code is still present only for backwards compatibility and will be removed in the future, is that correct @danieljanes ?

src/py/flwr/server/client_proxy.py
```diff
@@ -63,6 +63,7 @@ def fit(
         self,
         ins: FitIns,
         timeout: Optional[float],
+        group_id: Optional[str],
```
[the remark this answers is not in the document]
charlesbvll: @danieljanes I wasn't sure about that, my reasoning was that the concept of `group_id`, or `round`, would probably not be defined at the `get_parameters` or `get_properties` stage. But we could imagine some more exotic strategy where we use those functions after initialization, so I'll add them.

danieljanes (approved):

danieljanes (approved):

danieljanes (approved):

PR #2692, opened 2023-12-07
Move docs from __init__ to right after class defiition

adam-narozniak:
## Issue
Some strategies have comprehensive docstrings in __ init__ but also short docstrings just after the class definition. That approach is not consistent with PEP 257 and it leads to documentation doubling.

### Description
PEP 257 says that the class Docstrings should be (immediately following the class definition) - the most common option in our docs. The constructor (__init __) Docstrings is optional

## Related PRs
https://github.com/adap/flower/pull/2644 describes the issue from the documentation POV.

## Proposal
Move the docs from __ init__ to the right after the class def. This problem concerns Strategies (the rest of the code places the docstrings correctly).


src/py/flwr/server/strategy/fedavg_android.py
```diff
@@ -40,7 +40,34 @@
...
+    """Federated Averaging strategy.
+
+    Implementation based on https://arxiv.org/abs/1602.05629
+
+    Parameters
+    ----------
+    fraction_fit : Optional[float]
+        Fraction of clients used during training. Defaults to 0.1.
+    fraction_evaluate : Optional[float]
+        Fraction of clients used during validation. Defaults to 0.1.
```
charlesbvll, line 53: Those appear to be typos from the previous docstrings, it should be 1.0 in both cases

src/py/flwr/server/strategy/fedavgm.py
```diff
@@ -38,8 +38,40 @@
...
+    """Federated Averaging with Momentum strategy.
+
+    Implementation based on https://arxiv.org/pdf/1909.06335.pdf
+
+    Parameters
+    ----------
+    fraction_fit : float, optional
+        Fraction of clients used during training. Defaults to 0.1.
+    fraction_evaluate : float, optional
+        Fraction of clients used during validation. Defaults to 0.1.
```
charlesbvll, line 52: It is also the case here (it should also be 1.0)

src/py/flwr/server/strategy/fedtrimmedavg.py
```diff
@@ -35,10 +35,36 @@
...
 
-    Paper: https://arxiv.org/abs/1803.01498
+    Implemented based on: https://arxiv.org/abs/1803.01498
+
+    Parameters
+    ----------
+    fraction_fit : float, optional
+        Fraction of clients used during training. Defaults to 0.1.
+    fraction_evaluate : float, optional
+        Fraction of clients used during validation. Defaults to 0.1.
```
charlesbvll, line 49: Same here (0.1 should be 1.0)

src/py/flwr/server/strategy/fedyogi.py
```diff
@@ -36,12 +36,50 @@
...
+    accept_failures : bool, optional
+        Whether or not accept rounds containing failures. Defaults to True.
+    initial_parameters : Parameters
+        Initial global model parameters.
+    fit_metrics_aggregation_fn : Optional[MetricsAggregationFn]
+        Metrics aggregation function, optional.
+    evaluate_metrics_aggregation_fn: Optional[MetricsAggregationFn]
+        Metrics aggregation function, optional.
+    eta : float, optional
+        Server-side learning rate. Defaults to 1e-1.
```
charlesbvll, line 73:
```suggestion
        Server-side learning rate. Defaults to 1e-2.
```

src/py/flwr/server/strategy/fedyogi.py
```diff
@@ -36,12 +36,50 @@
...
+    eta_l : float, optional
+        Client-side learning rate. Defaults to 1e-1.
```
charlesbvll, line 75:
```suggestion
        Client-side learning rate. Defaults to 0.0316.
```

src/py/flwr/server/strategy/fedyogi.py
```diff
@@ -36,12 +36,50 @@
...
+    beta_1 : float, optional
+        Momentum parameter. Defaults to 0.9.
+    beta_2 : float, optional
+        Second moment parameter. Defaults to 0.99.
+    tau : float, optional
+        Controls the algorithm's degree of adaptability.
+        Defaults to 1e-9.
```
charlesbvll, line 82:
```suggestion
        Defaults to 1e-3.
```

src/py/flwr/server/strategy/krum.py
```diff
@@ -39,10 +39,40 @@
...
-    """Configurable Krum strategy implementation."""
+    """Krum strategy.
 
-    # pylint: disable=too-many-arguments,too-many-instance-attributes, line-too-long
+    Parameters
+    ----------
+    fraction_fit : float, optional
+        Fraction of clients used during training. Defaults to 0.1.
+    fraction_evaluate : float, optional
+        Fraction of clients used during validation. Defaults to 0.1.
```
charlesbvll, line 51: Same here (0.1 -> 1.0)

src/py/flwr/server/strategy/fedavgm.py
```diff
@@ -38,8 +38,40 @@
 from .fedavg import FedAvg
 
 
+# pylint: disable=line-too-long
 class FedAvgM(FedAvg):
-    """Configurable FedAvg with Momentum strategy implementation."""
+    """Federated Averaging with Momentum strategy.
+
+    Implementation based on https://arxiv.org/pdf/1909.06335.pdf
```
charlesbvll, line 45:
```suggestion
    Implementation based on https://arxiv.org/abs/1909.06335
```
We could use this opportunity to use the abstract instead of the pdf, although not completely necessary (the url works as is)

src/py/flwr/server/strategy/krum.py
```diff
@@ -39,10 +39,40 @@
 from .fedavg import FedAvg
 
 
+# pylint: disable=line-too-long
 class Krum(FedAvg):
-    """Configurable Krum strategy implementation."""
+    """Krum strategy.
```
charlesbvll, line 44:
```suggestion
    """Krum [Blanchard et al., 2017] strategy.
    
    Implementation based on https://arxiv.org/abs/1703.02757
```
We could also add the paper link there

charlesbvll (commented):
All of my comments are towards the old docstrings, but I feel like it's the right opportunity to make those fixes! Otherwise LGTM!

adam-narozniak:
@charlesbvll Thanks, good points. Let's make it consistent now.

What do you think about removing the "default from the docs"?

Re the links to the papers I think the best would be not to point directly to pdf but the arxiv.org/sth (there they can also see version etc). wdyt?

charlesbvll:
> @charlesbvll Thanks, good points. Let's make it consistent now.
> 
> What do you think about removing the "default from the docs"?
> 
> Re the links to the papers I think the best would be not to point directly to pdf but the arxiv.org/sth (there they can also see version etc). wdyt?

@adam-narozniak Looking at our docs, it seems like Sphinx does not indicate the default value in all cases, which means that removing it from the docstring might remove some useful information from the docs. If we find a way to make it work consistently without using the docstring, I agree that it would be ideal. It's would be in another PR anyways.

For the links, they're all pointing to the abstract instead of the pdf (except the one I pointed out), so it should be good!

charlesbvll (dismissed):

PR #2705, opened 2023-12-11
Fix build of docker base image

Robert-Steiner:
## Issue

The values (python versions) in the matrix were interpreted as floats. When the value `3.10` was passed to the build argument `PYTHON_VERSION` the zero was cut off. This lead the docker build command to fail because it tried to install python `3.1`. 

### Description

- convert python version to strings
- turned the value for `fail-fast` to `false`. if one build fails, we still can push the other.
- make artifact name unique (hash of the build args) to prevent jobs from overwriting each other's artifacts

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/.github/workflows/_docker-build.yml
+++ b/.github/workflows/_docker-build.yml
@@ -38,6 +38,8 @@ jobs:
     name: Build server image
     runs-on: ubuntu-22.04
     timeout-minutes: 60
+    outputs:
+      build-id: ${{ steps.build-id.outputs.id }}
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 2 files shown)


tanertopal:
My bad. I should have caught this in the review. We had that issue in the past, and I had forgotten it.

Robert-Steiner:
> My bad. I should have caught this in the review. We had that issue in the past, and I had forgotten it.

Don't worry, I should have tested it to make sure it works

Robert-Steiner:
ci passed this time: https://github.com/adap/flower/actions/runs/7169819131

Robert-Steiner:
there was another small error that i had fix. [if job upload an artifact with the same name, they will override each other.](https://github.com/actions/upload-artifact)

tanertopal (approved):

PR #2710, opened 2023-12-15
Fix automatic doc generation

charlesbvll:
### Description

The `build-versioned-docs` script no longer worked with the automatic doc generation.

### Related issues/PRs

N/A

### Explanation

Revert any action performed on each branch.

### Checklist

- [ ] Make CI checks pass

### Any other comments?

N/A


doc/build-versioned-docs.sh
```diff
@@ -76,7 +76,9 @@ END
 
     # Restore branch as it was to avoid conflicts
     git restore source/_templates
-    rm -rf source/ref-api
+    git restore source/_templates/autosummary || rm -rf source/_templates/autosummary
```
adam-narozniak, line 79: LGTM! Could you just explain behind this line?

charlesbvll: This line tries to restore using git if anything was changed on those files compared to the branch it is in and if it is not able to (because those files do not exist on that branch), it deletes them.
For a bit of context, this is needed because the versioning is done by checking out the different tagged versions, and so the current checked out branch shouldn't be changed before checking out the following one, otherwise it won't work.

adam-narozniak (approved):

PR #2735, opened 2023-12-20
Fix 404 error when docs version switching

adam-narozniak:
## Issue
When switching from `main` (generally later - a branch that has a new documentation style = a single doc per function/class/module) to an older style documentation the error - 404 page not found - appears.

### Description
e.g. from: https://flower.dev/docs/framework/main/en/ref-api/flwr.driver.html#module-flwr.driver
when the version gets switched to 1.6 - then such a page does not exist

## Proposal
Fall back to the reference section of the index.html when the page cannot be found. 

### Explanation
Employ a dynamic check:
If the page **exists** (in the requested new version), 
     change to it; 
 elseif the page does **not exist** 
     fall back to the reference section in the idex.html.

A javascript function is created to handle the check. (not the local testing might be harder because the `newUrl` prefix is fixed as https://flower.dev/docs/framework/ (see function in versioning).

Note that coming back from the old version, to the new version 
e.g. from https://flower.dev/docs/framework/v1.5.0/en/ref-api-flwr.html#module-flwr.server
to main (currently only the jump to main can be tested since that's the only version with the new type of docs), currently works by redirecting to the https://flower.dev/docs/framework/main/en/ref-api/flwr.html (which is the entry point of the documentation). I don't know why that mechanism works.


charlesbvll (approved):
Looks good to me!

tanertopal:
@danieljanes, I am okay with the solution to this issue. Wdyt?

danieljanes (approved):
LGTM!

PR #2738, opened 2023-12-20
Change runner for E2E tests to large GitHub hosted runners

tanertopal:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -17,7 +17,8 @@ env:
 
 jobs:
   wheel:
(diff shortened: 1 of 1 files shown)


Robert-Steiner (commented):
I really like the idea to use larger runners to increase the job concurrency limit. maybe we can also add a `label`  with the ubuntu version and # of cpu cores as shown [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-combining-groups-and-labels)

PR #2761, opened 2023-12-28
Update fedmedian.py docstring

edogab33:
Just a little fix to the docstring of FedMedian.


danieljanes (approved):
Thanks for the fix @edogab33 !

PR #2782, opened 2024-01-09
refactor: use alexnet to replace efficientnet

helin0815:
### Changelog entry

<examples>


jafermarq:
Hi @helin0815, I think instead of removing `efficientnet` what we can do is have both models and then let the user decide which one to use. For that, you can introduce a new argument to `argparse` (both for server and clients) and then do if/else and load the appropriate model. Does it sound good?

helin0815:
@jafermarq good.The problem I am currently facing is that if using Efficientnet, adding a little bit of noise during differential privacy noise addition can cause loss to become nan. When I try using Alexnet, I can add noise normally and I will add this part to Readme

helin0815:
@jafermarq hello,please review again.

examples/advanced-pytorch/README.md
```diff
@@ -69,3 +69,11 @@ but this can be changed by removing the `--toy` argument in the script. You can
...
 
 You can also manually run `python3 server.py` and `python3 client.py --client-id <ID>` for as many clients as you want but you have to make sure that each command is run in a different terminal window (or a different computer on the network).
+
+
+## About Differential Privacy
+If you want to achieve differential privacy, please use `alexnet` model, because efficientnet is particularly affected by noise. 
+
+If you add a little noise when using `efficientnet`, the loss will be Nan. This is issue:https://github.com/adap/flower/issues/2342
+
+Therefore, if you want to achieve differential privacy, please use `alexnet`
```
jafermarq, line 79: I'm not sure if we should mention this here since this example is not about DP... I would remove this.

jafermarq (commented):
Hi @helin0815,

I updated your branch with the latest `main`. I had to fix some merge conflicts because we recently updated the `advanced-pytorch` example. I also did some small modifications to how `--model` used. 

Currently your example with `--model alexnet` doesn't work because the CIFAR-10 batches are of shape `224x224` but your model expects them to be `32x32`... I'll get back to you about this early in the week. maybe it's a good idea to update the example to user `32x32`.

In the mean time, could you update the README and show how people using the example can change the model?

helin0815 (commented):
lgtm

examples/advanced-pytorch/README.md
helin0815: I just want to tell other users that if they want to use the example of advanced-pytorch to achieve differential privacy, please do not use efficientnet

jafermarq: We'll introduce new DP examples soon. I would suggest removing this just because this example is not about DP but to show how to use more features of Flower (compared to [quickstart-pytorch](https://github.com/adap/flower/tree/main/examples/quickstart-pytorch)).

helin0815: ok,that is sound good,i will close this pr

jafermarq: oh no need to close it

jafermarq: it's good to include both models. My comment was more about just including that section explaining the DP issue with efficientnet

helin0815: okok,i got it,i will remove the  the parts related to differential privacy in readme

helin0815:
@jafermarq hello bro,I just deleted the differential privacy related section in readme. What should I do next

jafermarq:
> @jafermarq hello bro,I just deleted the differential privacy related section in readme. What should I do next

I pushed with some changes. The main one is replacing `alexnet.py` with the model that `torchvision.models.AlexNet` provides. The reason for this is that this example is ready for 224x224 inputs (as opposed to 32x32, which is what the model in `alexnet.py` expected).

I added a couple of sentences at the end of the reamde to indicate how people can switch between efficientnet/alexent. Please take a look. Something else to include?

jafermarq (dismissed):
@helin0815 , Thanks for the additions to this example. This will be useful for users that want to see how easy it is to switch between ML models for Flower clients. 💯

jafermarq (approved):

helin0815:
@jafermarq Thank you for submitting the code and modifying the readme so that this PR can be merged smoothly. When you last replied to me, I was already asleep. In China, this is my sleeping time. I hope this PR can be helpful to other flower users. Thank you for your efforts

helin0815:
@jafermarq Hello, I would like to ask you a question. If you could answer it for me, I would be very grateful. I am currently adding code to calculate privacy budget during model training, but I encountered an issue after using it. The following are my code and the issues I encountered. Have you encountered similar problems or do you know how to solve them。
code:
`
def train(net, trainloader, valloader, epochs, config, device: str = "cpu"):
    """Train the network on the training set."""
    print("Starting training...")
    net.to(device)  # move model to GPU if available
    criterion = torch.nn.CrossEntropyLoss().to(device)
    optimizer = torch.optim.SGD(
        net.parameters(), lr=0.1, momentum=0.9, weight_decay=1e-4
    )
    net.train()
    from opacus import PrivacyEngine
    privacy_engine = PrivacyEngine(
        secure_mode=False  # Or True
    )
    net, optimizer, trainloader = privacy_engine.make_private(
        module=net,
        optimizer=optimizer,
        data_loader=trainloader,
        noise_multiplier=config["noise"],  
        max_grad_norm=config["clip_norm"],
        clipping="flat",
        grad_sample_mode="hooks",
    )
    for _ in range(epochs):
        index = 0
        for images, labels in trainloader:
            images, labels = images.to(device), labels.to(device)
            optimizer.zero_grad()
            loss = criterion(net(images), labels)
            loss.backward()
            optimizer.step()
            index = index + 1
            if index % 100 == 0:
                print("now index:", index, epochs)
                epsilon = privacy_engine.accountant.get_epsilon(delta=1e-5)
                print(f"Train Epoch: {epoch} "f"(ε = {epsilon:.2f}, δ = {1})")
    net.to("cpu")  # move model back to CPU
    train_loss, train_acc = test(net, trainloader)
    val_loss, val_acc = test(net, valloader)
    results = {
        "train_loss": train_loss,
        "train_accuracy": train_acc,
        "val_loss": val_loss,
        "val_accuracy": val_acc,
    }
    return results
`
and error:
`
torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 640.00 MiB (GPU 0; 23.65 GiB total capacity; 1.74 GiB already allocated; 11.11 GiB free; 2.36 GiB allowed; 1.78 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
`

PR #2794, opened 2024-01-15
Add InnerDirichletPartitioner

adam-narozniak:
## Proposal
Add Inner Dirichlet Partitioner based on Federated Learning Based on Dynamic Regularization https://arxiv.org/abs/2111.04263.

### Explanation
This variation of Dirchlet-based partitioning differs from the https://github.com/adap/flower/pull/2795.

This implementation does the following:
* Originally draws the number of samples - partition sizes - from the lognormal distribution, here it is parameterized by `partition_sizes` and required to be provided by a user.
* The Dirichlet distribution is used to create the fraction for each partition separately. There, it decided on the division of classes per partition (the probabilities from Dirichlet sum to one per each partition; central to the original Dirichlet where the fractions sum up to 1 along all partitions of class k).

How does it differ from the original Dirichlet from https://github.com/adap/flower/pull/2795? 
This implementation is used to choose to divide classes among partitions. The Dirichlet distribution is drawn such that the class `n` is split among `p` partitions. This is repeated `N` (number of unique classes) times. Therefore, there's no need to decide on the size of each partition in the original Dirichlet.

Also, what might help to understand the difference is to think about the shape of the alpha value (concentration) for the Dirichlet. 
Original: size is equal to the number of partitions (and is repeated the number of unique classes)
This implementation: size equals the number of unique classes (and is repeated the number of partitions). 

Note that it also means that at a certain point in this implementation, we run out of the samples from class `n` while sampling for later partitions, and the code needs to be adjusted for that (to give other samples in that case).


datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
```diff
@@ -0,0 +1,271 @@
...
+
+        Returns
+        -------
+        alpha : NDArrayFloat
+            Concentration parameter in a format ready to used in computation.
+        """
+        if self._initialized_alpha:
+            assert self._alpha is not None
+            return self._alpha
+        if isinstance(alpha, float):
```
jafermarq, line 133: I'm thinking we could relax this a bit an also allow for `int` alpha values. I think people would most likely write `alpha=1` as opposed to `alpha=1.0`.

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
```diff
@@ -0,0 +1,271 @@
...
+import datasets
+from flwr_datasets.common.typing import NDArrayFloat, NDArrayInt
+from flwr_datasets.partitioner.partitioner import Partitioner
+
+
+class InnerDirichletPartitioner(Partitioner):  # pylint: disable=R0902
+    """Partitioner based on Dirichlet distribution.
+
+    Implementation based on Federated Learning Based on Dynamic Regularization
+    https://arxiv.org/abs/2111.04263.
```
jafermarq, line 30: I wonder if we should format slightly the title of the paper. Maybe having it between quotes? Or something like:

```
Implementation based on: Federated Learning Based on Dynamic Regularization (https://arxiv.org/abs/2111.04263).
```

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
```diff
@@ -0,0 +1,271 @@
...
+            assert self._alpha is not None
+            return self._alpha
+        if isinstance(alpha, float):
+            assert self._num_unique_classes is not None
+            alpha = np.array([alpha], dtype=float).repeat(self._num_unique_classes)
+        elif isinstance(alpha, List):
+            if len(alpha) != self._num_unique_classes:
+                raise ValueError(
+                    "The alpha parameter needs to be of length of equal to the "
+                    "number of unique classes."
```
jafermarq, line 140:
```suggestion
                    "If passing alpha as an NDArray, its length needs to be of length "
                    "equal to the number of unique classes."
```
Because people can pass a single scalar also. That `ValueError` message follows a bit the style used in the [DirichletPartitioner](https://github.com/adap/flower/blob/3b686d5748ae84c5cfd46e3935638cd21c6773d4/datasets/flwr_datasets/partitioner/dirichlet_partitioner.py#L169). A couple more error messages below would need adjustment also.

jafermarq (commented):
Some initial comments

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
adam-narozniak: Agreed, done

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
adam-narozniak: It's consistent with how I describe other partitioners. Let's agree on the format here, and then I'll change it here and open another PR, changing it in other places.
Implementation based on [description in]: TITLE (URL). 
[description in] would be optional and used only if we want to emphasize that no code was available.
And I'd propose it should be the last paragraph in the docs (I was inconsistent regarding the placement). The top docs should explain how the algorithm works.
Wdyt?

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
adam-narozniak: Ok, fixed

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
jafermarq: `TITLE (URL)` looks ok. I agree placing this at the bottom is better.

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
```diff
@@ -0,0 +1,308 @@
...
+
+    Parameters
+    ----------
+    partition_sizes : Union[List[int], NDArrayInt]
+        The sizes of all partitions.
+    partition_by : str
+        Column name of the labels (targets) based on which Dirichlet sampling works.
+    alpha : Union[int, float, List[float], NDArrayFloat]
+        Concentration parameter to the Dirichlet distribution (a single value for
+        symmetric Dirichlet distribution, or a list/NDArray of lenght equal to the
```
jafermarq, line 48:
```suggestion
        symmetric Dirichlet distribution, or a list/NDArray of length equal to the
```

jafermarq (approved):

danieljanes (approved):

PR #2829, opened 2024-01-19
Migrate quickstart HuggingFace to flwr-datasets

adam-narozniak:
## Proposal

Migrate quickstart huggingface to flwr-datasets

* divide the dataset into 1000 iid partitions
* update script
* solve warnings
  * import AdamW from pytorch instead of transformer

### Changelog entry

<examples>


jafermarq (commented):
I'm thinking we can remove `[vision]`

jafermarq (approved):

PR #2831, opened 2024-01-19
Add `*Record` ProtoBuf messages and corresponding serde functions.

panh99:
- Defined 4 new messages in a new file `recordset.proto`
- Wrote serde functions for the 4 new messages
- Added unittests accordingly.
- Updated the `test_proto_file_count` (b/c we now have 6 proto files)


src/proto/flwr/proto/recordset.proto
```diff
@@ -0,0 +1,36 @@
...
+message Array {
+  string dtype = 1;
+  repeated int32 shape = 2;
+  string stype = 3;
+  bytes data = 4;
+}
+
+message ParametersRecord {
+  repeated string data_keys = 1;
+  repeated Array data_values = 2;
```
danieljanes, line 31: How about using a ProtoBuf `map<string, Array>` here?

src/proto/flwr/proto/recordset.proto
```diff
@@ -0,0 +1,36 @@
...
+}
+
+message MetricsRecord { map<string, Value> data = 1; }
```
danieljanes, line 34: `MetricsRecord` only allows `int` and `float` types (and lists of those two types)

src/proto/flwr/proto/recordset.proto
```diff
@@ -0,0 +1,36 @@
...
+
+message ConfigsRecord { map<string, Value> data = 1; }
```
danieljanes, line 36: `ConfigsRecord` doesn't allow `bool` yet. That's sth we should fix in a separate PR by adding it to `ConfigsRecord`.

The more general question: should we define ProtoBuf messages that follow the same naming scheme as their Python dataclass counterpart?

src/proto/flwr/proto/recordset.proto
panh99: `ParametersRecord.data` is of type `OrderedDict`. The `map` in ProtoBuf cannot keep the order, so we have to store keys and values separately.

src/proto/flwr/proto/recordset.proto
panh99: Yes, I get it.
I was thinking that, since we enforce type checks when instantiating the `MetricsRecord`, we can reuse `Value` to store them to avoid having repetitive code in proto files and in `serde.py`. But I also happy with copy-pasting part of the serde functions for `Value` and create new messages for `MetricsRecordValue`.

src/proto/flwr/proto/recordset.proto
panh99: The reason for reusing `Value` instead of creating a new message `ConfigsRecordValue` is the same as above (`MetricsRecord`).

Re: naming. IMO, we should use different names or even store them in a different way in a TaskIns. There's no benefits to keep them strictly equivalent. I think the ProtoBuf messages is solely to transfer contents over the wire, which is different from the purpose of introducing record types. And hence I think we don't need to follow the same naming scheme and even don't need to have a counterpart for each dataclass record.

src/proto/flwr/proto/recordset.proto
danieljanes: Got it, makes sense. I suspected that to be the case :)

My recommendation would be to create new messages that map closely to the Python code for two reasons:
1. Enable us to remove `Value` once we migrated everything to `RecordSet`
2. Make it easier to implement non-Python clients based on the ProtoBuf definitions

src/proto/flwr/proto/recordset.proto
danieljanes: Right! There was no strong reason for `OrderedDict`, but we decided to use it to stay closer to the PyTorch `state_dict` and to make the `Parameters` conversion easier b/c `Parameters` sorts layers by index.

danieljanes: Let's keep it as-is

src/proto/flwr/proto/recordset.proto
danieljanes: I agree that we do not need to aim for a strict 1:1 mapping from Python to ProtoBuf.

I do think however that there is an advantage to keeping them close. As stated in the other comment, we want to implement Flower clients in languages other than Python. This usually starts with compiling existing ProtoBuf messages. If those messages are close to the Python level, it will be easier for others to implement the Java/C++/... counterpart.

In addition to that, it will maintenance of the Python client easier as well. In the case of `MetricsRecord`, for example, we would not need to check for unsupported `Value` types if we have a `MetricsRecord` on the ProtoBuf level that supports the exact set of types that the Python `MetricsRecord` supports.

danieljanes (approved):

PR #2846, opened 2024-01-24
Handle ClientApp exception

danieljanes:


src/py/flwr/client/app.py
```diff
@@ -365,7 +367,23 @@ def _load_app() -> Flower:
                 app: Flower = load_flower_callable_fn()
 
                 # Handle task message
-                out_message = app(message=message, context=context)
+                try:
+                    out_message = app(message=message, context=context)
+                except Exception as ex:  # pylint: disable=broad-exception-caught
+                    log(ERROR, "FlowerCallable raised an exception", exc_info=ex)
```
panh99, line 373: Should we do the try-except here or in FlowerCallable?

src/py/flwr/client/app.py
```diff
@@ -365,7 +367,23 @@ def _load_app() -> Flower:
...
+
+                    # Don't update/change RunState
+                    # Return empty Message
+
+                    error_out_message = Message(
+                        metadata=message.metadata,
+                        message=RecordSet(),
+                    )
```
panh99, line 381: Though we don't send any error report for now, I think we can set the `task_type="error"` to indicate this is an error message, instead of sending an empty message.

src/py/flwr/driver/driver_client_proxy.py
```diff
@@ -156,8 +156,27 @@ def _send_receive_recordset(
...
+    if not task_res.task.HasField("recordset"):
+        raise ValueError("Invalid Task, field `recordset` missing")
+
+    rs = task_res.task.recordset
+    if (
+        (not rs.parameters.keys())
+        and (not rs.metrics.keys())
+        and (not rs.configs.keys())
+    ):
+        raise ValueError("Exception during client-side task execution")
```
panh99, line 182: As mentioned above, I think we can just check if `task_type == "error"`.

src/py/flwr/client/app.py
danieljanes: Good question - both are viable options, but I'm leaning towards keeping it here in case some user code (perhaps accidentally) modifies the `FlowerCallable` and causes it to raise still an `Exception` that should have been caught.

src/py/flwr/client/app.py
```diff
@@ -379,7 +381,23 @@ def _load_app() -> Flower:
                 app: Flower = load_flower_callable_fn()
 
                 # Handle task message
-                out_message = app(message=message, context=context)
+                try:
+                    out_message = app(message=message, context=context)
+                except Exception as ex:  # pylint: disable=broad-exception-caught
+                    log(ERROR, "FlowerCallable raised an exception", exc_info=ex)
+
+                    # Don't update/change RunState
```
danieljanes, line 389:
```suggestion
                    # Don't update/change NodeState
```

src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -168,8 +168,27 @@ def _send_receive_recordset(
...
+
+
+def validate_task_res(
+    task_res: task_pb2.TaskRes,  # pylint: disable=E1101
+) -> None:
+    """Validate if a TaskRes is empty or not."""
+    if not task_res.HasField("task"):
+        raise ValueError("Invalid TaskRes, field `task` missing")
+    if not task_res.task.HasField("recordset"):
+        raise ValueError("Invalid Task, field `recordset` missing")
```
jafermarq, line 186: A Error message will raise this since the content (i.e. `task.recordset` is empty)

src/py/flwr/client/app.py
danieljanes: I replaced this with the new error mechanism

src/py/flwr/server/compat/driver_client_proxy.py
danieljanes: @jafermarq i pushed an update

src/py/flwr/client/app.py
```diff
@@ -482,32 +482,43 @@ def _load_client_app() -> ClientApp:
...
+                    node_state.update_context(
+                        run_id=message.metadata.run_id,
+                        context=context,
+                    )
+                except Exception as ex:  # pylint: disable=broad-exception-caught
+                    log(ERROR, "ClientApp raised an exception", exc_info=ex)
+
+                    # Legacy grpc-bidi
+                    if transport in ["grpc-bidi", None]:
+                        # Rasie exception, crashes process
```
danieljanes, line 507:
```suggestion
                        # Raise exception, crash process
```

src/py/flwr/client/app.py
```diff
@@ -482,32 +482,43 @@ def _load_client_app() -> ClientApp:
...
+                        raise ex
+
+                    # Don't update/change NodeState
+
+                    # Create error message
+                    # Reason example: "<class 'ZeroDivisionError'>:<'division by zero'>"
+                    reason = str(type(ex)) + ":<'" + str(ex) + "'>"
+                    reply_message = message.create_error_reply(
+                        error=Error(code=0, reason=reason), ttl=message.metadata.ttl
```
danieljanes, line 516:
```suggestion
                        error=Error(code=0, reason=reason)
```

tanertopal (approved):

PR #2848, opened 2024-01-24
Format main

panh99:
- Formatted files via `./dev/format.sh`

### DIFF
--- a/examples/vertical-fl/README.md
+++ b/examples/vertical-fl/README.md
@@ -295,7 +295,7 @@ class ServerModel(nn.Module):
 
 It comprises a single linear layer that accepts the concatenated outputs from
 all client models as its input. The number of inputs to this layer equals the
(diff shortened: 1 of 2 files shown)


examples/vertical-fl/README.md
```diff
@@ -295,7 +295,7 @@ class ServerModel(nn.Module):
 
 It comprises a single linear layer that accepts the concatenated outputs from
 all client models as its input. The number of inputs to this layer equals the
-total number of outputs from the client models ( $3 \times 4 = 12$ ). After processing
+total number of outputs from the client models ( $3 \\times 4 = 12$ ). After processing
```
jafermarq, line 298: double \\ breaks the latex notation. I guess we can replace it with `3 x 4 = 12` and that's it?

jafermarq, line 298:
```suggestion
total number of outputs from the client models (3 x 4 = 12). After processing
```

PR #2871, opened 2024-01-30
Add SecAggPlus middleware

panh99:
- [x] Add the `secaggplus_middleware` function to `flwr/client/middleware/secure_aggregation/secaggplus_middleware.py`
- [x] Add unit tests accordingly
- [x] Remove `flwr/client/secure_aggregation/`
- [x] Update `secaggplus-mt` example


src/py/flwr/client/middleware/secure_aggregation/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Flower Labs GmbH. All Rights Reserved.
+# Copyright 2024 Flower Labs GmbH. All Rights Reserved.
```
danieljanes, line 1:
```suggestion
# Copyright 2023 Flower Labs GmbH. All Rights Reserved.
```

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
```diff
@@ -112,70 +113,115 @@ class SecAggPlusState:
...
+                        v_list.extend(b1_b2)
+                    ret[f"{k}:V"] = v_list
+                else:
+                    ret[f"{k}:V"] = list(v.values())
+        return ret
+
+
+def _get_fit_fn(
+    msg: Message, call_next: FlowerCallable, ctxt: Context
+) -> Callable[[], FitRes]:
```
danieljanes, line 154: Should we reorder these for consistency?
```suggestion
def _get_fit_fn(
    msg: Message, ctxt: Context, call_next: FlowerCallable
) -> Callable[[], FitRes]:
```

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
```diff
@@ -112,70 +113,115 @@ class SecAggPlusState:
...
+    call_next: FlowerCallable,
+) -> Message:
+    """Handle incoming message and return results, following the SecAgg+ protocol."""
+    # Ignore non-fit messages
+    if msg.metadata.task_type != TASK_TYPE_FIT:
+        return call_next(msg, ctxt)
+
+    # Retrieve state
+    if RECORD_KEY_STATE not in ctxt.state.configs:
+        ctxt.state.set_configs(RECORD_KEY_STATE, ConfigsRecord({}))
```
danieljanes, line 176: Should we do this check depending on the `current_stage`? If `current_stage` is not the initial stage, then having no `RECORD_KEY_STATE` key in `ctxt.state` would be an error, no?

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
```diff
@@ -112,70 +113,115 @@ class SecAggPlusState:
...
+
+
+def _get_fit_fn(
+    msg: Message, call_next: FlowerCallable, ctxt: Context
+) -> Callable[[], FitRes]:
+    """Get the fit function."""
+
+    def fit() -> FitRes:
+        out_msg = call_next(msg, ctxt)
+        return compat.recordset_to_fitres(out_msg.message, keep_input=False)
```
danieljanes, line 159: Middleware layers should only use `RecordSet`, but none of the legacy messages like `FitRes`.

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
```diff
@@ -417,18 +466,9 @@ def _collect_masked_input(
...
-        )
-        parameters_factor = fit_res.num_examples
-        parameters = parameters_to_ndarrays(fit_res.parameters)
-    elif isinstance(state.client, NumPyClient):
-        parameters, parameters_factor, _ = state.client.fit(parameters, {})
-    else:
-        log(ERROR, "Client %d: fit function is missing.", state.sid)
+    fit_res = fit()
+    parameters_factor = fit_res.num_examples
+    parameters = parameters_to_ndarrays(fit_res.parameters)
```
danieljanes, line 471: Same as above - we mustn't use `FitRes` and `Parameters` here since these are considered legacy messages

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
panh99: The current stage is stored in the `SecAggPlusState` as a `ConfigsRecord` in `Context.state`. We need to retrieve the current stage from `Context.state`. If no `RECORD_KEY_STATE` in the `ctxt.state`, we have to assume that the state is not initialised.

Do you mean the incoming stage we obtained from the server message? In this case, having no `RECORD_KEY_STATE` means the client is in the initial stage, and hence if the stage in the server message is not the first stage, an error will be raised.

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
panh99: This is b/c SA MW is a layer wrapping around the `fit` method. The SA MW needs to pass the message to the inner `fit` method to train the model and obtain the updated parameters. But for now, our `RecordSet` APIs do not allow us to easily and directly read/write parameters, so here it's a bit hard to directly modify the parameters inside `RecordSet`. Plus, the returned message is actually converted from a FitRes, so I converted it back to access the updated parameters.

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
panh99: I agree that SA MW should not use legacy messages. I plan to rewrite this part after we improve the `RecordSet` APIs and add utility functions to convert `ParametersRecord` to numpy arrays. But for now, I think we still have to work with legacy messages a little bit, as explained above.

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
danieljanes: Yes, that's what I meant. If we received a server message with at stage that is not the initial stage, and there is not `RECORD_KEY_STATE` present, then an error should be raised because any stage that's not the initial stage should rely on the assumption that the initial stage took care of the state initialization (`ctxt.state.set_configs(RECORD_KEY_STATE, ConfigsRecord({}))`).

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
danieljanes: This would mean that our SA middleware is incompatible with all future task handlers that do not return a `RecordSet` that was derived from a `FitRes`.

In the worst case, we can update the parameters in the `RecordSet` by constructing a new `RecordSet` that has all the same fields as the `RecordSet` we get from `call_next`, just with modified parameters.

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
```diff
@@ -112,70 +113,115 @@ class SecAggPlusState:
...
+    msg: Message,
+    ctxt: Context,
+    call_next: FlowerCallable,
+) -> Message:
+    """Handle incoming message and return results, following the SecAgg+ protocol."""
+    # Ignore non-fit messages
+    if msg.metadata.task_type != TASK_TYPE_FIT:
+        return call_next(msg, ctxt)
+
+    # Retrieve state
```
danieljanes, line 174:
```suggestion
    # Retrieve local state
```

src/py/flwr/client/middleware/secure_aggregation/secaggplus_middleware.py
```diff
@@ -112,70 +113,115 @@ class SecAggPlusState:
...
+    if RECORD_KEY_STATE not in ctxt.state.configs:
+        ctxt.state.set_configs(RECORD_KEY_STATE, ConfigsRecord({}))
+    state_dict = ctxt.state.get_configs(RECORD_KEY_STATE).data
+    state = SecAggPlusState(**state_dict)
+
+    # Retrieve configs
```
danieljanes, line 180:
```suggestion
    # Retrieve incoming configs
```

danieljanes (approved):

PR #2909, opened 2024-02-06
Central DP (server-side adaptive clipping)

mohammadnaseri:
### Description
Central DP, also known as user-level DP, enables differential privacy strategies in Flower with server-side adaptive clipping.

### Explanation
A new class, `DifferentialPrivacyServerSideAdaptiveClipping`, has been added. It can be wrapped around a strategy class to enable central DP with adaptive clipping. The clipping phase of clients' model updates is performed at the server side. A new clipping norm is computed on every round. 

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/py/flwr/server/strategy/__init__.py
```diff
@@ -16,7 +16,10 @@
...
 
 from .bulyan import Bulyan as Bulyan
-from .dp_adaptive_clipping import DifferentialPrivacyClientSideAdaptiveClipping
+from .dp_adaptive_clipping import (
+    DifferentialPrivacyClientSideAdaptiveClipping,
+    DifferentialPrivacyServerSideAdaptiveClipping,
+)
 from .dp_fixed_clipping import (
     DifferentialPrivacyClientSideFixedClipping,
     DifferentialPrivacyServerSideFixedClipping,
```
panh99, line 25: I think it would be better if we reformat them as `from xx import yyy as yyy`, in case that we want to change the class names after a stable release. Wdyt?

src/py/flwr/server/strategy/dp_adaptive_clipping.py
```diff
@@ -40,6 +51,199 @@
...
+    initial_clipping_norm : float
+        The initial value of clipping norm. Deafults to 0.1.
+        Andrew et al. recommends to set to 0.1.
+    target_clipped_quantile : float
+        The desired quantile of updates which should be clipped. Defaults to 0.5.
+    clip_norm_lr : float
+        The learning rate for the clipping norm adaptation. Defaults to 0.2.
+        Andrew et al. recommends to set to 0.2.
+    clipped_count_stddev : float
+        The stddev of the noise added to the count of updates currently below the estimate.
```
panh99, line 74:
```suggestion
        The standard deviation of the noise added to the count of updates currently below the estimate.
```

PR #2912, opened 2024-02-07
Add `node_id` to metadata

jafermarq:


src/py/flwr/common/serde_test.py
```diff
@@ -320,8 +322,12 @@ def test_message_to_and_from_taskins() -> None:
     taskins.run_id = metadata.run_id
     taskins.task_id = metadata.task_id
     taskins.group_id = metadata.group_id
+    taskins.task.consumer = metadata.node_id
```
panh99, line 325: `consumer` is of type `Node`. I think it should be updated to `taskins.task.consumer.node_id = metadata.node_id`.

panh99 (approved):

src/py/flwr/common/message.py
```diff
@@ -43,6 +46,7 @@ class Metadata:
     run_id: int
     task_id: str
     group_id: str
+    node_id: Optional[int]
```
danieljanes, line 49: Why is this optional, whereas `run_id`/`task_id` are not?

panh99 (approved):

danieljanes (approved):

PR #2941, opened 2024-02-13
`Message`-based Driver

panh99:


src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+
+        Returns
+        -------
+        Message
+            A new `Message` instance with the specified content and metadata.
+        """
+        _, run_id = self._get_grpc_driver_and_run_id()
+        metadata = Metadata(
+            run_id=run_id,
+            message_id="",
```
danieljanes, line 124:
```suggestion
            message_id="",  # Will be set by the server
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+        return [node.node_id for node in res.nodes]
 
-    def push_task_ins(self, task_ins_list: List[TaskIns]) -> List[str]:
-        """Schedule tasks."""
-        grpc_driver, run_id = self._get_grpc_driver_and_run_id()
+    def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
+        """Push messages to specified node IDs.
+
+        This method takes an iterable of messages and sends each message
+        to the node it specifies in `node_id`.
```
danieljanes, line 145:
```suggestion
        This method takes an iterable of messages and sends each message
        to the node specified in `dst_node_id`.
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+            The ID of the group to which this message is associated. In some settings,
+            this is used as the FL round.
+        ttl : str
+            Time-to-live for the round trip of this message, i.e., the time from sending
+            this message to receiving a reply. It specifies the duration for which the
+            message and its potential reply are considered valid.
+
+        Returns
+        -------
+        Message
```
danieljanes, line 118:
```suggestion
        message : Message
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+        Parameters
+        ----------
+        messages : Iterable[Message]
+            An iterable of messages to be sent.
 
+        Returns
+        -------
+        message_ids : Iterable[str]
+            An iterable of IDs for the messages that were sent, which can be used
+            to pull responses.
```
danieljanes, line 156: How do we indicate errors? For example, if two messages are sent via `push_messages`, but one of them has an invalid `dst_node_id` specified, what does the response look like?

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
 
-    def pull_task_res(self, task_ids: Iterable[str]) -> List[TaskRes]:
-        """Get task results."""
-        grpc_driver, _ = self._get_grpc_driver_and_run_id()
+    def pull_messages(self, message_ids: Iterable[str]) -> Iterable[Message]:
+        """Pull messages based on message IDs.
 
-        # Call GrpcDriver method
+        This method is used to collect responses or messages from the network
+        that correspond to a set of given task identifiers.
```
danieljanes, line 176:
```suggestion
        This method is used to collect messages from the SuperLink
        that correspond to a set of given message IDs.
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+
+        Parameters
+        ----------
+        message_ids : Iterable[str]
+            An iterable of task identifiers for which messages are to be retrieved.
```
danieljanes, line 181:
```suggestion
            An iterable of message IDs for which reply messages are to be retrieved.
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+
+        Returns
+        -------
+        Iterable[Message]
```
danieljanes, line 185:
```suggestion
        messages : Iterable[Message]
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+        grpc_driver, _ = self._get_grpc_driver_and_run_id()
+        # Pull TaskRes
         res = grpc_driver.pull_task_res(
-            PullTaskResRequest(node=self.node, task_ids=task_ids)
+            PullTaskResRequest(node=self.node, task_ids=message_ids)
         )
-        return list(res.task_res_list)
+        # Convert TaskRes to Message
+        msgs = [message_from_taskres(taskres) for taskres in res.task_res_list]
+        return msgs
```
danieljanes, line 195: In the docstring, we should describe what happens in different situations: successful reply found, no reply yet, error cases.

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+
+    def send_and_receive(
+        self,
+        messages: Iterable[Message],
+        *,
+        time_out: Optional[float] = None,
```
danieljanes, line 201: There's no `_` in other instances of `timeout` in the codebase:
```suggestion
        timeout: Optional[float] = None,
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+    ) -> Iterable[Message]:
+        """Push messages to specified node IDs and pull the responses.
```
danieljanes, line 203:
```suggestion
        """Push messages to specified node IDs and pull the replies.
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+
+        This method sends a list of messages to their target node IDs and then
+        waits for the responses. It continues to pull responses until either all
+        responses are received or the specified time out duration is exceeded.
```
danieljanes, line 207:
```suggestion
        This method sends a list of messages to their destination node IDs and then
        waits for the responses. It continues to pull replies until either all
        replies are received or the specified timeout duration is exceeded.
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+
+        Parameters
+        ----------
+        messages : Iterable[Tuple[Message, Iterable[int]]]
```
danieljanes, line 211:
```suggestion
        messages : Iterable[Message]
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+            An iterable of messages to be sent.
+        time_out : Optional[float], default=None
```
danieljanes, line 213:
```suggestion
        timeout : Optional[float] (default: None)
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+            The time out duration in seconds. If specified, the method will wait for
+            responses for this duration. If None, there is no time limit and the method
+            will wait until responses for all messages are received.
```
danieljanes, line 216:
```suggestion
            The timeout duration in seconds. If specified, the method will wait for
            responses for this duration. If `None`, there is no time limit and the method
            will wait until responses for all messages are received.
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+
+        Returns
+        -------
+        Iterable[Message]
```
danieljanes, line 220:
```suggestion
        messages : Iterable[Message]
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
+            An iterable of messages received.
+
+        Notes
+        -----
+        This method uses `push_messages` to send the messages and `pull_messages`
+        to collect the responses. If `time_out` is set, the method may not return
+        responses for all sent messages. A message remains valid until its TTL,
+        which is not affected by `time_out`.
```
danieljanes, line 228:
```suggestion
        This method uses `push_messages` to send the messages and `pull_messages`
        to collect the replies. If `timeout` is set, the method may not return
        responses for all sent messages. A message remains valid until its TTL,
        which is not affected by `timeout`.
```

src/py/flwr/server/driver/driver.py
```diff
@@ -69,44 +70,185 @@ def _get_grpc_driver_and_run_id(self) -> Tuple[GrpcDriver, int]:
...
-    def get_nodes(self) -> List[Node]:
+    def _check_message(self, message: Message) -> None:
+        # Check if the message is valid
+        if not (
+            message.metadata.run_id == self.run_id
+            and message.metadata.src_node_id == self.node.node_id
+            and message.metadata.message_id == ""
+            and message.metadata.reply_to_message == ""
+        ):
+            raise ValueError(f"Invalid message: {message}")
```
danieljanes, line 83: There should be a test for this. I should verify if an exception is being raised by using `push_messages`. It would pass valid and invalid messages to it and test if an exception is raised.

src/py/flwr/server/driver/driver.py
```diff
@@ -13,18 +13,20 @@
 # limitations under the License.
 # ==============================================================================
 """Flower driver service client."""
-
-
+import time
```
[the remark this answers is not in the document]
panh99: Added. Thanks

src/py/flwr/server/driver/driver.py
panh99: Added, thanks.

src/py/flwr/server/driver/driver.py
panh99: Thanks! Great catch

src/py/flwr/server/driver/driver.py
panh99: Added. Thanks.

src/py/flwr/server/driver/driver.py
panh99: For now, if one of the message has incorrect `Metadata` (which usually means the code is wrong), an exception will be raised. Should we do something instead of raising an exception to keep the ServerApp running?

src/py/flwr/server/driver/driver.py
panh99: Thanks, I've updated the doc string.

src/py/flwr/server/driver/driver.py
panh99: Yes. Thanks for pointing it out.

Regarding error cases, I think we may need to update the docstring after we have the error-handling mechanism (including reporting) implemented.

src/py/flwr/server/driver/driver.py
panh99: Thanks. I've added a test.

src/py/flwr/server/driver/driver.py
panh99: Changed, thanks.

src/py/flwr/server/driver/driver.py
panh99: Updated, thanks.

src/py/flwr/server/driver/driver.py
panh99: Updated, thanks.

src/py/flwr/server/driver/driver.py
panh99: updated, thanks.

danieljanes (approved):

src/py/flwr/server/driver/driver.py
danieljanes: True, what you mention is our client-side safety net. The SuperLink also has the capability to reject messages, in which case it returns an empty `str` as the message ID. That's compatible with the current design, so I'm resolving this comment.

danieljanes (approved):

PR #2961, opened 2024-02-15
Add new fields to `Metadata` class

panh99:
- Added `src_node_id`, `dst_node_id`, `reply_to_message` fields to `Metadata` class
- Made `metadata` field in `Message` and some fields in `Metadata` read-only
- Added message validity check in the client main loop


src/py/flwr/common/message.py
```diff
@@ -30,25 +32,73 @@ class Metadata:
         An identifier for the current run.
     message_id : str
         An identifier for the current message.
+    src_node_id : int
+        An identifier for the node sending this message.
+    dst_node_id : int
+        An identifier for the node receiving this message.
+    reply_to_message : str
```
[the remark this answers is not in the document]
panh99: That's a valid option. I keep it as `str` to be consistent with other fields. For example, the `ttl` field cannot set to `None`; instead we treat `""` as no TTL.

src/py/flwr/client/message_handler/message_handler.py
```diff
@@ -175,3 +155,20 @@ def _reconnect(
...
     return ClientMessage(disconnect_res=disconnect_res), sleep_duration
+
+
+def validate_out_message(out_message: Message, in_message_metadata: Metadata) -> bool:
+    """Validate the out message."""
+    out_meta = out_message.metadata
+    in_meta = in_message_metadata
+    if (  # pylint: disable-next=too-many-boolean-expressions
+        out_meta.run_id == in_meta.run_id
+        and out_meta.message_id == ""
```
jafermarq, line 166: might be good to leave a short comment here indicating why this is left as `""`.

src/py/flwr/common/constant.py
```diff
@@ -36,3 +36,12 @@
...
 MESSAGE_TYPE_EVALUATE = "evaluate"
+
+
+class MessageType:
+    """Message type."""
+
+    GET_PROPERTIES = MESSAGE_TYPE_GET_PROPERTIES
+    GET_PARAMETERS = MESSAGE_TYPE_GET_PARAMETERS
+    TRAIN = MESSAGE_TYPE_FIT
+    EVALUATE = MESSAGE_TYPE_EVALUATE
```
jafermarq, line 47: why not define them here? instead of referencing those a few lines above?

src/py/flwr/common/message.py
```diff
@@ -63,5 +149,47 @@ class Message:
...
+    def __init__(self, metadata: Metadata, content: RecordSet) -> None:
+        self._metadata = metadata
+        self.content = content
+
+    @property
+    def metadata(self) -> Metadata:
+        """A dataclass including information about the message to be executed."""
+        return self._metadata
+
+    def create_reply(self, content: RecordSet, ttl: str) -> "Message":
```
jafermarq, line 164: why `-> "Message":` instead of `-> Message:`?

panh99: I tried to use Message without quotes, but got error saying Message is not defined.

jafermarq: You can add at the top this line and then it will work w/o the `""`:
```
```

panh99: Cool, thanks

src/py/flwr/common/constant.py
panh99: Many existing code still uses `MESSAGE_TYPE_*`, and thus I referencing the above ones to avoid divergence.

src/py/flwr/client/message_handler/message_handler.py
panh99: Thanks. I've added a comment.

src/py/flwr/common/message.py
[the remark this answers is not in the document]
danieljanes: True, we should keep the approach consistent. We can do that here. Or should we make them both `Optional` to enable things like `if msg.ttl:` (instead of the more unintuitive version`if msg.ttl == "":`)?

src/py/flwr/common/constant.py
panh99: The `MessageType` will be introduced in a separate PR to make this PR smaller.

src/py/flwr/client/app.py
```diff
@@ -362,10 +366,13 @@ def _load_client_app() -> ClientApp:
...
                     time.sleep(3)  # Wait for 3s before asking again
                     continue
+                in_metadata = copy(message.metadata)
 
                 # Handle control message
                 out_message, sleep_duration = handle_control_message(message)
                 if out_message:
+                    if not validate_out_message(out_message, in_metadata):
+                        log(ERROR, "Invalid out message: %s", out_message)
                     send(out_message)
```
danieljanes, line 376: Should we still send the message if the validation fails? Or rather construct and send a different error message?

src/py/flwr/client/app.py
```diff
@@ -388,6 +395,8 @@ def _load_client_app() -> ClientApp:
                 )
 
                 # Send
+                if not validate_out_message(out_message, in_metadata):
+                    log(ERROR, "Invalid out message: %s", out_message)
                 send(out_message)
```
danieljanes, line 400: Should we still send the message if the validation fails? Or rather construct and send a different error message?

src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -165,25 +160,14 @@ def send(message: Message) -> None:
...
         # Construct TaskRes
         task_res = message_to_taskres(message)
 
-        # Check if fields to be set are not initialized
-        if not validate_task_res(task_res):
-            state[KEY_TASK_INS] = None
-            log(ERROR, "TaskRes has been initialized accidentally")
-
-        # Configure TaskRes
-        task_res = configure_task_res(task_res, task_ins, node)
```
danieljanes, line 185: It seems like we don't need `state` at all any more, right? We can remove it in this PR, or in a separate one.

panh99: I think we may want to keep it. We need this in `delete_node`. Plus, I think we need to add a san-check in `receive()` to verify if the consumer node id in received messages is this node.

src/py/flwr/client/app.py
panh99: Oh, my bad. I thought `log(ERROR)` will interrupt the process. Yes, I think we should construct and send an error message instead of the invalid one.

danieljanes (approved):

PR #2963, opened 2024-02-16
feat(datasets) Add DivideResplitter

adam-narozniak:
## Issue
Certain splits of the dataset need to be divided into smaller splits. No functionality exists to accomplish that that can be used in the FederatedDataset abstraction.

## Proposal
Create a `DivideResplitter` abstraction that solves the problem. See docstrings for more explanation.


datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+        multi-split configuration.
+    drop_remaining_splits: bool
+        In case of single level dictionary specification of `divide_config`, specifies
+        if the splits that are not divided are dropped.
+
+    Raises
+    ------
+    ValuesError if the specified name of a new split is already present in the dataset
+    and the `drop_remaining_splits` is False.
+    """
```
jafermarq, line 54: I think it would be good to have some minimal example here, as some other components have in their docstrings body (e.g. `MergeSplitter`). Because different args types to `divide_config` results three different (but related) behaviours this would complement the description of the `Parameters` above.

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+    and custom names.
+
+    Parameters
+    ----------
+    divide_config: Union[Dict[str, int], Dict[str, float], Dict[str, Dict[str,
+    int]], Dict[str, Dict[str, float]]]
+        If single level dictionary with keys - the new split names and the values of int
+        = number of samples, float - fraction of the split. The fractions do not have
+        to sum up to 1.0. The order of matter = the first key will get fraction_1
+        starting from the beginning of the dataset.
```
jafermarq, line 37:
```suggestion
        If single level dictionary, keys represent the split names. If values are: int, they
        represent the number of samples in each split; float, they represent the fraction
        of the total samples assigned to that split. These fractions do not have
        to sum up to 1.0. The order of values (either int or float) matter: the first key 
        will get the first split starting from the beginning of the dataset, and so on.
```

jafermarq: I tried to make it a bit clearer. But let me know if this is the exact meaning you had in mind.

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+        if self._config_type == "multiple-splits" and self._divide_split is not None:
+            warnings.warn(
+                "The `divide_split` was specified but the multiple split "
+                "configuration was given. The `divide_split` will be "
+                "ignored.",
+                stacklevel=1,
+            )
+
+
+def _determine_config_type(
```
jafermarq, line 262: I wonder if we could enhance this function to check not only if `divide_config` is single-split or multi-split but also if it's "valid" (i.e. to check that actually a single-split `divide_config` is of type `Union[Dict[str,int], Dict[str, float]]` and nothing else.

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+        else:
+            raise ValueError("Incorrect type of config.")
+
+        duplicates = [
+            item for item, count in collections.Counter(new_splits).items() if count > 1
+        ]
+        if duplicates:
+            raise ValueError(
+                "The specified values of the new splits in "
+                "`divide_config` are duplicated. Please specify"
```
jafermarq, line 182:
```suggestion
                f"`divide_config` contains duplicates ({duplicates}). Please specify"
```

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+        """Check duplicates along the new split values and dataset splits.
+
+        This check can happen only at the time this class is called (it does not have
+        access to the dataset prior to that).
+        """
+        if self._config_type == "single-split":
+            new_splits = list(self._single_split_config.keys())
+            all_splits = dataset_splits + new_splits
+            assert self._divide_split is not None
+            all_splits.pop(all_splits.index(self._divide_split))
```
jafermarq, line 198: Why do we need this? in case the `_divide_split` split is going to be re-split but keeping the same name?

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+        else:
+            raise ValueError("Incorrect type of config.")
+
+        duplicates = [
+            item for item, count in collections.Counter(all_splits).items() if count > 1
+        ]
+        if duplicates:
+            raise ValueError(
+                "The specified values of the new splits in "
+                "`divide_config` are duplicated with the split names of "
```
jafermarq, line 215:
```suggestion
                f"`divide_config` are duplicated ({duplicates}) with the split names of "
```

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+        dataset_splits: List[str] = list(dataset.keys())
+        # Change the "single-split" config to look like "multiple-split" config
+        if self._config_type == "single-split":
+            # First, if the `divide_split` is None determine the split
+            if self._divide_split is None:
+                if len(dataset_splits) != 1:
+                    raise ValueError(
+                        "When giving the config that is single level and working with "
+                        "dataset with more than one split you need to specify the "
+                        "`divide_split` but given None instead."
```
jafermarq, line 107:
```suggestion
                        "`divide_split` but current value is None."
```

datasets/flwr_datasets/resplitter/divide_resplitter.py
jafermarq: I noted afterwards you do this in `_check_size_values()` but I wonder if we should better do this check earlier. Let's keep it as is if it's better to do this checks each time upon `__call__`.

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+
+            elif all(isinstance(x, int) for x in new_split_dict.values()):
+                dataset_len = len(dataset[split_from])
+                len_from_divide_resplit = sum(new_split_dict.values())
+                if len_from_divide_resplit > dataset_len:
+                    raise ValueError(
+                        f"The sum of the sample numbers in `divide_config` must be "
+                        f"smaller than the split size. This is not the case for "
+                        f"{split_from} split which is of length {dataset_len} and the "
+                        f"sum in config is {len_from_divide_resplit}."
```
jafermarq, line 244:
```suggestion
                        f"sum in the supplied `divide_config` is {len_from_divide_resplit}."
```

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+        else:
+            # The remaining data is not kept (by simply not coping it=the reference)
+            pass
+
+        # Split the splits
+        for split_from, new_splits_dict in self._multiple_splits_config.items():
+            start_index = 0
+            end_index = 0
+            split_data = dataset[split_from]
+            assert isinstance(new_splits_dict, dict)
```
jafermarq, line 133: This assert will never be triggered, right?

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,286 @@
...
+            for new_split_name, size in new_splits_dict.items():
+                if isinstance(size, float):
+                    end_index += int(len(split_data) * size)
+                elif isinstance(size, int):
+                    end_index += size
+                else:
+                    raise ValueError(
+                        "The type of size value for the divide config must "
+                        "be int or float."
+                    )
```
jafermarq, line 143:
```suggestion
                else: # an int
                    end_index += size
```

jafermarq: from checks in `_check_size_values()`, we can simplify this if/else

jafermarq: similarly, the check immediately  below to this comment is not needed right?

datasets/flwr_datasets/resplitter/divide_resplitter.py
adam-narozniak: In this case (single-split) the divide_split dictates which split will be divided. Here I try to collect all (new) splits -> the situation that will be after the split will be performed. So if I have {"train": train-data, "test": test-data} and I say to split e.g. train (self._divide_split = "train") into {"train": 0.5, "valid":0.5}, then all splits (`all_splits` ) will be ["train" (the new one), "valid", "test"]. and in the code I'm merging the original (`dataset_splits`) ["train", "test"] with the new (`new_splits`) ["train", "valid"]. What I need to do is to remove the original "train" from the sum of splits (the original train will be divided so that's why I do the pop on the `_divide_split` which is "train"

datasets/flwr_datasets/resplitter/divide_resplitter.py
adam-narozniak: Ok, I'll add a separate check on the type validity.

adam-narozniak: Done

datasets/flwr_datasets/resplitter/divide_resplitter.py
adam-narozniak: Done

datasets/flwr_datasets/resplitter/divide_resplitter.py
adam-narozniak: Also, a thing we can consider in the future is adding the external type checker e.g. `from typeguard import typechecked` and then 
```@typechecked
def whatever(param: annotation) -> sth:
```
or e.g.`pydantic`

datasets/flwr_datasets/resplitter/divide_resplitter.py
jafermarq: ok, let's keep if they complain. w/ Py3.10 it seemed to be fine

datasets/flwr_datasets/resplitter/divide_resplitter_test.py
```diff
@@ -0,0 +1,228 @@
...
+            {"train_1": 10, "train_2": 30, "valid": 20, "test": 40},
+        ),
+        # As the first use case but drop the remaining splits
+        (
+            {"train_1": 0.25, "train_2": 0.75},
+            "train",
+            True,
+            {"train_1": 10, "train_2": 30},
+        ),
+        # Split does not sum to 1.0
```
jafermarq, line 43:
```suggestion
        # Split without ratios
        (
            {"train_1": 15, "train_2": 25},
            "train",
            False,
            {"train_1": 15, "train_2": 25, "valid": 20, "test": 40},
        ),
        # Split does not sum to 1.0
```

jafermarq: I'm thinking worth adding this other test that triggers a slightly different part of the `resplit` logic.

datasets/flwr_datasets/resplitter/divide_resplitter_test.py
```diff
@@ -0,0 +1,228 @@
...
+        # Completely custom names
+        (
+            {"test_a": 0.2, "asdfasdfsa": 0.4},
+            "test",
+            False,
+            {"test_a": 8, "asdfasdfsa": 16, "valid": 20, "train": 40},
+        ),
+        # Mirror copies of the first example but using multiple split
+        (
+            {"train": {"train_1": 0.25, "train_2": 0.75}},
```
jafermarq, line 59: would it be worth extending this to also resplit another split? (e.g. `valid`?)

datasets/flwr_datasets/resplitter/divide_resplitter_test.py
adam-narozniak: Sure, it's now extended by the example you gave

datasets/flwr_datasets/resplitter/divide_resplitter_test.py
adam-narozniak: Ok, added now

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,383 @@
...
+        self,
+        divide_config: Union[
+            Dict[str, float],
+            Dict[str, int],
+            Dict[str, Dict[str, float]],
+            Dict[str, Dict[str, int]],
+        ],
+    ) -> None:
+        if not isinstance(divide_config, dict):
+            raise ValueError("Provided input dictionary is not a dictionary")
```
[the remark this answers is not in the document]
adam-narozniak: I moved that part to the `_determine_config_type`

datasets/flwr_datasets/resplitter/divide_resplitter.py
```diff
@@ -0,0 +1,383 @@
...
+    >>>     divide_config={
+    >>>         "train": {
+    >>>             "train": 0.8,
+    >>>             "test": 0.2,
+    >>>         }
+    >>>     }
+    >>> )
+    >>> new_dataset_dict = resplitter(dataset_dict)
+    >>> # new_dataset_dict is
+    >>> # {"train": 80% of train, "valid": 20% of train, "test": test-data}
```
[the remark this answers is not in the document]
adam-narozniak: Yes, makes sense. I've updated the example

jafermarq (approved):

PR #2969, opened 2024-02-17
Add Flower Client App connection error handling

charlesbvll:
### Description

When starting a long-running Flower client app, if anything interrupts the connection between the server and the client, the client will crash (making it not so long running...).

### Related issues/PRs

N/A

### Explanation

Use the existing `RetryInvoker` class to wrap the communication function calls in order to catch any communication error and retry to call the given function after some time.

### Checklist

- [x] Make CI checks pass

### Changelog entry

The `flower-client-app` will now try to reconnect indefinitely to the `flower-superlink` in case of connection error. The arguments `--max-retries` and `--max-wait-time` can now be passed to the `flower-client-app`, `--max-retries` will define the number of tentatives the client should make before it gives up trying to reconnect to the Superlink, and, `--max-wait-time` defines the time before the client gives up trying to reconnect to the Superlink.

### Any other comments?

N/A


src/py/flwr/client/app.py
```diff
@@ -141,6 +147,21 @@ def _parse_args_run_client_app() -> argparse.ArgumentParser:
         default="0.0.0.0:9092",
         help="Server address",
     )
+    parser.add_argument(
+        "--retry_max_tries",
+        type=int,
+        default=1,
+        help="The maximum number of times the client will try to reconnect to the"
+        "server before giving up in case of a connection error. If set to None,"
+        "there is no limit to the number of tries.",
```
panh99, line 156: I think we may want to say that `--retry_max_tries` is the maximum number of times the client will try to connect (not reconnect) to the server. Because `--retry_max_tries 1` means that the client will work as before, trying to send the message once and then giving up. In this case, I think the client will not try to reconnect to the server. Wdyt?

src/py/flwr/client/app.py
```diff
@@ -141,6 +147,21 @@ def _parse_args_run_client_app() -> argparse.ArgumentParser:
...
+    )
+    parser.add_argument(
+        "--retry_max_time",
+        type=float,
+        help="The maximum total amount of time before the client stops trying to"
+        "reconnect to the server before giving up in case of connection error."
+        "If set to None, there is no limit to the total time.",
```
panh99, line 163: Same as before. Should we use `reconnect`, or `connect` (or other words)? Because this implies that the timer for retry_max_time starts when the client first encounters a connection error and begins attempting to reconnect, not from the initial connection attempt to the server.

src/py/flwr/client/app.py
```diff
@@ -213,6 +236,14 @@ class `flwr.client.Client` (default: None)
...
         - 'grpc-rere': gRPC, request-response (experimental)
         - 'rest': HTTP (experimental)
+    retry_max_tries: Optional[int]
+        The maximum number of times the client will try to reconnect to the
+        server before giving up in case of a connection error. If set to None,
+        there is no limit to the number of tries.
+    retry_max_time: Optional[float]
+        The maximum total amount of time before the client stops trying to
+        reconnect to the server before giving up in case of connection error.
+        If set to None, there is no limit to the total time.
```
panh99, line 246: Same as the comments for argparse part.

src/py/flwr/client/app.py
```diff
@@ -299,14 +334,22 @@ class `flwr.client.Client` (default: None)
...
```
panh99, line 352: Same as the comments for the argparse part.

src/py/flwr/client/app.py
```diff
@@ -141,6 +147,21 @@ def _parse_args_run_client_app() -> argparse.ArgumentParser:
         default="0.0.0.0:9092",
         help="Server address",
     )
+    parser.add_argument(
+        "--retry_max_tries",
```
Moep90, line 151: Is it possible to rename the flag to `--max-retries`, `--retries` or `--retry` I have a feeling that this might lead to confusion?

src/py/flwr/client/app.py
```diff
@@ -22,6 +22,9 @@
 from pathlib import Path
 from typing import Callable, ContextManager, Optional, Tuple, Union
 
+from grpc import RpcError
+from requests.exceptions import ConnectionError as RequestsConnectionError
```
danieljanes, line 26: We must not depend on `requests` here because most users won't have it installed. Instead of doing the type `if/else` in this module, we could return the error type to check for from the respective connection.

This line

```
 receive, send, create_node, delete_node = conn
```

would change to

```
 receive, send, create_node, delete_node, connection_error_type = conn
```

src/py/flwr/client/app.py
```diff
@@ -340,13 +383,50 @@ def _load_client_app() -> ClientApp:
     # Initialize connection context manager
     connection, address = _init_connection(transport, server_address)
 
+    retry_invoker = RetryInvoker(
+        exponential,
+        RequestsConnectionError if transport == "rest" else RpcError,
```
danieljanes, line 388: With the proposal above, we can change this to
```suggestion
        connection_error_type,
```

src/py/flwr/client/app.py
charlesbvll: Good catch! I'll do it in the `_init_connection` function

src/py/flwr/client/app.py
```diff
@@ -141,6 +147,21 @@ def _parse_args_run_client_app() -> argparse.ArgumentParser:
...
+        "server before giving up in case of a connection error. If set to None,"
+        "there is no limit to the number of tries.",
+    )
+    parser.add_argument(
+        "--max-wait-time",
+        type=float,
+        help="The maximum duration before the client stops trying to"
+        "connect to the server in case of connection error."
+        "If set to None, there is no limit to the total time.",
+    )
```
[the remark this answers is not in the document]
charlesbvll: The default is `None`, I will add it explicitly

danieljanes (approved):

PR #2989, opened 2024-02-21
Replace dead links, typo fixes and general improvements in framework docs

Svdvoort:
### Description

There were several outdated or misformatted links in the framework docs. See for example [here](https://flower.ai/docs/framework/v1.7.0/en/tutorial-quickstart-huggingface.html) at the bottom of the page and [here](https://flower.ai/docs/framework/contributor-tutorial-contribute-on-github.html) the "getting started for contributors" link. 
There were also several URLs that gave back 404s. 

### Explanation

I went through the documentation to find dead links and replace those with updated links (if available) if they were external. For the internal links (to other pages in the docs) used `:doc:` instead of hardcoded URLs. This should be more robust for future use as it will give a warning when a document is no longer available (e.g. after a name change). 

To check for dead llinks more generally I added a `make` command that will report dead links throughout the
documents.

While going through the documentation, I also did some smaller cleanup tasks: replacing flower.dev with flower.ai in several places, fixing small typos, and fixing an issue with a docstring. 

### Checklist

- [X] Write tests
- [ ] Make CI checks pass

### Changelog entry

- Links to internal documents have been replaced by `:doc:` instead of the hardcoded URL. Should make links to internal documents more consistent.
- Added a `make` command to check for broken links in the documentation: `make checklinks`
- Replaced dead or redirected (external) links with updated working links.
- Fixed several small typos in framework docs.
- Ensure proper docstring formatting in `driver.py`
- flower.dev links have been replaced with flower.ai

### Any other comments?

I replaced most external documents with `:doc:`. This does not allow for referencing specific sections, so those links have not been updated. However, this could be achieved by [adding targets to the documentation](https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#explicit-targets). Might also make it more robust for than `:doc:` in case of filename changes. However, this requires specifying targets, which is more work and might require some discussion. Is this something that would be of interest?


doc/source/contributor-tutorial-contribute-on-github.rst
```diff
@@ -256,36 +255,36 @@ Example of first contribution
...
 Solution
 ********
 
-This is a tiny change, but it’ll allow us to test your end-to-end setup. After cloning and setting up the Flower repo, here’s what you should do:
+This is a tiny change, but it'll allow us to test your end-to-end setup. After cloning and setting up the Flower repo, here's what you should do:
 
 - Find the source file in ``doc/source``
 - Make the change in the ``.rst`` file (beware, the dashes under the title should be the same length as the title itself)
-- Build the docs and check the result: `<https://flower.ai/docs/writing-documentation.html#edit-an-existing-page>`_
+- Build the docs and `check the result <contributor-how-to-write-documentation.html#edit-an-existing-page>`_
```
danieljanes, line 279: Should this also have `:doc:`?
```suggestion
- Build the docs and :doc:`check the result <contributor-how-to-write-documentation.html#edit-an-existing-page>`_
```

danieljanes (commented):
Wow, thanks for the great PR @Svdvoort ! This is a very nice improvement.

I only added one comment where I wasn't sure whether `:doc:` should be added or not.

Can I ask, how did you find all these whitespace issues? Did you use a tool for that? I'm wondering if we could add a tool to the CI to also prevent these in the future.

doc/source/contributor-tutorial-contribute-on-github.rst
Svdvoort: The problem with `:doc:` is that it can only refer to a document as a whole (so to the whole contributor-how-to-write-documentation page in this case), not to specific sections within that page (the edit-an-existing-page section in this case). 
That would require `:ref:` and a specific anchor on that section in the page. Personally I think that would be a good idea (even better than using `:doc:` in general), since even with name changes of documents the reference will always stay correct, or give an error if the section is removed. 

I didn't implement `:ref:` as I would need to define anchor names, but didn't see any guidelines for that. But if you think it would be better to go for `:ref:` everywhere from the start I can implement it.

Svdvoort:
With whitespace issues do you mean all the spaces at the end of lines and such? I have a VSCode plugin for that, [this one](https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces). In my personal repos [I use pre-commit trailing spaces](https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces) to ensure whitespace is rectified in all files, you can also add that for contributions.

doc/source/contributor-tutorial-contribute-on-github.rst
danieljanes: Thanks for the explanation. I'm resolving the comment, I think we can merge this as-is.

danieljanes:
Using `:ref:` and targets sounds like a good idea, we can do that in a separate PR after merging this one.

The trailing spaces tool also looks good, we should add something like this to the CI to prevent these from creeping in in the future.

danieljanes (approved):
Thanks again for the great PR @Svdvoort , approved!

PR #2991, opened 2024-02-21
Add conversion functions between `Array` and `NDArray`

panh99:


src/py/flwr/common/constant.py
```diff
@@ -36,3 +38,13 @@
 MESSAGE_TYPE_GET_PARAMETERS = "get_parameters"
 MESSAGE_TYPE_FIT = "fit"
 MESSAGE_TYPE_EVALUATE = "evaluate"
+
+
+class SType:
+    """Serialisation type."""
+
+    NUMPY = "builtin:numpy"
```
danieljanes, line 46: Is the idea here to add other static `SType` properties in the future?

For the value, I was thinking we could use the convention of using the fully qualified package/module name of the type this returns. For example `numpy.ndarray` or `torch.Tensor`.

panh99: Yes, exactly. That's the idea.

panh99: Sure, I think we can use `numpy.ndarray` to make it more consistent with their names.

danieljanes (approved):

PR #3005, opened 2024-02-23
Add `MessageType`

panh99:


src/py/flwr/common/constant.py
```diff
@@ -42,6 +42,19 @@
 MESSAGE_TYPE_EVALUATE = "evaluate"
 
 
+class MessageType:
+    """Message type."""
+
+    GET_PROPERTIES = "get_properties"
+    GET_PARAMETERS = "get_parameters"
+    FIT = "fit"
+    EVALUATE = "evaluate"
```
danieljanes, line 51:
```suggestion
    TRAIN = "train"
    EVALUATE = "evaluate"
```

src/py/flwr/common/constant.py
```diff
@@ -42,6 +42,19 @@
 MESSAGE_TYPE_EVALUATE = "evaluate"
 
 
+class MessageType:
+    """Message type."""
+
+    GET_PROPERTIES = "get_properties"
+    GET_PARAMETERS = "get_parameters"
```
danieljanes, line 49: Give these these are only necessary for legacy compatibility, should we remove these two?

danieljanes (approved):

src/py/flwr/common/constant.py
danieljanes: Good point. As discussed, we'll solve that with the new `MessageTypeLegacy`.

PR #3017, opened 2024-02-27
Add `Workflow`

panh99:


src/py/flwr/server/run_serverapp.py
```diff
@@ -107,7 +107,7 @@ def run_server_app() -> None:
     run(server_app_attr, driver, server_app_dir)
 
     # Clean up
-    del driver
+    driver.__del__()
```
panh99, line 110: This is a temporary fix. The fix of this problem is outside the scope of this PR.

The `del driver` will not trigger `driver.__del__` here, b/c there may be other references to the driver object. If the driver object is not cleaned up here, it will then cause an error in telemetry. I think we should fix this issue.

src/py/flwr/server/server_app.py
```diff
@@ -55,32 +57,39 @@ class ServerApp:
 
     def __init__(
         self,
-        server: Optional[Server] = None,
```
danieljanes, line 58: We should not remove the `Server` argument to enable users to use `ServerApp` with existing `Server` implementations. My perspective is that there are two modes to use `ServerApp`:
- Legacy mode via `Server`/`Strategy`/`ServerConfig`/`ClientManager`
- New mode via `Workflow` and/or `@app.main()`

src/py/flwr/server/compat/app_utils.py
```diff
@@ -0,0 +1,100 @@
...
+
+    New nodes will be added to the ClientManager via `client_manager.register()`,
+    and dead nodes will be removed from the ClientManager via
+    `client_manager.unregister()`.
+
+    Parameters
+    ----------
+    driver : Driver
+        The Driver object to use.
+    client_manager :ClientManager
```
danieljanes, line 45:
```suggestion
    client_manager : ClientManager
```

src/py/flwr/server/compat/app_utils.py
```diff
@@ -0,0 +1,100 @@
...
+        The ClientManager object to be updated.
+
+    Returns
+    -------
+    Tuple[threading.Thread, threading.Event]
+        The thread updating the ClientManager and the stop event.
+    """
```
danieljanes, line 52: The `threading.Event()` needs to be added here

src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -0,0 +1,347 @@
...
+        client_manager=context.client_manager
+    )
+    if parameters is not None:
+        log(INFO, "Using initial parameters provided by strategy")
+        paramsrecord = compat.parameters_to_parametersrecord(
+            parameters, keep_input=True
+        )
+
+    # Get initial parameters from one of the clients
+    else:
```
danieljanes, line 117:
```suggestion
    else:
        # Get initial parameters from one of the clients
```

src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -0,0 +1,346 @@
...
+            [
+                driver.create_message(
+                    content=content,
+                    message_type=MESSAGE_TYPE_GET_PARAMETERS,
+                    dst_node_id=random_client.node_id,
+                    group_id="",
+                    ttl="",
+                )
+            ]
+        )
```
danieljanes, line 131: Let's put `message = driver.create_message(...)` on a line before `driver.send_and_receive` to make it easier to read

src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -0,0 +1,346 @@
...
+            default_centralized_evaluation_workflow(driver, context)
+
+            # Evaluate round
+            self.evaluate_workflow(driver, context)
+
+        # Bookkeeping
+        end_time = timeit.default_timer()
+        elapsed = end_time - start_time
+        log(INFO, "FL finished in %s", elapsed)
+
```
danieljanes, line 94: After this, should we add the logs that are currently in `run_fl`?

src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -0,0 +1,346 @@
...
+    # Update the parameters and write history
+    if parameters_aggregated:
+        paramsrecord = compat.parameters_to_parametersrecord(
+            parameters_aggregated, True
+        )
+        context.state.parameters_records[PARAMS_RECORD_KEY] = paramsrecord
+
+    context.history.add_metrics_distributed_fit(
+        server_round=current_round, metrics=metrics_aggregated
+    )
```
danieljanes, line 265: In `Server`, this is only done if parameters are also aggregated. Should we keep it like that, or always update it?

src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -0,0 +1,346 @@
...
+            group_id="",
+            ttl="",
+        )
+        for proxy, evalins in client_instructions
+    ]
+
+    # Send instructions to clients and
+    # collect `evaluate` results from all clients participating in this round
+    messages = list(driver.send_and_receive(out_messages))
+    del out_messages
```
danieljanes, line 316: Same comment as for `default_fit_workflow`

src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -0,0 +1,346 @@
...
+            group_id="",
+            ttl="",
+        )
+        for proxy, fitins in client_instructions
+    ]
+
+    # Send instructions to clients and
+    # collect `fit` results from all clients participating in this round
+    messages = list(driver.send_and_receive(out_messages))
+    del out_messages
```
danieljanes, line 234: One option here would be to replace this part with `fit_clients` from `server.py`. `fit_clients` operates on `ClientProxy` instances. What we get from `strategy.configure_fit` are `DriverClientProxy` instances that already have a reference to our `Driver` object. This means that `fit_clients` should work transparently on these objects.

src/py/flwr/server/__init__.py
```diff
@@ -22,18 +22,22 @@
...
 from .history import History as History
 from .run_serverapp import run_server_app as run_server_app
 from .server import Server as Server
 from .server_app import ServerApp as ServerApp
 from .server_config import ServerConfig as ServerConfig
+from .workflow import DefaultWorkflow as DefaultWorkflow
 
 __all__ = [
     "ClientManager",
+    "DefaultWorkflow",
```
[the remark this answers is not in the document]
panh99: make sense.

src/py/flwr/server/compat/app_utils.py
panh99: Updated

src/py/flwr/server/workflow/default_workflows.py
panh99: Yes. I've just added the logs. Btw, may I know what `app_fit` means in the logs?

src/py/flwr/server/workflow/default_workflows.py
panh99: My bad. We should keep it like in `Server`.

src/py/flwr/server/workflow/default_workflows.py
danieljanes: it's just a historic artifact. i think it refers to the fact that this was once placed in the `app` module. we should update it at some point.

danieljanes (approved):

PR #3018, opened 2024-02-27
Handle loading of non-existing `ClientApp` in `RayBackend`.

jafermarq:


src/py/flwr/server/superlink/fleet/vce/backend/test/__init__.py
jafermarq, line 1: @danieljanes one option would be moving this `test/` directory to right under `vce/`.

danieljanes (approved):

danieljanes (approved):

PR #3030, opened 2024-02-28
Add Newton-Raphson strategy example

charlesbvll:
The Newton-Raphson strategy is based on Newton-Raphson distributed method (see https://docs.substra.org/en/stable/substrafl_doc/api/strategies.html#newton-raphson). It leads to a faster convergence than `FedAvg`, however it can only be used on convex problems.

See [here](https://en.wikipedia.org/wiki/Newton%27s_method_in_optimization) for more details.

In this example we will use it in order to train a linear model on this [FLamby dataset](https://github.com/owkin/FLamby/tree/main/flamby/datasets/fed_heart_disease) that contains tabular data about patients with heart diseases. Note that the dataset already provides a partitioning of the data for 4 different clients (or 'centers' as they represent medical centers).


jeandut:
Hello @charlesbvll nice work !
I would also be glad to see another flower example in FLamby alongside substra, fedbiomed and fedml: https://github.com/owkin/FLamby/tree/main/integration

PR #3032, opened 2024-02-28
Update constants for SecAgg+ protocol

panh99:


src/py/flwr/common/secure_aggregation/secaggplus_constants.py
```diff
@@ -14,33 +14,55 @@
...
+    """Stages for the SecAgg+ protocol."""
+
+    SETUP = "setup"
+    SHARE_KEYS = "share_keys"
+    COLLECT_MASKED_INPUT = "collect_masked_input"
+    UNMASK = "unmask"
+    _stages = (SETUP, SHARE_KEYS, COLLECT_MASKED_INPUT, UNMASK)
+
+    @classmethod
+    def all(cls) -> tuple[str, str, str, str]:
```
danieljanes, line 34: If we use a `tuple` here, we'll have to change the type if new elements are introduces. We could also use a `Set[str]`

panh99: I think we prolly don't need to change it. The stage follows the SecAgg+ protocol. It's not defined by us.

danieljanes (approved):

PR #3034, opened 2024-02-28
Define `Error` structure in `common.Message`.

jafermarq:
New protobufs merged from https://github.com/adap/flower/pull/3033 (but needed to have `serde` working)


src/py/flwr/common/message.py
```diff
@@ -169,10 +199,12 @@ class Message:
 
     _metadata: Metadata
     _content: RecordSet
+    _error: Error
```
danieljanes, line 202: Should these both be `Optional` and mutually exclusive?

src/py/flwr/common/message.py
```diff
@@ -169,10 +199,12 @@ class Message:
...
 
     def __init__(self, metadata: Metadata, content: RecordSet) -> None:
         self._metadata = metadata
         self._content = content
+        self._error = Error(code=0) # TODO: decide about codes, init with default "NO_ERROR" code?
```
danieljanes, line 207: We don't need this if we make `Error` optional

src/py/flwr/common/message.py
```diff
@@ -189,6 +221,42 @@ def content(self, value: RecordSet) -> None:
...
+    def error(self) -> Error:
+        """Error captured by this message."""
+        return self._error
+
+    @error.setter
+    def error(self, value: Error) -> None:
+        """Set error."""
+        self._error = value
+
+    def construct_error_message(
```
danieljanes, line 234: If we want to keep the `create_error_message`, then we should rename `create_reply` to `create_reply_message`:
```suggestion
    def create_error(
```

src/py/flwr/common/message.py
```diff
@@ -189,6 +221,42 @@ def content(self, value: RecordSet) -> None:
...
+        self,
+        error_code: int,
```
danieljanes, line 236: Should this be `error: Error` instead?

src/py/flwr/common/message.py
```diff
@@ -189,6 +221,42 @@ def content(self, value: RecordSet) -> None:
...
+        ttl: str,
+        error_reason: str | None = None,
+        content: RecordSet | None = None,
```
danieljanes, line 239: If we make `Error` and `RecordSet` mutually exclusive, then we don't need this

src/py/flwr/common/message.py
```diff
@@ -154,6 +154,36 @@ def partition_id(self, value: int) -> None:
...
 
+@dataclass
+class Error:
+    """A dataclass that stores information about an error that ocurred.
+
+    Parameters
+    ----------
+    code : int
+        An identifier for the error.
+    reasong : Optional[str]
```
danieljanes, line 165:
```suggestion
    reason : Optional[str]
```

src/py/flwr/common/message.py
```diff
@@ -182,14 +226,67 @@ def metadata(self) -> Metadata:
     @property
     def content(self) -> RecordSet:
         """The content of this message."""
+        if self._content is None:
+            raise ValueError(
+                "Message content is None. Use <message>.has_content() "
+                "if you'd like to check first if a message has content."
```
danieljanes, line 232:
```suggestion
                "to check if a message has content."
```

src/py/flwr/common/message.py
```diff
@@ -182,14 +226,67 @@ def metadata(self) -> Metadata:
...
+        else:
+            raise ValueError("A message with an error set cannot have content.")
+
+    @property
+    def error(self) -> Error:
+        """Error captured by this message."""
+        if self._error is None:
+            raise ValueError(
+                "Message error is None. Use <message>.has_error() "
+                "if you'd like to check first if a message carries an error."
```
danieljanes, line 250:
```suggestion
                "to check first if a message carries an error."
```

src/py/flwr/server/driver/driver_test.py
```diff
@@ -157,7 +163,12 @@ def test_send_and_receive_messages_complete(self) -> None:
         # Prepare
         mock_response = Mock(task_ids=["id1"])
         self.mock_grpc_driver.push_task_ins.return_value = mock_response
-        mock_response = Mock(task_res_list=[TaskRes(task=Task(ancestry=["id1"]))])
+        # The response message must include either `content` (i.e. a recordset) or
+        # an `Error`. We choose the later in this case
```
danieljanes, line 167:
```suggestion
        # an `Error`. We choose the latter in this case
```

src/py/flwr/common/message.py
```diff
@@ -182,14 +226,67 @@ def metadata(self) -> Metadata:
...
+
+    def has_content(self) -> bool:
+        """Return True if message has content, else False."""
+        return self._content is not None
+
+    def has_error(self) -> bool:
+        """Return True if message has an error, else False."""
+        return self._error is not None
+
+    def create_error(
```
panh99, line 269: Could we rename this to `create_error_reply`?  This name immediately informs the developer that the method is for creating a reply, specifically one that contains an error. The `create_error` may make the developer think it creates an `Error`. The new method name is also consistent with our existing `create_reply` method, indicating that both methods are aimed to create a reply message to the current message.

src/py/flwr/common/message.py
```diff
@@ -182,14 +226,67 @@ def metadata(self) -> Metadata:
...
+            Time-to-live for this message.
+        """
+        # Create reply without content
+        message = self.create_reply(ttl=ttl)
+        # Set error
+        message.error = error
+        return message
 
-    def create_reply(self, content: RecordSet, ttl: str) -> Message:
+    def create_reply(self, ttl: str, content: RecordSet | None = None) -> Message:
```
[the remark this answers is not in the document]
panh99: I agree that `content` should be a required argument and should be placed as the first argument to be consistent with `create_error` method above.

src/py/flwr/common/message.py
```diff
@@ -182,14 +226,67 @@ def metadata(self) -> Metadata:
...
+    def has_error(self) -> bool:
+        """Return True if message has an error, else False."""
+        return self._error is not None
+
+    def create_error(
+        self,
+        error: Error,
+        ttl: str,
+    ) -> Message:
+        """Construct valid response message indicating an error happened.
```
panh99, line 274:
```suggestion
        """Construct valid reply message indicating an error happened.
```

Would you mind changing "response" to "reply"? Both words are valid, actually. But it would be better if we can use the same words, since the "reply message" is widely used in `create_reply` and in `Driver`.

src/py/flwr/common/message.py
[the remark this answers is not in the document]
jafermarq: Makes sense

src/py/flwr/common/message.py
jafermarq: sure, let's do it.

panh99 (approved):

danieljanes (approved):

PR #3064, opened 2024-03-06
Add `LegacyMessageType`

panh99:


src/py/flwr/common/constant.py
```diff
@@ -38,7 +38,7 @@
 
 MESSAGE_TYPE_GET_PROPERTIES = "get_properties"
 MESSAGE_TYPE_GET_PARAMETERS = "get_parameters"
-MESSAGE_TYPE_FIT = "fit"
+MESSAGE_TYPE_FIT = "train"
```
panh99, line 41: This is to make sure that messages of type `MESSAGE_TYPE_FIT` and `MessageType.TRAIN` can be directed to `client.fit` and work with mods.

danieljanes (approved):

PR #3075, opened 2024-03-07
Handle `ClientApp` exception simulation

jafermarq:
- Upon message execution exception, return `error_reply` message.
 - Attempt loading `ClientApp` before launching simulation backend (mirroring more closely in this way what happens in the `SupperNode`)
 - Simplifications related to `LoadClientAppError` in the backend as a result.


src/py/flwr/server/superlink/fleet/vce/backend/raybackend.py
```diff
@@ -160,13 +159,15 @@ async def process_message(
 
             return out_mssg, updated_context
 
-        except LoadClientAppError as load_ex:
+        except Exception as ex:
```
danieljanes, line 162: I guess we don't care about reporting `LoadClientAppError` back to the SuperLink in the Simulation Engine, right? In that case, the current approach is ok. I just want us to be aware of the semantic difference between Simulation Engine and Deployment Engine.

src/py/flwr/server/superlink/fleet/vce/vce_api.py
```diff
@@ -82,24 +85,25 @@ async def worker(
...
             )
 
-            # Convert to TaskRes
-            task_res = message_to_taskres(out_mssg)
-            # Store TaskRes in state
-            state.store_task_res(task_res)
-
         except asyncio.CancelledError as e:
-            log(DEBUG, "Async worker: %s", e)
+            log(DEBUG, "Terminating Async worker: %s", e)
```
danieljanes, line 89:
```suggestion
            log(DEBUG, "Terminating async worker: %s", e)
```

src/py/flwr/server/superlink/fleet/vce/backend/raybackend.py
jafermarq: Any potential `LoadClientAppError` will be raised earlier (before starting the actual simulation), we won't be reporting it to the superlink, just trigger a termination of all components involved during a `flower-simulation`.

src/py/flwr/server/superlink/fleet/vce/vce_api.py
```diff
@@ -82,24 +85,25 @@ async def worker(
...
+            error = Error(code=0, reason=reason)
+            out_mssg = message.create_error_reply(error=error)
+
+        finally:
+            if out_mssg:
+                # Convert to TaskRes
+                task_res = message_to_taskres(out_mssg)
+                # Store TaskRes in state
+                task_res.task.pushed_at = time.time()
+                state.store_task_res(task_res)
```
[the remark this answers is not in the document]
jafermarq: not when the Simulation (either gracefully or because something else crashed -- e.g. the ServerApp) is ending which requires stopping the `Asyncio` workers (this triggers a `asyncio.CancelledError` which is what we use to `break`).

danieljanes (approved):

PR #3080, opened 2024-03-07
Add `SecAggPlusWorkflow`

panh99:


src/py/flwr/server/workflow/secure_aggregation/secaggplus_workflow.py
```diff
@@ -0,0 +1,575 @@
...
+
+from ..default_constant import MAIN_CONFIGS_RECORD, MAIN_PARAMS_RECORD
+from ..default_constant import Key as DefaultKey
+
+
+@dataclass
+class WorkflowState:  # pylint: disable=R0902
+    """The state of the SecAgg+ protocol."""
+
+    nid_to_fitins: dict[int, RecordSet] = field(default_factory=dict)
```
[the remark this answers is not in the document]
panh99: Yes, the latter one is correct. The dict holds recordsets that are converted from `FitIns`.

src/py/flwr/server/workflow/secure_aggregation/secaggplus_workflow.py
```diff
@@ -0,0 +1,575 @@
...
+    - `num_shares`, `reconstruction_threshold`, and the quantization parameters
+    (`clipping_range`, `target_quantization_range`, `modulus_range`) play critical roles
+    in balancing privacy, robustness, and efficiency within the SecAgg+ protocol.
+    """
+
+    def __init__(  # pylint: disable=R0913
+        self,
+        num_shares: int | float,
+        reconstruction_threshold: int | float,
+        *,
```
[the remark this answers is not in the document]
panh99: Yes, it is. I am not expecting users to change the arguments below `*` here unless they are very sure what they want to do. For most of the time, users only needs to specify `num_shares` and `reconstruction_threshold`.

examples/app-secaggplus/server.py
```diff
@@ -0,0 +1,46 @@
...
+from flwr.server.workflow import SecAggPlusWorkflow
+from workflow_with_log import SecAggPlusWorkflowWithLogs
+
+
+# Define strategy
+strategy = fl.server.strategy.FedAvg(
+    fraction_fit=1.0,  # Select all available clients
+    fraction_evaluate=0.0,  # Disable evaluation
+    min_available_clients=3,
+    fit_metrics_aggregation_fn=lambda _: {},
```
jafermarq, line 13: Do we need this?

src/py/flwr/server/workflow/secure_aggregation/secaggplus_workflow.py
```diff
@@ -0,0 +1,598 @@
...
+                raise ValueError("`num_shares` as an integer must be greater than 2.")
+            if self.num_shares > self.modulus_range / self.quantization_range:
+                log(
+                    WARN,
+                    "A `num_shares` larger than `mod_range / target_range` "
+                    "will potentially cause overflow when computing the aggregated "
+                    "model parameters.",
+                )
+            if self.num_shares == 1:
+                self.num_shares = 1.0
```
jafermarq, line 222: will this ever be reached?

panh99: I think so. This is when users pass `self.num_share=1`, and the int `1` will be converted to float `1.0` here.

examples/app-secaggplus/client.py
```diff
@@ -0,0 +1,31 @@
+import time
+
+import numpy as np
+
+import flwr as fl
+from flwr.client.mod import secaggplus_mod
+
+
+# Define Flower client with the SecAgg+ protocol
```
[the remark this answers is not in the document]
danieljanes: Yes, let's make this a real example.

examples/app-secaggplus/server.py
```diff
@@ -0,0 +1,46 @@
...
+            num_shares=3,
+            reconstruction_threshold=2,
+            timeout=5,
+        )
+        # # Use the following code in practice
+        # fit_workflow=SecAggPlusWorkflow(
+        #     num_shares=3,
+        #     reconstruction_threshold=2,
+        #     timeout=5,
+        # )
```
danieljanes, line 42: Should we add logs to the built-in `SecAggPlusWorkflow` and set the log level to `DEBUG`? That way they won't show up for regular users when they don't change the log level

src/py/flwr/client/mod/secure_aggregation/secaggplus_mod.py
```diff
@@ -390,25 +391,25 @@ def _share_keys(
...
             )
             ciphertext = encrypt(shared_key, plaintext)
-            srcs.append(state.sid)
-            dsts.append(sid)
+            srcs.append(state.nid)
+            dsts.append(nid)
             ciphertexts.append(ciphertext)
 
-    log(INFO, "Client %d: stage 1 completes. uploading key shares...", state.sid)
+    log(INFO, "Client %d: stage 1 completes. uploading key shares...", state.nid)
```
danieljanes, line 412:
```suggestion
    log(INFO, "Node %d: stage 1 completes. uploading key shares...", state.nid)
```

src/py/flwr/client/mod/secure_aggregation/secaggplus_mod.py
```diff
@@ -418,7 +419,7 @@ def _collect_masked_input(
     configs: ConfigsRecord,
     fit: Callable[[], FitRes],
 ) -> Dict[str, ConfigsRecordValues]:
-    log(INFO, "Client %d: starting stage 2...", state.sid)
+    log(INFO, "Client %d: starting stage 2...", state.nid)
```
danieljanes, line 422:
```suggestion
    log(INFO, "Node %d: starting stage 2...", state.nid)
```

src/py/flwr/client/mod/secure_aggregation/secaggplus_mod.py
```diff
@@ -492,20 +509,19 @@ def _collect_masked_input(
...
     state: SecAggPlusState, configs: ConfigsRecord
 ) -> Dict[str, ConfigsRecordValues]:
-    log(INFO, "Client %d: starting stage 3...", state.sid)
+    log(INFO, "Node %d: starting stage 3...", state.nid)
 
-    active_sids = cast(List[int], configs[Key.ACTIVE_SECURE_ID_LIST])
-    dead_sids = cast(List[int], configs[Key.DEAD_SECURE_ID_LIST])
+    active_nids = cast(List[int], configs[Key.ACTIVE_NODE_ID_LIST])
+    dead_nids = cast(List[int], configs[Key.DEAD_NODE_ID_LIST])
     # Send private mask seed share for every avaliable client (including itclient)
```
danieljanes, line 516: What does `itclient` mean?

src/py/flwr/client/mod/secure_aggregation/secaggplus_mod.py
```diff
@@ -492,20 +509,19 @@ def _collect_masked_input(
...
-    sids += dead_sids
-    shares += [state.sk1_share_dict[sid] for sid in dead_sids]
+    all_nids, shares = [], []
+    all_nids = active_nids + dead_nids
+    shares += [state.rd_seed_share_dict[nid] for nid in active_nids]
+    shares += [state.sk1_share_dict[nid] for nid in dead_nids]
 
-    log(INFO, "Client %d: stage 3 completes. uploading key shares...", state.sid)
-    return {Key.SECURE_ID_LIST: sids, Key.SHARE_LIST: shares}
+    log(INFO, "Client %d: stage 3 completes. uploading key shares...", state.nid)
```
danieljanes, line 526:
```suggestion
    log(INFO, "Node %d: stage 3 completes. uploading key shares...", state.nid)
```

src/py/flwr/server/workflow/secure_aggregation/secaggplus_workflow.py
```diff
@@ -0,0 +1,598 @@
...
+    num_shares: int = 0
+    threshold: int = 0
+    clipping_range: float = 0.0
+    quantization_range: int = 0
+    mod_range: int = 0
+    max_weight: float = 0.0
+    nid_to_neighbours: dict[int, set[int]] = field(default_factory=dict)
+    nid_to_publickeys: dict[int, list[bytes]] = field(default_factory=dict)
+    forward_srcs: dict[int, list[int]] = field(default_factory=dict)
+    forward_ciphertexts: dict[int, list[bytes]] = field(default_factory=dict)
```
danieljanes, line 84: Is there a reason for using `dict` vs `Dict` here? Asking b/c the rest of the codebase uses `Dict`

src/py/flwr/server/workflow/secure_aggregation/secaggplus_workflow.py
danieljanes: should we add a test for that?

jafermarq: But @panh99 ,given the `self.num_shares <= 2` check above, the one i'm pointing to here is not going to be evaluated ever because a `ValueError` is going to be triggered.

examples/app-secaggplus/server.py
panh99: I would not recommend doing so. The logs are tailored for this example only. In practice, it will print too much info that can occupy the whole console.

However, I think we can move some logs to to the built-in `SecAggPlusWorkflow`, e.g., the logs notifying the start of each stage.

src/py/flwr/server/workflow/secure_aggregation/secaggplus_workflow.py
panh99: Oh I get it now! Thanks, I will fix it.

examples/app-secaggplus/README.md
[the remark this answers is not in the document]
panh99: That makes sense. I've updated the example.

examples/app-secaggplus/client.py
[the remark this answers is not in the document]
panh99: Thanks! Fixed.

examples/app-secaggplus/pyproject.toml
```diff
@@ -3,9 +3,9 @@ requires = ["poetry-core>=1.4.0"]
...
 
 [tool.poetry]
-name = "secaggplus-mt"
+name = "app-secaggplus"
 version = "0.1.0"
-description = "Secure Aggregation with Driver API"
+description = "Example usage of the SecAggPlusWorkflow."
 authors = ["The Flower Authors <hello@flower.ai>"]
 
 [tool.poetry.dependencies]
```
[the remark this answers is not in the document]
panh99: no, extras are not needed actually. I aligned this with the `app-pytorch` example.

examples/app-secaggplus/server.py
panh99: For demonstration purposes, evaluation is disabled. This follows the `app-pytorch` example.

src/py/flwr/server/workflow/secure_aggregation/secaggplus_workflow.py
panh99: No specific reasons... I've updated all type hints.

src/py/flwr/client/mod/secure_aggregation/secaggplus_mod.py
panh99: Great catch. It's a typo - should be `itself`.

examples/app-secaggplus/pyproject.toml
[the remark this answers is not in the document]
danieljanes: the dependencies in `pyproject.toml` and in `requirements.txt` should be exactly the same. using `flwr-nightly` seems to be the right approach before we release 1.8. we can immediately open a draft pr that changes `flwr-nightly` to `flwr>=1.8.0`, and merge it right after the release.

examples/app-secaggplus/pyproject.toml
```diff
@@ -3,11 +3,11 @@ requires = ["poetry-core>=1.4.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
-name = "secaggplus-mt"
+name = "app-secaggplus"
```
danieljanes, line 6: about about changing the name to leave room for other SA protocols in the future?
```suggestion
name = "app-secagg"
```

panh99: That makes sense. I will change it to `app-secure-aggregation`. The name `secagg` is more or less specifically for the `SecAgg` protocol.

danieljanes: good idea

examples/app-secaggplus/server.py
danieljanes: if we don't use eval, we can remove the entire line

danieljanes (approved):

PR #3081, opened 2024-03-08
Check if object references exist at startup

charlesbvll:
### Description

Currently, we only check the validity of the client/server modules after messages are exchanged. This is because sometimes import times can be long.

### Related issues/PRs

N/A

### Explanation

Instead of importing the modules on startup, we can look at the abstract syntax tree from source to check if the attribute exists on startup.

### Checklist

- [x] Make CI checks pass

### Changelog entry

<skip>

### Any other comments?

N/A


src/py/flwr/client/app.py
```diff
@@ -97,8 +97,18 @@ def run_client_app() -> None:
...
+        raise LoadClientAppError(error_msg) from None
+
     def _load() -> ClientApp:
-        client_app: ClientApp = load_client_app(getattr(args, "client-app"))
+        client_app = load_app(getattr(args, "client-app"), LoadClientAppError)
+
+        if not isinstance(client_app, ClientApp):
+            raise LoadClientAppError(
+                f"Attribute {getattr(args, 'client-app')} is not of type {ClientApp}",
+            ) from None
```
danieljanes, line 110: Given that we have this check after each of our three call to `load_app`, we can also move it into `load_app` by passing `ClientApp` or `ServerApp` as and additional argument.

charlesbvll: This created issues with circular imports (for type hints)

src/py/flwr/cli/flower_toml_test.py
```diff
@@ -18,11 +18,12 @@
 import textwrap
 from typing import Any, Dict
 
+from flwr.common.object_ref import validate
```
tanertopal, line 21: Tests for validate should be in `object_ref_test.py`

danieljanes (dismissed):

src/py/flwr/common/object_ref.py
```diff
@@ -0,0 +1,140 @@
...
+import importlib
+from importlib.util import find_spec
+from typing import Any, Optional, Tuple, Type
+
+OBJECT_REF_HELP_STR = """
+\nThe object reference string should have the form <module>:<attribute>. Valid
+examples include `client:app` and `project.package.module:wrapper.app`. It must
+refer to a module on the PYTHONPATH and the module needs to have the specified
+attribute.
+"""
```
danieljanes, line 28:
```suggestion
OBJECT_REF_HELP_STR = """
\n\nThe object reference string should have the form <module>:<attribute>. Valid
examples include `client:app` and `project.package.module:wrapper.app`. It must
refer to a module on the PYTHONPATH and the module needs to have the specified
attribute.
"""
```

danieljanes (approved):

PR #3083, opened 2024-03-08
Add `count_bytes` method to `RecordSets`

jafermarq:
All record sets get the `count_bytes` method which, as the name suggests, returns the number of bytes in their `.data`.


src/py/flwr/common/record/configsrecord.py
```diff
@@ -85,3 +85,34 @@ def __init__(
...
+                var_bytes = 4
+            if isinstance(value, bool):
+                var_bytes = 1
+            if isinstance(value, (str, bytes)):
+                var_bytes = len(value)
+            return var_bytes
+
+        num_bytes = 0
+
+        for v in self.values():
```
panh99, line 107: I think we should count the bytes of dict keys, right?

src/py/flwr/common/record/metricsrecord.py
```diff
@@ -84,3 +84,15 @@ def __init__(
...
+    def count_bytes(self) -> int:
+        """Return number of Bytes stored in this object."""
+        num_bytes = 0
+
+        for v in self.values():
+            if isinstance(v, List):
+                # both int and float take 4 bytes
+                num_bytes += 4 * len(v)
+            else:
+                num_bytes += 4
```
panh99, line 97: Same as above, I think we should use `8` instead of `4`. Plus, we should count the bytes of keys.

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -117,3 +117,17 @@ def __init__(
                 self[k] = array_dict[k]
                 if not keep_input:
                     del array_dict[k]
+
+    def count_bytes(self) -> int:
+        """Return number of Bytes stored in this object."""
+        num_bytes = 0
```
panh99, line 123: I think we can just count the bytes by counting the bytes of each array and key, right?

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -117,3 +117,17 @@ def __init__(
...
+
+        for v in self.values():
+            if v.dtype and v.shape:
+                num_bytes += int(np.prod(v.shape)) * np.dtype(v.dtype).itemsize
+            else:
+                # If datatype and shape aren't encoding during serialization
+                # we need to deserialize the data to accurately count its
+                # footprint in Bytes.
+                num_bytes += v.numpy().nbytes
```
panh99, line 132: One potential issue is that `numpy()` method does not always convert an `Array` to `NDArray`. It can only complete the conversion when `stype` is `SType.NUMPY` (which indicates the `data` contains serialized ndarrays).

src/py/flwr/common/record/configsrecord.py
```diff
@@ -85,3 +85,34 @@ def __init__(
...
+    def count_bytes(self) -> int:
+        """Return number of Bytes stored in this object.
+
+        This function counts booleans as ocupying 1 Byte.
+        """
+
+        def get_var_bytes(value: ConfigsScalar) -> int:
+            """Return Bytes of value passed."""
+            if isinstance(value, (int, float)):
+                var_bytes = 4
```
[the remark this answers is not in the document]
jafermarq: do we need to use 64bits for an int?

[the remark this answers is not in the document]
panh99: Not necessary for most of time, I think. But it would be a little bit difficult for us to support all int32, int64, float32 and float64. For now, as defined in .proto files, we only use int64 and float64.

src/py/flwr/common/record/parametersrecord.py
jafermarq: done. we're counting now the size of the buffer array

src/py/flwr/common/record/configsrecord.py
```diff
@@ -85,3 +85,39 @@ def __init__(
                 self[k] = configs_dict[k]
                 if not keep_input:
                     del configs_dict[k]
+
+    def count_bytes(self) -> int:
+        """Return number of Bytes stored in this object.
+
+        This function counts booleans as ocupying 1 Byte.
```
danieljanes, line 92:
```suggestion
        This function counts booleans as occupying 1 Byte.
```

src/py/flwr/common/record/configsrecord.py
```diff
@@ -85,3 +85,39 @@ def __init__(
...
+            if isinstance(value, (str, bytes)):
+                var_bytes = len(value)
+            return var_bytes
+
+        num_bytes = 0
+
+        for k, v in self.items():
+            if isinstance(v, List):
+                if isinstance(v[0], (bytes, str)):
+                    # not all str are of equal length necesarily
```
danieljanes, line 112:
```suggestion
                    # not all str are of equal length necessarily
```

danieljanes (approved):

PR #3090, opened 2024-03-10
feat(framework) Allow clients to exit gracefully

charlesbvll:
### Description

- Currently we have to press Ctrl+C 2 times to exit the `flower-client-app` in some cases
- The `flower-client-app` doesn't exit gracefully

### Related issues/PRs

N/A

### Explanation

Catch the interrupt exception and stop the program gracefully.

### Checklist

- [x] Make CI checks pass

### Changelog entry

<skip>

### Any other comments?

N/A


src/py/flwr/client/app.py
```diff
@@ -263,6 +265,31 @@ def _load_client_app() -> ClientApp:
         transport, server_address
     )
 
+    run_tracker = _RunTracker()
```
jafermarq, line 268: I note that if we don't explicitly do:
```
del run_tracker
```
at the end of `_start_client_internal()`, the function will never exit after a `ctrl+c` (it will hang). I tested this by having just a `SuperLink` and a single `flower-client-app` connected.

charlesbvll: Thanks! I just updated it

jafermarq (dismissed):
works great! let's have a followup PR that calls `delete_node()` when processing a `ctrl+c` event?

jafermarq (dismissed):

src/py/flwr/client/app.py
```diff
@@ -278,6 +278,7 @@ def _on_sucess(retry_state: RetryState) -> None:
             )
 
     def _on_backoff(retry_state: RetryState) -> None:
+        app_state_tracker.is_connected = False
```
panh99, line 281: There's one small (but rare) issue. If the user starts a ClientApp with `max_tries=1`, the callback `on_backoff` won't be called even if the connection fails. This means the `app_state_tracker.is_connected` will not be set back to `False`. I am not sure if this would impact the `app_state_tracker`'s functionality.

charlesbvll: @panh99 I just tested it, in this case, as soon as the client loses the connection with the `SuperLink`, it errors out and shuts down properly, so I don't think it's an issue for the `app_state_tracker`

image

danieljanes (approved):

PR #3116, opened 2024-03-12
Update quickstart-pytorch example to use `app` model

tabdar-khan:
## Issue
***quickstart-pytorch*** example is outdated as it is not using the new app model from Flower Next.

### Description

- Updated the `client.py` and `server.py` files in the *quickstart-pytorch* example to use the new `app` model. 
- Refactored some of the imports as well

### Changelog entry
<examples> Update quickstart-pytorch example to use `app` model

PR #3117, opened 2024-03-12
Update quickstart-tensorflow example to use `app` model

tabdar-khan:
## Issue
***quickstart-tensorflow*** example is outdated as it is not using the new app model from Flower Next.

### Description

- Updated the `client.py` and `server.py` files in the *quickstart-tensorflow* example to use the new `app` model. 
- Refactored some of the imports as well

### Changelog entry
<examples> Update quickstart-tensorflow example to use `app` model


jafermarq (commented):
@tabdar-khan , I applied similar edits to those in PR #3116 (see how i handle now argparse) so to make both Flower Next and old version compatible. I also made some unrelated edits to the README of this example, to make it more consistent with those in other examples. Let me know what you think.

tabdar-khan:
@jafermarq lgtm! 🙌

jafermarq (approved):

PR #3133, opened 2024-03-13
Make `PyTorch` `flwr run` template use `Flower Datasets`

jafermarq:


src/py/flwr/cli/new/templates/app/code/client.pytorch.py.tpl
```diff
@@ -31,10 +32,17 @@ class FlowerClient(NumPyClient):
         return loss, len(self.valloader.dataset), {"accuracy": accuracy}
 
 
+# Load config
+cfg, *_ = load_and_validate_with_defaults()
```
jafermarq, line 36: One drawback of this is that it gets executed each time the clientapp below is loaded/imported. An update to the `Simulation Engine` should import the client app module once and cache it.

danieljanes (approved):

PR #3150, opened 2024-03-15
Add a pre-commit hook

tabdar-khan:
## Proposal
Add a dynamic pre-commit hook (formatting and testing) to the repo for better productivity.
### Explanation
Adding a pre-commit hook that works in multiple ways so that everyone can use it according to their style of work. 
Currently, it uses the `./dev/format.sh` and `./dev/test.sh` scripts.

**There are multiple ways you can use this:**

1. You can install the pre-commit hook to your local git directory by simply running `pre-commit install`. 
     - Every time you commit something, it will run the formatting and test scripts.
     - If you are in a hurry, you can bypass this by adding a `--no-verify` flag to you commit command.
2. If you don't want to install the hook permanently, you can just run `pre-commit run --all-files` and it will run the formatting and tests on all the files but it will not stop you from committing.
. . .

### Changelog entry
<skip>


pyproject.toml
```diff
@@ -127,6 +127,7 @@ check-wheel-contents = "==0.4.0"
 GitPython = "==3.1.32"
 PyGithub = "==2.1.1"
 licensecheck = "==2024"
+pre-commit = "3.5.0"
```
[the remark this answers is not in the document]
tabdar-khan: @Moep90 
- `Pre-commit >=3.6.0` requires `python >=3.9` while we are using python 3.8.x. 
- `pre-commit = 3.5.0` is the latest distribution for python 3.8.x

tanertopal (commented):
lgtm! @danieljanes wdyt?

pyproject.toml
```diff
@@ -127,6 +127,7 @@ check-wheel-contents = "==0.4.0"
 GitPython = "==3.1.32"
 PyGithub = "==2.1.1"
 licensecheck = "==2024"
+pre-commit = "^3.5.0"
```
danieljanes, line 130: We use `==` for dev dependencies to ensure a uniform dev environment for everyone:
```suggestion
pre-commit = "==3.5.0"
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+repos:
+  - repo: local
+    hooks:
+      - id: format-code
+        name: Format Code
+        entry: ./dev/format.sh
+        language: script
+        # Ensures the script runs from the repository root:
+        pass_filenames: false
+        stages: [commit]
```
danieljanes, line 10: We should only run `test.sh`. Imagine the following scenario:
- Stashed files are not properly formatted
- Developer runs git commit
- Pre-commit runs `format.sh` -> files are now properly formatted
- Pre-commit runs `test.sh` -> passes, since files are properly formatted
- `git commit` still commits files that have the wrong formatting since the changes caused by `format.sh` are not staged for commit

tabdar-khan: @danieljanes once `format.sh` runs, pre-commit works in a way that if there was something changed by `format.sh`, the commit is blocked from being committed.
Screenshot 2024-03-28 at 13 54 49

pyproject.toml
tabdar-khan: Will do this.

.pre-commit-config.yaml
tabdar-khan: You can then add the new changes caused by `format.sh` and then commit again.

danieljanes: Oh nice! Let's keep it as-is then.

doc/source/contributor-tutorial-get-started-as-a-contributor.rst
```diff
@@ -102,6 +102,33 @@ Run Linters and Tests
...
+
+   ::
+      
+      $ pre-commit install
+
+   - Each ``git commit`` will trigger the execution of formatting and linting/test scripts.
+   - If in a hurry, bypass the hook using ``--no-verify`` with the ``git commit`` command.
+     ::
+          
+       $ git commit --no-verify -m "My flowery commit"
```
danieljanes, line 122:
```suggestion
       $ git commit --no-verify -m "Add new feature"
```

danieljanes (approved):

PR #3156, opened 2024-03-17
feat(*): update ubuntu 22.04 to alpine 3.19

Moep90:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/src/docker/base/Dockerfile
+++ b/src/docker/base/Dockerfile
@@ -1,45 +1,38 @@
 # Copyright 2023 Flower Labs GmbH. All Rights Reserved.
 
-ARG UBUNTU_VERSION=22.04
-FROM ubuntu:$UBUNTU_VERSION as base
+ARG PYTHON_VERSION=3.10
+ARG ALPINE_VERSION=alpine3.19
+FROM python:${PYTHON_VERSION}-${ALPINE_VERSION} as base
 
-ENV DEBIAN_FRONTEND noninteractive
+ENV DEBIAN_FRONTEND=noninteractive \
 # Send stdout and stderr stream directly to the terminal. Ensures that no
 # output is retained in a buffer if the application crashes.
-ENV PYTHONUNBUFFERED 1
+  PYTHONUNBUFFERED=1 \
 # Typically, bytecode is created on the first invocation to speed up following invocation.
 # However, in Docker we only make a single invocation (when we start the container).
 # Therefore, we can disable bytecode writing.
-ENV PYTHONDONTWRITEBYTECODE 1
+  PYTHONDONTWRITEBYTECODE=1 \
 # Ensure that python encoding is always UTF-8.
-ENV PYTHONIOENCODING UTF-8
-ENV LANG C.UTF-8
-ENV LC_ALL C.UTF-8
+  PYTHONIOENCODING=UTF-8 \
+  LANG=C.UTF-8 \
+  LC_ALL=C.UTF-8
 
 # Install system dependencies
-RUN apt-get update \
-    && apt-get -y --no-install-recommends install \
-    clang-format git unzip ca-certificates openssh-client liblzma-dev \
-    build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev wget\
-    libsqlite3-dev curl llvm libncursesw5-dev xz-utils tk-dev libxml2-dev \
-    libxmlsec1-dev libffi-dev liblzma-dev \
-    && rm -rf /var/lib/apt/lists/*
-
-# Install PyEnv and Python
-ARG PYTHON_VERSION
-ENV PYENV_ROOT /root/.pyenv
-ENV PATH $PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
-# https://github.com/hadolint/hadolint/wiki/DL4006
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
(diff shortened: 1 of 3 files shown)


src/docker/base/Dockerfile
```diff
@@ -1,45 +1,38 @@
...
+    pip==${PIP_VERSION:-24.0} \
+    setuptools==${SETUPTOOLS_VERSION:-69.2.0} \
+    # Add User
+    && adduser -D -u 49999 app \
+    && mkdir -p /app \
+    && chown -R app:app /app
+
+WORKDIR /app
+USER app
+ENV HOME=/app
```
Robert-Steiner, line 38: Can we move the user creation to the server and client docker file as the last step before the `ENTRYPOINT`?
This will allow us to install `apk` packages that are specific to the server/client implementation and we can use the poerty scripts.

src/docker/base/Dockerfile
```diff
@@ -1,45 +1,38 @@
...
 
 # Install specific version of pip
 ARG PIP_VERSION
-RUN python -m pip install --no-cache-dir pip==$PIP_VERSION
-
-# Install specific version of setuptools
 ARG SETUPTOOLS_VERSION
-RUN python -m pip install --no-cache-dir setuptools==$SETUPTOOLS_VERSION
+RUN pip install --no-cache-dir \
+    pip==${PIP_VERSION:-24.0} \
```
Robert-Steiner, line 29: Just my personal preference, but I would avoid using default settings here. I'm fine with having defaults for Alpine and Python version as those only belong/are relevant in the Dockerfile, but for PIP_VERSION, SETUPTOOLS_VERSION, etc. we already specify the version [here](https://github.com/adap/flower/blob/main/.github/actions/bootstrap/action.yml) and pass them when building the [image](https://github.com/adap/flower/blob/main/.github/workflows/docker-base.yml).

src/docker/server/Dockerfile
```diff
@@ -1,12 +1,14 @@
...
+ARG ALPINE_VERSION=alpine3.19
+FROM ${BASE_REPOSITORY}:py${PYTHON_VERSION}-${ALPINE_VERSION} as server
 
-WORKDIR /app
 ARG FLWR_VERSION
-RUN python -m pip install -U --no-cache-dir flwr[rest]==${FLWR_VERSION}
+RUN pip install -U --no-cache-dir \
+    flwr[rest]==${FLWR_VERSION:-1.7.0}
+
 ENTRYPOINT ["python", "-c", "from flwr.server import run_superlink; run_superlink()"]
```
Robert-Steiner, line 12: We can now update this to 
```suggestion
ENTRYPOINT ["flower-server"]
```
Note: it will not work if create the non-root user before we install flwr.

src/docker/client/Dockerfile
```diff
@@ -1,8 +1,10 @@
...
-ARG BASE_IMAGE_TAG
-FROM $BASE_REPOSITORY:$BASE_IMAGE_TAG
+ARG PYTHON_VERSION=3.10
+ARG ALPINE_VERSION=alpine3.19
+FROM ${BASE_REPOSITORY}:py${PYTHON_VERSION}-${ALPINE_VERSION} as client
 
 ARG FLWR_VERSION
-RUN python -m pip install -U --no-cache-dir flwr[rest]==${FLWR_VERSION}
+RUN pip install -U --no-cache-dir \
+    flwr[rest]==${FLWR_VERSION:-1.7.0}
```
Robert-Steiner, line 10: @jafermarq do you see any issue when we switch the client image to alpine?

src/docker/base/Dockerfile
Moep90: I found it easier to not specify all versions manually for local building and testing, although the versions are defined in the GitHub action.

We can use this as a builder image and then copy over all build modules to the final container.

Moep90: See a similar use case [here](https://github.com/neXenio/kapitan/blob/master/Dockerfile).

src/docker/base/Dockerfile
Moep90: It is still possible by utilizing the `USER root` ... `RUN apk` ... `USER app` Instructions.  
By doing it directly in the base image, we ensure a safe and secure foundation to build upon.

src/docker/base/Dockerfile
```diff
@@ -1,45 +1,38 @@
...
-ARG PYTHON_VERSION
-ENV PYENV_ROOT /root/.pyenv
-ENV PATH $PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
-# https://github.com/hadolint/hadolint/wiki/DL4006
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-RUN curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
-RUN pyenv install ${PYTHON_VERSION} \
-    && pyenv global ${PYTHON_VERSION} \
-    && pyenv rehash
+RUN apk add --no-cache \
```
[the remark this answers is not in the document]
Moep90: We can use this as a builder image and then copy over all build modules to the final container.

See a similar use case [here](https://github.com/neXenio/kapitan/blob/master/Dockerfile).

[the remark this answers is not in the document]
danieljanes: Is it still necessary to compile `grpcio` with newer releases? For macOS, they started to release pre-built packages.

https://pypi.org/project/grpcio/1.62.1/#files

PR #3166, opened 2024-03-21
Make `ttl` a `float`

jafermarq:
- make `ttl` a `float`
- define a default `ttl` in `common.message` as `DEFAULT_TTL=3600`
- replace all instances of `ttl=""` with `ttl=DEFAULT_TTL`
- Update logic that checks for validity of `taskins/taskres` and `message`


examples/app-pytorch/client_low_level.py
```diff
@@ -20,16 +20,16 @@ def hello_world_mod(msg, ctx, call_next) -> Message:
 @app.train()
 def train(msg: Message, ctx: Context):
     print("`train` is not implemented, echoing original message")
-    return msg.create_reply(msg.content, ttl="")
+    return msg.create_reply(msg.content, ttl=DEFAULT_TTL)
```
panh99, line 23: Will this be the future use case? I mean, import `DEFAULT_TTL` and use it? Should we set the default value for `ttl` in `Message.create_reply`?

src/py/flwr/server/driver/driver.py
```diff
@@ -110,7 +111,7 @@ def create_message(  # pylint: disable=too-many-arguments
         group_id : str
             The ID of the group to which this message is associated. In some settings,
             this is used as the FL round.
-        ttl : str
+        ttl : int
             Time-to-live for the round trip of this message, i.e., the time from sending
             this message to receiving a reply. It specifies the duration for which the
             message and its potential reply are considered valid.
```
panh99, line 117: I think we can mention here that `ttl` is the `Time-to-live in seconds`.

src/py/flwr/server/superlink/state/in_memory_state.py
```diff
@@ -50,15 +51,14 @@ def store_task_ins(self, task_ins: TaskIns) -> Optional[UUID]:
...
+        # Create task_id and created_at
         task_id = uuid4()
         created_at: datetime = now()
-        ttl: datetime = created_at + timedelta(hours=24)
 
         # Store TaskIns
         task_ins.task_id = str(task_id)
         task_ins.task.created_at = created_at.isoformat()
-        task_ins.task.ttl = ttl.isoformat()
+        task_ins.task.ttl += time.time_ns()
```
panh99, line 61: I think `ttl` is in seconds, but `time.time_ns()` is in nanoseconds, right?

src/py/flwr/server/superlink/state/in_memory_state.py
```diff
@@ -113,15 +113,14 @@ def store_task_res(self, task_res: TaskRes) -> Optional[UUID]:
...
+        # Create task_id and created_at
         task_id = uuid4()
         created_at: datetime = now()
-        ttl: datetime = created_at + timedelta(hours=24)
 
         # Store TaskRes
         task_res.task_id = str(task_id)
         task_res.task.created_at = created_at.isoformat()
-        task_res.task.ttl = ttl.isoformat()
+        task_res.task.ttl += time.time_ns()
```
panh99, line 123: Same as above

examples/app-pytorch/client_low_level.py
jafermarq: I added `DEFAULT_TTL` as the default value, set to 1h should be fine for most cases.

src/py/flwr/common/message.py
```diff
@@ -283,22 +285,22 @@ def _create_reply_metadata(self, ttl: str) -> Metadata:
...
+        ttl: float,
     ) -> Message:
         """Construct a reply message indicating an error happened.
 
         Parameters
         ----------
         error : Error
             The error that was encountered.
-        ttl : str
+        ttl : int
```
danieljanes, line 296:
```suggestion
        ttl : float
```

src/py/flwr/common/message.py
```diff
@@ -309,7 +311,7 @@ def create_reply(self, content: RecordSet, ttl: str) -> Message:
         ----------
         content : RecordSet
             The content for the reply message.
-        ttl : str
+        ttl : int
```
danieljanes, line 314:
```suggestion
        ttl : float
```

src/py/flwr/common/message.py
```diff
@@ -283,22 +285,22 @@ def _create_reply_metadata(self, ttl: str) -> Metadata:
...
     ) -> Message:
         """Construct a reply message indicating an error happened.
 
         Parameters
         ----------
         error : Error
             The error that was encountered.
-        ttl : str
+        ttl : int
             Time-to-live for this message.
```
danieljanes, line 297:
```suggestion
        ttl : float
            Time-to-live for this message in seconds.
```

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -105,7 +105,7 @@ def _wrap_recordset_in_message(
                 src_node_id=0,
                 dst_node_id=int(self.cid),
                 reply_to_message="",
-                ttl=str(timeout) if timeout else "",
+                ttl=int(timeout) if timeout else DEFAULT_TTL,
```
danieljanes, line 108:
```suggestion
                ttl=float(timeout) if timeout else DEFAULT_TTL,
```

src/py/flwr/common/message.py
```diff
@@ -283,22 +285,22 @@ def _create_reply_metadata(self, ttl: str) -> Metadata:
...
         """
         # Create reply with error
         message = Message(metadata=self._create_reply_metadata(ttl), error=error)
         return message
 
-    def create_reply(self, content: RecordSet, ttl: str) -> Message:
+    def create_reply(self, content: RecordSet, ttl: float = DEFAULT_TTL) -> Message:
```
danieljanes, line 303: Shouldn't the default be the TTL of the instruction message?

We'd need to make `ttl` an optional argument here and also add a check below that sets `ttl` to the TTL of the instruction message if the user does not set the `ttl` argument. We should probably also check that a user-provided TTL does not exceed the instruction message TTL.

```suggestion
    def create_reply(self, content: RecordSet, ttl: Optional[float] = None) -> Message:
```

jafermarq: my understanding was that the`SuperNode`-side `TTL` was to specify the max duration the `TaskRes` is allowed to remain at the `SuperLink` before it gets pulled by the `ServerApp` (so they are different TTLs)

jafermarq: I've made the change to make it optional. But to keep things consistent, in this PR if the `ttl` is unset, it will set `DEFAULT_TTL`. In a follow up PR we'll replace it to what we agreed upon (the remaining time based on message creation and it's ttl from serverapp)

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -54,7 +54,7 @@
     consumer_node_id        INTEGER,
     created_at              TEXT,
     delivered_at            TEXT,
-    ttl                     TEXT,
+    ttl                     FLOAT,
```
panh99, line 57: According to https://www.sqlite.org/datatype3.html, I think it would be better to use `REAL` instead of `FLOAT` (which will be interpreted as `REAL` under the hood.)

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -74,7 +74,7 @@
...
```
panh99, line 77: Same as above.

src/py/flwr/server/superlink/state/sqlite_state.py
jafermarq: i did this. thanks!

panh99 (approved):

danieljanes (approved):

PR #3173, opened 2024-03-26
Add `pushed_at` into `Task`

jafermarq:
- Adds `pushed_at` into `Task` protobuf definition
- Its value is set at the `SuperLink` (same behaviour as `created_at` but now using timestamps as opposed to iso-dates)
- `created_at` remains, since a follow up PR (https://github.com/adap/flower/pull/3174) will replace it:
    - will be defined upon `Message creation`
    - All tests associated with existing `created_at` will need to be revisited then.

### DIFF
--- a/src/py/flwr/server/superlink/state/in_memory_state.py
+++ b/src/py/flwr/server/superlink/state/in_memory_state.py
@@ -17,7 +17,6 @@
 
 import os
 import threading
-import time
 from datetime import datetime
 from logging import ERROR
 from typing import Dict, List, Optional, Set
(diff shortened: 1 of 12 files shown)


src/py/flwr/server/superlink/state/in_memory_state.py
```diff
@@ -53,10 +54,13 @@ def store_task_ins(self, task_ins: TaskIns) -> Optional[UUID]:
         # Create task_id and created_at
         task_id = uuid4()
         created_at: datetime = now()
+        # Timestamp in seconds with nanosecond resolution
+        pushed_at = time.time_ns() / 1e9
```
danieljanes, line 58: Doesn't `time.time()` give us that float value?

panh99: According to https://docs.python.org/3/library/time.html#time.time, and some discussions on some forums, `time.time()` is less precise than `time.time_ns() /1e9`. The precision of `time.time()` is around `1e-6` second. 

Thus, both `time.time()` and `time.time_ns() /1e9` are sufficient for our cases.

PR #3180, opened 2024-03-28
Update default `TTL` logic in message replies

jafermarq:
Updates:
- logic in `Message.create_reply` regarding default `ttl`
- logic in `Message.create_error_reply` and always defaulting to default ttl.
- Added test

### DIFF
--- a/src/py/flwr/common/message.py
+++ b/src/py/flwr/common/message.py
@@ -297,22 +297,33 @@ class Message:
             partition_id=self.metadata.partition_id,
         )
 
-    def create_error_reply(
-        self,
-        error: Error,
-    ) -> Message:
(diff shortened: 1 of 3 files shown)


src/py/flwr/common/message.py
```diff
@@ -300,17 +300,17 @@ def _create_reply_metadata(self, ttl: float) -> Metadata:
     def create_error_reply(
         self,
         error: Error,
-        ttl: float,
```
danieljanes, line 303: Do we want to enable the user to set a custom TTL? We could use the same `ttl: float | None = None` approach as in `create_reply`

danieljanes (dismissed):

danieljanes (approved):

PR #3181, opened 2024-03-28
Heartbeat client app

panh99:


src/py/flwr/client/heartbeat.py
```diff
@@ -0,0 +1,44 @@
...
+    """Start a ping loop in a separate thread.
+
+    This function initializes a new thread that runs a ping loop, allowing for
+    asynchronous ping operations. The loop can be terminated through the provided stop
+    event.
+    """
+    thread = threading.Thread(target=_ping_loop, args=(ping, stop_event))
+    thread.start()
+
+    return thread
```
danieljanes, line 44: Small idea to improve readability:
```suggestion
def start_ping_loop(
    ping_fn: Callable[[], None], stop_event: threading.Event
) -> threading.Thread:
    """Start a ping loop in a separate thread.

    This function initializes a new thread that runs a ping loop, allowing for
    asynchronous ping operations. The loop can be terminated through the provided stop
    event.
    """
    thread = threading.Thread(target=_ping_loop, args=(ping_fn, stop_event))
    thread.start()

    return thread
```

src/py/flwr/client/heartbeat.py
```diff
@@ -0,0 +1,44 @@
...
+
+import grpc
+
+
+def _ping_loop(ping: Callable[[], None], stop_event: threading.Event) -> None:
+    while not stop_event.is_set():
+        try:
+            ping()
+        except grpc.RpcError:
+            pass
```
danieljanes, line 29:
```suggestion
def _ping_loop(ping_fn: Callable[[], None], stop_event: threading.Event) -> None:
    while not stop_event.is_set():
        try:
            ping_fn()
        except grpc.RpcError:
            pass
```

src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -107,47 +115,83 @@ def grpc_request_response(
...
+            log(ERROR, "Node instance missing")
+            if not ping_stop_event.is_set():
+                ping_stop_event.wait(PING_CALL_TIMEOUT)
+            return
+
+        # Construct the ping request
+        req = PingRequest(node=node, ping_interval=PING_DEFAULT_INTERVAL)
+
+        # Call FleetAPI
+        res: PingResponse = stub.Ping(req, timeout=PING_CALL_TIMEOUT)
```
danieljanes, line 142: If an `RpcError` is raised here, we're not waiting until the `next_interval`. This is desired for cases where the ping call fails because of a network failure. But if, for example, the server is down, then an immediate retry would lead to a number of client nodes retrying in a busy loop. Once the server restarts, it could be brought down immediately because all nodes will retry their ping requests at the same time. To prevent that, we should probably have one immediate retry followed by retries on an exponential backoff schedule.

src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -107,47 +115,83 @@ def grpc_request_response(
...
+
+        # Check if success
+        if not res.success:
+            raise RuntimeError("Ping failed unexpectedly.")
```
danieljanes, line 146: Do we want this to crash the client? Or just trigger an immediate retry (subject to exponential backoff after the immediate retry)?

src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -107,47 +115,83 @@ def grpc_request_response(
...
         # Get Node
-        if node_store[KEY_NODE] is None:
+        nonlocal node
+        if node is None:
             log(ERROR, "Node instance missing")
             return
-        node: Node = cast(Node, node_store[KEY_NODE])
 
+        # Stop the ping-loop thread
+        ping_stop_event.set()
```
danieljanes, line 178: Should we join the `ping_thread` here to ensure that it's done before calling `DeleteNode`?

danieljanes (approved):

PR #3190, opened 2024-04-02
Set `ping_interval` in `State.create_node`

panh99:


src/py/flwr/server/superlink/fleet/vce/vce_api.py
```diff
@@ -43,7 +43,7 @@ def _register_nodes(
     nodes_mapping: NodeToPartitionMapping = {}
     state = state_factory.state()
     for i in range(num_nodes):
-        node_id = state.create_node()
+        node_id = state.create_node(ping_interval=1e300)
```
danieljanes, line 46: Should we use the 64bit float MAX_VALUE here?

panh99: I don't think so. There's no max value of python float, and thus if we use the de facto max value of float64 (1.7976931348623157e+308), we might cause an overflow error when we set the value to a field in gRPC messages. Because, under the hood, we compute `online_until = time.time() + ping_interval`, which can exceed the max value if we set `ping_interval` to max value.

danieljanes (dismissed):

PR #3191, opened 2024-04-02
Introduce `ClientAppException`

jafermarq:
Introduces `ClientAppException` and when raised, the associated error code is used.


src/py/flwr/client/app.py
```diff
@@ -507,11 +508,15 @@ def _load_client_app() -> ClientApp:
...
+                    e_code = ErrorCode.UNKNOWN
+                    if isinstance(ex, ClientAppException):
+                        e_code = ErrorCode.CLIENT_APP_RAISED_EXCEPTION
+
                     # Create error message
                     # Reason example: "<class 'ZeroDivisionError'>:<'division by zero'>"
                     reason = str(type(ex)) + ":<'" + str(ex) + "'>"
                     reply_message = message.create_error_reply(
-                        error=Error(code=0, reason=reason)
+                        error=Error(code=e_code, reason=reason)
```
panh99, line 519: I think it's generally Okay but I am not sure how industry guys see this. Should we mention clearly in the change log that the error will be reported to the server? I think some people may consider the exception message as privacy.

danieljanes: does the `reason` carry any specific info? we could change this to just use `LOAD_CLIENT_APP_EXCEPTION` and a generic `reason`

src/py/flwr/client/client_app.py
```diff
@@ -28,6 +28,15 @@
 from .typing import ClientAppCallable
 
 
+class ClientAppException(Exception):
+    """Exepction raised when an exception is raised while executing a ClientApp."""
```
danieljanes, line 32:
```suggestion
    """Exception raised when an exception is raised while executing a ClientApp."""
```

src/py/flwr/client/app.py
```diff
@@ -491,14 +492,11 @@ def _load_client_app() -> ClientApp:
...
                     reply_message = client_app(message=message, context=context)
-                    # Update node state
-                    node_state.update_context(
-                        run_id=message.metadata.run_id,
-                        context=context,
-                    )
+
                 except Exception as ex:  # pylint: disable=broad-exception-caught
-                    log(ERROR, "ClientApp raised an exception", exc_info=ex)
+                    log(ERROR, "SuperNode raised an exception", exc_info=ex)
```
danieljanes, line 499: Exception is coming from the `ClientApp`, no?
```suggestion
                    log(ERROR, "ClientApp raised an exception", exc_info=ex)
```

src/py/flwr/client/app.py
```diff
@@ -491,14 +492,11 @@ def _load_client_app() -> ClientApp:
...
                     client_app: ClientApp = load_client_app_fn()
 
+                    # Execute ClientApp
                     reply_message = client_app(message=message, context=context)
-                    # Update node state
-                    node_state.update_context(
-                        run_id=message.metadata.run_id,
-                        context=context,
-                    )
+
```
danieljanes, line 497:
```suggestion
```

src/py/flwr/client/client_app.py
```diff
@@ -28,6 +28,15 @@
 from .typing import ClientAppCallable
 
 
+class ClientAppException(Exception):
+    """Exception raised when an exception is raised while executing a ClientApp."""
+
+    def __init__(self, message: str):
+        ex_name = self.__class__.__name__
+        self.message = f"\nA {ex_name} occurred." + message
```
danieljanes, line 36:
```suggestion
        self.message = f"\nException {ex_name} occurred. Message: " + message
```

src/py/flwr/client/app.py
```diff
@@ -491,14 +492,10 @@ def _load_client_app() -> ClientApp:
...
+                    # Execute ClientApp
                     reply_message = client_app(message=message, context=context)
-                    # Update node state
-                    node_state.update_context(
-                        run_id=message.metadata.run_id,
-                        context=context,
-                    )
                 except Exception as ex:  # pylint: disable=broad-exception-caught
-                    log(ERROR, "ClientApp raised an exception", exc_info=ex)
+                    log(ERROR, "SuperNode raised an exception", exc_info=ex)
```
danieljanes, line 498: Isn't the common case here `ClientApp` raising an exception?
```suggestion
                    log(ERROR, "ClientApp raised an exception", exc_info=ex)
```

jafermarq: It's the most likely scenario but since we are also having the `load_client_app_fn()` inside the `try`, that can potentially trigger its own exception.

danieljanes: should me make the message conditional based on the type of exception we get?

src/py/flwr/client/app.py
```diff
@@ -483,35 +484,49 @@ def _load_client_app() -> ClientApp:
...
 
-                    # Create error message
+                    e_code = ErrorCode.CLIENT_APP_RAISED_EXCEPTION
                     # Reason example: "<class 'ZeroDivisionError'>:<'division by zero'>"
                     reason = str(type(ex)) + ":<'" + str(ex) + "'>"
+                    exc_entity = "ClientApp"
+                    if isinstance(ex, LoadClientAppError):
+                        reason = (
+                            "An exception was raised when attempting to load "
+                            "the ClientApp module"
```
danieljanes, line 514:
```suggestion
                            "`ClientApp`"
```

danieljanes (approved):

PR #3195, opened 2024-04-02
Warn once if `TTL` is set

jafermarq:
- Warns once if TTL is set in `create_message()` or `create_reply`.
- `create_message` now accepts optional `ttl` (mirroring `create_reply` signature and logic to set `DEFAULT_TTL`).

### DIFF
--- a/src/py/flwr/common/message.py
+++ b/src/py/flwr/common/message.py
@@ -314,8 +314,9 @@ class Message:
         """
         if ttl:
             warnings.warn(
-                "A custom TTL was set, but note that the SuperLink does not enforce the TTL yet."
-                "The SuperLink will start enforcing the TTL in a future version of Flower.",
+                "A custom TTL was set, but note that the SuperLink does not enforce "
+                "the TTL yet. The SuperLink will start enforcing the TTL in a future "
+                "version of Flower.",
                 stacklevel=2,
             )
         # If no TTL passed, use default for message creation (will update after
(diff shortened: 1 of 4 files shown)


src/py/flwr/common/message.py
```diff
@@ -311,6 +312,12 @@ def create_error_reply(self, error: Error, ttl: float | None = None) -> Message:
 
             ttl = msg.meta.ttl - (reply.meta.created_at - msg.meta.created_at)
         """
+        if ttl:
+            warnings.warn(
+                "TTL in a Message's Metadata is not being used. Messages do not "
+                "expire. This behaviour will change in a future version of Flower",
```
danieljanes, line 318: Wording proposal:
```suggestion
                "A custom TTL was set, but note that the SuperLink does not enforce the TTL yet."
                "The SuperLink will start enforcing the TTL in a future version of Flower.",
```

src/py/flwr/server/driver/driver.py
```diff
@@ -111,25 +111,34 @@ def create_message(  # pylint: disable=too-many-arguments
         group_id : str
             The ID of the group to which this message is associated. In some settings,
             this is used as the FL round.
-        ttl : float (default: common.DEFAULT_TTL)
+        ttl : Optional[float] (default: None)
             Time-to-live for the round trip of this message, i.e., the time from sending
             this message to receiving a reply. It specifies in seconds the duration for
-            which the message and its potential reply are considered valid.
+            which the message and its potential reply are considered valid. If unset
+            the default TTL (i.e. common.DEFAULT_TTL) will be used.
```
danieljanes, line 118:
```suggestion
            which the message and its potential reply are considered valid. If unset,
            the default TTL (i.e. common.DEFAULT_TTL) will be used.
```

danieljanes (approved):

PR #3197, opened 2024-04-02
Generate `TaskRes` containing error message for unavailable nodes

panh99:


src/py/flwr/server/superlink/state/utils.py
```diff
@@ -0,0 +1,49 @@
...
+from flwr.proto.task_pb2 import Task, TaskIns, TaskRes  # pylint: disable=E0611
+
+NODE_UNAVAILABLE_ERROR_REASON = (
+    "Error: Node Unavailable - The destination node is currently unavailable."
+)
+
+
+def make_node_unavailable_taskres(ref_taskins: TaskIns) -> TaskRes:
+    """Create a TaskRes containing the node available error based on the reference
+    TaskIns."""
```
danieljanes, line 33: First sentence of a docstring must fit on a single line

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -0,0 +1,49 @@
...
+    return TaskRes(
+        task_id=str(uuid4()),
+        group_id=ref_taskins.group_id,
+        run_id=ref_taskins.run_id,
+        task=Task(
+            producer=Node(node_id=0, anonymous=True),
+            consumer=Node(node_id=ref_taskins.task.producer.node_id, anonymous=False),
+            created_at=time.time(),
+            ttl=0,
```
jafermarq, line 42: Should we make use of `common.DEFAULT_TTL`?

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -0,0 +1,49 @@
...
+import time
+from uuid import uuid4
+
+from flwr.common.constant import ErrorCode
+from flwr.proto.error_pb2 import Error  # pylint: disable=E0611
+from flwr.proto.node_pb2 import Node  # pylint: disable=E0611
+from flwr.proto.task_pb2 import Task, TaskIns, TaskRes  # pylint: disable=E0611
+
+NODE_UNAVAILABLE_ERROR_REASON = (
+    "Error: Node Unavailable - The destination node is currently unavailable."
```
jafermarq, line 27:
```suggestion
    "Error: Node Unavailable - The destination node is currently unavailable. " \
    "It didn't send a heartbeat."
```

jafermarq: I think it would be informative this is due to heartbeat?

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+def make_node_unavailable_taskres(ref_taskins: TaskIns) -> TaskRes:
+    """Create a TaskRes containing the node available error based on the reference
+    TaskIns."""
+    return TaskRes(
+        task_id=str(uuid4()),
+        group_id=ref_taskins.group_id,
+        run_id=ref_taskins.run_id,
+        task=Task(
+            producer=Node(node_id=0, anonymous=True),
```
danieljanes, line 39: Open question: should we set the `Node` to anonymous here, or use the `node_id` of the Node that became unavailable?

src/py/flwr/server/superlink/state/utils.py
danieljanes, line 42: Should we set the TTL based on the `ref_taskins` TTL, similar to how we do it in `Message.create_reply`?

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -408,6 +410,54 @@ def get_task_res(self, task_ids: Set[UUID], limit: Optional[int]) -> List[TaskRe
...
+            FROM node
+            WHERE node_id IN ({placeholders})
+            AND online_until < :time;
+        """
+        data = {f"id_{i}": str(node_id) for i, node_id in enumerate(node_ids)}
+        data["time"] = time.time()
+        offline_node_ids = [int(row["node_id"]) for row in self.query(query, data)]
+
+        # 3. Query: Select TaskIns for offline nodes
+        placeholders = ",".join([f":id_{i}" for i in range(len(offline_node_ids))])
```
jafermarq, line 442: is this correct? we are ignoring the values of `offline_node_ids`.

src/py/flwr/server/superlink/state/utils.py
danieljanes: Yes, we can add this. But let's call it ping everywhere.

src/py/flwr/server/superlink/state/utils.py
danieljanes: I'm in favour of setting it using the same logic as in `create_reply`

panh99: Thanks. I've updated to using `DEFAULT_TTL`.

src/py/flwr/server/superlink/state/utils.py
panh99: I also had that question. I am inclined to set the `Node` to anonymous here, because the de facto "producer" of this Task is the Superlink. If we set the field to the unavailable node, it would break the convention that the producer_node_id is the place where the task is created. But I might understand the `producer_node_id` wrongly.

Wdyt? @danieljanes @jafermarq

src/py/flwr/server/superlink/state/utils.py
panh99: Sure, that makes sense.

src/py/flwr/server/superlink/state/sqlite_state.py
panh99: I am not sure if I follow. What do you mean by "ignoring the values"?

jafermarq: false alarm 😅

danieljanes (approved):

PR #3209, opened 2024-04-03
Hide members and make `RecordSet` picklable.

panh99:
We previously expose both public and internal members of `RecordSet`. Users were being presented with pairs of properties for managing parameters, metrics, and configurations (`parameters_records`, `_parameters_records`; `metrics_records`, `_metrics_records`; `configs_records`, `_configs_records`). This redundancy not only cluttered the autocomplete suggestions but also appeared unprofessional and potentially confusing for users interacting with the API.

The PR solves this issue by hiding the internal members. In addition, it makes the `RecordSet` work with `pickle`.

Previous:
image
image

Improved:
image


jafermarq:
I wonder if at least a super minimal test could be added to `recordset_test.py`:

```
def test_record_is_picklable():
    pickle.dumps(RecordSet())
```

This passes with the updates in this PR.

danieljanes (approved):

PR #3218, opened 2024-04-04
refactor(framework:skip) Hide private `TypedDict` members and inherit from `MutableMapping`

panh99:
After inheriting `MutableMapping`, our `TypedDict` can be recognized by IDE, mypy and pylint.


jafermarq (approved):
looks good!

PR #3228, opened 2024-04-05
Feature/remove poetry from quickstart tensorflow

chongshenng:
### Description

`poetry` and `pip` does not manage the environments consistently. This PR removes `poetry` from the `quickstart-tensorflow/` example.

### Explanation

Use `pyproject.toml` to manage dependencies and install with the `hatchling` backend.

### Checklist

- [X] Make CI checks pass
- [X] Ping maintainers on [Slack](https://flower.ai/join-slack/) (channel `#contributions`)

### Changelog entry

<examples>


examples/quickstart-tensorflow/pyproject.toml
```diff
@@ -1,16 +1,20 @@
 [build-system]
-requires = ["poetry-core>=1.4.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["hatchling"]
```
charlesbvll, line 2: Does that require an install?

charlesbvll: From my testing it doesn't seem like it... But does it work on all systems?

examples/quickstart-tensorflow/README.md
```diff
@@ -15,39 +15,28 @@ This will create a new directory called `quickstart-tensorflow` containing the f
...
+# From a new python environment, run:
+pip install .
 ```
 
-Poetry will install all your dependencies in a newly created virtual environment. To verify that everything works correctly you can run the following command:
+Then, to verify that everything works correctly you can run the following command:
 
 ```shell
-poetry run python3 -c "import flwr"
+python -c "import flwr"
```
charlesbvll, line 35: @danieljanes @jafermarq Have we taken a decision for replacing `python3` with `python`?

jafermarq: I think the last time we discussed you recommended using `python3`? I always use `python` though

examples/quickstart-tensorflow/pyproject.toml
chongshenng: > Does that require an install? 

I don't believe it does.

> But does it work on all systems?

What systems did you have in mind? My initial thought was that this is a simple example to allow others to build on. We could equivalently use only `requirements.txt` and avoid `pyproject.toml` altogether. Perhaps it's a matter of opinion? 

One possible benefit with `pyproject.toml` that I can see here is that after running `pip install .` in each example, we effectively build a package (perhaps in an atypical way). This means that when we query `pip show {name-of-example}`, it will display the Flower authors and other metadata.

examples/quickstart-tensorflow/README.md
chongshenng: I would assume that the majority of users would have `python` aliased to `python3` by default. I'm in favour of just specifying `python` for simplicity since it's one character shorter 😄 .

chongshenng: Updated to `python3`.

jafermarq (approved):

PR #3236, opened 2024-04-08
Sending parameters beyond the 2GB protobuf limit

brianlck:
Currently, the Flower framework relies on Protobuf which has a message size limit of 2GB. For large modern ML models, the total size of parameters easily exceed 2GB. This PR extends the Flower GRPC protocol to workaround the limit using one of the two strategies:

1. Uploading parameter to AWS bucket
2. Chunking parameters into a consecutive stream of RPC packets

On the technical side, this PR:
1. Suggests a new wire protocol for grpc-bidi to support streaming
2. Updates the Python API to select between two strategies

Looking forward, if this wire protocol is appropriate, the low level C++ and Kotlin API should be updated.

This PR is an extension of a project submission to the Bachelor level Federated Learning course at Cambridge.


jafermarq:
Hey @brianlck, thanks for opening this pr and sorry for the delay ! (I see there was no github issue associated with it) . The good news is that large-weights support is not built-in into Flower 1.20.0 ! 🚀  It was a huge refactoring effort but I think result it's pretty good (streaming and broadcasting of models comes built-in too). Check out the release blogpost: https://flower.ai/blog/2025-07-29-announcing-flower-1.20-release

PR #3240, opened 2024-04-09
feat(flwr) Introduce flower-supernode

danieljanes:


src/py/flwr/client/supernode/app.py
```diff
@@ -0,0 +1,107 @@
...
+
+    event(EventType.RUN_SUPERNODE_ENTER)
+
+    args = _parse_args_run_supernode().parse_args()
+
+    log(
+        DEBUG,
+        "Flower will load ClientApp `%s`",
+        getattr(args, "client-app"),
+    )
```
tanertopal, line 37: @danieljanes, how does this relate to the multiapp support?

danieljanes: It doesn't. Once we merge the implementation of this command, the main way to load apps will be from the app install dir. We might want to keep support for this arg though to make dev/prototyping easier.

PR #3253, opened 2024-04-12
Add `GetRun` rpc to FleetAPI.

panh99:


src/proto/flwr/proto/fleet.proto
```diff
@@ -68,4 +70,11 @@ message PushTaskResResponse {
   map<string, uint32> results = 2;
 }
 
+// GetRun messages
+message GetRunRequest { sint64 run_id = 1; }
+message GetRunResponse {
+  string fab_id = 1;
+  string fab_version = 2;
+}
```
danieljanes, line 78: Should we create a `Run` message and embed it as a single field in `GetRunResponse`? It'd have the current fields plus `run_id`. If we don't want `Run`, then we should probably add the `run_id` to `GetRunResponse`.

panh99: That makes sense. Sure, I will update the proto.

danieljanes (approved):

PR #3256, opened 2024-04-12
feat(framework:cli): Add `flwr build` command

charlesbvll:
### Checklist

- [x] Make CI checks pass


panh99 (commented):
The PR looks great! Would you mind adding a short example in the PR description about how to use `flwr build`?

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,156 @@
...
+
+    It can also be ran with a directory argument:
+
+        `flwr build --directory ./docs/target_dir`
+
+    And finally, the `--signed` argument can also be used to sign the FAB:
+
+        `flwr build --signed`
+
+    Note that this will prompt the user for its private key.
```
danieljanes, line 57:
```suggestion
    """Build a Flower project into a Flower App Bundle (FAB).

    You can run `flwr build` without any arguments to bundle the current directory:

        `flwr build`

    You can also build a specific directory:

        `flwr build --directory ./projects/flower-hello-world`

    And finally, to sign the resulting `.fab` file, use the `--signed` argument:

        `flwr build --sign`

    Note that this will prompt you for its private key to use for signing.
```

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,156 @@
...
+from .utils import is_valid_project_name
+
+
+# pylint: disable=too-many-locals
+def build(
+    directory: Annotated[
+        Optional[Path],
+        typer.Option(help="The Flower project directory to bundle into a FAB"),
+    ] = None,
+    signed: Annotated[bool, typer.Option(help="Flag to sign the FAB")] = False,
```
danieljanes, line 38: Let's use imperative form here (`signed` -> `sign`):

```suggestion
    sign: Annotated[bool, typer.Option(help="Flag to sign the FAB")] = False,
```

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,156 @@
...
+
+    if not is_valid_project_name(directory.name):
+        typer.secho(
+            f"❌ The project name {directory.name} is invalid, "
+            "a valid project name must start with a letter or an underscore, "
+            "and can only contain letters, digits, and underscores.",
+            fg=typer.colors.RED,
+            bold=True,
+        )
+        raise typer.Exit(code=1)
```
danieljanes, line 79: This is too restrictive. The directory name of the project should allow `-` (quite common in Python library names). The Python package inside will then use `_` instead of `-` to make it a valid Python package name.
Example:
- Project name: `flwr-datasets` (e.g., `pip install flwr-datasets`)
- Package name: `flwr_datasets`

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,156 @@
...
+
+    if validate_project_dir(directory) is None:
+        raise typer.Exit(code=1)
+
+    # Load .gitignore rules if present
+    ignore_spec = _load_gitignore(directory)
+
+    # Set the name of the zip file
+    fab_filename = f"{directory.name}.fab"
```
danieljanes, line 88: We should probably include the FAB version here as well. Do we also want to include the publisher?

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,156 @@
...
+                archive_path = file_path.relative_to(directory)
+                fab_file.write(file_path, archive_path)
+
+                # Calculate file info
+                sha256_hash = _get_sha256_hash(file_path)
+                file_size_bits = os.path.getsize(file_path) * 8  # size in bits
+                list_file_content += f"{archive_path},{sha256_hash},{file_size_bits}\n"
+
+        # Add LIST and LIST.jwt to the zip file
+        fab_file.writestr(".info/LIST", list_file_content)
```
danieljanes, line 115: How about naming this `.info/CONTENT`?

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Optionally sign the LIST file
+        if signed:
+            secret_key = typer.prompt(
+                "Enter the secret key to sign the LIST file", hide_input=True
```
danieljanes, line 120:
```suggestion
                "Enter the path to the secret key to sign the LIST file", hide_input=True
```

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,156 @@
...
+            )
+            signed_token = _sign_content(list_file_content, secret_key)
+            fab_file.writestr(".info/LIST.jwt", signed_token)
+
+    typer.secho(
+        f"🎊 Bundled FAB into {fab_filename}.", fg=typer.colors.GREEN, bold=True
```
danieljanes, line 126:
```suggestion
        f"🎊 Successfully built {fab_filename}.", fg=typer.colors.GREEN, bold=True
```

src/py/flwr/cli/build.py
charlesbvll: I added the FAB version, let's see for the publisher (what should the format be? `flwrlabs-fedgpt-1.0.0.fab`?)

src/py/flwr/cli/build.py
charlesbvll: I think the input is the key, not the path, I will check. Should I modify the function to extract the key from a path?

src/py/flwr/cli/build.py
danieljanes: We can already make the change here. All projects generated by `flwr new` would be compatible.

src/py/flwr/cli/build.py
danieljanes: Isn't the common approach to pass a file path? With Android, you pass the path to the key store file: https://developer.android.com/build/building-cmdline#sign_manually

src/py/flwr/cli/build.py
charlesbvll: Here is the PR: #3309

pyproject.toml
```diff
@@ -72,6 +72,8 @@ pycryptodome = "^3.18.0"
 iterators = "^0.0.2"
 typer = { version = "^0.9.0", extras=["all"] }
 tomli = "^2.0.1"
+pathspec = "^0.12.1"
+pyjwt = "^2.8.0"
```
tanertopal, line 76: Where is this used? I can't find an import.

tanertopal: nvm found it.

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,192 @@
...
+            fg=typer.colors.RED,
+            bold=True,
+        )
+
+    # Load .gitignore rules if present
+    ignore_spec = _load_gitignore(directory)
+
+    # Set the name of the zip file
+    fab_filename = (
+        f"{conf['project']['publisher']}"
```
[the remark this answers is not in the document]
charlesbvll: @jafermarq Correct! Good catch

src/py/flwr/cli/build.py
```diff
@@ -0,0 +1,211 @@
...
+    You can also build a specific directory:
+
+        `flwr build --directory ./projects/flower-hello-world`
+
+    And finally, to sign the resulting `.fab` file, use the `--sign` argument:
+
+        `flwr build --sign`
+
+    Note that this will prompt you for the path of the private RSA key to use for
+    signing the FAB file, instead, you can directly provide a `--key_path`
```
jafermarq, line 60:
```suggestion
    signing the FAB file. Alternatively, you can directly provide a `--key_path`
```

jafermarq (approved):

danieljanes (approved):

PR #3261, opened 2024-04-15
Add `State.get_run`

panh99:
- Added `State.get_run`.
- Extended `State.creat_run`.
- Implemented `SqliteState.get_run` and `InMemoryState.get_run`.
- Passed `fab_id="None/None", fab_version="None"` to `State.create_run` in `DriverServicer`.


src/py/flwr/server/superlink/state/state.py
```diff
@@ -150,9 +150,26 @@ def get_nodes(self, run_id: int) -> Set[int]:
         """
 
     @abc.abstractmethod
-    def create_run(self) -> int:
+    def create_run(self, fab_id: str, fab_version: str) -> int:
         """Create one run."""
```
danieljanes, line 154: Should we update the docstring here?

src/py/flwr/server/superlink/state/state.py
```diff
@@ -150,9 +150,26 @@ def get_nodes(self, run_id: int) -> Set[int]:
...
+        ----------
+        run_id : int
+            The identifier of the run.
+
+        Returns
+        -------
+        Tuple[str, str]
+            A tuple containing two strings:
+            - `fab_id`: The identifier of the Flower Fabric used in the specified run.
+            - `fab_version`: The version of the Flower Fabric used in the specified run.
```
danieljanes, line 170: Let's add the `run_id` here as well. We can use a tuple for now, but we should probably switch to a dataclass `Run` in the future.

src/py/flwr/server/superlink/state/state.py
panh99: Sure, I've updated.

src/py/flwr/server/superlink/state/state.py
panh99: I am okay with adding `run_id` here, but just a little bit confused. To me, it seems like `get_run` is the method that retrieving the details of a run given its ID, so naturally the definition of the method would be ID -> details. May I know why we also return run_id?

src/py/flwr/server/superlink/state/state.py
```diff
@@ -150,8 +150,26 @@ def get_nodes(self, run_id: int) -> Set[int]:
...
+        run_id : int
+            The identifier of the run.
+
+        Returns
+        -------
+        Tuple[int, str, str]
+            A tuple containing three elements:
+            - `run_id`: The identifier of the run, same as the specified `run_id`.
+            - `fab_id`: The identifier of the Flower Fabric used in the specified run.
+            - `fab_version`: The version of the Flower Fabric used in the specified run.
```
danieljanes, line 171:
```suggestion
            - `fab_id`: The identifier of the Flower App Bundle used in the specified run.
            - `fab_version`: The version of the Flower App Bundle used in the specified run.
```

src/py/flwr/server/superlink/state/state_test.py
```diff
@@ -43,6 +43,19 @@ def state_factory(self) -> State:
...
+        """Test if create_run and get_run work correctly."""
+        # Prepare
+        state: State = self.state_factory()
+        run_id = state.create_run("Mock/mock", "v1.0.0")
+
+        # Execute
+        _, fab_id, fab_version = state.get_run(run_id)
+
+        # Assert
+        assert fab_id == "Mock/mock"
```
danieljanes, line 56:
```suggestion
        # Execute
        actual_run_id, fab_id, fab_version = state.get_run(run_id)

        # Assert
        assert actual_run_id == run_id
        assert fab_id == "Mock/mock"
```

danieljanes (approved):

src/py/flwr/server/superlink/state/state.py
danieljanes: You're right, we could equally well not do this. There's no strong reason for doing this. It's more of a convenience thing. When the `Run` object gets more complex in the future, we can just return the full object and then serialize it and send it back to the SuperNode.

src/py/flwr/server/superlink/state/state.py
```diff
@@ -150,8 +150,26 @@ def get_nodes(self, run_id: int) -> Set[int]:
...
+            The identifier of the run.
+
+        Returns
+        -------
+        Tuple[int, str, str]
+            A tuple containing three elements:
+            - `run_id`: The identifier of the run, same as the specified `run_id`.
+            - `fab_id`: The identifier of the Flower App Bundle used in the specified run.
+            - `fab_version`: The version of the Flower App Bundle used in the specified run.
+        """
```
danieljanes, line 172:
```suggestion
            - `fab_id`: The identifier of the FAB used in the specified run.
            - `fab_version`: The version of the FAB used in the specified run.
        """
```

danieljanes (approved):

PR #3276, opened 2024-04-17
Improve speed of NaturalIdPartitioner

adam-narozniak:
## Issue
When adding and testing the `femnist` dataset on `NaturalIdPartitioner`, I discovered that partitioning takes a very long time.
### Description

The `feminist` dataset consists of over 800,000 samples corresponding to over 3,500 unique writers. 

The current implementation filters (using `Dataset.filter`) the dataset for each unique writer (when there's a request for a particular partition, it filters that for a single unique value). It took about 1 minute to filter the data for a single unique writer_id (it takes 3600 minutes to do that for all = 60 hours). Also, the filtering takes all columns instead of a single column.

## Proposal
Use a numpy function, and filter based on `ints` not the `strings` directly in the dataset.

### Changelog entry
<skip>

### DIFF
--- a/datasets/flwr_datasets/partitioner/natural_id_partitioner.py
+++ b/datasets/flwr_datasets/partitioner/natural_id_partitioner.py
@@ -14,10 +14,12 @@
 # ==============================================================================
 """Natural id partitioner class that works with Hugging Face Datasets."""
 
-
 from typing import Dict
 
+import numpy as np
+
 import datasets
+from flwr_datasets.common.typing import NDArrayInt
 from flwr_datasets.partitioner.partitioner import Partitioner
 
 
@@ -30,6 +32,8 @@ class NaturalIdPartitioner(Partitioner):
     ):
         super().__init__()
         self._partition_id_to_natural_id: Dict[int, str] = {}
+        self._natural_id_to_partition_id: Dict[str, int] = {}
+        self._partition_id_to_indices: Dict[int, NDArrayInt] = {}
         self._partition_by = partition_by
 
     def _create_int_partition_id_to_natural_id(self) -> None:
@@ -41,6 +45,26 @@ class NaturalIdPartitioner(Partitioner):
         self._partition_id_to_natural_id = dict(
             zip(range(len(unique_natural_ids)), unique_natural_ids)
         )
+
+    def _create_natural_id_to_int_partition_id(self) -> None:
+        """Create a mapping from unique client ids from dataset to int indices.
+
+        Natural ids come from the column specified in `partition_by`. This object is
+        inverse of the `self._partition_id_to_natural_id`. This method assumes that
+        `self._partition_id_to_natural_id` already exist.
+        """
+        self._natural_id_to_partition_id = {
+            value: key for key, value in self._partition_id_to_natural_id.items()
+        }
+
+    def _create_partition_id_to_indices(self) -> None:
+        """Create an assignment of indices to the partition indices."""
+        natural_ids = self.dataset[self._partition_by]
+        unique_natural_ids, inverse = np.unique(natural_ids, return_inverse=True)
+
+        for i, natural_id in enumerate(unique_natural_ids):
+            partition_id = self._natural_id_to_partition_id[natural_id]
+            self._partition_id_to_indices[partition_id] = np.where(inverse == i)[0]
 
     def load_partition(self, partition_id: int) -> datasets.Dataset:
         """Load a single partition corresponding to a single `partition_id`.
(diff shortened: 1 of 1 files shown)


datasets/flwr_datasets/partitioner/natural_id_partitioner.py
```diff
@@ -87,3 +151,17 @@ def partition_id_to_natural_id(self, value: Dict[int, str]) -> None:
         raise AttributeError(
             "Setting the partition_id_to_natural_id dictionary is not allowed."
         )
+
+    def _check_supported_type_of_value_in_partition_by(self) -> None:
+        values = self.dataset[self._partition_by]
```
jafermarq, line 156: I note this can take quite long. For example for our `"flwrlabs/shakespeare"` it takes ~9s. I'd propose only loading the first row only. wdyt?:
```
values = self.dataset.select(range(1))[self._partition_by]
```

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
```diff
@@ -42,6 +47,61 @@ def _create_int_partition_id_to_natural_id(self) -> None:
...
+        `self._partition_id_to_natural_id` already exist.
+        """
+        self._natural_id_to_partition_id = {
+            value: key for key, value in self._partition_id_to_natural_id.items()
+        }
+
+    def _create_partition_id_to_indices(self) -> None:
+        """Create an assignment of indices to the partition indices."""
+        natural_ids = np.array(self.dataset[self._partition_by])
+        unique_natural_ids = self.dataset.unique(self._partition_by)
```
jafermarq, line 64: Could we do this without having to load the whole dataset into memory?

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
```diff
@@ -42,6 +47,61 @@ def _create_int_partition_id_to_natural_id(self) -> None:
...
+            none_present = True
+            dtype = self.dataset.features[self._partition_by].dtype
+            none_replacement: Union[int, str]
+            if dtype == "string":
+                none_replacement = "None"
+                # Ensure the replacement is not in the dataset
+                while True:
+                    if none_replacement not in unique_natural_ids:
+                        break
+                    none_replacement += "1"
```
jafermarq, line 77: This will result in a `none_replacement` in the form `"None1111111111...."`. Should we instead do the `+= 1` and only when ending the `if` statement convert it to `str`? -- I understand this is just for internal use in the scope of this method though.

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
adam-narozniak: Yes, we can do that. In that case, the `np.object_` will be detected only if the `None` will be the first object, but that should work anyway.

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
adam-narozniak: To make it work quicker we need to load the column specified by `self._partition_by` to the memory. It's required later `np.unique(natural_ids, return_inverse=True)` (line 96) which returns the alternative representation (ints instead of e.g. strings of this column and at the end it speeds things up).

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
adam-narozniak: I need to change it into the string in each loop to check if it's not present in the `unique_natural_ids` but I can append based on counter so instead of `None1111` (if None1 None2 None3 strings were in the partition_by column) we might have `None4` (if None1, None2, None3 were in partition_by column). 
Both of them are very unlikely. This is to protect ourselves from an unlikely situation that the column of `partition_by` contains "None" (str) and None (None type). And then theoretically we might go further (maybe "None1" string is in the column so it can be changed further.

So, I'll change it to the counter.

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
adam-narozniak: I removed that function now

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
```diff
@@ -17,19 +17,52 @@
...
+    >>> partitioner = NaturalIdPartitioner(partition_by="character_id")
+    >>> fds = FederatedDataset(dataset="flwrlabs/shakespeare",
+    >>>                        partitioners={"train": partitioner})
+    >>> partition = fds.load_partition(0)
+
+    "sentiment140" (aka Twitter) dataset
+    >>> from flwr_datasets import FederatedDataset
+    >>> from flwr_datasets.partitioner import NaturalIdPartitioner
+    >>>
+    >>> partitioner = NaturalIdPartitioner(partition_by="character_id")
```
jafermarq, line 52:
```suggestion
    >>> partitioner = NaturalIdPartitioner(partition_by="user")
```

datasets/flwr_datasets/partitioner/natural_id_partitioner.py
```diff
@@ -17,19 +17,52 @@
...
+from tqdm import tqdm
+
 import datasets
+from flwr_datasets.common.typing import NDArrayInt
 from flwr_datasets.partitioner.partitioner import Partitioner
 
 
 class NaturalIdPartitioner(Partitioner):
-    """Partitioner for dataset that can be divided by a reference to id in dataset."""
+    """Partitioner for dataset that can be divided by a reference to id in dataset.
```
jafermarq, line 29:
```suggestion
    """Partitioner for a dataset that can be divided by a column with partition ids.
```

jafermarq (approved):

PR #3281, opened 2024-04-17
Use abstract `Driver`

jafermarq:
This PR:
- Renames the original `Driver` class to `GrpcDriver` and makes it derive from the abstract `Driver`
- The content of the abstract `Driver` is in `driver.py`
- The content of the original driver (now `GrpcDriver`) is now in `grpc_driver.py`
- Previous instantiation of `Driver` is now replaced with `GrpcDriver`


jafermarq:
Show we shorten the docstrings for `GrpcDriver`? since they are very much based on what the base `Driver` already contains...

danieljanes:
> Should we shorten the docstrings for `GrpcDriver`? since they are very much based on what the base `Driver` already contains...

Yes, that'd be good. The only docstrings users should refer to / be exposed to are the ones in the abstract `Driver` class.

danieljanes (approved):

PR #3284, opened 2024-04-18
feat(framework:cli) Add HuggingFace template for `flwr new`

charlesbvll:
### Description

We have no template for `HuggingFace` when using `flwr new`.

### Related issues/PRs

N/A

### Explanation

Add a new template that create a `HuggingFace` project.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


src/py/flwr/cli/new/new.py
```diff
@@ -140,6 +141,7 @@ def new(
     # Depending on the framework, generate task.py file
     frameworks_with_tasks = [
         MlFramework.PYTORCH.value.lower(),
+        MlFramework.HUGGINGFACE.value.lower(),
```
adam-narozniak, line 144: Could you explain why this needs to be here? I don't quite get the purpose of this python file and the top-level docs doesn't help much :(

src/py/flwr/cli/new/templates/app/code/client.hf.py.tpl
```diff
@@ -0,0 +1,55 @@
...
+        loss, accuracy = test(self.net, self.testloader)
+        return float(loss), len(self.testloader), {"accuracy": accuracy}
+
+
+def client_fn(cid):
+    # Load model and data
+    net = AutoModelForSequenceClassification.from_pretrained(
+        CHECKPOINT, num_labels=2
+    ).to(DEVICE)
+    trainloader, valloader = load_data(int(cid), 2)
```
adam-narozniak, line 46: How does the `cid` is passed here? I think in the original example we had `argparse`?

src/py/flwr/cli/new/new.py
charlesbvll: @adam-narozniak The `task.py` file contains most of the "non-flower dependent" functions needed to perform training and evaluation. Here, we list the frameworks that contain a `task.py` file in their template. Note that this approach might change in the future, as we might create a template folder for each framework, instead of assigning the template files based on the filenames

src/py/flwr/cli/new/templates/app/code/client.hf.py.tpl
adam-narozniak: Ok, got it. It'd be good to have that information somewhere, but since it relates to all simulation examples, maybe there's one common place that we can add it? 
Maybe it'd also be worth having the information that the example can't be run "manually" = launching each file yourself.

src/py/flwr/cli/new/new.py
adam-narozniak: Thanks for the explanation. It makes sense.

adam-narozniak (dismissed):
LGTM!

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
```diff
@@ -0,0 +1,31 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "$package_name"
```
danieljanes, line 6: Don't we use the `$project_name` here?

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
```diff
@@ -0,0 +1,31 @@
...
+version = "1.0.0"
+description = ""
+authors = [
+    { name = "The Flower Authors", email = "hello@flower.ai" },
+]
+license = {text = "Apache License (2.0)"}
```
danieljanes, line 12: Spacing is off

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
```diff
@@ -0,0 +1,31 @@
...
+dependencies = [
+    "flwr[simulation]>=1.8.0,<2.0",
```
danieljanes, line 14: Let's fix this (and all of the other examples) to `>=1.9.0,<2.0.0`

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
```diff
@@ -0,0 +1,31 @@
...
+    "flwr-datasets>=0.0.2,<1.0.0",
```
danieljanes, line 15: Should we go to `>=0.1.0`?

src/py/flwr/cli/new/templates/app/code/server.hf.py.tpl
```diff
@@ -0,0 +1,16 @@
+"""$project_name: A Flower / HuggingFace Transformers app."""
+
+import flwr as fl
+
+
+# Define strategy
+strategy = fl.server.strategy.FedAvg(
```
danieljanes, line 7: Let's use the shorter import format here and below (imports need to change):
```suggestion
strategy = FedAvg(
```

src/py/flwr/cli/new/templates/app/code/client.hf.py.tpl
```diff
@@ -0,0 +1,55 @@
...
+
+    # Return Client instance
+    return FlowerClient(net, trainloader, valloader).to_client()
+
+
+# Flower ClientApp
+app = fl.client.ClientApp(
```
danieljanes, line 53: (Imports need to change)
```suggestion
app = ClientApp(
```

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
charlesbvll: @danieljanes Can we move this to another PR? Same with the comment below, as they apply to all templates

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
charlesbvll: @danieljanes Same as above

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
danieljanes: yes

src/py/flwr/cli/new/templates/app/pyproject.hf.toml.tpl
danieljanes: yes

PR #3307, opened 2024-04-22
feat(framework) Add `InMemoryDriver`

jafermarq:
This PR introduces a new type of driver that communicates in-memory (as opposed to via `gRPC`).

Then https://github.com/adap/flower/pull/3355 makes this `InMemoryDriver` the one to be used for simulations.


src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -0,0 +1,187 @@
...
+
+        This method is used to collect messages from the SuperLink that correspond to a
+        set of given message IDs.
+        """
+        msg_ids = {UUID(msg_id) for msg_id in message_ids}
+        # Pull TaskRes
+        task_res_list = self.state.get_task_res(task_ids=msg_ids, limit=len(msg_ids))
+        # Convert TaskRes to Message
+        msgs = [message_from_taskres(taskres) for taskres in task_res_list]
+        return msgs
```
panh99, line 157: I think we need to delete delivered TaskRes and the TaskIns they reply to after we get them from the `state`. Otherwise, the TaskIns and TaskRes will pile up in the `state.

panh99 (dismissed):
lgtm!

src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -0,0 +1,189 @@
...
+from .driver import Driver
+
+
+class InMemoryDriver(Driver):
+    """`InMemoryDriver` class provides an interface to the Driver API.
+
+    Parameters
+    ----------
+    state_factory : StateFactory
+        A StateFactory embedding a state designed for in-memory communication.
```
danieljanes, line 37: Can't this use any kind of `State` implementation created by the `StateFactory`?

src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -0,0 +1,189 @@
...
+        self.fab_id = fab_id if fab_id is not None else ""
+        self.fab_version = fab_version if fab_version is not None else ""
+        self.node = Node(node_id=0, anonymous=True)
+        state = state_factory.state()
+
+        if not isinstance(state, InMemoryState):
+            raise NotImplementedError(
+                f"{self.__class__.__name__} is only compatible with an `InMemoryState` "
+                f"state. You used {type(state)}"
+            )
```
danieljanes, line 60: Why is this necessary?

src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -0,0 +1,189 @@
...
+        return self.run_id
+
+    def create_message(  # pylint: disable=too-many-arguments
+        self,
+        content: RecordSet,
+        message_type: str,
+        dst_node_id: int,
+        group_id: str,
+        ttl: Optional[float] = None,
+    ) -> Message:
```
danieljanes, line 93: Perhaps not in this PR, but in a later one: should we create a `BaseDriver` that has an implementation of `create_message` that is shared across all `Driver` implementations?

src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        task_ids: List[str] = []
+        for msg in messages:
+            # Check message
+            self._check_message(msg)
+            # Convert Message to TaskIns
+            taskins = message_to_taskins(msg)
+            # Store in state
+            taskins.task.pushed_at = time.time()
+            res = self.state.store_task_ins(taskins)
```
danieljanes, line 140: Let's rename this for clarity:
```suggestion
            task_id = self.state.store_task_ins(taskins)
```

src/py/flwr/server/driver/inmemory_driver.py
jafermarq: removed

src/py/flwr/server/driver/inmemory_driver.py
jafermarq: yes

danieljanes (approved):

PR #3326, opened 2024-04-24
Refactor `DriverClientProxy`

jafermarq:
This PR:
- modifies the `DriverClientProxy` so it now receives a full `Driver` object
   - It's `_send_receive_recordset()` therefore makes use of the Driver's object it embeds, operating directly on `common.Message` objects
- The `GrpcDriver`'s helper is renamed to `driver_helper` (with the idea of other driver instances to have helpers as well)
- The tests in `DriverClientProxyTestCase` have been updated to reflect then changes above

### DIFF
--- a/src/py/flwr/server/compat/driver_client_proxy.py
+++ b/src/py/flwr/server/compat/driver_client_proxy.py
@@ -16,16 +16,15 @@
 
 
 import time
-from typing import List, Optional
+from typing import Optional
 
 from flwr import common
-from flwr.common import DEFAULT_TTL, MessageType, MessageTypeLegacy, RecordSet
+from flwr.common import DEFAULT_TTL, Message, MessageType, MessageTypeLegacy, RecordSet
 from flwr.common import recordset_compat as compat
-from flwr.common import serde
-from flwr.proto import driver_pb2, node_pb2, task_pb2  # pylint: disable=E0611
+from flwr.proto import task_pb2  # pylint: disable=E0611
 from flwr.server.client_proxy import ClientProxy
 
-from ..driver.grpc_driver import GrpcDriverHelper
+from ..driver.driver import Driver
 
 SLEEP_TIME = 1
 
@@ -33,9 +32,7 @@ SLEEP_TIME = 1
 class DriverClientProxy(ClientProxy):
     """Flower client proxy which delegates work using the Driver API."""
 
-    def __init__(
-        self, node_id: int, driver: GrpcDriverHelper, anonymous: bool, run_id: int
-    ):
+    def __init__(self, node_id: int, driver: Driver, anonymous: bool, run_id: int):
         super().__init__(str(node_id))
         self.node_id = node_id
         self.driver = driver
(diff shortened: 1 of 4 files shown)


src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -16,26 +16,23 @@
...
-from typing import List, Optional
+from typing import Optional
 
 from flwr import common
-from flwr.common import DEFAULT_TTL, MessageType, MessageTypeLegacy, RecordSet
+from flwr.common import DEFAULT_TTL, Message, MessageType, MessageTypeLegacy, RecordSet
 from flwr.common import recordset_compat as compat
-from flwr.common import serde
-from flwr.proto import driver_pb2, node_pb2, task_pb2  # pylint: disable=E0611
+from flwr.proto import task_pb2  # pylint: disable=E0611
```
danieljanes, line 24: This import shouldn't be necessary

src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -116,67 +113,33 @@ def _send_receive_recordset(
...
+            group_id=str(group_id) if group_id else "",
+            ttl=DEFAULT_TTL,
         )
 
-        # Send TaskIns to Driver API
-        push_task_ins_res = self.driver.push_task_ins(req=push_task_ins_req)
-
-        if len(push_task_ins_res.task_ids) != 1:
+        # Push message
+        task_id_list = list(self.driver.push_messages(messages=[message]))
```
danieljanes, line 127: Should we rename this for clarity?
```suggestion
        message_ids = list(self.driver.push_messages(messages=[message]))
```

src/py/flwr/server/driver/grpc_driver_test.py
```diff
@@ -55,7 +55,7 @@ def tearDown(self) -> None:
     def test_check_and_init_grpc_driver_already_initialized(self) -> None:
         """Test that GrpcDriverHelper doesn't initialize if run is created."""
         # Prepare
-        self.driver.grpc_driver_helper = self.mock_grpc_driver
+        self.driver.driver_helper = self.mock_grpc_driver
```
danieljanes, line 58: Should we rename this for clarity?
```suggestion
        self.driver.driver_helper = self.mock_grpc_driver_helper
```

src/py/flwr/server/compat/driver_client_proxy.py
jafermarq: True, I moved the function `validate_task_res` that made use of it to the `driver_client_proxy_tests.py` (since we only made use of it in there)

src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -116,80 +112,34 @@ def _send_receive_recordset(
...
 
-        push_task_ins_req = driver_pb2.PushTaskInsRequest(  # pylint: disable=E1101
-            task_ins_list=[task_ins]
+        # Create message
+        message = self.driver.create_message(
+            content=recordset,
+            message_type=task_type,
+            dst_node_id=self.node_id,
+            group_id=str(group_id) if group_id else "",
+            ttl=DEFAULT_TTL,
```
danieljanes, line 122: Would it make sense to set this based on the provided `timeout`?

src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -116,80 +112,34 @@ def _send_receive_recordset(
...
         )
 
-        # Send TaskIns to Driver API
-        push_task_ins_res = self.driver.push_task_ins(req=push_task_ins_req)
-
-        if len(push_task_ins_res.task_ids) != 1:
+        # Push message
+        message_ids = list(self.driver.push_messages(messages=[message]))
+        if len(message_ids) != 1:
             raise ValueError("Unexpected number of task_ids")
```
danieljanes, line 129:
```suggestion
        if len(message_ids) != 1:
            raise ValueError("Unexpected number of message_ids")

```

src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -116,80 +112,34 @@ def _send_receive_recordset(
...
 
-        task_id = push_task_ins_res.task_ids[0]
+        task_id = message_ids[0]
```
danieljanes, line 130:
```suggestion
        message_id = message_ids[0]
```

src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -116,80 +112,34 @@ def _send_receive_recordset(
...
         if task_id == "":
             raise ValueError(f"Failed to schedule task for node {self.node_id}")
```
danieljanes, line 132:
```suggestion
        if message_id == "":
            raise ValueError(f"Failed to send message to node {self.node_id}")
```

src/py/flwr/server/compat/driver_client_proxy_test.py
```diff
@@ -76,13 +77,27 @@ def _make_task(
...
+
 class DriverClientProxyTestCase(unittest.TestCase):
     """Tests for DriverClientProxy."""
 
     def setUp(self) -> None:
         """Set up mocks for tests."""
-        self.driver = MagicMock()
-        self.driver.get_nodes.return_value = (
+        self.driver = GrpcDriver()
+        self.driver.driver_helper = MagicMock()
```
danieljanes, line 98: Wouldn't it be better to remove everything that's gRPC-related? Would could use just a mock `Driver` b/c that's what `DriverClientProxy` relies on

src/py/flwr/server/compat/driver_client_proxy.py
jafermarq: i think so. added.

danieljanes: IIRC, `timeout` is not a timestamp, but just the seconds for this to time out. We'll have to add the current timestamp seconds to it to get the TTL

src/py/flwr/server/driver/driver.py
```diff
@@ -16,14 +16,17 @@
...
+from typing import Any, Iterable, List, Optional
 
 from flwr.common import Message, RecordSet
 
 
 class Driver(ABC):
     """Abstract base Driver class for the Driver API."""
 
+    driver_helper: Any
+    run_id: Optional[int]
```
danieljanes, line 28: Should we create an abstract getter for this?

src/py/flwr/server/compat/driver_client_proxy.py
```diff
@@ -116,80 +112,34 @@ def _send_receive_recordset(
...
-                task_res = task_res_list[0]
-
-                # This will raise an Exception if task_res carries an `error`
-                validate_task_res(task_res=task_res)
-
-                return serde.recordset_from_proto(task_res.task.recordset)
+            messages = list(self.driver.pull_messages(message_ids))
+            if len(messages) == 1:
+                msg: Message = messages[0]
+                return msg.content
```
danieljanes, line 141: We need to raise an exception here if message has an error (similar to what was previously done on line 177)

danieljanes (approved):

PR #3356, opened 2024-04-29
ci(*:skip) Improve automated changelog generation

charlesbvll:
### Description

Our changelog generation script didn't work automatically with our new PR title format.

### Related issues/PRs

N/A

### Explanation

Use the PR title to categorise the PRs.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A

### DIFF
--- a/dev/update_changelog.py
+++ b/dev/update_changelog.py
@@ -18,10 +18,14 @@
 
 import pathlib
 import re
-import tomllib
+
+try:
+    import tomllib
+except ModuleNotFoundError:
+    import tomli as tomllib
 from datetime import date
 from sys import argv
-from typing import Dict, Optional, Set, Tuple
+from typing import Optional
 
 from github import Github
 from github.PullRequest import PullRequest
@@ -49,14 +53,14 @@ PATTERN_TEMPLATE = CONFIG["pattern_template"]
 PATTERN = PATTERN_TEMPLATE.format(types=TYPES, projects=PROJECTS, scope=SCOPE)
 
 
-def _get_latest_tag(gh_api: Github) -> Tuple[Repository, Optional[Tag]]:
+def _get_latest_tag(gh_api: Github) -> tuple[Repository, Optional[Tag]]:
     """Retrieve the latest tag from the GitHub repository."""
     repo = gh_api.get_repo(REPO_NAME)
     tags = repo.get_tags()
     return repo, tags[0] if tags.totalCount > 0 else None
 
 
-def _add_shorlog(new_version: str, shortlog: str) -> None:
+def _add_shortlog(new_version: str, shortlog: str) -> None:
     """Update the markdown file with the new version or update existing logs."""
     token = f"<!---TOKEN_{new_version}-->"
     entry = (
@@ -87,7 +91,7 @@ def _add_shorlog(new_version: str, shortlog: str) -> None:
 
 def _get_pull_requests_since_tag(
     repo: Repository, tag: Tag
-) -> Tuple[str, Set[PullRequest]]:
+) -> tuple[str, set[PullRequest]]:
     """Get a list of pull requests merged into the main branch since a given tag."""
     commit_shas = set()
     contributors = set()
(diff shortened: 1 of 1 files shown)


dev/update_changelog.py
```diff
@@ -0,0 +1,274 @@
...
+from github import Github
+from github.PullRequest import PullRequest
+from github.Repository import Repository
+from github.Tag import Tag
+
+REPO_NAME = "adap/flower"
+CHANGELOG_FILE = "doc/source/ref-changelog.md"
+CHANGELOG_SECTION_HEADER = "### Changelog entry"
+
+# Load the YAML configuration
```
danieljanes, line 35:
```suggestion
# Load the TOML configuration
```

danieljanes (commented):
Instead of moving `update_changelog` to `dev`, it would be better to keep it in `flwr_tool` and eventually move `flwr_tool` into a separate project

charlesbvll:
> Instead of moving `update_changelog` to `dev`, it would be better to keep it in `flwr_tool` and eventually move `flwr_tool` into a separate project

Shouldn't this be done in a separate PR? I feel like in terms of consistency, it makes more sense to have everything under `dev/` (as `flwr_tool` currently only contains the `protoc.py` script, whereas `dev/` contains all dev scripts similar to the `update_changelog.py` script)

danieljanes:
The "move `flwr_tool` into a new project" should happen separately. But we had `update_changelog.py` in `flwr_tool` already (this PR removes it), so I'd recommend to leave it there and make the changes in the version in `flwr_tool`.

charlesbvll:
> The "move `flwr_tool` into a new project" should happen separately. But we had `update_changelog.py` in `flwr_tool` already (this PR removes it), so I'd recommend to leave it there and make the changes in the version in `flwr_tool`.

@danieljanes But this new script uses the same `changelog_config.toml` file as does `check_pr_title` script. This is why it was moved to `dev/` so we don't have to do some relative path handling

doc/source/contributor-how-to-release-flower.rst
```diff
@@ -9,7 +9,7 @@ During the release
 The version number of a release is stated in ``pyproject.toml``. To release a new version of Flower, the following things need to happen (in that order):
 
 1. Run ``python3 src/py/flwr_tool/update_changelog.py <YOUR_GH_TOKEN>`` in order to add every new change to the changelog (feel free to make manual changes to the changelog afterwards until it looks good).
```
danieljanes, line 11:
```suggestion
1. Run ``python3 ./dev/update_changelog.py <YOUR_GH_TOKEN>`` in order to add every new change to the changelog (feel free to make manual changes to the changelog afterwards until it looks good).
```

dev/update_changelog.py
```diff
@@ -0,0 +1,290 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Update the changelog using PR titles."""
+
+
+import pathlib
+import re
+import tomllib
```
panh99, line 21: I think this won't work in Python <3.11. We should use `import tomli as tomllib` when the python version is less than 3.11.

dev/update_changelog.py
```diff
@@ -0,0 +1,290 @@
...
+    _update_changelog(prs)
+
+    new_version = _bump_minor_version(latest_tag)
+    if not new_version:
+        print("Wrong tag format.")
+        return
+
+    _add_shorlog(new_version, shortlog)
+
+    print("Changelog updated succesfully.")
```
panh99, line 286: Maybe we should always print this? If there's no `## Unreleased` tag, we will see:

```
Unreleased header not found in the changelog.
Changelog updated succesfully.
```

dev/update_changelog.py
```diff
@@ -0,0 +1,290 @@
...
+    gh_api = Github(argv[1])
+    repo, latest_tag = _get_latest_tag(gh_api)
+    if not latest_tag:
+        print("No tags found in the repository.")
+        return
+
+    shortlog, prs = _get_pull_requests_since_tag(repo, latest_tag)
+    _update_changelog(prs)
+
+    new_version = _bump_minor_version(latest_tag)
```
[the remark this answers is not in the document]
charlesbvll: No, and this is intentional, this script is meant to be ran multiple times and to only modify the changelog

PR #3360, opened 2024-04-29
Use four digits after the decimal point for DP logging

mohammadnaseri:
## Issue
Currently, DP log entries do not impose any limit on the number of digits after the decimal point.

### Description
This pull request updates the DP logging functionality to restrict the number of digits after the decimal point to four.
<!--
Describe the problem addressed by this PR.


jafermarq (approved):

tanertopal (approved):
lgtm!

PR #3361, opened 2024-04-29
Hide `run_simulation_from_cli`

jafermarq:
w/o this change users can do `flwr.simulation.run...` and it will autocomplete with options `run_simulation` and `run_simulation_from_cli`. The latter now is hidden.


jafermarq:
closed in favour of #3372

PR #3383, opened 2024-05-01
Add How-to guide for using Flower Driver API

chongshenng:
### Description
Our Flower Driver APIs lack official feature documentation. 

### Related issues/PRs
N/A

### Explanation
Add a how-to guide to using the features of Flower Driver APIs.

### Checklist

- ~~[ ] Write tests~~
- [ ] Make CI checks pass

### Changelog entry

Add a How-to guide on the Flower Driver API feature.

### Any other comments?
N/A

### DIFF
--- a/doc/source/how-to-use-driver-api.rst
+++ b/doc/source/how-to-use-driver-api.rst
@@ -1,0 +1,206 @@
+Use Driver API
+==============
+
+This guide explains how you can use Flower framework's |driverapi_link|_ API to customize the :code:`ServerApp` for *any* server workloads in your federated learning system. The Driver APIs are a specific set of server-side APIs that provide a robust way to communicate with client nodes in a Flower federated learning system. They include methods to:
+
+* `Select nodes`_
+* `Push messages to nodes`_
+* `Pull messages from nodes`_
+
+..
+    Generate link text as literal.
+
+.. |driverapi_link| replace:: ``Driver()``
+.. |recordset_link| replace:: ``RecordSet()``
+.. _driverapi_link: ref-api/flwr.server.Driver.html
+.. _recordset_link: ref-api/flwr.common.RecordSet.html
+
+Setup
+-----
+
+Before diving into the :code:`Driver` API, we need to setup your :code:`server.py` module. In a file :code:`server.py`, import :code:`flwr` and create a :code:`ServerApp`:
+
+.. code-block:: python
+
+    import flwr
+    from flwr.server import ServerApp, Driver
+    from flwr.common import Context, log
+
+    # Run via `flower-server-app server:app`
+    app = ServerApp()
+
+Next, we initialize the :code:`main()` function using :code:`@app.main()`. Within :code:`main()`, we define the workload that we want. For example, in federated learning workloads, we define a number of server rounds (:code:`server_round`) and loop over them throughout training:
+
+.. code-block:: python
+
+    @app.main()
+    def main(driver: Driver, context: Context):
+        num_rounds = 3
+    
+        for server_round in range(num_rounds):
+            print(f"Commencing server round {server_round + 1}")
+            
+            # Do something using Driver APIs
+
+Now that we have completed the setup, let’s walk through how we can use the Driver APIs for a federated learning example.
+
+
+Select nodes
+------------
+
+The first stage of a federated learning workflow is to select client nodes. For simplicity, we will refer to client nodes as nodes throughout this guide. 
+
+To select nodes in each server round, we run :code:`Driver.get_node_ids()` to retrieve all connected node IDs. Occasionally, the Driver API may not immediately return enough node IDs (e.g. due to real-world network issues), so we loop and wait until the minimum number of nodes are available. If we have a large pool of nodes which we want to sample from - as is commonly the case for federated learning systems - we can randomly sample node IDs using, for instance, :code:`random.sample(all_node_ids, num_client_nodes_per_round)`. One example is shown below:
+
+.. code-block:: python
+
+    # In `main()`:
+    ...
+    num_client_nodes_per_round = 3
+
+    for server_round in range(num_rounds):
+        log(INFO, f"Commencing server round {server_round + 1}")
+
+        # List of sampled node IDs in this round
+        sampled_nodes: List[int] = []
+
+        # Loop and wait until enough nodes are available.
+        while True:
+            all_node_ids = driver.get_node_ids()
+            log(INFO,f"Got {len(all_node_ids)} client nodes: {all_node_ids}")
+            if len(all_node_ids) >= num_client_nodes_per_round:
+                # Sample client nodes
+                sampled_nodes = random.sample(all_node_ids, num_client_nodes_per_round)
+                break
+            time.sleep(3)
+
+        # Log sampled node IDs
+        log(INFO,f"Sampled {len(sampled_nodes)} node IDs: {sampled_nodes}")
+
+        # Do something else
+    ...
+
+
+Push messages to nodes
+----------------------
+
+Now that we have a list of node IDs, we can push information to them to execute a task such as training or evaluating a model. To achieve this, we will first use :code:`Driver.create_message()` to create a :code:`Message` and then use :code:`Driver.push_message()` to push the :code:`Messages` to the list of node IDs. 
+
+.. admonition:: Note
+    :class: note
+
+    Each :code:`Message` contains a |recordset_link|_ object. It contains :code:`parameters_records`, :code:`metrics_records`, and :code:`configs_records` attributes, which are - unsurprisingly - parameters, metrics, and configurations that are used by a node to execute a task. 
+
+Here is an example of how to push a PyTorch model and instructions to a set of client node IDs to train the model. First, we define a utility function to convert a PyTorch model into a :code:`ParametersRecord` and create a PyTorch model in :code:`main()` (we’ve omitted the implementation details for :code:`Net()`, but you can refer to :doc:`this quickstart tutorial <tutorial-quickstart-pytorch>` for an example):
+
+.. code-block:: python
+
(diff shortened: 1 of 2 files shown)


doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+
+Now that we have completed the setup, let’s walk through how we can use the Driver APIs for a federated learning example.
+
+
+Select nodes
+------------
+
+The first stage of a federated learning workflow is to select client nodes. For simplicity, we will refer to client nodes as nodes throughout this guide. 
+
+To select nodes in each server round, we run :code:`Driver.get_node_ids()` to retrieve all connected node IDs. Occasionally, the Driver API may not immediately return enough node IDs (e.g. due to real-world network issues), so we loop and wait until the minimum number of nodes are available. If we have a large pool of nodes which we want to sample from - as is commonly the case for federated learning systems - we can randomly sample node IDs using, for instance, :code:`random.sample(all_node_ids, num_client_nodes_per_round)`. One example is shown below:
```
jafermarq, line 57: "real-world network issues" might be a bit vague for some readers without background about FL. A couple of examples that come to mind that we can mention: some nodes went offline, a node is taking longer than expected to startup.

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+        log(INFO,f"Sampled {len(sampled_nodes)} node IDs: {sampled_nodes}")
+
+        # Do something else
+    ...
+
+
+Push messages to nodes
+----------------------
+
+Now that we have a list of node IDs, we can push information to them to execute a task such as training or evaluating a model. To achieve this, we will first use :code:`Driver.create_message()` to create a :code:`Message` and then use :code:`Driver.push_message()` to push the :code:`Messages` to the list of node IDs.
```
jafermarq, line 91:
```suggestion
Now that we have a list of node IDs, we can push information to them to execute a task such as training or evaluating a model. To achieve this, we will first use :code:`Driver.create_message()` to create a :code:`Message` and then use :code:`Driver.push_message()` to push the :code:`Message` to the `SuperLink` (which will relay it to the appropiate `SuperNode`). 
```

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+----------------------
+
+Now that we have a list of node IDs, we can push information to them to execute a task such as training or evaluating a model. To achieve this, we will first use :code:`Driver.create_message()` to create a :code:`Message` and then use :code:`Driver.push_message()` to push the :code:`Messages` to the list of node IDs. 
+
+.. admonition:: Note
+    :class: note
+
+    Each :code:`Message` contains a |recordset_link|_ object. It contains :code:`parameters_records`, :code:`metrics_records`, and :code:`configs_records` attributes, which are - unsurprisingly - parameters, metrics, and configurations that are used by a node to execute a task. 
+
+Here is an example of how to push a PyTorch model and instructions to a set of client node IDs to train the model. First, we define a utility function to convert a PyTorch model into a :code:`ParametersRecord` and create a PyTorch model in :code:`main()` (we’ve omitted the implementation details for :code:`Net()`, but you can refer to :doc:`this quickstart tutorial <tutorial-quickstart-pytorch>` for an example):
```
jafermarq, line 98:
```suggestion
Here is an example of how to push a PyTorch model and instructions to a set of node IDs to train the model. First, we define a utility function to convert a PyTorch model into a :code:`ParametersRecord` and create a PyTorch model in :code:`main()` (we’ve omitted the implementation details for :code:`Net()`, but you can refer to :doc:`this quickstart tutorial <tutorial-quickstart-pytorch>` for an example):
```

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+    # Create a RecordSet
+    recordset = RecordSet()
+
+    # Add model parameters to the RecordSet
+    recordset.parameters_records["my_model"] = pytorch_to_parameter_record(global_model)
+
+    # Add a training configuration for 1 epoch only to the RecordSet
+    recordset.configs_records["my_config"] = ConfigsRecord({"epochs": 1})
+
+Next, we create a list of :code:`Messages`, one for each node ID. To do so, we loop over all node IDs and run :code:`Driver.create_message()` with the :code:`recordset` as the content of the message:
```
jafermarq, line 131:
```suggestion
Next, we create a list of :code:`Message` objects, one for each node ID. To do so, we loop over all node IDs and run :code:`Driver.create_message()` with the :code:`recordset` as the content of the message: 
```

jafermarq: I'm not sure what's better to do:
- `Message` is the class name. If we want to use plural, "`Message`s" looks terrible. I think "`Message` objects" is a better way.
- Saying "`Messages`" I think is not ideal either because there is not a class named `Messages`.

What do you think? What's the preferred approach?

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+
+.. code-block:: python
+
+    # File: server.py
+
+    messages = []
+    for node_id in node_ids:
+        message = driver.create_message(
+            content=recordset,
+            message_type=MessageType.TRAIN,
```
jafermarq, line 141:
```suggestion
            message_type=MessageType.TRAIN,     # <-------- will tell the `ClientApp` to run its train method.
```

jafermarq: I wonder if mentioning it like this or better in the sentence before this code block.

jafermarq: This is a good point to highlight that this exact `Message` object created is what the `ClientApp` will receive in its  `@app.train()`. Maybe linking, once more the other docs page about the `ClientApp`? #3510

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+
+Finally, we use :code:`Driver.push_messages()` to push the list of :code:`Messages` containing the encapsulated parameters and configurations to the nodes.
+
+.. code-block:: python
+
+    # File: server.py
+
+    message_ids = driver.push_messages(messages)
+
+:code:`Driver.push_messages()` yields an iterable list of message IDs. In some real-world scenarios, you may encounter situations where only some :code:`Messages` can be pushed, so it is good practice to filter out empty message IDs:
```
jafermarq, line 156: I think we should ellaborate on these "real-world" scenarios. For now, I'm only aware that a returned message_id will be empty if the `Message` is not well-formed (i.e. there are some inconsistencies in its `Metadata`). What are the other scenarios where `push_messages(0)` can return `""`, @panh99 ?

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+    # File: server.py
+
+    # Wait for results, ignore empty message_ids
+    message_ids = [message_id for message_id in message_ids if message_id != ""]
+
+
+Pull messages from nodes
+------------------------
+
+Once messages are successfully sent to the nodes, we can use the associated message IDs to get results from these nodes. To do so, we continuously run :code:`Driver.pull_messages()` with the list of :code:`message_ids` until all of the :code:`Messages` from the nodes are received.
```
jafermarq, line 169: I think it's worth mentioning that what's happening here is that the `ServerApp` is requesting the `SuperLink` for any result from the `Messages` previously pushed (of which we have retained the IDs).

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
...
+        for msg in all_replies
+        if msg.has_content()
+    ]
+    print(f"Received {len(all_replies)} results")
+
+
+Process results from nodes
+--------------------------
+
+Now that we have a results from the nodes in the form of :code:`Messages`, we can access their content and use them for any subsequent server-side tasks. Here is how we print the :code:`metrics_records` for each node in a for-loop:
```
jafermarq, line 204: similarly to my previous comment, should we stick to `Messages` (with that `s`) or change it?

jafermarq (commented):
@chongshenng Looks good overall. My main suggestion is to use the term `SuperNode` instead of "client nodes" throughout and to mention in a couple of key places that the `SuperLink` (which is not mentioned in the page) plays an important role (i.e. `push_message` and `pull_message` requests are processed by it)

doc/source/how-to-use-driver-api.rst
```diff
@@ -0,0 +1,254 @@
+Use Driver API
+==============
+
+This guide explains how you can use Flower framework's |driverapi_link|_ API to customize the :code:`ServerApp` for *any* server workloads in your federated learning system. The Driver APIs are a specific set of server-side APIs that provide a robust way to communicate with client nodes in a Flower federated learning system. They include methods to:
```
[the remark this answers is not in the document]
chongshenng: Added!

doc/source/how-to-use-driver-api.rst
chongshenng: @jafermarq 💯 I prefer the first approach. Will update throughout.

doc/source/how-to-use-driver-api.rst
chongshenng: I've updated them to `Message` objects.

doc/source/how-to-use-driver-api.rst
chongshenng: @jafermarq @panh99 How about for now, I emphasize that this is a "good practice" step. Later we can open another PR to elaborate on the scenarios if required? I can reword the sentence to this: 

"In some rare scenarios, you may encounter situations where not all `Message` objects can be pushed, e.g. if a `Message` is not well formed, so it is good practice to filter out empty message IDs:"

What do you think?

PR #3389, opened 2024-05-02
Prepare `State` to enable `node_id` verification

danielnugraha:
### Description

Current client authentication implementation does not verify whether `node_id` is assigned to a given public key.

### Explanation

Implement functions in `State` that enables verification of `node_id` assigned to a given public key.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/src/py/flwr/server/superlink/state/in_memory_state.py
+++ b/src/py/flwr/server/superlink/state/in_memory_state.py
@@ -225,7 +225,7 @@ class InMemoryState(State):  # pylint: disable=R0902,R0904
             if public_key is not None:
                 if (
                     public_key in self.public_key_to_node_id
-                    or self.public_key_to_node_id.get(public_key) == node_id
+                    or node_id in self.public_key_to_node_id.values()
                 ):
                     log(ERROR, "Unexpected node registration failure.")
                     return 0
@@ -244,7 +244,7 @@ class InMemoryState(State):  # pylint: disable=R0902,R0904
             if public_key is not None:
                 if (
                     public_key not in self.public_key_to_node_id
-                    and self.public_key_to_node_id.get(public_key) != node_id
+                    or node_id not in self.public_key_to_node_id.values()
                 ):
                     raise ValueError("Public key or node_id not found")
 
--- a/src/py/flwr/server/superlink/state/in_memory_state.py
+++ b/src/py/flwr/server/superlink/state/in_memory_state.py
@@ -30,19 +30,24 @@ from flwr.server.utils import validate_task_ins_or_res
 from .utils import make_node_unavailable_taskres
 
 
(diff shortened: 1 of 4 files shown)


src/py/flwr/server/superlink/state/in_memory_state.py
```diff
@@ -205,23 +210,46 @@ def num_task_res(self) -> int:
...
-        return 0
+            if node_id in self.node_ids:
+                log(ERROR, "Unexpected node registration failure.")
+                return 0
 
-    def delete_node(self, node_id: int) -> None:
+            if public_key is not None:
+                if (
+                    public_key in self.public_key_to_node_id
+                    or self.public_key_to_node_id.get(public_key) == node_id
```
danieljanes, line 228: This line seems to be unreachable. If the part after the `or` is true, doesn't it imply that the part before `or` must also be true?

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -534,26 +535,52 @@ def delete_tasks(self, task_ids: Set[UUID]) -> None:
...
-        query = "DELETE FROM node WHERE node_id = :node_id;"
-        self.query(query, {"node_id": node_id})
+        query = "SELECT node_id FROM node WHERE node_id = :node_id;"
+        row = self.query(query, {"node_id": node_id})
+
+        if len(row) == 0:
+            raise ValueError(f"Node {node_id} not found")
+
+        if public_key is not None:
+            query = "SELECT node_id FROM node WHERE public_key = :public_key;"
```
danieljanes, line 576: Do we need the initial two `SELECT` queries, or could we just have a single `DELETE WHERE` (either with or without the `public_key` part?

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -534,26 +535,54 @@ def delete_tasks(self, task_ids: Set[UUID]) -> None:
...
+        if self.conn is None:
+            raise AttributeError("State is not initialized.")
+
+        try:
+            with self.conn:
+                rows = self.conn.execute(query, params)
+                if rows.rowcount < 1:
+                    raise ValueError("Public key or node_id not found")
+        except KeyError as exc:
+            log(ERROR, {"query": query, "data": params, "exception": exc})
```
danieljanes, line 585: Should we also raise an exception here?

danielnugraha: The self.query function logs it. And the error comes from the query that doesn't match its params, which is pretty unlikely since the unit test passes. We can just leave it and not catch the exception at all.

danieljanes (approved):

PR #3404, opened 2024-05-06
ci(framework:skip) Add reconnection tests

charlesbvll:
### Description

We currently have no way to test the reliability of the SuperLink and SuperNodes when connection issues arise.

### Related issues/PRs

N/A

### Explanation

Add E2E test for testing reconnection between the SuperLink and SuperNodes.

### Checklist

- [x] Write tests
- [x] Make CI checks pass

### Changelog entry

<skip>

### Any other comments?

N/A


e2e/test_reconnection.sh
```diff
@@ -0,0 +1,75 @@
...
+sleep 3
+
+timeout 2m flower-client-app client:app --insecure $rest_arg --server $server_address &
+cl1_pid=$!
+sleep 3
+
+timeout 2m flower-client-app client:app --insecure $rest_arg --server $server_address &
+cl2_pid=$!
+sleep 3
+
```
chongshenng, line 38: I think it could be useful to echo a message here, so that we know what stage of the test we are at?

e2e/test_reconnection.sh
```diff
@@ -0,0 +1,75 @@
...
+# Kill superlink
+kill $sl_pid
+sleep 3
+
+# Restart superlink
+timeout 2m flower-superlink --insecure $db_arg $rest_arg &
+sl_pid=$!
+sleep 20
+
```
chongshenng, line 47: Same comment about the echo.

e2e/test_reconnection.sh
```diff
@@ -0,0 +1,75 @@
...
+sleep 1
+
+# Restart first client
+timeout 2m flower-client-app client:app --insecure $rest_arg --server $server_address &
+cl1_pid=$!
+
+wait $pid
+res=$?
+
+if [[ "$res" = "0" ]];
```
chongshenng, line 71: I wonder if it would it be better to use [`bats`](https://github.com/bats-core/bats-core) to test this shell script rather than relying on the exit statement? There's an example in their CI [here](https://github.com/bats-core/bats-core/blob/master/.github/workflows/tests.yml#L200-L212).

chongshenng (changes requested):
@charlesbvll I think the logic is fine, but I'm not familiar enough with the expected behaviour of killing and restarting the superlink/clients. Can you add some comments in the script to explain the expected behaviour? Also, I commented about using `bats` to test the shell script, but I appreciate that it could be out scope of this PR.

e2e/test_reconnection.sh
charlesbvll: I didn't know about this library, it looks interesting! I think this should be left for another PR though (as it would apply to many places in our codebase, and we might also ditch bash for most of our tests)

e2e/test_reconnection.sh
```diff
@@ -0,0 +1,88 @@
...
+    server_address="127.0.0.1:9092"
+    server_app_address="127.0.0.1:9091"
+    db_arg="--database $(date +%s).db"
+    ;;
+  *)
+    rest_arg=""
+    server_address="127.0.0.1:9092"
+    server_app_address="127.0.0.1:9091"
+    db_arg="--database :flwr-in-memory-state:"
+    ;;
```
panh99, line 22: Do we need this part if in `e2e.yml`, we only have `./../test_reconnection.sh sqlite`?

e2e/test_reconnection.sh
```diff
@@ -0,0 +1,88 @@
...
+sleep 5
+
+# We start the server-app to begining the training
+timeout 2m flower-server-app server:app --insecure $dir_arg $rest_arg --server $server_app_address &
+pid=$!
+echo "Starting server-app to start training"
+
+# Kill first client as soon as the training starts,
+# the server-app should just receive a failure in this case and continue the rounds
+# when enough clients are connected
```
panh99, line 71: Just tested locally. It seems that, the server app will wait for enough clients. It won't receive any failure. Is this also as expected?

panh99 (commented):
Looks good to me, generally. Just a quick q: may I know the purpose of this test? Is it to ensure that no exception will be raised after killing/restarting clients?

e2e/test_reconnection.sh
charlesbvll: It was in case we want to extend the test in the future (currently `in_memory` state won't work)

e2e/test_reconnection.sh
charlesbvll: Yes! Maybe I should update the comment...

charlesbvll:
> Looks good to me, generally. Just a quick q: may I know the purpose of this test? Is it to ensure that no exception will be raised after killing/restarting clients?

Yes exactly, we want to simulate unstable connections and make sure that the different component do not crash when they lose the connection with the others

tanertopal (approved):
Lgtm!

chongshenng (approved):
LGTM

PR #3422, opened 2024-05-08
fix(framework:skip) Handle no response for params init in workflow

charlesbvll:
### Description

Currently, if the sampled client for getting initial parameters fails or doesn't reply, we will get this error:

```py
  File "/Users/main/.pyenv/versions/flower-3.10/lib/python3.10/site-packages/flwr/server/workflow/default_workflows.py", line 63, in __call__
    default_init_params_workflow(driver, context)
  File "/Users/main/.pyenv/versions/flower-3.10/lib/python3.10/site-packages/flwr/server/workflow/default_workflows.py", line 135, in default_init_params_workflow
    paramsrecord = next(iter(msg.content.parameters_records.values()))
  File "/Users/main/.pyenv/versions/flower-3.10/lib/python3.10/site-packages/flwr/common/message.py", line 255, in content
    raise ValueError(
ValueError: Message content is None. Use <message>.has_content() to check if a message has content.
```

### Related issues/PRs

N/A

### Explanation

We instead want to resample another available client and send a get parameters request until it is successful.

### Checklist

- [x] Make CI checks pass

### Changelog entry

<skip>

### Any other comments?

N/A

### DIFF
--- a/src/py/flwr/server/workflow/default_workflows.py
+++ b/src/py/flwr/server/workflow/default_workflows.py
@@ -106,6 +106,7 @@ def default_init_params_workflow(driver: Driver, context: Context) -> None:
     if not isinstance(context, LegacyContext):
         raise TypeError(f"Expect a LegacyContext, but get {type(context).__name__}.")
 
+    paramsrecord = None
     parameters = context.strategy.initialize_parameters(
         client_manager=context.client_manager
     )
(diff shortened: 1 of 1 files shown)


src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -115,40 +116,47 @@ def default_init_params_workflow(driver: Driver, context: Context) -> None:
...
+                [
+                    driver.create_message(
+                        content=content,
+                        message_type=MessageTypeLegacy.GET_PARAMETERS,
+                        dst_node_id=random_client.node_id,
+                        group_id="0",
+                    )
+                ]
+            )
+            log(INFO, "Received initial parameters from one random client")
```
jafermarq, line 135: Should we move this inside the `if msg.has_content()`?

src/py/flwr/server/workflow/default_workflows.py
```diff
@@ -115,40 +116,47 @@ def default_init_params_workflow(driver: Driver, context: Context) -> None:
...
-    log(INFO, "Evaluating initial global parameters")
-    parameters = compat.parametersrecord_to_parameters(paramsrecord, keep_input=True)
-    res = context.strategy.evaluate(0, parameters=parameters)
-    if res is not None:
-        log(
-            INFO,
-            "initial parameters (loss, other metrics): %s, %s",
-            res[0],
-            res[1],
+    if paramsrecord is not None:
```
jafermarq, line 142: Can this ever be not `None`?

panh99: Maybe when the client returns an `Error`?

jafermarq: we'll never exit the `while True` if that's the case since we check for `.has_content()`

charlesbvll: I only added this so the variable wouldn't be unbound. It might still pass the tests if I remove it

src/py/flwr/server/workflow/default_workflows.py
charlesbvll: Good point

PR #3426, opened 2024-05-10
fix(examples) Update XGBoost quick-start example for independent object instantiation

yan-gao-GY:
### Description
The xgboost-quickstart example instantiates xgb object at the first round and persist over FL rounds. This is ok for long-running clients in deployment mode within cross-silo settings. However, it is not feasible to use Flower Simulation and migrate to Flower Next.

## Proposal
Implement the same Xgb client as the comprehensive xgb example does, where the xgb object is instantiated independently every single round. No need to persist the xgb object over FL rounds.

### Changelog entry
<examples>


examples/xgboost-quickstart/client.py
```diff
@@ -106,41 +118,41 @@ def get_parameters(self, ins: GetParametersIns) -> GetParametersRes:
...
-        for i in range(num_local_round):
-            self.bst.update(train_dmatrix, self.bst.num_boosted_rounds())
+        for i in range(self.num_local_round):
+            bst_input.update(self.train_dmatrix, bst_input.num_boosted_rounds())
 
-        # Extract the last N=num_local_round trees for sever aggregation
-        bst = self.bst[
-            self.bst.num_boosted_rounds()
-            - num_local_round : self.bst.num_boosted_rounds()
+        # Bagging: extract the last N=num_local_round trees for sever aggregation
```
chongshenng, line 126: Small typo: sever > server

examples/xgboost-quickstart/requirements.txt
chongshenng, line 1: I'd suggest removing `requirements.txt` and use only `pyproject.toml` for the installation.

examples/xgboost-quickstart/pyproject.toml
chongshenng, line 1: I suggest using this `pyproject.toml` to install the dependencies for the project - it no longer depends on `poetry`:

```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
name = "quickstart-xgboost"
version = "0.1.0"
description = "XGBoost Federated Learning Quickstart with Flower"
authors = [
    { name = "The Flower Authors", email = "hello@flower.ai" },
]
dependencies = [
    "flwr>=1.8.0,<2.0",
    "flwr-datasets>=0.1.0,<1.0.0",
    "xgboost>=2.0.0,<3.0.0",
]

[tool.hatch.build.targets.wheel]
packages = ["."]
```

chongshenng: I also bumped the minimum version of Flower to 1.8.

chongshenng (changes requested):
Overall looks fine, thanks @yan-gao-GY. I think one more PR is needed to convert this fully to use `ClientApp` and `ServerApp`. (We should place `start_client` and `start_server` within a `if __name__ == "__main__"` for this). Let's draft another PR for this upgrade. 

You'll also need to update the PR title to conform to the new format. I think something like `fix(examples) Update XGBoost quickstart examples for independent object instantiation` could work.

examples/xgboost-quickstart/pyproject.toml
yan-gao-GY: Ok! so we don't mention that `we recommend Poetry to install those dependencies` any more in readme right? Is this the thing we want to do for all examples? @jafermarq

jafermarq: yes, we are gradually removing the need for poetry in the examples. An examples of an updated example is [quickstart-pytorch](https://github.com/adap/flower/tree/main/examples/quickstart-pytorch#installing-dependencies).

jafermarq:
> @chongshenng Thanks a lot for your comments and suggestions!
> 
> Do we want to fully replace `start_client` and `start_server` now? I saw other examples keep both and using Flower Next with a `preview` label.

Let's do that later in a different PR.

jafermarq (approved):

chongshenng (approved):
lgtm!

PR #3432, opened 2024-05-11
docs(*) Remove placeholder search text to Flower documentation

chongshenng:
Remove `search.html` that's preventing the correct version of docs to be built.


tanertopal (approved):
lgtm! Thank you for the revert.

PR #3438, opened 2024-05-13
doc(framework:skip) Fix versioned docs by forcing checkout

charlesbvll:
### Description

The current implementation of `build-version-docs.sh` can break silently and make the framework docs outdated in production.

### Related issues/PRs

N/A

### Explanation

Make sure that an error will make the CI fail, and force checkout to discard changes.

### Checklist

- [x] Implement proposed change
- [x] Update the changelog entry below
- [x] Make CI checks pass
- [x] Ping maintainers on [Slack](https://flower.ai/join-slack/) (channel `#contributions`)

### Changelog entry

<skip>

### Any other comments?

N/A

PR #3441, opened 2024-05-14
doc(framework:skip) Clean all git history before checkout

charlesbvll:
### Description

We still have an error on checkout because files are not removed properly.

### Related issues/PRs

#3438 

### Explanation

Run `git clean -fd` before checkout.

### Checklist

- [x] Make CI checks pass

### Changelog entry

<skip>

### Any other comments?

N/A

### DIFF
--- a/doc/build-versioned-docs.sh
+++ b/doc/build-versioned-docs.sh
@@ -73,6 +73,8 @@ END
     # Actually building the docs for a given language and version
     sphinx-build -b html source/ build/html/${current_version}/${current_language} -A lang=True -D language=${current_language}
 
+    git clean -fd
+
   done
 done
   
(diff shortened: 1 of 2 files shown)


doc/build-versioned-docs.sh
```diff
@@ -73,6 +73,8 @@ END
     # Actually building the docs for a given language and version
     sphinx-build -b html source/ build/html/${current_version}/${current_language} -A lang=True -D language=${current_language}
 
+    git clean -fd
```
chongshenng, line 76: Should we include a comment here about what this does? If I understand correctly, the purpose is to remove any unwanted files that's not part of the checked-out repository tag. Since we start from the oldest branch and progressively check out newer tags, any folder/files structure that differ from newer branches will cause a `The following untracked working tree files would be overwritten by checkout:` error. I'm still fuzzy on the details.

charlesbvll: @chongshenng I added a small comment, wdyt? Should I be more verbose?

chongshenng: @charlesbvll Yes, that works!

chongshenng (approved):
LGTM!

tanertopal (approved):
lgtm!

PR #3447, opened 2024-05-14
Add initial pytest for compose.yaml

chongshenng:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/e2e/docker/test.py
+++ b/e2e/docker/test.py
@@ -1,0 +1,10 @@
+from testcontainers.compose import DockerCompose
+
+
+def test_compose():
+
+    # Create a tc-DockerCompose object
+    services = DockerCompose(context=".", compose_file_name="compose.yaml")
+
+    containers = services.get_containers(include_all=True)
+    assert len(containers) == 4


e2e/docker/test.py
```diff
@@ -0,0 +1,84 @@
...
+        services.stop()
+
+
+def test_compose_logs():
+    """Test if Flower containers can successfully finish training."""
+    # create the services
+    services = DockerCompose(context=".", compose_file_name=COMPOSE_MANIFEST)
+
+    with services:
+        sleep(45)  # generate some logs
```
Robert-Steiner, line 73: Is that just a placeholder value, or does it really only take 45 seconds on your machine?

e2e/docker/test.py
```diff
@@ -0,0 +1,84 @@
...
+        assert containers[1].Service == "superlink"
+        assert containers[2].Service == "supernode"
+        assert containers[3].Service == "supernode"
+
+        # stop the services but don't remove containers,
+        # networks, volumes, and images
+        services.stop(down=False)
+
+        # Check the state for ServerApp and SuperLink.
+        # TODO: Add test for SuperNode services  # pylint: disable=W0511
```
Robert-Steiner, line 49: It's a bit disappointing that `get_container` doesn't return an array when you have multiple services with the same `Service` value but a different `Name`.

Robert-Steiner (commented):
Look good to me!
Just added one question.

e2e/docker/test.py
chongshenng: It took 45 secs on my machine 😅. Perhaps there's a better way to do it than this heuristic. Alternatively, we can set `num_rounds=1` in `server.py`, which will reduce this `sleep` counter significantly.

e2e/docker/test.py
chongshenng: Yes, it's a small drawback. I _could_ [parameterize the test fixture](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) but I'm planning to revisit this later. IMO, I think these 2 tests are sufficient for now.

e2e/docker/test.py
```diff
@@ -0,0 +1,82 @@
...
+        services.stop()
+
+
+def test_compose_logs():
+    """Test if Flower containers can successfully finish training."""
+    # create the services
+    services = DockerCompose(context=".", compose_file_name=COMPOSE_MANIFEST)
+
+    with services:
+        sleep(35)  # add delay to ensure training completes
```
Robert-Steiner, line 72: For some reason the test takes 107 secs on my machine. Maybe it is because of my network.
What about replicating the [`wait_for_logs`](https://github.com/testcontainers/testcontainers-python/blob/9eabb79f213cfb6d8e60173ff4c40f580ae0972a/core/testcontainers/core/waiting_utils.py#L80) functionality for docker compose e.g.

```python
def test_compose_logs():
    """Test if Flower containers can successfully finish training."""
    # create the services
    services = DockerCompose(context=".", compose_file_name=COMPOSE_MANIFEST)

    with services:
        wait_for_logs(services, RUN_COMPLETE_STR, 120)

def wait_for_logs(
    services: "DockerCompose",
    predicate: Union[Callable, str],
    timeout: float = 60,
    interval: float = 1,
) -> float:
    """
    Wait for the services to emit logs satisfying the predicate.

    Args:
        services: Services whose logs to wait for.
        predicate: Predicate that should be satisfied by the logs. If a string, then it is used as
        the pattern for a multiline regular expression search.
        timeout: Number of seconds to wait for the predicate to be satisfied. Defaults to wait
            indefinitely.
        interval: Interval at which to poll the logs.

    Returns:
        duration: Number of seconds until the predicate was satisfied.
    """
    if isinstance(predicate, str):
        predicate = re.compile(predicate, re.MULTILINE).search
    start = time.time()
    while True:
        duration = time.time() - start
        stdout, stderr = services.get_logs()
        if predicate(stdout) or predicate(stderr):
            return duration
        if duration > timeout:
            raise TimeoutError(
                f"Services did not emit logs satisfying predicate in {timeout:.3f} "
                "seconds"
            )
        time.sleep(interval)
```
In this way, we make the execution a little more resilient by allowing the test to take a little longer if something irregular happens, such as a slow network connection.

chongshenng: Great idea! I somehow missed that implementation. Have included it. Thanks!

e2e/docker/test.py
```diff
@@ -0,0 +1,110 @@
+"""Tests for Flower Docker containers."""
+
+import re
+import time
+from typing import Callable, Union
+
+import pytest
+from testcontainers.compose import ContainerIsNotRunning, DockerCompose
```
tanertopal, line 8: Where is this dependency installed? I could not find it in the `pyproject.toml`.

chongshenng: Good point 😅! I've added it now to `pyproject.toml`.

Robert-Steiner (dismissed):
LGTM🚀

PR #3452, opened 2024-05-15
ci(*:skip) Add Python script to check PR titles

charlesbvll:
### Description

We want to enforce the following PR title format:

```
<type>(<project>) <subject>
```

(or `<type>(<project>:skip) <subject>` to ignore the PR in the changelog)

Where `<type>` $\in$ `{ci, fix, feat, docs, refactor, break}`, `<project>` $\in$ `{framework, baselines, datasets, examples, '*'}`, and `<subject>` starts with a capitalised verb in the imperative mood.

Valid examples:

- `feat(framework) Add flwr build CLI command`
- `refactor(examples:skip) Improve quickstart-pytorch logging`
- `ci(*:skip) Enforce PR title format`

Invalid examples:

- `feat(framework): Add flwr build CLI command` (extra `:`)
- `feat(skip) Add flwr build CLI command` (missing `<project>`)
- `feat(framework) add flwr build CLI command` (non capitalised verb)
- `feat(framework) Add flwr build CLI command.` (dot at the end)
- `Add flwr build CLI command` (missing `<type>(<project>)`)

### Related issues/PRs

N/A

### Explanation

- [x] Create a configuration file defining the defined rules
- [x] Write a Python script to check if a title has the correct format using this config file
- [x] Add a CI check to enforce the new format using this script, the CI should fail if the format is not respected.
- [x] Update the contributor docs

### Checklist

- [x] Implement proposed change
- [x] Write tests
- [x] Update [documentation](https://flower.ai/docs/writing-documentation.html)
- [x] Update the changelog entry below
- [x] Make CI checks pass
- [x] Ping maintainers on [Slack](https://flower.ai/join-slack/) (channel `#contributions`)

### Changelog entry

<skip>

### Any other comments?

N/A

PR #3461, opened 2024-05-16
feat(framework) Log a warning if the server fails to get initial parameters

panh99:
# Problem
Currently, in both `server.py` and `default_workflows.py`, the server side will not log anything when it fails to get initial parameters from `Strategy` and `Client`, and will proceed with empty parameters.

# Solution

> After a few rounds of discussion with @jafermarq , we think this is one of the cleanest solution to fix the issue.

Check the status code of the `GetParametersRes`. Log a warning if the status code is not `Code.OK`.


jafermarq (dismissed):
looks good to me. You'll need to update the title of the PR to follow the format in here #3452 . I guess the below would do:

```
feat(framework) Log a warning if the server fails to get initial parameters.
```

danieljanes (dismissed):

danieljanes (approved):

PR #3476, opened 2024-05-20
break(datasets) Rename resplitter parameter and type to preprocessor

adam-narozniak:
## Issue
What is currently named as `Resplitter= Callable[[DatasetDict], DatasetDict]` and in FederatedDataset(..., resplittter: Union[Resplitter, ...]` is not the most accurate description what can happen.

### Description

The `resplitter` can currently serve more as a `preprocessor` (it seems like a broader term to better capture the essence of the object). 
After the dataset is downloaded, more operations that resplitting can happen. I'll list a few of them:
* removal of some data (not meeting certain criteria)
* transformation e.g., different date format
* creation of new features
Currently, even though such operations can happen, they are captured as `Resplitter` (a callable that accomplishes the goal).

## Proposal

Rename the `resplitter` parameter of FederatedDataset and `Resplitter` the callable to `preprocessor` and `Preprocessor` to better capture the essence of what can happen.
Rename the `resplitter` directory to `preprocessor`.

Also simplify the following names:
* `MergeResplitter` to `Merger` 
* `DivideResplitter` to `Divider`


datasets/flwr_datasets/resplitter/merge_resplitter_test.py
```diff
@@ -25,7 +25,7 @@
 
 
 class TestResplitter(unittest.TestCase):
```
jafermarq, line 27: Should we rename to `TestPreprocessor`?

datasets/flwr_datasets/resplitter/__init__.py
```diff
@@ -12,13 +12,11 @@
...
 # ==============================================================================
-"""Resplitter package."""
+"""Preprocessor package."""
 
 
 from .merge_resplitter import MergeResplitter
-from .resplitter import Resplitter
 
 __all__ = [
     "MergeResplitter",
```
jafermarq, line 21: Should we introduce a `MergePreprocessor` class that basically does the same as `MergeResplitter`. Then, keep `MergeResplitter` as a dummy sub-class from the new `MergePreprocessor` that does nothing new but throws a deprecated warning encouraging users to use `MergePreprocessor` instead? wdyt?

datasets/flwr_datasets/federated_dataset.py
```diff
@@ -45,9 +45,11 @@ class FederatedDataset:
     subset : str
         Secondary information regarding the dataset, most often subset or version
         (that is passed to the name in datasets.load_dataset).
-    resplitter : Optional[Union[Resplitter, Dict[str, Tuple[str, ...]]]]
-        `Callable` that transforms `DatasetDict` splits, or configuration dict for
-        `MergeResplitter`.
+    preprocessor : Optional[Union[Preprocessor, Dict[str, Tuple[str, ...]]]]
+        `Callable` that transforms `DatasetDict` by resplitting, removing
+        features, creating new features, performing any other preprocessing operation,
+        or configuration dict for `MergeResplitter`. Applied after shuffling. If None,
```
jafermarq, line 51: following up with the idea in another comment, should we rename to `MergePreprocessor`?

jafermarq (commented):
@adam-narozniak looks good to me, some comments below. I wonder if it would make sense to rename the `resplitter` directory to `preprocessor`.

datasets/flwr_datasets/resplitter/__init__.py
adam-narozniak: I thought this could be left as it is. Resplitting is the main action this class (also DivideResplitter) does, but conceptually, it's a Preprocessor (but it does not have that in its name). Wdyt?

adam-narozniak:
> @adam-narozniak looks good to me, some comments below. I wonder if it would make sense to rename the `resplitter` directory to `preprocessor`.

We can do that, and I can move the preprocessor definition there, too. Maybe in the future, we will need to have more subdirs there, e.g., if we had more types of preprocessors.

datasets/flwr_datasets/resplitter/merge_resplitter_test.py
adam-narozniak: I think for that I should divide the file so there would be one for merge_resplitter and one for preprocessor in general. I'll do after we confirm the directory renaming.

adam-narozniak:
Based on the discussion off the github I did the following:
* rename the `resplitter` dir to `preprocessor`
* moved back the `preprocessor.py` (renamed from `resplitter.py`) 
* rename `Resplitter` to `Preprocessor`
* simplify the naming of *Resplitter to drop it from the names (so we have now `Merger` and `Divider`)
(the description of the PR is updated accordingly)

jafermarq (approved):

PR #3478, opened 2024-05-20
ci(*:skip) Enforce the skip flag on wildcard PRs

charlesbvll:
### Description

Currently the wildcard can be used for describing PRs modifying multiple projects, but the changelog doesn't support such PRs.

### Related issues/PRs

N/A

### Explanation

Those PRs should be marked as `:skip`, we therefore enforce the `:skip` flag to be used when `*` is used in the title.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A

### DIFF
--- a/dev/check_pr_title.py
+++ b/dev/check_pr_title.py
@@ -43,16 +43,32 @@ if __name__ == "__main__":
 
     # Check for the pattern in the first argument given to the script
     match = re.search(pattern, pr_title)
-    if match and match.group(4).split()[0] in allowed_verbs:
-        print("PR title is valid")
-        sys.exit(0)
+
+    valid = True
+    error = "it doesn't have the correct format"
+
+    if not match:
+        valid = False
     else:
+        if not match.group(4).split()[0] in allowed_verbs:
+            valid = False
+            error = "the <PR_SUBJECT> doesn't start with a verb in the imperative mood"
+        elif match.group(2) == "*" and match.group(3) is None:
+            valid = False
+            error = "the <PR_PROJECT> cannot be '*' without using the ':skip' flag"
+
+    if not valid:
         print(
-            f"PR title `{pr_title}` is invalid, it should be of the form:\n\n\t<PR_TYPE>(<PR_PROJECT>) "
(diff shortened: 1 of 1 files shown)


dev/check_pr_title.py
```diff
@@ -42,16 +42,32 @@
...
+            f"<PR_SUBJECT>\n\nOr, if the PR shouldn't appear in the changelog:\n\n\t"
+            f"<PR_TYPE>(<PR_PROJECT>:skip) <PR_SUBJECT>\n\nwith <PR_TYPE> in [{types}],\n"
+            f"<PR_PROJECT> in [{'|'.join(config['project']) + '|*'}] (where '*' is used "
+            "when modifying multiple projects and should be used in "
+            "conjunction with the ':skip' flag),\nand <PR_SUBJECT> starting with "
             "a capitalized verb in the imperative mood and without any punctuation at the end.\n\n"
-            "A valid example is:\n\n\t`feat(framework) Add flwr build CLI command`\n"
+            "A valid example is:\n\n\t`feat(framework) Add flwr build CLI command`\n\n"
+            "Or, if the PR shouldn't appear in the changelog:\n\n\t"
+            "`feat(framework:skip) Add new option to build CLI`\n"
```
Robert-Steiner, line 71: Maybe I'm missing something, but if the title is valid, we should exit with 0, shouldn't we?

Robert-Steiner (commented):
Minor question rest looks good to me!

dev/check_pr_title.py
charlesbvll: @Robert-Steiner This is what implicitly happens yes (when reaching the end of the file), but if you think it is clearer, I can make the call explicit at the end

Robert-Steiner: I understand. In the diff view it looked to me like `sys.exit(1)` is outside the if block, but when I look at the full view it's clear. I think we can leave it as it is.

Robert-Steiner (approved):

tanertopal (approved):
lgtm!

PR #3480, opened 2024-05-21
ci(framework:skip) Add caching to speed up E2E CI tests

chongshenng:
### Description

This PR improves the speed of the CI for E2E tests.

### Related issues/PRs

This PR is inspired by #1975 and [this blogpost](https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d) by Evan Pete Walsh. The CI should be triggered couple of times after that has happened to compare the impact of the caching. 

## Explanation

The CI cache works as follows:

### `main` branch

Whenever a new commit or PR is made to the main branch the `Cache Python location` job in `.github/workflows/e2e.yml` will run and cache the following directory:

* `${{ env.pythonLocation }}` (Note that caching only `~/.cache/pip` will cache the wheels, but not the installation itself)

The cache will be overwritten if there is a cache miss. It is retrieved by jobs in pull requests by the following key:

`pip-${{ runner.os }}-${{ matrix.directory }}-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml') }}`

We ensure that the job retrieving the cache must meet the following requirements:
1. Same OS
2. Same matrix directory (important because matrix is used in the E2E test)
3. Same Python location (which looks something like `/opt/hostedtoolcache/Python/3.8.18/x64`)
4. No changes to `pyproject.toml` in the project folders.

One significant update in this PR is that all `pyproject.toml`s in the E2E folder have been migrated to pure `pip` installation with `hatchling` backend without relying on `poetry` and `setuptools`.

In this PR, we only rely on `actions@cache` for caching the Python location instead of `actions@setup-python` because the latter does not yet work with matrix configurations due to cache key conflicts. This is explained by the `setup-python` maintainers in [this thread](https://github.com/actions/setup-python/issues/826#issuecomment-2021999109). 

### Pull Requests

Pull requests will try to use the cache by doing a lookup using the previously described key and pull the cache is available. 

### Preliminary results

E2E timings can reduce down to ~4 min 50+ secs by caching the Python location, compared to ~5 min 20+ secs if we cache only `~/.cache/pip`, and ~5 min 40+ secs without caching (see [this workflow run](https://github.com/adap/flower/actions/runs/9191558539)). 

### Checklist

- ~~[ ] Write tests~~
- [X] Make CI checks pass

### Changelog entry

<skip>

### Any other comments?

N/A

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -131,7 +131,7 @@ jobs:
           path: |
             ~/.config/pypoetry
             ~/.local
-          key: poetry-local-${{ runner.os }}-3.8-${{ hashFiles('**/poetry.lock') }}-${{ hashFiles('.github/workflows/*.yml') }}
+          key: poetry-local-${{ runner.os }}-3.8-${{ matrix.directory }}-${{ hashFiles('.github/workflows/*.yml') }}
           restore-keys: |
             poetry-local-${{ runner.os }}-3.8-
       - name: Install dependencies
@@ -147,7 +147,7 @@ jobs:
           path: |
             ~/.fastai
             ./data
-          key: datasets-${{ runner.os }}
+          key: datasets-${{ runner.os }}-${{ matrix.directory }}-${{ hashFiles('.github/workflows/*.yml') }}
           restore-keys: |
             datasets-
       - name: Download dataset
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -122,7 +122,6 @@ jobs:
         uses: ./.github/actions/bootstrap
         with:
           python-version: 3.8
-      # virtualenv cache should depends on OS, Python version and `poetry.lock` (and optionally workflow files).
       - name: Cache Packages
         id: cache-packages
         uses: actions/cache@v4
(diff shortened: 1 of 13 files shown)


.github/workflows/e2e.yml
```diff
@@ -118,12 +118,26 @@ jobs:
...
+        id: cache-python
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.pythonLocation }}
+          key: pythonloc-${{ runner.os }}-${{ matrix.directory }}-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml') }}
+          restore-keys: |
+            pythonloc-${{ runner.os }}-${{ matrix.directory }}-${{ env.pythonLocation }}
       - name: Install dependencies
-        run: python -m poetry install
+        run: python -m pip install --upgrade .
```
Moep90, line 140: Is there a reason why we do an upgrade while having [exact versions](https://github.com/adap/flower/pull/3480/files#diff-676c487c6f2c63c5d04f661140ceecdc515a4f04e074ec6415cb26b52439b1e1R12) ?

chongshenng: Yes, in some `pyproject.toml`, we provide an upper limit to the versions allowed, such as `"torchvision>=0.14.1,<0.15.0"`. When the cache goes stale, `--upgrade` will install the latest allowed versions. At least this is my understanding from [this blogpost](https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d).

tanertopal (approved):
LGTM

PR #3501, opened 2024-05-23
fix(*:skip) Update `ray` and unpin `setuptools`

jafermarq:
Updates ray to `2.10.0` and reverts pinning from #3494


tanertopal (approved):
lgtm!

PR #3525, opened 2024-05-28
docs(framework): Add how-to guide for using `SecAgg`

chongshenng:
### Checklist

- ~~[ ] Write tests~~
- [ ] Make CI checks pass

### DIFF
--- a/doc/source/how-to-use-secagg.rst
+++ b/doc/source/how-to-use-secagg.rst
@@ -1,0 +1,150 @@
+Use Secure Aggregation
+======================
+
+What is Secure Aggregation?
+---------------------------
+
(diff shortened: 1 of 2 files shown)


doc/source/how-to-use-secagg.rst
```diff
@@ -0,0 +1,150 @@
+Use Secure Aggregation
+======================
+
+What is Secure Aggregation?
+---------------------------
+
+Secure Aggregation is a class of Secure Multi-Party Computation algorithm that provides a way for an aggregated value (such as the sum of data) from multiple parties to be computed without revealing their individual inputs to one another.
```
panh99, line 7: Should we use different terminology instead of "the sum of data"? For example, "the sum of parameters." This might help prevent users from thinking they are sending out data.

chongshenng: Good point, I've updated it.

PR #3528, opened 2024-05-29
feat(framework) Enable saving of `SuperLink` logs and introduce `verbose` flag

jafermarq:
Pass `--log-file` to save the `SuperLink` logs to disk. Use `--verbose` to include `DEBUG`-level logs.


src/py/flwr/server/app.py
```diff
@@ -727,6 +732,13 @@ def _add_args_common(parser: argparse.ArgumentParser) -> None:
         type=str,
         help="The SuperLink's public key (as a path str) to enable authentication.",
     )
+    parser.add_argument(
+        "--log-file",
+        type=str,
+        default="",
+        help="Log file where SuperLink's logs are written. "
```
charlesbvll, line 739:
```suggestion
        help="Log file where SuperLink's debug logs are written. "
```

Should we add the mention that `DEBUG` logs will be saved?

charlesbvll (approved):

PR #3534, opened 2024-05-30
refactor(framework) Add `adapter` argument to `grpc_request_response`

panh99:


src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -73,6 +73,7 @@ def grpc_request_response(  # pylint: disable=R0913, R0914, R0915
     authentication_keys: Optional[
         Tuple[ec.EllipticCurvePrivateKey, ec.EllipticCurvePublicKey]
     ] = None,
+    adapter: Optional[Type[FleetStub]] = None,
```
danieljanes, line 76: Should we name this something like `adappter_cls` to indicate that it must refer to a class that can be instantiated?
```suggestion
    adapter_cls: Optional[Type[FleetStub]] = None,
```

danieljanes (approved):

PR #3557, opened 2024-06-07
feat(*:skip) Add general NLP task to FlowerTune LLM leaderboard

yan-gao-GY:
## Proposal
This PR adds General NLP task as the first challenge to FlowerTune LLM Leaderboard.


yan-gao-GY:
We use `flwr new` to create template now, so no need to add training code to the repo.

PR #3568, opened 2024-06-10
feat(examples) Add Tabular Dataset Example

mohammadnaseri:
This PR introduces a new example about the use of the Adult Census Income Tabular dataset in Flower.


examples/fl-tabular/centralized.py
```diff
@@ -0,0 +1,119 @@
...
+    "capital_gain",
+    "capital_loss",
+    "hours_per_week",
+    "native_country",
+    "income",
+]
+
+data = pd.read_csv(
+    url, header=None, names=column_names, na_values=" ?", sep=",\s", engine="python"
+)
```
jafermarq, line 31: why not use flower datasets?

examples/fl-tabular/pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[project]
+name = "fl-tabular"
+version = "0.1.0"
+description = "Adult Census Income Tabular Dataset and Federated Learning in Flower"
+authors = [
+    { name = "The Flower Authors", email = "hello@flower.ai" },
+]
+dependencies = [
+    "flwr>=1.8.0,<2.0",
```
jafermarq, line 13:
```suggestion
    "flwr[simulation]>=1.9.0,<2.0",
```

examples/fl-tabular/federated.py
```diff
@@ -0,0 +1,186 @@
...
+        return get_weights(self.net), len(self.trainloader), {}
+
+    def evaluate(self, parameters, config):
+        set_weights(self.net, parameters)
+        loss, accuracy = evaluate(self.net, self.testloader)
+        return loss, len(self.testloader), {"accuracy": accuracy}
+
+
+def client_fn(cid: str) -> Client:
+    train_loader, test_loader, input_dim = load_data(partition_id=int(cid))
```
jafermarq, line 166: Preferably we don't re-create the FDS object w/ each call to `client_fn()`. I think it's better to create the FDS object once, then pass it to the `client_fn` so partitions (previously created) can be loaded when calling this `client_fn` function. One way of doing this is how the `examples/simulation_pytorch` does it.

examples/fl-tabular/federated.py
```diff
@@ -0,0 +1,186 @@
...
+import numpy as np
+import random
+
+# Set seed for reproducibility
+SEED = 42
+random.seed(SEED)
+np.random.seed(SEED)
+torch.manual_seed(SEED)
+if torch.cuda.is_available():
+    torch.cuda.manual_seed(SEED)
```
jafermarq, line 25: has the example been tested in GPU? My guess is that the way `initialize_model` has been designed it won't work since it's not sent to device.

jafermarq: maybe we don't need CUDA for this small dataset/example?

examples/fl-tabular/README.md
```diff
@@ -0,0 +1,43 @@
...
+### 1. Centralized
+
+```bash
+python centralized.py
+```
+
+### 2. Federated Using Flower Simulation
+
+```bash
+python federated.py
```
jafermarq, line 42: why not use `flower-simulation`?

jafermarq (commented):
Hey @mohammadnaseri , i left a few comments below. In addition, I see quite a few warnings in the form of:
```
DeprecationWarning: np.find_common_type is deprecated.  Please use `np.result_type` or `np.promote_types`.
```
It would be good to find the line causing it (maybe the `.select_dtypes`?)

examples/fl-tabular/centralized.py
mohammadnaseri: I did that intentionally to demonstrate how Flower datasets can be utilized compared to not using them.

adam-narozniak: Sorry, but it does not make full sense to me. It's a point that can confuse users. Also, while theoretically it shouldn't, it brings the possibility of data discrepancy that is unintentional (since it comes from a different source). + Double data download.

examples/fl-tabular/centralized.py
```diff
@@ -0,0 +1,119 @@
...
+
+data.dropna(inplace=True)
+
+label_encoders = {}
+for column in data.select_dtypes(include=["object"]).columns:
+    label_encoders[column] = LabelEncoder()
```
adam-narozniak, line 37: `LabelEncoder` is not meant to be used on input data (X) but instead on target (y).
`OrdinalEncoder` should be used instead.

It also brings problems of different preprocessing of the dataset in centralized and federated settings. Why would be the reason for that? Also, no info about that in the README.

examples/fl-tabular/federated.py
```diff
@@ -0,0 +1,184 @@
...
+            outputs = model(X_batch)
+            batch_loss = criterion(outputs, y_batch)
+            loss += batch_loss.item()
+            predicted = (outputs > 0.5).float()
+            total += y_batch.size(0)
+            correct += (predicted == y_batch).sum().item()
+    accuracy = correct / total
+    loss = loss / len(test_loader)
+    print(f"Accuracy: {accuracy:.2f}")
+    return loss, accuracy
```
adam-narozniak, line 126: That full code seems to be duplicated

adam-narozniak (changes requested):

examples/fl-tabular/centralized.py
mohammadnaseri: @adam-narozniak @jafermarq in that case, I think it is better to remove the `centralized.py`.  My idea was to show how centralized code (without the FL framework and dataset) can be transitioned to FL code. But it is getting confusing and I guess better to just remove it. wdyt?

jafermarq: how about using HF Datasets to get the dataset. (this is what FDS would use under the hood). This will likely allow to move some of the common functionality (e.g. doing the one-hot stuff, model definition, train/test functions) to a separate Python file. (as you did a while back for the [`examples/fl-dp-sa`](https://github.com/adap/flower/tree/main/examples/fl-dp-sa/fl_dp_sa)) ?

mohammadnaseri: @jafermarq I restructured it and removed `centralized.py`. Please have a look.

examples/fl-tabular/README.md
```diff
@@ -0,0 +1,37 @@
+# Flower Example on Adult Census Income Tabular Dataset
+
+Federated Learning on a Tabular Dataset with Flower Framework
+This code exemplifies a federated learning setup using the Flower framework on tabular dataset, based on the "Adult Census Income" dataset. The "Adult Census Income" dataset contains demographic information such as age, education, occupation, etc., with the target attribute being income level (\<=50K or >50K). The dataset is partitioned into subsets, simulating a federated environment with 5 clients, each holding a distinct portion of the data. Categorical variables are one-hot encoded, and the data is split into training and testing sets. Federated learning is conducted using the FedAvg strategy for 5 rounds.
```
jafermarq, line 4:
```suggestion
This code exemplifies a federated learning setup using the Flower framework on the "Adult Census Income" tabular dataset. The "Adult Census Income" dataset contains demographic information such as age, education, occupation, etc., with the target attribute being income level (\<=50K or >50K). The dataset is partitioned into subsets, simulating a federated environment with 5 clients, each holding a distinct portion of the data. Categorical variables are one-hot encoded, and the data is split into training and testing sets. Federated learning is conducted using the FedAvg strategy for 5 rounds.
```

examples/fl-tabular/README.md
```diff
@@ -0,0 +1,37 @@
...
+git clone --depth=1 https://github.com/adap/flower.git && mv flower/examples/fl-tabular . && rm -rf flower && cd fl-tabular
+```
+
+This will create a new directory called `fl-tabular` containing the following files:
+
+```shell
+-- pyproject.toml
+-- centralized.py
+-- federated.py
+-- README.md
```
jafermarq, line 20: needs update after the restructuring.

examples/fl-tabular/README.md
jafermarq: I removed the first sentence

examples/fl-tabular/task.py
```diff
@@ -0,0 +1,126 @@
...
+    optimizer = optim.Adam(model.parameters(), lr=0.0001)
+    model.train()
+    for epoch in range(num_epochs):
+        for X_batch, y_batch in train_loader:
+            optimizer.zero_grad()
+            outputs = model(X_batch)
+            loss = criterion(outputs, y_batch)
+            loss.backward()
+            optimizer.step()
+        print(f"Epoch [{epoch+1}/{num_epochs}], Loss: {loss.item():.4f}")
```
jafermarq, line 95: are you ok leaving the `print` here ?

examples/fl-tabular/task.py
```diff
@@ -0,0 +1,126 @@
...
+from collections import OrderedDict
+
+NUMBER_OF_CLIENTS = 5
+
+
+def load_data(partition_id: int):
+    fds = FederatedDataset(
+        dataset="scikit-learn/adult-census-income",
+        partitioners={"train": NUMBER_OF_CLIENTS},
+    )
```
jafermarq, line 17: wouldn't it be better to remove this out of this function ? in theory creating the federated dataset is only needed once, leaving `load_data` to simply load a particular client partition.

jafermarq (commented):
Some extra comments. We should also list the example in the top-level `README.md`

examples/fl-tabular/centralized.py
adam-narozniak: Abstracting from this example, maybe the partitioners parameter could also be None (in the future releases of FDS) to make it easier to use the dataset for the centralized ML (then load_split could be used)?

We can't show all the examples from the different starting points for people who decide to change from centralized to federate. But if there's starting from zero we can show to easily create both.

jafermarq (approved):
🔥

adam-narozniak (approved):

PR #3576, opened 2024-06-11
feat(framework) Enable setting `run_id` when starting simulation

jafermarq:
The `SuperExec` tracks the id of a run. We need a mechanism to set the `run_id` when launching the simulation. We do this by means of a flag.


src/py/flwr/simulation/run_simulation.py
```diff
@@ -413,5 +430,10 @@ def _parse_args_run_simulation() -> argparse.ArgumentParser:
         "ClientApp and ServerApp from there."
         " Default: current working directory.",
     )
+    parser.add_argument(
+        "--run-id",
+        type=int,
+        help="Sets the id of the run started by the Simulation Engine.",
```
danieljanes, line 436:
```suggestion
        help="Sets the ID of the run started by the Simulation Engine.",
```

danieljanes (approved):

PR #3577, opened 2024-06-11
feat(framework) Add `flwr log`

chongshenng:
### Description
With the `SuperExec` service, the `ServerApp` logs are not easily accessible. This CLI enables a request to be sent to the service. 
  

### Related issues/PRs
The companion PR below adds log capturing to the SuperExec service. This PR can be merged independently of it.
* #3584 

Blocks:
* #3611 
* #3613

### Explanation
`flwr log` sends a requests to the `SuperExec` service to get the log stream and receives a response from it. The response will be handled in two ways:
1. Run `flwr log <RUN_ID> <DIRECTORY> <FEDERATION_NAME>`. The default is `stream=True` (`--stream`). This streams the logs and periodically refreshes the gRPC connection.
2. Run `flwr log <RUN_ID> <DIRECTORY> <FEDERATION_NAME> --show`. This prints the logs and closes the connection gracefully. Every subsequent execution of this command always prints the logs from the beginning of the run.

### Checklist

- [X] Make CI checks pass


src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,74 @@
...
+from logging import DEBUG, INFO
+
+import grpc
+import typer
+from typing_extensions import Annotated
+
+from flwr.common.grpc import GRPC_MAX_MESSAGE_LENGTH, create_channel
+from flwr.common.logger import log as logger
+
+STREAM_DURATION: int = 60
```
jafermarq, line 27: Is this related to the frequency at which the logs are streamed/pulled (i.e. every this many seconds)? if so maybe we should rename it?

chongshenng: @jafermarq Yes! It is related, though I'd say it's a matter of perspective. 

Duration (or period) being the allowable elapsed time, whereas frequency is the inverse of it (once every `int` elapsed time). I favour `duration`/`period` or something like that vs `frequency` which to me reads the units of `Hz`, which in the example above would say 60 times per second.

Maybe I'm overthinking it, lmk @jafermarq.

jafermarq: I prefer "`period`" over "`duration`". With "`duration`" it gives the impression the logs are streamed for that many seconds and then it stops. Refreshing after 60s might not be frequent enough? Should we set it as an input arg right from the start?

chongshenng: Sure, I'll change it to `period`. From my tests, 60s might be too short - it depends on how fast clients train and send replies to the ServerApp. I'd say, use 60s as the default and users can set a different duration if required. Wdyt @jafermarq?

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,79 @@
...
+    ] = True,
+) -> None:
+    """Get logs from Flower run."""
+
+    def on_channel_state_change(channel_connectivity: str) -> None:
+        """Log channel connectivity."""
+        logger(DEBUG, channel_connectivity)
+
+    # pylint: disable=unused-argument
+    def stream_logs(run_id: int, channel: grpc.Channel, duration: int) -> None:
```
jafermarq, line 52:
```suggestion
    def stream_logs(run_id: int, channel: grpc.Channel, period: int) -> None:
```

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,79 @@
...
+        """Stream logs from the beginning of a run with connection refresh."""
+
+    # pylint: disable=unused-argument
+    def print_logs(run_id: int, channel: grpc.Channel, timeout: int) -> None:
+        """Print logs from the beginning of a run."""
+
+    channel = create_channel(
+        server_address="127.0.0.1:9093",
```
jafermarq, line 60: Maybe we could also leave this as an input arg? Also, I note that `:9093` is the default port for REST api (https://github.com/adap/flower/blob/c21c913bb248325dbbb91c47c1094a6bf9f8343b/src/py/flwr/server/app.py#L69). I wonder if we should use something else for the logs as default? @danielnugraha

chongshenng: Yes, this can be an input arg. I'm conscious that the SuperExec address will also need to be set in `flwr run`. The address should be set in a consistent way later. In the meantime, I've added it to the input arg.

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,111 @@
...
+        Optional[str],
+        typer.Option(case_sensitive=False, help="The address of the SuperExec server"),
+    ] = None,
+    period: Annotated[
+        int,
+        typer.Option(
+            case_sensitive=False,
+            help="Use this to set connection refresh time period (in seconds)",
+        ),
+    ] = 60,
```
tanertopal, line 57: I think it's better not to include this parameter in the public API initially and see if we can do it without it even if we internally use a reasonable default.

chongshenng: Updated! I've set it as a constant for now.

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+    channel.subscribe(on_channel_state_change)
+
+    if follow:
+        try:
+            while True:
+                logger(INFO, "Starting logstream for run_id `%s`", run_id)
+                stream_logs(run_id, channel, CONN_REFRESH_PERIOD)
+                time.sleep(2)
+                logger(INFO, "Reconnecting to logstream")
```
tanertopal, line 181: Why are we reconnecting every 60 seconds? 

This might look weird to the user having the output of `logger(INFO, "Reconnecting to logstream")`

Whats the technical reason for reconnecting every 60 seconds? Apart from that if we have to do it then at least lets not log the reconnecting message on INFO level but rather DEBUG

chongshenng: Set logging level to `DEBUG`

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,194 @@
...
+        typer.Argument(help="Path of the Flower project to run"),
+    ] = Path("."),
+    federation_name: Annotated[
+        Optional[str],
+        typer.Argument(help="Name of the federation to run the app on"),
+    ] = None,
+    stream: Annotated[
+        bool,
+        typer.Option(
+            "--stream/--show",
```
Moep90, line 60:
```suggestion
            "--stream, -s",
```

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,194 @@
...
+            "-s/-p",
```
Moep90, line 61:
```suggestion
            "--print, -p",
```

src/py/flwr/cli/log.py
Moep90: Usually, options with short perspectives are printed similar to:

```
Options supported by ....:
  -c or --capitalize
	Yoda Yoda description
  -A or --no-capitalize
	Yoda Yoda other description
  -n or --numerals
```

src/py/flwr/cli/log.py
Moep90: In the `help` section, it states `print`. We should strive for consistency in naming to avoid confusion.

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,194 @@
...
+                "is set to `True`.",
+                fg=typer.colors.RED,
+                bold=True,
+            )
+            raise typer.Exit(code=1)
+    else:
+        root_certificates_bytes = None
+        if insecure_str is None:
+            typer.secho(
+                "❌ To disable TLS, set `insecure = true` in `pyproject.toml`.",
```
Moep90, line 153: The same statements are written here:
https://github.com/adap/flower/pull/3611/files#r1696499600

https://github.com/adap/flower/pull/3611/files#r1696502558

src/py/flwr/cli/log.py
chongshenng: The abbreviations are the source of the confusion, so I'll remove them.

src/py/flwr/cli/log.py
chongshenng: @Moep90 Wdym by this?

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,193 @@
...
+            + "\n".join(available_feds),
+            fg=typer.colors.RED,
+            bold=True,
+        )
+        raise typer.Exit(code=1)
+
+    if "address" in federation:
+        _log_with_superexec(federation, run_id, stream)
+    else:
+        pass
```
jafermarq, line 123: Should we log something in this case? like a message saying that `flwr logs` only makes sense when the run runs via a superexec? I presume this line would only be hit a user that passes a made up `run_id`?

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,193 @@
...
+        interceptors=None,
+    )
+    channel.subscribe(on_channel_state_change)
+
+    if stream:
+        try:
+            while True:
+                logger(INFO, "Starting logstream for run_id `%s`", run_id)
+                stream_logs(run_id, channel, CONN_REFRESH_PERIOD)
+                time.sleep(2)
```
jafermarq, line 179: Would it make sense to allow setting this sleep value as an optional arg to `flwr log` when `--stream` is used?

src/py/flwr/cli/log.py
jafermarq: For example if i go to `examples/quickstart-pytorch` and do:
```bash
flwr log 1234
# the below prints
Loading project configuration... 
Success
```

src/py/flwr/cli/log.py
chongshenng: No. For starters, and as previously suggested by @tanertopal, we want to add sensible constants here for users.

chongshenng:
@jafermarq About your comments [here](https://github.com/adap/flower/pull/3577#discussion_r1717328815), we need to refactor `flwr log` and `flwr run` together soon since they share a lot of similar code. It's substantial work worthy of another PR.

src/py/flwr/cli/log.py
chongshenng: Actually, when all the PRs are merged, any made up `run_id` will throw an error. But it does happen later when a request is sent to the `SuperExec`. I've added a `sys.exit()` and a `typer` error for now.

jafermarq: but what about if `address` is not defined in the `[federation]` block of the `pyproject.toml`? there won't be `SuperExec` to interface with in that case

chongshenng: Yes, the `sys.exit()` happens in the `else` block. Specifically, I've update it to:
```python
    if "address" in federation:
        _log_with_superexec(federation, run_id, stream)
    else:
        typer.secho(
            "❌ `flwr log` currently works with `SuperExec`. Ensure that the correct"
            "`SuperExec` address is provided in the `pyproject.toml`.",
            fg=typer.colors.RED,
            bold=True,
        )
        sys.exit()
```

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,198 @@
...
+# pylint: disable=unused-argument
+def print_logs(run_id: int, channel: grpc.Channel, timeout: int) -> None:
+    """Print logs from the beginning of a run."""
+
+
+def log(
+    run_id: Annotated[
+        int,
+        typer.Argument(help="The Flower run ID to query"),
+    ],
```
danieljanes, line 47: @chongshenng are we sure about this argument order? in `flwr run`, `app` and `federation` are the first two arguments

chongshenng: Good point, I've reorganised it.

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,197 @@
...
+    if "address" in federation_config:
+        _log_with_superexec(federation_config, run_id, stream)
+    else:
+        typer.secho(
+            "❌ `flwr log` currently works with `SuperExec`. Ensure that the correct"
+            "`SuperExec` address is provided in the `pyproject.toml`.",
+            fg=typer.colors.RED,
+            bold=True,
+        )
+        sys.exit()
```
danieljanes, line 131: Let's make `_log_with_superexec` the last line in this function and have an `if "address" not in ...` to fail if a precondition is not met

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+# pylint: disable-next=too-many-branches
+def _log_with_superexec(
+    federation_config: dict[str, str],
+    run_id: int,
+    stream: bool,
+) -> None:
+
+    insecure_str = federation_config.get("insecure")
```
danieljanes, line 141:
```suggestion
) -> None:
    insecure_str = federation_config.get("insecure")
```

danieljanes (approved):

PR #3580, opened 2024-06-11
feat(framework) Implement DriverAPI `GetRun`

panh99:
Depends on #3578

- Added `run` abstract property to `Driver` class.
- Implemented the property in `GrpcDriver` and `InMemoryDriver`.
- Made the `run_id` argument mandatory for `GrpcDriver` and `InMemoryDriver`.
- Moved the run creation process out from the drivers.


danieljanes (dismissed):

src/py/flwr/server/superlink/driver/driver_servicer.py
```diff
@@ -134,7 +138,15 @@ def GetRun(
...
-        raise NotImplementedError
+        log(DEBUG, "DriverServicer.GetRun")
+
+        # Init state
+        state: State = self.state_factory.state()
+
+        # Retrieve run information
+        run = state.get_run(request.run_id)
+        run_proto = None if run is None else Run(**vars(run))
+        return GetRunResponse(run=run_proto)
```
danieljanes, line 149: Should we set the gRPC status code depending on the outcome here?

panh99: We can't set the gRPC status code here. The status code must be set in the `context` in the servicer, e.g.:

``` python
class ExampleServiceServicer(example_pb2_grpc.ExampleServiceServicer):
    def ExampleMethod(self, request, context):
        if request.name == "":
            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
            context.set_details('Name field cannot be empty')
            return example_pb2.ExampleResponse()

        response = example_pb2.ExampleResponse(message=f'Hello, {request.name}!')
        return response
```

panh99: Alternatively, I set `run` to `None` if the `run_id` is invalid, which can be detected on the other end by `GetRunResponse.HasField`.

src/py/flwr/server/run_serverapp.py
```diff
@@ -149,13 +151,15 @@ def run_server_app() -> None:
...
-        driver_service_address=args.superlink,
-        root_certificates=root_certificates,
-        fab_id=args.fab_id,
-        fab_version=args.fab_version,
+    # Create run
+    stub = GrpcDriverStub(
+        driver_service_address=args.superlink, root_certificates=root_certificates
     )
+    req = CreateRunRequest(fab_id=args.fab_id, fab_version=args.fab_version)
+    res = stub.create_run(req)
```
danieljanes, line 159: We should have some error handling here and also check the gRPC status code

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -157,39 +159,50 @@ class GrpcDriver(Driver):
...
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
-        driver_service_address: str = DEFAULT_SERVER_ADDRESS_DRIVER,
-        root_certificates: Optional[bytes] = None,
-        fab_id: Optional[str] = None,
-        fab_version: Optional[str] = None,
+        run_id: int,
+        stub: Optional[GrpcDriverStub] = None,
```
danieljanes, line 165: Looks good! The docstring above needs to be updated to reflect this

src/py/flwr/server/run_serverapp.py
panh99: But for now `create_run` won't raise any exception. Only `get_run` can raise an error when `run_id` can't be found.

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -157,39 +159,50 @@ class GrpcDriver(Driver):
...
-        self.addr = driver_service_address
-        self.root_certificates = root_certificates
-        self.driver_helper: Optional[GrpcDriverHelper] = None
-        self.run_id: Optional[int] = None
-        self.fab_id = fab_id if fab_id is not None else ""
-        self.fab_version = fab_version if fab_version is not None else ""
+        self.stub = stub
+        self._run_id = run_id
+        self._fab_id = ""
+        self._fab_ver = ""
```
danieljanes, line 170: It'd be better to change the type to `Optional[str]` here instead of using a "magic" empty value

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -157,39 +159,50 @@ class GrpcDriver(Driver):
...
-            req = CreateRunRequest(fab_id=self.fab_id, fab_version=self.fab_version)
-            res = self.driver_helper.create_run(req)
-            self.run_id = res.run_id
-        return self.driver_helper, self.run_id
+            if self.stub is None:
+                self.stub = GrpcDriverStub()
+
+            # Get the run info
+            req = GetRunRequest(run_id=self._run_id)
+            res = self.stub.get_run(req)
```
danieljanes, line 193: We should add more error handling / status code checking here

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -157,39 +159,50 @@ class GrpcDriver(Driver):
...
+            if not res.HasField("run"):
+                raise RuntimeError(f"Cannot find the run with ID: {self._run_id}")
+            self._fab_id = res.run.fab_id
+            self._fab_ver = res.run.fab_version
+            self._has_initialized = True
```
danieljanes, line 198: We could replace these three lines with just `self.run` of type `Optional[Run]`

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -46,92 +46,94 @@
...
-    """`GrpcDriverHelper` provides access to the gRPC Driver API/service."""
+class GrpcDriverStub(DriverStub):
+    """`GrpcDriverStub` provides access to the gRPC Driver API/service."""
 
     def __init__(
         self,
         driver_service_address: str = DEFAULT_SERVER_ADDRESS_DRIVER,
         root_certificates: Optional[bytes] = None,
     ) -> None:
+        event(EventType.DRIVER_CONNECT)
```
danieljanes, line 62: Constructors should not have side effects (for example, making network requests), so it's better to move this into a separate call that happens when the object is being used. This was one of the motivations behind the previous `connect` function.

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -46,92 +46,94 @@
...
-        """Connect to the Driver API."""
-        event(EventType.DRIVER_CONNECT)
-        if self.channel is not None or self.stub is not None:
-            log(WARNING, "Already connected")
-            return
         self.channel = create_channel(
             server_address=self.driver_service_address,
             insecure=(self.root_certificates is None),
             root_certificates=self.root_certificates,
         )
```
danieljanes, line 69: Similar to `event` above, this should not happen in the constructor

src/py/flwr/server/driver/grpc_driver.py
panh99: Thanks! I've updated the doc string

src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -43,37 +44,47 @@ class InMemoryDriver(Driver):
...
+
+    @property
+    def run(self) -> Run:
+        """Run ID."""
+        self._init_run()
+        return Run(
+            run_id=self._run_id,
+            fab_id=self._fab_id,
+            fab_version=self._fab_ver,
+        )
```
jafermarq, line 85: a higher level question: why not creating the `Run` obj in `__init__()`? will help simplify this logic a bit imo

jafermarq: I just find having to go through `_init_run()` (even if returning immediately) each time this class access `self.run` a bit unnecessary. Or is this done in this way to support some upcoming functionality?

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -46,92 +46,94 @@
...
         # Check if channel is open
-        if self.stub is None:
+        if self.channel is None:
             log(ERROR, ERROR_MESSAGE_DRIVER_NOT_CONNECTED)
-            raise ConnectionError("`GrpcDriverHelper` instance not connected")
+            raise ConnectionError("`GrpcDriverStub` instance not connected")
 
         # Call Driver API
-        res: CreateRunResponse = self.stub.CreateRun(request=req)
+        res: CreateRunResponse = self.CreateRun(request=req)
```
danieljanes, line 92: The subclassing leads creates a `GrpcDriverStub` that exposes both `create_run` and `CreateRun` as methods. Would it perhaps be better to keep the internal `stub` object to have only `create_run`? One alternative - perhaps for a different PR - would be to get rid of `GrpcDriverStub`/`GrpcDriverHelper` altogether and just give `GrpcDriver` and internal `stub` object (without an additional layer of indirection)

src/py/flwr/server/driver/grpc_driver.py
danieljanes: `driver_service_address` and `certificates` need to be removed as well

src/py/flwr/simulation/run_simulation.py
```diff
@@ -201,8 +201,11 @@ def _main_loop(
...
     try:
+        # Create run (with empty fab_id and fab_version)
+        run_id = state_factory.state().create_run("", "")
+
         # Initialize Driver
-        driver = InMemoryDriver(state_factory)
+        driver = InMemoryDriver(run_id=run_id, state_factory=state_factory)
 
         if run_id:
             _init_run_id(driver, state_factory, run_id)
```
jafermarq, line 211: The idea behind having this logic was so the `Simulation Engine` uses a `run_id` passed from an external component (e.g. the `SuperExec`). With the changes done a few lines above, this behaviour is changed (since `run_id` passed will be overridden by what `create_run()` returns.

jafermarq: Would this be enough to retain the functionality?
```
    try:
        # Create run (with empty fab_id and fab_version)
        run_id_ = state_factory.state().create_run("", "")

        # Initialize Driver
        driver = InMemoryDriver(state_factory)
        driver = InMemoryDriver(run_id=run_id_, state_factory=state_factory)
```

src/py/flwr/server/driver/grpc_driver.py
panh99: Cool! Thanks for the suggestion.

src/py/flwr/server/driver/grpc_driver.py
panh99: Ok, sure. I've reverted the changes.

src/py/flwr/server/driver/inmemory_driver.py
panh99: I think it's a rather common property implementation (though maybe not in Python?). We also had the same workflow before this PR, e.g., `run_id = self._get_run_id()`. Regarding creating a `Run` object in the `run` property each time, it's to prevent the `Run` object from being modified externally.

As for creating the `Run` object in `__init__()`, I think we should postpone the initialization until necessary, following the same workflow we had previously.

jafermarq: yeah, we have the same logic before this PR. I thought it would make it unnecessarily complex. But i see the value of _"prevent the Run object from being modified externally."_. I'll resolve this comment.

src/py/flwr/server/run_serverapp.py
danieljanes: Any call that involves networking can raise an exception if the network is down

src/py/flwr/server/superlink/driver/driver_servicer.py
danieljanes: That's better than the previous version.

I'm afraid I don't fully understand the first comment though. This is a servicer method and we have access to the `context`. So we can set the status code using `context` if the `run_id` wasn't valid just before returning from the `GetRunResponse`, similar to how we do it in `PullTaskRes` above.

src/py/flwr/simulation/run_simulation.py
```diff
@@ -201,11 +194,11 @@ def _main_loop(
     f_stop = asyncio.Event()
     serverapp_th = None
     try:
-        # Initialize Driver
-        driver = InMemoryDriver(state_factory)
+        # Create run (with empty fab_id and fab_version)
+        run_id_ = state_factory.state().create_run("", "")
```
danieljanes, line 198: Should this be conditional on whether a `run_id` was passed to `_main_loop`?

src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -33,47 +34,47 @@ class InMemoryDriver(Driver):
...
+        run = self.state.get_run(self._run_id)
+        if run is None:
+            raise RuntimeError(f"Cannot find the run with ID: {self._run_id}")
+        self._run = run
+
+    @property
+    def run(self) -> Run:
+        """Run ID."""
+        self._init_run()
+        return Run(**vars(cast(Run, self._run)))
```
danieljanes, line 77: Why do we create a new object here?

src/py/flwr/server/run_serverapp.py
panh99: Oh, I see what you mean. Yes, we should have error handling and retry mechanism here and in other driver methods. But I would like to open a separate PR to do this, i.e., making `GrpcDriver` reliable.

src/py/flwr/simulation/run_simulation.py
panh99: Do we still need the `run_id` arg after making the change you proposed?

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -46,13 +48,28 @@
...
-    """`GrpcDriverHelper` provides access to the gRPC Driver API/service."""
+class GrpcDriverStub:
+    """`GrpcDriverStub` provides access to the gRPC Driver API/service.
+
+    Parameters
+    ----------
+    driver_service_address : Optional[str]
+        The IPv4 or IPv6 address of the Driver API server.
+        Defaults to `"[::]:9091"`.
+    certificates : bytes (default: None)
```
danieljanes, line 64: Both name and type are not in line with the name/type in `__init__`. The description is not in line with either of them, it suggests that the type would have to be `Optional[Tuple[bytes, bytes, bytes]]`.

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -308,8 +327,8 @@ def send_and_receive(
 
     def close(self) -> None:
         """Disconnect from the SuperLink if connected."""
-        # Check if GrpcDriverHelper is initialized
-        if self.driver_helper is None:
+        # Check if is initialized
+        if self._run is None:
```
danieljanes, line 331: Shouldn't this be removed? In `__init__`, we ensure that `self.stub` is never `None`, so we don't need to check for it

src/py/flwr/server/driver/grpc_driver.py
panh99: Oh, yes. Thanks for pointing it out. I just moved the docstring from the `GrpcDriver`. Didn't notice that it was wrong originally.

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -308,8 +323,8 @@ def send_and_receive(
 
     def close(self) -> None:
         """Disconnect from the SuperLink if connected."""
-        # Check if GrpcDriverHelper is initialized
-        if self.driver_helper is None:
+        # Check if is initialized
```
danieljanes, line 326:
```suggestion
        # Check if `connect` was called before
```

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -140,56 +168,52 @@ class GrpcDriver(Driver):
...
-            self.run_id = res.run_id
-        return self.driver_helper, self.run_id
+
+        return self.stub, self._run.run_id
 
     def _check_message(self, message: Message) -> None:
         # Check if the message is valid
         if not (
-            message.metadata.run_id == self.run_id
+            message.metadata.run_id == cast(Run, self._run).run_id
```
danieljanes, line 216: Why is `cast` necessary here but not in `InMemoryDriver`?

src/py/flwr/server/driver/inmemory_driver.py
```diff
@@ -33,47 +34,47 @@ class InMemoryDriver(Driver):
...
+        self._run_id = run_id
+        self._run: Optional[Run] = None
         self.state = state_factory.state()
+        self.node = Node(node_id=0, anonymous=True)
 
     def _check_message(self, message: Message) -> None:
         # Check if the message is valid
         if not (
-            message.metadata.run_id == self.run_id
+            message.metadata.run_id == self.run.run_id
```
danieljanes, line 56: Should this be:
```suggestion
            message.metadata.run_id == self._run.run_id
```

src/py/flwr/server/driver/grpc_driver.py
panh99: `InMemoryDriver` uses the property `self.run : Run` instead of the variable `self._run: Run | None`, and thus it doesn't need `cast`.

danieljanes (approved):

PR #3584, opened 2024-06-11
feat(framework) Add SuperExec logcatcher

chongshenng:
### Description
Logs from the subprocess that launches the ServerApp needs caught in the SuperExec service.

### Related issues/PRs

The companion PR is the following.
* #3577

Blocks:
* #3611 
* #3613

### Explanation

The logging mechanism is split into two parts:
* Capture `stdout` and `stderr` to memory
* Stream logs from memory as the response to the client request

### In-memory logs
Logstreams can be requested at any stage of a run via `flwr log`. Each call to `StartRun` starts log capturing for a `run_id`. Logs for a `run_id` are stored in `self.runs[run_id].logs`.

### Checklist

- [x] Make CI checks pass


src/py/flwr/superexec/exec_servicer.py
```diff
@@ -38,13 +41,14 @@ class ExecServicer(exec_pb2_grpc.ExecServicer):
     def __init__(self, executor: Executor) -> None:
         self.executor = executor
         self.runs: Dict[int, RunTracker] = {}
+        self.lock = threading.Lock()
```
tanertopal, line 44: Can you elaborate on why we need a global lock (global in the context of all runs)?

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -53,13 +57,66 @@ def StartRun(
...
+        # Start a background thread to capture the log output
+        capture_thread = threading.Thread(
+            target=self._capture_logs, args=(run,), daemon=True
+        )
+        capture_thread.start()
+
         return StartRunResponse(run_id=run.run_id)
 
-    def StreamLogs(
+    def _capture_logs(
```
tanertopal, line 68: Can we make this a standalone function instead of a class method? I think this would improve testability.

tanertopal: On that front - can we include the tests in this PR?

src/py/flwr/superexec/exec_servicer.py
chongshenng: Good question. The idea behind this is so that only the `capture_thread` that acquires `self.lock` can at any one time append to the `.logs` member of the `.runs[run.run_id]` dictionary. The reason was to ensure thread-safe updates to the dictionary (which I wasn't sure about). 

But, I dug deeper and found from [this SO discussion](https://stackoverflow.com/a/54681070) that updating Python dictionaries are thread safe if the keys are builtin hashable types, which in this case it is since we use signed int64 for `run.run_id`. So, I can also remove `self.lock` and the `with self.lock` lines - the log dictionary will be updated by their respective `capture_thread`.

src/py/flwr/superexec/exec_servicer.py
chongshenng: Sure! This makes sense especially if we remove `self.lock` and set `self.select_timeout` as a constant. I'll update this.

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -56,13 +61,65 @@ def StartRun(
...
+            # Exit if `run_id` not found
+            if request.run_id not in self.runs:
+                context.abort(grpc.StatusCode.NOT_FOUND, "Run ID not found")
+
+            # Yield n'th row of logs, if n'th row < len(logs)
+            logs = self.runs[request.run_id].logs
+            if last_sent_index < len(logs):
+                for i in range(last_sent_index, len(logs)):
+                    yield StreamLogsResponse(log_output=logs[i])
+                last_sent_index = len(logs)
```
tanertopal, line 89:
```suggestion
              for i in range(last_sent_index, len(logs)):
                  yield StreamLogsResponse(log_output=logs[i])
              last_sent_index = len(logs)
```

tanertopal: I'm not sure, but is the if-not redundant?

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -56,13 +61,65 @@ def StartRun(
...
+
+            # Shutdown context if process has completed. Previously stored
+            # logs will still be printed.
+            if self.runs[request.run_id].proc.poll() is not None:
+                log(INFO, "Run ID `%s` completed", request.run_id)
+                context.cancel()
+
+            time.sleep(0.1)  # Sleep briefly to avoid busy waiting
```
tanertopal, line 97:
```suggestion
            time.sleep(1.0)  # Sleep briefly to avoid busy waiting
```

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -56,13 +61,65 @@ def StartRun(
...
+
+
+def _capture_logs(
+    run: RunTracker,
+) -> None:
+    while not run.stop_event.is_set():
```
tanertopal, line 103: Can we check if we need the `stop_event` here? I think we could do without if we use a `while True` as we later use `break` anyway.

chongshenng: Checked and yes, it works using just `while True`. Much easier!

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -59,13 +64,66 @@ def StartRun(
...
+            logs = self.runs[request.run_id].logs
+            for i in range(last_sent_index, len(logs)):
                 yield StreamLogsResponse(log_output=logs[i])
+            last_sent_index = len(logs)
+
+            # Wait for and continue to yield more log responses only if the
+            # run isn't completed yet. If the run is finished, the entire log
+            # is returned at this point and the server ends the stream.
+            if self.runs[request.run_id].proc.poll() is not None:
+                log(INFO, "Run ID `%s` completed", request.run_id)
```
danieljanes, line 97:
```suggestion
                log(INFO, "All logs for run ID `%s` returned", request.run_id)
```

danieljanes (approved):

PR #3585, opened 2024-06-12
docs(datasets) Update Flower Datasets docs

adam-narozniak:
## Proposal
Update FederatedDataset docs 
* outdated partitioning information,
* outdated references to resplitting,
* fix newlines (bc of too long text width)
* add new docstrings example

README:
* add label distribution plot
* add missing partitioners to the list

INDEX:
* add label distribution plot
* add visualization to the list of key features

Others:
* edits in the list of key features the info about partitioning schemes (previously, only the customization was mentioned).

Note: how-to-guide on visualization is added in a separate PR: https://github.com/adap/flower/pull/3672


datasets/doc/source/index.rst
```diff
@@ -67,6 +74,19 @@ Thanks to using Hugging Face's ``datasets`` used under the hood, Flower Datasets
 - Jax
 - Arrow
 
+Here are a few of the ``Partitioner`` s that are available: (for a full list see `link <https://flower.ai/docs/datasets/ref-api/flwr_datasets.partitioner.html#module-flwr_datasets.partitioner>`_ )
```
jafermarq, line 77:
```suggestion
Here are a few of the ``Partitioner`` s that are available: (for a full list see `link <ref-api/flwr_datasets.partitioner.html#module-flwr_datasets.partitioner>`_ )
```

jafermarq: In the framework docs we use most of the times (or always?) relative links like the one I suggest. I believe it's better for caching etc.

datasets/doc/source/index.rst
```diff
@@ -67,6 +74,19 @@ Thanks to using Hugging Face's ``datasets`` used under the hood, Flower Datasets
...
+* Dirichlet partitioning ``DirichletPartitioner(num_partitions, partition_by, alpha)``
+* InnerDirichlet partitioning ``InnerDirichletPartitioner(partition_sizes, partition_by, alpha)``
+* Natural ID partitioner ``NaturalIdPartitioner(partition_by)``
+* Size partitioner (the abstract base class for the partitioners dictating the division based the number of samples) ``SizePartitioner``
+* Linear partitioner ``LinearPartitioner(num_partitions)``
+* Square partitioner ``SquarePartitioner(num_partitions)``
+* Exponential partitioner ``ExponentialPartitioner(num_partitions)``
+* more to come in the future releases (contributions are welcome).
+
 Install
```
jafermarq, line 90: I wonder if the `Install` block should go higher up in the page.. or maybe it should be removed since there is an installation `How-To` guide?

datasets/README.md
```diff
@@ -22,14 +25,23 @@ Thanks to using Hugging Face's `datasets` used under the hood, Flower Datasets i
 * Arrow.
 
 Create **custom partitioning schemes** or choose from the **implemented partitioning schemes**:
```
jafermarq, line 27:
```suggestion
Create **custom partitioning schemes** or choose from the **implemented [partitioning schemes](https://flower.ai/docs/datasets/ref-api/flwr_datasets.partitioner.html#module-flwr_datasets.partitioner)**:
```

datasets/README.md
```diff
@@ -22,14 +25,23 @@ Thanks to using Hugging Face's `datasets` used under the hood, Flower Datasets i
...
+* Linear partitioner `LinearPartitioner(num_partitions)`
+* Square partitioner `SquarePartitioner(num_partitions)`
+* Exponential partitioner `ExponentialPartitioner(num_partitions)`
+* more to come in the future releases (contributions are welcome).
+<p align="center">
+  <img src="./doc/source/_static/readme/comparison_of_partitioning_schemes.png" alt="Comparison of partitioning schemes."/>
+  <br>
+  <em>Comparison of Partitioning Schemes on CIFAR10</em>
+</p>
+PS: This plot was generated using a library function (see flwr_datasets.visualization package for more).
```
jafermarq, line 44:
```suggestion
PS: This plot was generated using a library function (see [flwr_datasets.visualization](https://flower.ai/docs/datasets/ref-api/flwr_datasets.visualization.html) package for more).
```

datasets/README.md
```diff
@@ -22,14 +25,23 @@ Thanks to using Hugging Face's `datasets` used under the hood, Flower Datasets i
...
 
 # Installation
```
jafermarq, line 46: I wonder if we should move this to the top of the `README.md`

datasets/README.md
```diff
@@ -67,11 +79,11 @@ Here's a basic quickstart example of how to partition the MNIST dataset:
 from flwr_datasets import FederatedDataset
 
 # The train split of the MNIST dataset will be partitioned into 100 partitions
-mnist_fds = FederatedDataset("mnist", partitioners={"train": 100}
+fds = FederatedDataset("mnist", partitioners={"train": 100}
```
jafermarq, line 82:
```suggestion
fds = FederatedDataset("mnist", partitioners={"train": 100})
```

datasets/README.md
jafermarq: or maybe we want to update this later after the visualization tutorial page is up?

datasets/doc/source/index.rst
adam-narozniak: Yes, thanks. That makes more sense.

datasets/README.md
adam-narozniak: I thought about it now, and it makes sense to leave the installation below. Here's my logic: typically, you install the library once (then you can update it), so the entry point for most people is to use find the tutorials, etc. as long as the installation is somewhere low, it should be fine. 

But one can argue that the conversion rate of install can be higher if it's left at the top. But then it's a question who enters the page, if someone who doesn't know what the library does, then the install should be lower, if someone who already knows what the library should do it should be moved up.

So, I'd keep it as it is. But wdyt? if you think that it'd work better to have it at the top of the page then let's move it. 

Or maybe we can add a single line at the top that installs both images and audio, no explanation. Seems like a good compromise? (we'd also leave the full section as it is now). I'd say the same applies to the index.rst

datasets/doc/source/index.rst
```diff
@@ -7,6 +7,15 @@ learning/analytics/evaluation. It is created by the ``Flower Labs`` team that al
 Flower Datasets Framework
 -------------------------
 
+Install
+~~~~~~~
+
+.. code-block:: bash
+
+  python -m pip install flwr-datasets[vision]
```
jafermarq, line 15:
```suggestion
  python -m pip install "flwr-datasets[vision]"
```

jafermarq: this change is needed for `zsh` shells. We should probably also update the installation how-to page.

datasets/doc/source/index.rst
```diff
@@ -7,6 +7,15 @@ learning/analytics/evaluation. It is created by the ``Flower Labs`` team that al
...
+
+Check out all the details on the download in :doc:`how-to-install-flwr-datasets`.
```
jafermarq, line 17:
```suggestion
Check out all the details on how to install Flower Datasets in :doc:`how-to-install-flwr-datasets`.
```

datasets/README.md
jafermarq: yes, single line makes sense to me too. Let's do it after this PR

jafermarq (approved):

PR #3553, opened 2024-06-06
feat(framework:skip) Add get_fab_metadata function

charlesbvll:
### Description

We have no function to easily obtain the `fab_id` and `fab_version` from a FAB file.

### Related issues/PRs

N/A

### Explanation

Introduce a `get_fab_metadata` function that returns a `fab_version`, `fab_id` tuple given a FAB file as a path or as `bytes`.

### Checklist

- [x] Implement proposed change
- [x] Make CI checks pass
- [x] Ping maintainers on [Slack](https://flower.ai/join-slack/) (channel `#contributions`)

### Any other comments?

N/A

PR #3589, opened 2024-06-12
feat(framework) Add simulation engine `SuperExec` plugin

jafermarq:
Minimal Simulation Engine plugin for the `SuperExec`.

Built on top of #3555 , expects #3553 

---------------------------

Lauch (without TLS):
```
flower-superexec --executor flwr.superexec.simulation:executor --insecure --executor-config 'num-supernodes=100'
```

Then start the run from a FAB dir
```
flwr run --use-superexec 
```


src/py/flwr/superexec/simulation_plugin.py
```diff
@@ -0,0 +1,75 @@
...
+                    str(run_id),
+                ],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                text=True,
+            ),
+        )
+
+
+simulation_plugin = SimulationEngine()
```
danieljanes, line 75:
```suggestion
executor = SimulationEngine()
```

src/py/flwr/superexec/simulation_plugin.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import Optional
+
+from flwr.cli.config_utils import get_fab_metadata
+from flwr.cli.install import install_from_fab
+
+from .executor import Executor, RunTracker
+
+
+class SimulationEngine(Executor):
+    """Simulation engine executor plugin."""
```
danieljanes, line 31:
```suggestion
"""Simulation engine executor."""


import os
import subprocess
import sys
from pathlib import Path
from typing import Optional

from flwr.cli.config_utils import get_fab_metadata
from flwr.cli.install import install_from_fab

from .executor import Executor, RunTracker


class SimulationEngine(Executor):
    """Simulation engine executor."""
```

src/py/flwr/__init__.py
```diff
@@ -17,13 +17,14 @@
...
 
-from . import client, common, server, simulation
+from . import client, common, server, simulation, superexec
 
 __all__ = [
     "client",
     "common",
     "server",
     "simulation",
+    "superexec",
```
danieljanes, line 27: Why is this needed?

jafermarq: was added a long ago, now removed.

src/py/flwr/superexec/simulation.py
```diff
@@ -0,0 +1,139 @@
...
+    ) -> None:
+        """Set executor config arguments.
+
+        Parameters
+        ----------
+        config : Dict[str, str]
+            A dictionary for configuration values.
+            Supported configuration key/value pairs:
+            - "num-supernodes": str
+                Number of nodes to register for the Simulation.
```
danieljanes, line 63:
```suggestion
                Number of nodes to register for the simulation.
```

src/py/flwr/superexec/simulation.py
```diff
@@ -0,0 +1,139 @@
...
+        """Start run using the Flower Simulation Engine."""
+        try:
+            if self.num_supernodes is None:
+                log(
+                    ERROR,
+                    "To start a run with the simulation plugin, please specify "
+                    "the number of supernodes. You can do this by using the "
+                    "`--executor-config` argument when launching the SuperExec.",
+                )
+                raise ValueError("Need to set `num-supernodes`.")
```
danieljanes, line 83: This should never happen b/c the SuperExec will guarantee that `set_config` gets called once before it calls `start_run`. Should we do this check in `set_config`?

src/py/flwr/superexec/simulation.py
```diff
@@ -0,0 +1,139 @@
...
+
+            if override_config:
+                raise ValueError(
+                    "Overriding the run config is not yet supported with the "
+                    "simulation plugin.",
+                )
```
danieljanes, line 89:
```suggestion
                raise ValueError(
                    "Overriding the run config is not yet supported with the "
                    "simulation executor.",
                )
```

src/py/flwr/superexec/simulation.py
```diff
@@ -0,0 +1,139 @@
...
+            # Install FAB to flwr dir
+            _, fab_id = get_fab_metadata(fab_file)
+            fab_path = install_from_fab(fab_file, None, True)
+
+            # Install FAB Python package
+            subprocess.check_call(
+                [sys.executable, "-m", "pip", "install", str(fab_path)],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
```
danieljanes, line 100: Can we unify this with the implementation in `DeploymentExecutor`? For example, in `DeploymentExecutor`, we now added `--no-deps`.

src/py/flwr/superexec/simulation.py
```diff
@@ -0,0 +1,140 @@
...
+            log(INFO, "Created run %s", str(run_id))
+
+            fab_name = Path(fab_id).name
+
+            # Start Simulation
+            proc = subprocess.Popen(  # pylint: disable=consider-using-with
+                [
+                    "flower-simulation",
+                    "--client-app",
+                    f"{fab_name}.client:app",
```
danieljanes, line 114: This needs to be read from the FAB itself (key: `flower.components.clientapp`). Users might rename `client` or `app`, or move the module in a sub-package.

src/py/flwr/superexec/simulation.py
```diff
@@ -0,0 +1,140 @@
...
+                    "--server-app",
+                    f"{fab_name}.server:app",
```
danieljanes, line 116: This needs to be read from the FAB as well.

danieljanes (approved):

PR #3610, opened 2024-06-14
refactor(framework:skip) Use absolute paths when setting `sys.path` and replace `os.path.abspath` with `pathlib.Path.absolute`

panh99:


charlesbvll (commented):
@panh99 I would suggest using `Path.absolute` from `pathlib` as it is the way forward of dealing with paths in Python (instead of `os.path`)

danieljanes (approved):

PR #3611, opened 2024-06-14
feat(framework) Add log functions to `flwr log`

chongshenng:
This PR is blocked by
* #3577
* #3584 

This PR blocks:
* #3613


src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,221 @@
...
+            # pylint: disable=E1101
+            if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
+                break
+            if e.code() == grpc.StatusCode.NOT_FOUND:
+                logger(ERROR, "Invalid run_id `%s`, exiting", run_id)
+                break
+            if e.code() == grpc.StatusCode.CANCELLED:
+                break
+
+    channel.close()
```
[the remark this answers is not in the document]
Robert-Steiner: My bad! I missed the while loop. Maybe we can wrap it in another try, finally?
e.g.:
```python
try:
    while True:
        try:
            fun()
        except Exception:
            break
finally:
    channel.close()
```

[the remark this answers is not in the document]
chongshenng: @Robert-Steiner The reason `channel.close()` is outside of the `try-except` loop is to close the channel when the connection hits the `timeout` on this line:
```
for res in stub.StreamLogs(req, timeout=timeout):
...
```

So, when we only print the logs with `flwr log -F -- <RUN_ID> <APP_DIR> <FEDERATION_NAME>`, the `print_logs()` function will display the logs and wait until the `timeout` is hit and then it goes to `channel.close()` and returns to shell. We should not need to hit CTRL-C here.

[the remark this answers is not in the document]
chongshenng: This is slightly clunky, but the reason is because we are using the streaming RPC response for printing.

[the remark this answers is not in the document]
chongshenng: Another possibly cleaner option is to break from the `yield StreamLogsResponse(log_outputs=logs[i])` if it receives a `print` request. Something like:

```python
file: exec_servicer.py
# Other code
for i in range(last_sent_index, len(logs)):
    yield StreamLogsResponse(log_output=logs[i])
last_sent_index = len(logs)
if request.print:
    context.cancel()
```

Wdyt @Robert-Steiner @tanertopal?

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,221 @@
...
+        typer.Argument(help="Path of the Flower project to run"),
+    ] = Path("."),
+    federation_name: Annotated[
+        Optional[str],
+        typer.Argument(help="Name of the federation to run the app on"),
+    ] = None,
+    follow: Annotated[
+        bool,
+        typer.Option(
+            "--follow/--no-follow",
```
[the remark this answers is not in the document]
chongshenng: That's a valid point. I initially used the `docker` CLI style, but switched to the `typer` style of `--param/--no-param`. Can see why it's confusing. 

How about this experience:
* `flwr run <APP_DIR> <FEDERATION_NAME>`: Executes run, follows logstream (default)
* `flwr run --silent/-s <APP_DIR> <FEDERATION_NAME>`: Executes run without showing the logstream. 

And when users want to retrieve logs after executing `flwr run ...`:
* `flwr log -- <RUN_ID> <APP_DIR> <FEDERATION_NAME>`: Streams logs (default) if run is still running.
* `flwr log --print/-p -- <RUN_ID> <APP_DIR> <FEDERATION_NAME>`: Print logs.

(Note that `--` is necessary for `flwr log` because of the negative values of `<RUN_ID>`.)

Wdyt @Robert-Steiner (cc @danieljanes @tanertopal)?

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,221 @@
...
+from flwr.cli.config_utils import load_and_validate
+from flwr.common.grpc import GRPC_MAX_MESSAGE_LENGTH, create_channel
+from flwr.common.logger import log as logger
+from flwr.proto.exec_pb2 import StreamLogsRequest  # pylint: disable=E0611
+from flwr.proto.exec_pb2_grpc import ExecStub
+
+CONN_REFRESH_PERIOD = 60  # Connection refresh period for log streaming (seconds)
+
+
+def stream_logs(run_id: int, channel: grpc.Channel, duration: int) -> None:
```
tanertopal, line 36: @chongshenng can you add tests for this and the `print_logs` function?

src/py/flwr/cli/log.py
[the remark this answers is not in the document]
Robert-Steiner: I'm not so familiar with the code but i like your idea of having the gRPC server notify/signal the client when there are no more messages remaining, rather than relying on a client-side timeout.

I still would suggest wrapping the while loop in a try/finally block to guarantee that the channel is closed in the event of an exception, ensuring a clean exit regardless of the outcome. The user may want to terminate the process if the server sends an excessive amount of logs.

src/py/flwr/cli/log.py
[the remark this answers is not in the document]
Robert-Steiner: I like that idea! Maybe `--quite/-q` instead of  `silent`? 
For `log` we can also do it the other way around:
- `flwr log -- <RUN_ID> <APP_DIR> <FEDERATION_NAME>` Print logs.
- `flwr log --follow/-f -- <RUN_ID> <APP_DIR> <FEDERATION_NAME>` Streams logs if run is still running.

is the `--` required if you put the negative value in quotes? e.g. `flwr log "-123" ...`

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,221 @@
...
+def print_logs(run_id: int, channel: grpc.Channel, timeout: int) -> None:
+    """Print logs from the beginning of a run."""
+    stub = ExecStub(channel)
+    req = StreamLogsRequest(run_id=run_id)
+
+    while True:
+        try:
+            # Enforce timeout for graceful exit
+            for res in stub.StreamLogs(req, timeout=timeout):
+                print(res.log_output)
```
tanertopal, line 57: Why don't we use `logger(INFO, res.log_output)` here?

src/py/flwr/cli/log.py
tanertopal, line 68:
```suggestion
    try:
        while True:
            try:
                # Enforce timeout for graceful exit
                for res in stub.StreamLogs(req, timeout=timeout):
                    print(res.log_output)
            except grpc.RpcError as e:
                # pylint: disable=E1101
                if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
                    break
                if e.code() == grpc.StatusCode.NOT_FOUND:
                    logger(ERROR, "Invalid run_id `%s`, exiting", run_id)
                    break
                if e.code() == grpc.StatusCode.CANCELLED:
                    break
    
        channel.close()
    except KeyboardInterrupt:
        logger(DEBUG, "Stream interrupted by user")
    finally:
        channel.close()
        logger(DEBUG, "Channel closed")
```

@Robert-Steiner I guess you mean something like this? @chongshenng I am not 100% sure but I believe this would slightly improve the behaviour.

Robert-Steiner: @tanertopal yes exactly. To avoid calling `channel.close()` twice, we should ensure it's called only in the `finally` block.
```python
try:
    while True:
        try:
            # Enforce timeout for graceful exit
            for res in stub.StreamLogs(req, timeout=timeout):
                print(res.log_output)
        except grpc.RpcError as e:
            # pylint: disable=E1101
            if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
                break
            if e.code() == grpc.StatusCode.NOT_FOUND:
                logger(ERROR, "Invalid run_id `%s`, exiting", run_id)
                break
            if e.code() == grpc.StatusCode.CANCELLED:
                break
except KeyboardInterrupt:
    logger(DEBUG, "Stream interrupted by user")
finally:
    channel.close()
    logger(DEBUG, "Channel closed")
```

src/py/flwr/cli/log.py
chongshenng: Good question. Our logger has been pre-formatted and we'll get a duplicate `INFO  : INFO : ...` printed to screen.

src/py/flwr/cli/log.py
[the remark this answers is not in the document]
chongshenng: Re: `--quiet/-q`, good idea! It's a consistent pattern with `pip` and `poetry`. I prefer this.

To keep the user experience the same, IMO it's preferable for `flwr log -- ...` to stream the log like in `flwr run <APP_DIR> <FEDERATION_NAME>`. 

IIRC, quotes doesn't work too. Will check.

src/py/flwr/cli/log.py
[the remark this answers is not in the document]
chongshenng: Good idea, will adopt this.

chongshenng: Yes, this makes sense, especially in edge cases where there's lots of logs to print. I'll add it.

src/py/flwr/cli/log.py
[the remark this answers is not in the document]
chongshenng: @Robert-Steiner The quotes doesn't work.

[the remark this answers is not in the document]
Robert-Steiner: That is a petty! But thanks for checking it.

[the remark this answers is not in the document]
Robert-Steiner: I will resolve the conversation as the flag design will be discussed elsewhere.

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,225 @@
...
+    def on_channel_state_change(channel_connectivity: str) -> None:
+        """Log channel connectivity."""
+        logger(DEBUG, channel_connectivity)
+
+    insecure_str = federation.get("insecure")
+    if root_certificates := federation.get("root-certificates"):
+        root_certificates_bytes = Path(root_certificates).read_bytes()
+        if insecure := bool(insecure_str):
+            typer.secho(
+                "❌ `root_certificates` were provided but the `insecure` parameter"
```
Moep90, line 174: If I understand correctly, the application will exit if a `root_certificate` is provided while the federation is set to `insecure` right?

I would prefer to log this as a warning and ignore the provided parameter to avoid disrupting the user experience.

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,225 @@
...
+                "is set to `True`.",
+                fg=typer.colors.RED,
+                bold=True,
+            )
+            raise typer.Exit(code=1)
+    else:
+        root_certificates_bytes = None
+        if insecure_str is None:
+            typer.secho(
+                "❌ To disable TLS, set `insecure = true` in `pyproject.toml`.",
```
Moep90, line 184: Can we assume that the user intends to connect to an `insecure` environment when they do not provide a `root_certificate`?

src/py/flwr/cli/log.py
```diff
@@ -0,0 +1,225 @@
...
+        typer.Argument(help="Path of the Flower project to run"),
+    ] = Path("."),
+    federation_name: Annotated[
+        Optional[str],
+        typer.Argument(help="Name of the federation to run the app on"),
+    ] = None,
+    stream: Annotated[
+        bool,
+        typer.Option(
+            "--stream/--show",
```
Moep90, line 91: Same statement as I posted here:

https://github.com/adap/flower/pull/3577/files#r1696506854

Moep90, line 91: Same statement I posted here:

https://github.com/adap/flower/pull/3577/files#r1696506854

chongshenng: Resolved in #3577.

chongshenng: Resolved in #3577.

src/py/flwr/cli/log.py
chongshenng: This is a valid point. I suggest that we refactor this together in a separate PR with the `flwr run` CLI since `flwr log` also uses a lot of common code. cc @charlesbvll for info.

src/py/flwr/cli/log.py
chongshenng: Same as above, cc @charlesbvll .

src/py/flwr/cli/log.py
chongshenng: Completed.

src/py/flwr/cli/log.py
charlesbvll: I agree that this discussion should be moved to a separate PR, but I think the goal of those checks is so that users are very deliberate about those security choices

src/py/flwr/cli/log.py
charlesbvll: Related to my above comment, this is a case where we shouldn't modify the current implementation as we don't want users to start an insecure connection by error, so we must add as much friction as possible

src/py/flwr/cli/log.py
```diff
@@ -26,18 +26,57 @@
...
+        print(res.log_output)
+        if time.time() - start_time > duration:
+            break
 
-# pylint: disable=unused-argument
-def print_logs(run_id: int, channel: grpc.Channel, timeout: int) -> None:
-    """Print logs from the beginning of a run."""
+
+def print_logs(
+    run_id: int, channel: grpc.Channel, timeout: int, is_test: bool = False
```
Robert-Steiner, line 48: We can avoid introducing `is_test` here by throwing a keyboard exception in the test.
I'm not so familiar with testing grpc message, but the following test seems to work:

```python
    def test_flwr_log_print_method(self) -> None:
        """Test print_logs."""

        class MockStreamLogsResponse(object):
            @property
            def log_output(self):
                raise KeyboardInterrupt

        mock_obj = Mock()
        mock_obj.StreamLogs.side_effect = (
            [StreamLogsResponse(log_output=f"result_{i}") for i in range(1, 4)]
            + [
                MockStreamLogsResponse(),
            ],
        )

        with patch("flwr.cli.log.ExecStub", return_value=mock_obj):
            with patch("builtins.print") as mock_print:
                print_logs(run_id=123, channel=Mock(), timeout=0)
                mock_print.assert_any_call("result_1")
                mock_print.assert_any_call("result_2")
                mock_print.assert_any_call("result_3")
```

chongshenng: Thanks! `Mock.side_effect` is very useful here.

Robert-Steiner (approved):
lgtm!

danieljanes (approved):

PR #3613, opened 2024-06-14
feat(framework) Add logstream functionality to `flwr run`

chongshenng:
This PR is blocked by:
* #3577
* #3611 
* #3584


src/py/flwr/cli/run/run.py
```diff
@@ -54,6 +60,14 @@ def run(
             "inside the `pyproject.toml` in order to be properly overriden.",
         ),
     ] = None,
+    follow: Annotated[
+        bool,
+        typer.Option(
+            "--follow/--no-follow",
```
Moep90, line 66: Same statement I posted here:

https://github.com/adap/flower/pull/3577/files#r1696506854

chongshenng: Resolved in #3577

src/py/flwr/cli/run/run.py
```diff
@@ -62,6 +68,13 @@ def run(
             "inside the `pyproject.toml` in order to be properly overriden.",
         ),
     ] = None,
+    stream: Annotated[
+        bool,
+        typer.Option(
+            "--stream/--show",
```
danieljanes, line 74: Should we use `--log` here (true when set, false otherwise)? `--show` seems misleading b/c nothing would be shown

src/py/flwr/cli/run/run.py
```diff
@@ -183,6 +197,22 @@ def _run_with_superexec(
...
+                time.sleep(2)
+                log(INFO, "Reconnecting to logstream")
+        except KeyboardInterrupt:
+            log(INFO, "Exiting logstream")
+        except grpc.RpcError as e:
+            # pylint: disable=E1101
+            if e.code() == grpc.StatusCode.CANCELLED:
+                pass
+        finally:
+            channel.close()
```
danieljanes, line 214: Can we put this into a shared function to reduce duplication with `flwr log`?

src/py/flwr/cli/run/run.py
chongshenng: Good point. You're right that `--show` isn't intuitive here. But to me, I read the word `log` more naturally as a noun instead of a verb. Maybe this should just be `--stream` and is true when set, false otherwise. Then the help text is "Use `--stream` with `flwr run` to display logs; omit to suppress them".

danieljanes: We can start with that. Just maybe not the "suppress" part since it's technically wrong.

src/py/flwr/cli/run/run.py
```diff
@@ -62,6 +66,14 @@ def run(
             "inside the `pyproject.toml` in order to be properly overriden.",
         ),
     ] = None,
+    stream: Annotated[
+        bool,
+        typer.Option(
+            "--stream",
+            help="Use `--stream` with `flwr run` to display logs;\n "
+            "logs are hidden by default.",
```
danieljanes, line 74:
```suggestion
            "logs are not streamed by default.",
```

danieljanes (approved):

PR #3621, opened 2024-06-16
fix(framework) Pass `superlink` address when starting `supernode`

jafermarq:
We have deprecated `--server` for the supernodes in favour of `--superlink`. Without this change, when launching the supernode via `flower-supernode` if the address of the superlink is specified via the expected `--superlink` the internal variable associated with `--server` is used (which isn't updated from that passed via the cli)


tanertopal (approved):
lgtm!

PR #3623, opened 2024-06-17
feat(datasets) Add pathological partitioner

adam-narozniak:
## Issue
There's no support for what is commonly known as a `Pathological` partitioner.
### Description

`PathologicalPartitioner` (the version that does not take the distribution as an input) provides a way of creating partition IDs such that the number of unique classes is constrained to the number specified by users. Note that it does not imply that each class is divided evenly (see more in docs).

This method is used in our baselines in:
* fedstar,
* fedvssl,
* heterofl,
* niiid_bench.

## Proposal
Provide an implementation of `PathologicalPartitioner`. For details on how the method works, please refer to the documentation and ask questions directly there.

### Explanation
Note: if reproducing these images, they will look slightly different since a change that sorts the unique labels was applied after the images were generated.

Here's the visualization of the results that can be obtained using `PathologicalPartitioner` (not available in the docstrings but can help to understand the method):

comparison_pathological_partitioner_class_assignment_mode_2_10
comparison_pathological_partitioner_class_assignment_mode_2_20
comparison_pathological_partitioner_class_assignment_mode_3_10
comparison_pathological_partitioner_class_assignment_mode_3_20


datasets/flwr_datasets/partitioner/pathological_partitioner.py
```diff
@@ -0,0 +1,293 @@
...
+                    label = self._rng.choice(self._unique_labels, size=1)[0]
+                    if label not in self._partition_id_to_unique_labels[partition_id]:
+                        self._partition_id_to_unique_labels[partition_id].append(label)
+        elif self._class_assignment_mode == "deterministic":
+            for partition_id in range(self._num_partitions):
+                labels = []
+                for i in range(self._num_classes_per_partition):
+                    label = self._unique_labels[
+                        (partition_id + i) % len(self._unique_labels)
+                    ]
```
jafermarq, line 242:
```suggestion
                    label = (partition_id + i) % len(self._unique_labels)
```

jafermarq: is this what we want instead?

datasets/flwr_datasets/partitioner/pathological_partitioner.py
```diff
@@ -0,0 +1,293 @@
...
+from flwr_datasets.common.typing import NDArray
+from flwr_datasets.partitioner.partitioner import Partitioner
+
+
+# pylint: disable=too-many-arguments, too-many-instance-attributes
+class PathologicalPartitioner(Partitioner):
+    """Partition dataset such that each partition has a chosen number of classes.
+
+    Implementation based on Federated Learning on Non-IID Data Silos: An Experimental
+    Study https://arxiv.org/pdf/2102.02079.
```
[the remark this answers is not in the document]
adam-narozniak: I checked and both on the webiste for other links in docs and in my locally build version of the docs it works fine.

datasets/flwr_datasets/partitioner/pathological_partitioner.py
adam-narozniak: No. It should stay as it is because a label can be of different types (or with non-zero offset in case of int)

datasets/flwr_datasets/partitioner/pathological_partitioner.py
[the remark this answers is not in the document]
jafermarq: ok, then let's leave it as is 👍

datasets/flwr_datasets/partitioner/pathological_partitioner.py
```diff
@@ -0,0 +1,295 @@
...
+                    replace=False,
+                ).tolist()
+                self._partition_id_to_unique_labels[partition_id] = labels
+        else:
+            raise ValueError(
+                f"The supported class_assignment_mode are: 'random', 'deterministic', "
+                f"'first-deterministic'. You provided: {self._class_assignment_mode}."
+            )
+
+    def _determine_unique_label_to_times_used(self) -> None:
```
jafermarq, line 261: maybe not critical but could we rename this method to `construct_labels_histogram` and the  `self._unique_label_to_times_used_counter` to simply `self._labels_histogram` ? or something similar. It might help with readability and when maintaining this class.

jafermarq: What this method does is construct a histogram right?

datasets/flwr_datasets/partitioner/pathological_partitioner.py
```diff
@@ -0,0 +1,295 @@
...
+        for unique_label in self._unique_labels:
+            self._unique_label_to_times_used_counter[unique_label] = 0
+        for unique_labels in self._partition_id_to_unique_labels.values():
+            for unique_label in unique_labels:
+                self._unique_label_to_times_used_counter[unique_label] += 1
+
+    def _check_correctness_of_unique_label_to_times_used_counter(
+        self, labels: NDArray
+    ) -> None:
+        """Check if the number of times the label is possible to execute.
```
jafermarq, line 276:
```suggestion
        """Check if the number of times the label assigned to a partition.
```

jafermarq: i think it's a bit clearer like this, but I'll leave it up to you. Probably it can be tweaked further

adam-narozniak:
```suggestion
        "Check if partitioning is possible given the presence requirements.
```

adam-narozniak: I can propose alternatively this.

datasets/flwr_datasets/partitioner/pathological_partitioner.py
adam-narozniak: Resolved based on the offline conversation. 
Solution: 
* rename the fnc to improve clarity
* update the docstring

datasets/flwr_datasets/partitioner/pathological_partitioner.py
jafermarq: looks good!

jafermarq (approved):

PR #3632, opened 2024-06-17
feat(framework) Make `RecordSet` easier to use

panh99:


src/py/flwr/common/record/recordset.py
```diff
@@ -15,14 +15,63 @@
...
+            orig_value = data.metrics_records[key]
+        elif key in data.configs_records:
+            orig_value = data.configs_records[key]
+
+        if orig_value is not None and not isinstance(orig_value, self._type):
+            raise TypeError(
+                f"Key '{key}' is already associated with "
+                f"a '{type(orig_value).__name__}', but a value "
+                f"of type '{self._type.__name__}' was provided."
+            )
```
jafermarq, line 63: is this going to prevent a user from updating the value of an existing key?

src/py/flwr/common/constant.py
```diff
@@ -103,3 +103,26 @@ class ErrorCode:
...
+
+class Metric:
+    """Keys for metrics in a MetricsRecord."""
+
+    NUM_EXAMPLES = "num_examples"
+    LOSS = "loss"
+
+    def __new__(cls) -> Metric:
+        """Prevent instantiation."""
+        raise TypeError(f"{cls.__name__} cannot be instantiated.")
```
jafermarq, line 128: doesn't seem to be used anywhere atm. Maybe in a follow up PR?

jafermarq: maybe in a follow up, it sounds like a lot of changes :)

PR #3635, opened 2024-06-17
feat(framework) Check TTL when retrieving TaskRes

mohammadnaseri:
### Description
This PR checks if the corresponding `TaskIns` exists and remains valid (not expired) when retrieving `TaskRes`.
Please merge after #3620


src/py/flwr/server/superlink/state/in_memory_state.py
```diff
@@ -91,11 +91,13 @@ def get_task_ins(
                     and task_ins.task.consumer.anonymous is False
                     and task_ins.task.consumer.node_id == node_id
                     and task_ins.task.delivered_at == ""
+                    and task_ins.task.created_at + task_ins.task.ttl > time.time()
                 ) or (
                     node_id is None  # Anonymous
                     and task_ins.task.consumer.anonymous is True
                     and task_ins.task.consumer.node_id == 0
                     and task_ins.task.delivered_at == ""
+                    and task_ins.task.created_at + task_ins.task.ttl > time.time()
```
danieljanes, line 100: Let's do the same thing as we did in the previous PR here

src/py/flwr/server/superlink/state/in_memory_state.py
```diff
@@ -189,6 +189,15 @@ def get_task_res(self, task_ids: set[UUID], limit: Optional[int]) -> list[TaskRe
             replied_task_ids: set[UUID] = set()
             for _, task_res in self.task_res_store.items():
                 reply_to = UUID(task_res.task.ancestry[0])
+                task_ins = self.task_ins_store.get(reply_to)
+                if task_ins is None:
+                    log(ERROR, "TaskIns with task_id %s does not exist.", reply_to)
+                    return []
```
danieljanes, line 195: We should not return `[]` here. Some of the requested `task_ids` might have responses and we should return those back to the caller. In the future, we need to introduce a way to return errors for those IDs that are either not existing or expired.

src/py/flwr/server/superlink/state/in_memory_state.py
```diff
@@ -189,6 +189,15 @@ def get_task_res(self, task_ids: set[UUID], limit: Optional[int]) -> list[TaskRe
...
+
+                if task_ins.task.created_at + task_ins.task.ttl <= time.time():
+                    log(ERROR, "TaskIns with task_id %s is expired.", reply_to)
+                    return []
```
danieljanes, line 199: Same as above

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -451,6 +452,36 @@ def get_task_res(self, task_ids: set[UUID], limit: Optional[int]) -> list[TaskRe
...
+
+        task_ins_data = {}
+        for index, task_id in enumerate(task_ids):
+            task_ins_data[f"id_{index}"] = str(task_id)
+
+        task_ins_rows = self.query(query, task_ins_data)
+
+        if not task_ins_rows:
+            log(ERROR, "TaskIns does not exist.")
+            return []
```
danieljanes, line 473: Same as above

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -451,6 +452,36 @@ def get_task_res(self, task_ids: set[UUID], limit: Optional[int]) -> list[TaskRe
...
+
+        for row in task_ins_rows:
+            # Convert values from sint64 to uint64
+            convert_sint64_values_in_dict_to_uint64(
+                row, ["run_id", "producer_node_id", "consumer_node_id"]
+            )
+            task_ins = dict_to_task_ins(row)
+            if task_ins.task.created_at + task_ins.task.ttl <= time.time():
+                log(ERROR, "TaskIns with task_id %s is expired.", task_ins.task_id)
+                return []
```
danieljanes, line 483: Same as above

src/py/flwr/server/superlink/state/sqlite_state.py
mohammadnaseri: I think we need to return empty here is this indicates that `task_ins_rows` is empty so to `task_res` to return.

danieljanes: the two state implementations should not behave inconsistently. let's add a test that covers this particular scenarios (some ids are valid, while others are either expired or unknown) and then align the two implementations.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #3653, opened 2024-06-20
feat(datasets) Add partition_id_to_indices as property

adam-narozniak:
## Issue
There's no common way to capture the results of the partitioning.

### Description
The result of partitioning can always be in `partition_id_to_indices` and exposed to the user for saving data.

## Proposal
* add attribute `_partition_id_to_indices` 
* expose `_partition_id_to_indices` publicly via property `partition_id_to_indices`

Implementation:
* Add attribute `_partition_id_to_indices` to `Partitioner`
* Add abstract property `_partition_id_to_indices`
* Update the `IidPartitioner` to create `_partition_id_to_indices`
* Update all partitioners to have `partition_id_to_indices`. Make it a trigger (like load_partition load_split and num_partitions) of the partitioning


datasets/flwr_datasets/partitioner/iid_partitioner.py
```diff
@@ -47,11 +50,32 @@ def load_partition(self, partition_id: int) -> datasets.Dataset:
         dataset_partition : Dataset
             single dataset partition
         """
-        return self.dataset.shard(
-            num_shards=self._num_partitions, index=partition_id, contiguous=True
-        )
+        self._determine_partition_id_to_indices_if_needed()
+        return self.dataset.select(self._partition_id_to_indices[partition_id])
```
jafermarq, line 54: the `.shard()` approach above had to set `contiguous` to `True`, which is not the default value. Do we need some special handling because we are now using [`.select()`](https://huggingface.co/docs/datasets/v2.20.0/en/package_reference/main_classes#datasets.Dataset.select). I guess no?

datasets/flwr_datasets/partitioner/inner_dirichlet_partitioner.py
```diff
@@ -129,6 +128,17 @@ def num_partitions(self) -> int:
         self._determine_partition_id_to_indices_if_needed()
         return self._num_partitions
 
+    @property
+    def partition_id_to_indices(self) -> Dict[int, List[int]]:
```
jafermarq, line 132: I wonder if it's worth it moving most of this logic to a separate aux method that's also called by `num_partitions()`, `load_partition()`. The code is repeated (except for the `return` statement)

jafermarq: This `3x` repetition happens also in `shard_`, `inner_dirichtlet_`, `dirichlet_`, and `natural_id` partitioners.

datasets/flwr_datasets/partitioner/iid_partitioner.py
adam-narozniak: No, the change I did matches the behavior of shard with `contiguous=True` and it's based on the implemention of `shard`. So the order of samples will remain the same after this change.

WilliamLindskog:

PR #3658, opened 2024-06-20
feat(framework) Allow `flower-server-app` to start with `run_id`

panh99:


src/py/flwr/server/run_serverapp.py
```diff
@@ -228,5 +249,22 @@ def _parse_args_run_server_app() -> argparse.ArgumentParser:
...
+        "--run-id",
+        default=None,
+        type=int,
+        help="The identifier of the run.",
+    )
+    parser.add_argument(
+        "--flwr-dir",
+        default=None,
+        help="""The path containing installed Flower Apps.
+    By default, this value isequal to:
```
danieljanes, line 262:
```suggestion
    By default, this value is equal to:
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
             cert_path,
         )
 
-    log(
-        DEBUG,
-        "Flower will load ServerApp `%s`",
-        getattr(args, "server-app"),
+    if not (getattr(args, "server-app") is None) ^ (args.run_id is None):
```
danieljanes, line 140: We should call this only once and then use the read value below:
```suggestion
    server_app_ref: str = getattr(args, "server-app")
    if not (server_app_ref is None) ^ (args.run_id is None):
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+        raise sys.exit(
+            "Invalid options: Either `server-app` or `--run-id` "
+            "should be set but not both."
```
danieljanes, line 143:
```suggestion
            "Invalid options: Either `--server-app` or `--run-id` "
            "should be set but not both."
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+
+    # Initialize GrpcDriver
+    driver = GrpcDriver(run_id=run_id, stub=stub)
+
+    # Dynamically obtain ServerApp path based on run_id
+    if args.run_id is not None:
+        if args.flwr_dir is None:
+            flwr_dir = get_flwr_dir()
+        else:
+            flwr_dir = Path(args.flwr_dir).absolute()
```
danieljanes, line 166: nit: we could move this if/else into `get_flwr_dir()` so that getting the flwr dir becomes a consistent one-liner everywhere:
```suggestion
        flwr_dir = get_flwr_dir(args.flwr_dir)
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+        server_app_dir = str(
+            get_project_dir(driver.run.fab_id, driver.run.fab_version, flwr_dir)
+        )
```
danieljanes, line 169: nit: would it be better to get the `Run` once before this line, and then pass `run.fab_id` and `run.fab_version`?

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+        config = get_project_config(server_app_dir)
+        server_app_attr = config["flower"]["components"]["serverapp"]
+    else:
+        server_app_dir = str(Path(args.dir).absolute())
+        server_app_attr = getattr(args, "server-app")
```
danieljanes, line 174: Related to the very first comment above:
```suggestion
```

src/py/flwr/server/run_serverapp.py
charlesbvll: Here it is: #3661

danieljanes: Thanks, merged

src/py/flwr/server/run_serverapp.py
panh99: Should we add `--` before the positional argument `server-app`? I am asking this because we will run the CLI like `flower-server-app server:app` or `flower-server-app --run-id 123`.

danieljanes: Good question. I added it to make it clear that this is an argument, but it might also mislead users into thinking that it needs to be passed with `--server-app`. Just `server-app` looked a bit off. Is there a different way we could put this?

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,46 @@ def run_server_app() -> None:
...
+            "Please provide either a ServerApp reference or a Run ID, but not both. "
+            "For more details, use: ``flower-server-app -h``"
+        )
+
+    stub = GrpcDriverStub(
+        driver_service_address=args.superlink, root_certificates=root_certificates
     )
+    # Create run if run_id is not provided
+    if args.run_id is None:
+        stub.connect()
```
danieljanes, line 152:
```suggestion
        # User provided `server-app`, but not `--run-id`
        stub.connect()
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,46 @@ def run_server_app() -> None:
...
+        req = CreateRunRequest(fab_id=args.fab_id, fab_version=args.fab_version)
+        res = stub.create_run(req)
+        run_id = res.run_id
+    else:
+        run_id = args.run_id
```
danieljanes, line 157:
```suggestion
        # User provided `--run-id`, but not `server-app`
        run_id = args.run_id
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,46 @@ def run_server_app() -> None:
...
+
+    # Dynamically obtain ServerApp path based on run_id
+    if args.run_id is not None:
+        flwr_dir = get_flwr_dir(args.flwr_dir)
+        run_ = driver.run
+        server_app_dir = str(get_project_dir(run_.fab_id, run_.fab_version, flwr_dir))
+        config = get_project_config(server_app_dir)
+        server_app_attr = config["flower"]["components"]["serverapp"]
+    else:
+        server_app_dir = str(Path(args.dir).absolute())
```
danieljanes, line 170:
```suggestion
        # User provided `server-app`, but not `--run-id`
        server_app_dir = str(Path(args.dir).absolute())
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+    # Create run if run_id is not provided
+    if args.run_id is None:
+        # User provided `server-app`, but not `--run-id`
+        stub.connect()
+        req = CreateRunRequest(fab_id=args.fab_id, fab_version=args.fab_version)
+        res = stub.create_run(req)
+        run_id = res.run_id
+    else:
+        # User provided `--run-id`, but not `server-app`
+        run_id = args.run_id
```
danieljanes, line 159:
```suggestion
    if args.run_id is not None:
        # User provided `--run-id`, but not `server-app`
        run_id = args.run_id
    else:
        # User provided `server-app`, but not `--run-id`
        stub.connect()
        req = CreateRunRequest(fab_id=args.fab_id, fab_version=args.fab_version)
        res = stub.create_run(req)
        run_id = res.run_id
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+    # Create run if run_id is not provided
+    if args.run_id is not None:
+        # User provided `--run-id`, but not `server-app`
+        run_id = args.run_id
+    else:
+        # User provided `server-app`, but not `--run-id`
+        stub.connect()
+        req = CreateRunRequest(fab_id=args.fab_id, fab_version=args.fab_version)
+        res = stub.create_run(req)
+        run_id = res.run_id
```
danieljanes, line 159:
```suggestion
    if args.run_id is not None:
        # User provided `--run-id`, but not `server-app`
        run_id = args.run_id
    else:
        # User provided `server-app`, but not `--run-id`
        # Create run if run_id is not provided
        stub.connect()
        req = CreateRunRequest(fab_id=args.fab_id, fab_version=args.fab_version)
        res = stub.create_run(req)
        run_id = res.run_id
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+
+    # Initialize GrpcDriver
+    driver = GrpcDriver(run_id=run_id, stub=stub)
+
+    # Dynamically obtain ServerApp path based on run_id
+    if args.run_id is not None:
+	# User provided `--run-id`, but not `server-app`
+        flwr_dir = get_flwr_dir(args.flwr_dir)
+        run_ = driver.run
```
danieljanes, line 168: Can this be just `run`?
```suggestion
        run = driver.run
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -136,32 +137,50 @@ def run_server_app() -> None:
...
+        server_app_dir = str(get_project_dir(run_.fab_id, run_.fab_version, flwr_dir))
+        config = get_project_config(server_app_dir)
```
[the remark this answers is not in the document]
panh99: `get_project_config` can accept a path, but the `run` function does not accept a `str` type `server_app_dir`.

danieljanes (dismissed):

danieljanes (approved):

PR #3659, opened 2024-06-20
feat(datasets) Add to_config[_file] and from_config[_file] to Partitioners

adam-narozniak:
## Issue
Config-based loading and saving is a utily that does not exist but can save time and simply the experiments for many people.

## Proposal
Enable config-based loading and saving via the following 4 methods.

Loading:
* from_config
* from_config_file
Saving:
* to_config
* to_config_file

They enable object creation based on config and optionally state (`partition_id_to_indices`) recreation.


WilliamLindskog:

PR #3663, opened 2024-06-21
feat(datasets) Add semantic partitioner

KarhouTam:
## New Feature

Implement semantic partitioning scheme (named `SemanticPartitioner`). 

Semantic partitioning scheme is proposed by [What Do We Mean by Generalization in Federated Learning?](https://arxiv.org/abs/2110.14216 ) (accepted by ICLR 2022)

(Cited from the paper)
Semantic partitioner's goal is to reverse-engineer the federated dataset-generating process so that each client possesses semantically similar data.

### Checklist

- [x] Write tests
- [ ] Make CI checks pass

### Any other comments?
I have to mention that this partitioner has a strong dependency on the PyTorch library, so I don't know if this is allowed or not.

It's my first time to contribute flower with new feature.

Happy to see contributors's comments and suggestions ❤.


KarhouTam:
Check failed due to 

```
- flake8: start
flwr_datasets/partitioner/semantic_partitioner.py:251:29: E203 whitespace before ':'
Error: Process completed with exit code 1.
```

The target code is
```python
images[i : i + self._batch_size]
```

`flake8` means that format should be `images[i: i + self._batch_size]`. However, this format cannot pass the `black` format check.

Means it's impossible for my codes to pass all the checks. Sad.

Anyway, I've tried my best to format my codes to pass as many checks as possible.

KarhouTam:
Some checks failed due to the strong dependency on `pytorch`, `sklearn` and `scipy`.

jafermarq:
Hi @KarhouTam , this looks amazing! we'll get back to you early in the week.

KarhouTam:
Hi, @jafermarq 
Any progress on the feedback?

datasets/flwr_datasets/partitioner/semantic_partitioner.py
```diff
@@ -0,0 +1,493 @@
...
+
+import numpy as np
+
+import datasets
+from flwr_datasets.common.typing import NDArrayFloat
+from flwr_datasets.partitioner.partitioner import Partitioner
+
+
+# pylint: disable=R0902, R0912, R0914
+class SemanticPartitioner(Partitioner):
```
adam-narozniak, line 29:
```suggestion
class ImageSemanticPartitioner(Partitioner):
```

adam-narozniak: Since it works only with images, maybe it'll be more descriptive (I assume in the future there could be as well, e.g. `AudioSemanticPartitioner`. Or maybe some reference to the fact that labels are also considered can be included in the name (e.g., Supervised), but that might be too long, wdyt?

datasets/flwr_datasets/partitioner/semantic_partitioner.py
```diff
@@ -0,0 +1,493 @@
...
+    gmm_init_params: str
+        The initialization method for the GMM algorithm.
+        Options: ["random", "kmeans", "k-means++"]
+    use_cuda: bool
+        Whether to use CUDA for computation acceleration.
+    shuffle: bool
+        Whether to randomize the order of samples. Shuffling applied after the
+        samples assignment to partitions.
+    seed: int
+        Seed used for dataset shuffling. It has no effect if `shuffle` is False.
```
adam-narozniak, line 94:
```suggestion
        Seed used for dataset shuffling and as random_state in PCA and GMM. 
        It has no effect on shuffling if `shuffle` is False. It is always passed to 
        the models mentioned before.
```

adam-narozniak: Alternatively, we can have two additional parameters specifying the seed values for the PCA and GMM.

datasets/flwr_datasets/partitioner/semantic_partitioner.py
```diff
@@ -0,0 +1,493 @@
...
+        The maximum number of iterations for the GMM algorithm.
+    gmm_init_params: str
+        The initialization method for the GMM algorithm.
+        Options: ["random", "kmeans", "k-means++"]
+    use_cuda: bool
+        Whether to use CUDA for computation acceleration.
+    shuffle: bool
+        Whether to randomize the order of samples. Shuffling applied after the
+        samples assignment to partitions.
+    seed: int
```
adam-narozniak, line 93:
```suggestion
    seed: Optional[int]
```

datasets/flwr_datasets/partitioner/semantic_partitioner.py
```diff
@@ -0,0 +1,493 @@
...
+    # partitioner = SemanticPartitioner(
+    #     num_partitions=5, partition_by="label", pca_components=128
+    # )
+    # fds = FederatedDataset(dataset="cifar10", partitioners={"train": partitioner})
+    # partition = fds.load_partition(0)
+    # print(partition[0])  # Print the first example
+    # partition_sizes = partition_sizes = [
+    #     len(fds.load_partition(partition_id)) for partition_id in range(5)
+    # ]
+    # print(sorted(partition_sizes))
```
adam-narozniak, line 492: We can add it as an example, or remove it

datasets/flwr_datasets/partitioner/semantic_partitioner.py
```diff
@@ -0,0 +1,493 @@
...
+            raise ValueError(
+                "The gmm_init_params needs to be in [kmeans, k-means++, random]"
+            )
+        if self._gmm_max_iter <= 0:
+            raise ValueError("The gmm max iter needs to be greater than zero.")
+        if self._pca_components <= 0:
+            raise ValueError("The pca components needs to be greater than zero.")
+
+
+if __name__ == "__main__":
```
adam-narozniak, line 461: Let's drop the `if __name__ == "__main__":` section

datasets/flwr_datasets/partitioner/semantic_partitioner.py
```diff
@@ -0,0 +1,493 @@
...
+    >>> fds = FederatedDataset(dataset="mnist", partitioners={"train": partitioner})
+    >>> partition = fds.load_partition(0)
+    >>> print(partition[0])  # Print the first example
+    {'image': <PIL.PngImagePlugin.PngImageFile image mode=L
+    size=28x28 at 0x7FCF49741B10>, 'label': 3}
+    >>> partition_sizes = partition_sizes = [
+    >>>     len(fds.load_partition(partition_id)) for partition_id in range(5)
+    >>> ]
+    >>> print(sorted(partition_sizes))
+    [3163, 5278, 5496, 6320, 9522]
```
adam-narozniak, line 118: A more general comment: I'm profiling the code to understand better what takes time (I'll check the memory perf, too). For data bigger than 1k, it takes some time on the local machine. We should signal that progress is being made (not to confuse people that the code froze). The lack of a proper logger makes it hard, so I'll suggest a print for now.

adam-narozniak (commented):
Hi! First of all, sorry for the long wait time. Overall, outstanding contribution and a really high-quality code!!! Thanks!

I left a few minor comments, suggestions/questions. 

From my side, I'm still checking the perf (we can signalize why the method will take longer than simpler label skew-based heterogeneity simulation) + I'll handle the CI to to have the proper libs installed for the test for this partitioner.

KarhouTam:
Hi, @adam-narozniak. 
Thanks for your reviewing. Very happy to see Flower's appreciation! 
I will check these suggestions later this week

Let's make Flower better and better. 🥳

datasets/flwr_datasets/partitioner/semantic_partitioner.py
```diff
@@ -0,0 +1,493 @@
...
+                )
+
+    def _check_data_type_if_needed(self) -> None:
+        """Test whether data is image-like."""
+        if not self._partition_id_to_indices_determined:
+            features_dict = self.dataset.features.to_dict()
+            self._data_column_name = list(features_dict.keys())[0]
+            try:
+                data = np.array(
+                    self.dataset[self._data_column_name][0], dtype=np.float32
```
adam-narozniak, line 415: This part can be replaced with checking if the column type is `datasets.Image`, if not (for reasons I can't see right now) we should index first based on the row id then the column name. The current indexing pulls all the images to memory first (since it's indexed first by a column first).

datasets/flwr_datasets/partitioner/semantic_partitioner.py
KarhouTam: Agree. Each type of data can have its own semantic information. So it seems practical to partition for other types of data as well. Perhaps the only difference is the way pretrained models are used to extract semantic information.

I will also change the python filenames to `image_semantic_partitioner.py` and `image_semantic_partitioner_test.py` as well.

datasets/flwr_datasets/partitioner/semantic_partitioner.py
KarhouTam: Agree. I will add corresponding test case too.

datasets/flwr_datasets/partitioner/semantic_partitioner.py
KarhouTam: Agree with most of it. But I think the column type should not be strict. 
Since maybe some users want to use their private datasets by `Dataset.from_dict()` or other ways and the feature dict of that dataset maybe not set properly. 
So I think in this place, we can throw a warning says the process maybe failed due to the column type of `image_column_name` is not `datasets.Image`.

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
```diff
@@ -0,0 +1,505 @@
...
+        self._efficient_output_size = 1280  # fixed in EfficientNet class
+        self._pca_components = pca_components
+        self._gmm_max_iter = gmm_max_iter
+        self._gmm_init_params = gmm_init_params
+        self._use_cuda = use_cuda
+        self._shuffle = shuffle
+        if pca_seed is None:
+            pca_seed = sample_seed
+        if gmm_seed is None:
+            gmm_seed = sample_seed
```
adam-narozniak, line 160: I think if it's None, then it should stay None. These would be a surprising side effect if done as currently. Though, let's add the default values to them in the init.

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
```diff
@@ -0,0 +1,505 @@
...
+        num_partitions: int,
+        partition_by: str,
+        efficient_net_type: int = 3,
+        batch_size: int = 32,
+        pca_components: int = 128,
+        gmm_max_iter: int = 100,
+        gmm_init_params: str = "kmeans",
+        use_cuda: bool = False,
+        shuffle: bool = True,
+        sample_seed: int = 42,
```
adam-narozniak, line 140: It's more of `shuffle_seed`. It should be also `Optional`

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
```diff
@@ -0,0 +1,505 @@
...
+        pca_seed: Optional[int] = None,
+        gmm_seed: Optional[int] = None,
+        image_column_name: Optional[str] = None,
```
adam-narozniak, line 143: According to, unfortunately, unwritten convention, let's keep the shuffling parameters last and move this parameter anywhere before the shuffle.

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
```diff
@@ -0,0 +1,505 @@
...
+            ):
+                warnings.warn(
+                    "Image semantic partitioner only supports image-like data. "
+                    f"But column '{self._image_column_name}' is not datasets.Image. "
+                    "So the partition might be failed.",
+                    stacklevel=1,
+                )
+            try:
+                image = np.array(
+                    self.dataset.take(1)[self._image_column_name][0], dtype=np.float32
```
adam-narozniak, line 454:
```suggestion
                    self.dataset[0][self._image_column_name], dtype=np.float32
```

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
```diff
@@ -0,0 +1,505 @@
...
+        # Use EfficientNet to extract embeddings
+        images = self._preprocess_dataset_images()
+        embedding_list = []
+        with torch.no_grad():
+            for i in range(0, images.shape[0], self._batch_size):
+                idxs = list(range(i, min(i + self._batch_size, images.shape[0])))
+                batch = torch.tensor(images[idxs], dtype=torch.float, device=device)
+                if batch.shape[1] == 1:
+                    batch = batch.broadcast_to((batch.shape[0], 3, *batch.shape[2:]))
+                embedding_list.append(efficient_net(batch).cpu().numpy())
```
adam-narozniak, line 271: I'm curious, what is the reason that you handle batches "by hand" instead of using `DataLoader`?

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
```diff
@@ -0,0 +1,505 @@
...
+
+        # Use EfficientNet to extract embeddings
+        images = self._preprocess_dataset_images()
+        embedding_list = []
+        with torch.no_grad():
+            for i in range(0, images.shape[0], self._batch_size):
+                idxs = list(range(i, min(i + self._batch_size, images.shape[0])))
+                batch = torch.tensor(images[idxs], dtype=torch.float, device=device)
+                if batch.shape[1] == 1:
+                    batch = batch.broadcast_to((batch.shape[0], 3, *batch.shape[2:]))
```
adam-narozniak, line 270: Let's add a comment here that it's for the purpose of changing the grayscale images to meet the RGB shape.

adam-narozniak:
@KarhouTam I left a few more comments and questions.

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
KarhouTam: Due to this seed is used by `self._rng_numpy` for index sampling and shuffling through the whole partition process. How about `partition_seed`?

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
KarhouTam: I think using `DataLoader` needs to trans `images` to `torch.utils.data.Dataset` first and is better used in multiple round data loading. Here I just want to traverse `images` one time. Anyway, just my habit.

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
```diff
@@ -0,0 +1,505 @@
...
+        # Shuffle the indices not to have the datasets with targets in sequences like
+        # [00000, 11111, ...]) if the shuffle is True
+        if self._shuffle:
+            for indices in partition_id_to_indices.values():
+                # In place shuffling
+                self._rng_numpy.shuffle(indices)
+        self._partition_id_to_indices = partition_id_to_indices
+        self._partition_id_to_indices_determined = True
+
+    def _preprocess_dataset_images(self) -> NDArrayFloat:
```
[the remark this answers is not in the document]
KarhouTam: You're right. We should load data batch and process it one by one.

datasets/flwr_datasets/partitioner/image_semantic_partitioner.py
adam-narozniak: Sampling can be more about selecting a subset of samples we basically reorder = shuffle the samples. I'm also ok with just seed for cross-partitioners consistency but `partition_seed` doesn't clarify anything to me. Also, this seed is used only if the `shuffle` is True because it's a seed used when shuffling, that's also the name of the function that is used `self._rng_numpy.shuffle`.

KarhouTam: But `self._rng_numpy` is also used for sampling:

L207, 329, 335:
```
        idx_samples = self._rng_numpy.choice(
            len(embeddings), num_samples, replace=False
        )
```
```
latest_matched_label = self._rng_numpy.choice(unmatched_labels) # type: ignore
```
```
label_to_match = self._rng_numpy.choice(unmatched_labels) # type: ignore
```

😐

adam-narozniak: Ok, sorry my bad, so `rng_seed` seems appropriate

adam-narozniak:
I wanted to merge the code, but checking the results visually on the mnist dataset doesn't seem to produce results such that the same partition has images in the same style (D.3 Visualization of Semantically Partitioned MNIST Dataset from the paper). I'm gonna investigate it further but we should be able to produce something similar.

KarhouTam:
Maybe some hyperparameter settings are not the same. I remember that I reproduced this partitioner by the official repo: https://github.com/google-research/federated/tree/master/generalization.

adam-narozniak:
Hi, I haven't found the problem; however, before resolving that, we won't be able to merge this PR. We need to ensure that we can reproduce these results.

KarhouTam:
Hi, @adam-narozniak.
I've followed the source and fix some different process codes in the partitioner.

Also I evaluate the partitioner with the same settings in the paper and here is the graph:
image

adam-narozniak:
What parameters did you use? In the paper, directly, I saw just information about the number of clients = 300 clients and it gives me this plot (which is not convincing) 
image

KarhouTam:
I reviewed the source code from Google and noticed some parameter discrepancies between the paper and the code. I’ve updated them based on the official source code.

Could you please help test the latest version with 300 clients on MNIST? My laptop doesn’t have enough memory to handle the EfficientNet-B7 model with 300 partition settings. Most parameters don’t require any changes.

PR #3665, opened 2024-06-21
refactor(datasets) Rename the private attributes in InnerDirichlet

adam-narozniak:
## Issue
The naming of the class attributes resulting from the constructor paramters is not consistent.
The `alpha` is named `_initial_alpha`.

### Description
There are two _alpha attributes: initial alpha (which can be as simple as an int or float) and fully transformed alpha (proper shape for sampling from dirichlet).
The issue with inconsistent naming is trouble with saving this class in automatic manner to a config. The `_alpha` is save which is `None` but instead the `_initial_alpha` should be saved (to be albe to later on instantiate the object based on the config).

## Proposal

Rename:
* `_alpha` to `_full_alpha`
* `_initial_alpha` to `_alpha`

Related PRs
This is needed for: https://github.com/adap/flower/pull/3659


WilliamLindskog:

PR #3669, opened 2024-06-21
fix(examples:skip) Fix main README

mohammadnaseri:
### Description
It removes the extra space in main README file.


jafermarq (approved):
Good catch!

PR #3675, opened 2024-06-23
fix(*:skip) Update visualization guide how-to

jafermarq:
Small suggestions for #3672:
- wrap pip install with `""`
- install `audio` extension is needed for speechcommands at the end of the guide
- Added a bit extra context in some parts of the guide.


adam-narozniak (approved):
Looks great, thanks for the changes and for opening this PR

PR #3695, opened 2024-06-29
feat(framework) Make `NodeState` capture `partition-id`

jafermarq:
- The `NodeState` carries the `partition-id` that the `Context` should use
- `SuperNode` gains `--partition-id` flag
- `<context>.partition_id` is unused in this PR -- it will be used in a follow up PR

:warning: TBD if we want to throw a warning when passing `--partition-id` when launching a `SuperNode`. The original motivation for doing so was so to make it clear that thinking about data partitions (in the sense of "one client gets one of the N locally available partitions") does not quite fit how deployment setups work.

### DIFF
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -191,6 +191,7 @@ def _start_client_internal(
     ] = None,
     max_retries: Optional[int] = None,
     max_wait_time: Optional[float] = None,
+    partition_id: Optional[int] = None,
 ) -> None:
     """Start a Flower client node which connects to a Flower server.
 
@@ -234,6 +235,9 @@ def _start_client_internal(
         The maximum duration before the client stops trying to
         connect to the server in case of connection error.
         If set to None, there is no limit to the total time.
+    partitioni_id: Optional[int] (default: None)
+        The data partition index associated to this process. Better suited for
+        prototyping purposes.
     """
     if insecure is None:
         insecure = root_certificates is None
@@ -309,7 +313,7 @@ def _start_client_internal(
         on_backoff=_on_backoff,
     )
 
-    node_state = NodeState()
+    node_state = NodeState(partition_id=partition_id)
     # run_id -> (fab_id, fab_version)
     run_info: Dict[int, Tuple[str, str]] = {}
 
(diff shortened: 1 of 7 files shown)


src/py/flwr/client/app.py
```diff
@@ -234,6 +235,9 @@ class `flwr.client.Client` (default: None)
         The maximum duration before the client stops trying to
         connect to the server in case of connection error.
         If set to None, there is no limit to the total time.
+    partitioni_id: Optional[int] (default: None)
+        The data partition index associated to this process. Better suited for
```
danieljanes, line 239:
```suggestion
        The data partition index associated with this node. Better suited for
```

src/py/flwr/client/node_state.py
```diff
@@ -15,22 +15,25 @@
...
+from typing import Any, Dict, Optional
 
 from flwr.common import Context, RecordSet
 
 
 class NodeState:
     """State of a node where client nodes execute runs."""
 
-    def __init__(self) -> None:
+    def __init__(self, partition_id: Optional[int] = None) -> None:
```
danieljanes, line 26: Better to set the default only in one place and then force setting the value everywhere else:
```suggestion
    def __init__(self, partition_id: Optional[int]) -> None:
```

src/py/flwr/client/supernode/app.py
```diff
@@ -373,6 +374,13 @@ def _parse_args_common(parser: argparse.ArgumentParser) -> None:
         type=str,
         help="The SuperNode's public key (as a path str) to enable authentication.",
     )
+    parser.add_argument(
+        "--partition-id",
+        type=int,
+        help="The data partition index associated to this SuperNode. Better suited "
```
danieljanes, line 380:
```suggestion
        help="The data partition index associated with this SuperNode. Better suited "
```

src/py/flwr/client/supernode/app.py
```diff
@@ -373,6 +374,13 @@ def _parse_args_common(parser: argparse.ArgumentParser) -> None:
...
+        "for prototyping purposes where a SuperNode might only load a fraction of a "
```
danieljanes, line 381:
```suggestion
        "for prototyping purposes where a SuperNode might only load a fraction of an "
```

danieljanes (approved):

PR #3697, opened 2024-06-29
feat(framework) Introduce new `client_fn` signature

jafermarq:
Modify `client_fn` signature. Support previous signature through an adaptor (see #3700 ). The old signature for `client_fn` is deprecated.


src/py/flwr/client/app.py
```diff
@@ -248,7 +248,8 @@ class `flwr.client.Client` (default: None)
         if client_fn is None:
             # Wrap `Client` instance in `client_fn`
             def single_client_factory(
-                cid: str,  # pylint: disable=unused-argument
+                node_id: int,  # pylint: disable=unused-argument
+                partition_id: Optional[int],  # pylint: disable=unused-argument
             ) -> Client:
                 if client is None:  # Added this to keep mypy happy
```
danieljanes, line 254: Unrelated, but this check should happen before `def single_client_factory(...)`

src/py/flwr/client/message_handler/message_handler.py
```diff
@@ -93,7 +92,12 @@ def handle_legacy_message_from_msgtype(
     client_fn: ClientFn, message: Message, context: Context
 ) -> Message:
     """Handle legacy message in the inner most mod."""
-    client = client_fn(str(context.partition_id))
+    try:
+        client = client_fn(message.metadata.dst_node_id, context.partition_id)
+    except Exception as ex:  # pylint: disable=broad-exception-caught
+        log(DEBUG, ex)
+        # Attempt execution `client_fn` as it was done before flwr 1.10.0
+        client = client_fn(str(context.partition_id))  # type: ignore
```
danieljanes, line 100: This should not happen in the `except` block b/c the body of a `client_fn` could raise an exception that is unrelated to the signature of `client_fn`. Instead of this, we should inspect the signature of the `client_fn` and call it with either one or two arguments (and no retry in `except`).

jafermarq: right

src/py/flwr/client/typing.py
```diff
@@ -15,14 +15,14 @@
...
 
-from typing import Callable
+from typing import Callable, Optional
 
 from flwr.common import Context, Message
 
 from .client import Client as Client
 
 # Compatibility
-ClientFn = Callable[[str], Client]
```
danieljanes, line 25: Should we keep this as `ClientFnLegacy` or so?

This type is still used in `start_simulation` and `start_client`, or not?

src/py/flwr/simulation/ray_transport/ray_client_proxy_test.py
```diff
@@ -50,8 +50,8 @@
 class DummyClient(NumPyClient):
     """A dummy NumPyClient for tests."""
 
-    def __init__(self, cid: str) -> None:
-        self.cid = int(cid)
+    def __init__(self, cid: int) -> None:
```
danieljanes, line 53: It'd be better to either use `node_id: int` or `cid: str` to prevent confusion

src/py/flwr/simulation/ray_transport/ray_client_proxy_test.py
```diff
@@ -64,9 +64,13 @@ def get_properties(self, config: Config) -> Dict[str, Scalar]:
...
 
-def get_dummy_client(cid: str) -> Client:
+def get_dummy_client(
+    node_id: int, partition_id: Optional[int]  # pylint: disable=unused-argument
+) -> Client:
     """Return a DummyClient converted to Client type."""
-    return DummyClient(cid).to_client()
+    if partition_id is None:
+        raise ValueError("`partition_id` is not set.")
+    return DummyClient(partition_id).to_client()
```
danieljanes, line 73: Should we start to pass `node_id` instead of `partition_id` to `Client`/`NumPyClient` objects?

jafermarq: This is just a test, but I could amend it to use `node_id`. 

Do you mean injecting `node_id` to all `Client/NumPyClient` ? How does doing so via the `Context`, which is already being injected into the clients (see https://github.com/adap/flower/blob/5fe288aec10386a9f479750d5f968f47236dece6/src/py/flwr/client/message_handler/message_handler.py#L108)

src/py/flwr/client/typing.py
jafermarq: we can keep the previous `ClientFn` as `ClientFnLegacy`. 

Not sure if I understand the second question, both go through the same same logic in `message_handler` so they will both fetch the `partition_id` from the `Context` and `node_id` from the `Message`. However, since `start_client` isn't triggered via `flower-supernode`, `partition_id` will remain unset. As for `node_id`, it defaults to `0` for `start_client` (probably we can changes things in our compatibility layer so it uses a randomly generate value). in case of `start_simulation` it will use an UUID once #3699 is merged.

danieljanes: Let me rephrase. Existing projects that use either `start_simulation` or `start_client` might depend on `ClientFn` if they use type declarations. After more digging I noticed that it's exported and part of the public API. I'm therefore retracting my comment above - we must keep it as-is and introduce the new signature under a new name (`ClientFn...`).

danieljanes: The `message_handler` question is more difficult. It seems like there is no issue with everything related to Flower Next. It would be nice to find a better solution for `start_client` (something other than passing 0), but it's also not too critical. `node_id 0` and `partition_id None` might be the best solution - it'd be confusing to see a different random `node_id` on the client and on the server.

src/py/flwr/simulation/ray_transport/ray_client_proxy_test.py
danieljanes: My comment was more about consistency. `node_id` and `cid` refer to the same overall concept, whereas `partition_id` does not.

It would therefore be more consistent to use
```suggestion
    return DummyClient(cid=str(node_id)).to_client()
```
if a `cid: str` is expected.

src/py/flwr/client/message_handler/message_handler.py
jafermarq: resolved in #3700

src/py/flwr/simulation/ray_transport/ray_client_proxy_test.py
jafermarq: I'll make the change in #3699 once this PR is merged. I think it's easier that way

src/py/flwr/client/client_app.py
```diff
@@ -15,19 +15,43 @@
...
 from flwr.common import Context, Message, MessageType
-from flwr.common.logger import warn_preview_feature
+from flwr.common.logger import warn_deprecated_feature, warn_preview_feature
 
 from .typing import ClientAppCallable
 
 
+def _inspect_maybe_adapt_client_fn_signature(client_fn: ClientFnExt) -> ClientFnExt:
+
+    client_fn_args = inspect.signature(client_fn).parameters
```
danieljanes, line 35:
```suggestion
def _inspect_maybe_adapt_client_fn_signature(client_fn: ClientFnExt) -> ClientFnExt:
    client_fn_args = inspect.signature(client_fn).parameters
```

src/py/flwr/simulation/app.py
```diff
@@ -92,16 +92,16 @@ def start_simulation(
...
-        `str` argument called `cid`. It should return a single client instance
-        of type Client. Note that the created client instances are ephemeral
-        and will often be destroyed after a single method invocation. Since client
-        instances are not long-lived, they should not attempt to carry state over
-        method invocations. Any state required by the instance (model, dataset,
-        hyperparameters, ...) should be (re-)created in either the call to `client_fn`
-        or the call to any of the client methods (e.g., load evaluation data in the
-        `evaluate` method itself).
+    client_fn : ClientFnExt
+        A function creating client instances. The function must has signature
```
danieljanes, line 96:
```suggestion
        A function creating Client instances. The function must has signature
```

src/py/flwr/simulation/app.py
```diff
@@ -92,16 +92,16 @@ def start_simulation(
...
-        `str` argument called `cid`. It should return a single client instance
-        of type Client. Note that the created client instances are ephemeral
-        and will often be destroyed after a single method invocation. Since client
-        instances are not long-lived, they should not attempt to carry state over
-        method invocations. Any state required by the instance (model, dataset,
-        hyperparameters, ...) should be (re-)created in either the call to `client_fn`
-        or the call to any of the client methods (e.g., load evaluation data in the
-        `evaluate` method itself).
+    client_fn : ClientFnExt
+        A function creating Client instances. The function must has signature
```
danieljanes, line 96:
```suggestion
        A function creating Client instances. The function must have the signature
```

src/py/flwr/simulation/app.py
```diff
@@ -92,16 +92,16 @@ def start_simulation(
...
+        `client_fn(node_id: int, partition_id: Optional[int]). It should return
+        a single client instance of type Client. Note that the created client
+        instances are ephemeral and will often be destroyed after a single method
+        invocation. Since clientinstances are not long-lived, they should not attempt
```
danieljanes, line 100:
```suggestion
        invocation. Since client instances are not long-lived, they should not attempt
```

src/py/flwr/simulation/ray_transport/ray_client_proxy_test.py
```diff
@@ -50,12 +50,12 @@
...
 
-    def __init__(self, cid: str) -> None:
-        self.cid = int(cid)
+    def __init__(self, node_id: int) -> None:
+        self.node_id = node_id
 
     def get_properties(self, config: Config) -> Dict[str, Scalar]:
         """Return properties by doing a simple calculation."""
-        result = int(self.cid) * pi
+        result = int(self.node_id) * pi
```
danieljanes, line 58:
```suggestion
        result = self.node_id * pi
```

src/py/flwr/client/app.py
```diff
@@ -51,7 +51,7 @@
 
 
 def _check_actionable_client(
-    client: Optional[Client], client_fn: Optional[ClientFn]
+    client: Optional[Client], client_fn: Optional[Union[ClientFn, ClientFnExt]]
```
danieljanes, line 54:
```suggestion
    client: Optional[Client], client_fn: Optional[ClientFnExt]
```

danieljanes (approved):

danieljanes (approved):

PR #3699, opened 2024-06-29
break(framework) Remove support for `client_ids` in `start_simulation`

jafermarq:
`RayClientProxy` objects now receive a `node-id` and `partition-id` from a mapping generated in the same way as in the Flower-Next compatible `Simulation Engine` (excluding the fact there's no `SuperLink` state).

Subsequent PRs should remove the usage of `cid` in `simulation/ray_actor.py`


src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -44,16 +44,22 @@ class RayActorClientProxy(ClientProxy):
     """Flower client proxy which delegates work using Ray."""
 
     def __init__(
-        self, client_fn: ClientFn, cid: str, actor_pool: VirtualClientEngineActorPool
+        self,
+        client_fn: ClientFn,
+        partition_id: int,
+        node_id: int,
```
danieljanes, line 50: nit: let's keep the order the same everywhere
```suggestion
        node_id: int,
        partition_id: int,
```

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -44,16 +44,22 @@ class RayActorClientProxy(ClientProxy):
...
+        actor_pool: VirtualClientEngineActorPool,
     ):
-        super().__init__(cid)
+        super().__init__(cid=str(partition_id))
```
danieljanes, line 53: Or should this be `node_id`?

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
```diff
@@ -44,16 +44,22 @@ class RayActorClientProxy(ClientProxy):
...
+        self.node_id = node_id
+        self.partition_id = partition_id
 
         def _load_app() -> ClientApp:
             return ClientApp(client_fn=client_fn)
 
         self.app_fn = _load_app
         self.actor_pool = actor_pool
-        self.proxy_state = NodeState(partition_id=int(self.cid))
+        self.proxy_state = NodeState(partition_id=int(self.partition_id))
```
danieljanes, line 62:
```suggestion
        self.proxy_state = NodeState(partition_id=self.partition_id)
```

src/py/flwr/simulation/app.py
```diff
@@ -70,6 +72,22 @@
 
 """
 
+NodeToPartitionMapping = Dict[int, int]
+
+
+def _create_node_id_to_partition_mapping(
+    partition_ids: List[str],
```
danieljanes, line 79: Shouldn't these be integers?
```suggestion
    partition_ids: List[int],
```

src/py/flwr/simulation/app.py
```diff
@@ -70,13 +72,29 @@
...
+    return nodes_mapping
+
 
 # pylint: disable=too-many-arguments,too-many-statements,too-many-branches
 def start_simulation(
     *,
     client_fn: ClientFn,
     num_clients: Optional[int] = None,
-    clients_ids: Optional[List[str]] = None,
+    clients_ids: Optional[Union[List[int], List[str]]] = None,
```
danieljanes, line 97: Should we add the `List[int]` variant as a separate argument `node_ids`? And deprecate `client_ids`?

src/py/flwr/simulation/app.py
```diff
@@ -105,10 +123,11 @@ def start_simulation(
     num_clients : Optional[int]
         The total number of clients in this simulation. This must be set if
         `clients_ids` is not set and vice-versa.
-    clients_ids : Optional[List[str]]
+    clients_ids : OptionalUnion[List[int],List[str]]]
         List `client_id`s for each client. This is only required if
         `num_clients` is not set. Setting both `num_clients` and `clients_ids`
         with `len(clients_ids)` not equal to `num_clients` generates an error.
+        If list contains `str` values, they will be converted to `int`.
```
danieljanes, line 130: This implies a breaking change, right?

src/py/flwr/simulation/app.py
jafermarq: handled slightly different in latest commit

src/py/flwr/simulation/app.py
jafermarq: yes, as discussed. The PR title is updated to `break(...)`

src/py/flwr/simulation/ray_transport/ray_client_proxy.py
jafermarq: Done for a equivalent comment in #3697

src/py/flwr/simulation/app.py
```diff
@@ -87,6 +115,7 @@ def start_simulation(
     actor_type: Type[VirtualClientEngineActor] = ClientAppActor,
     actor_kwargs: Optional[Dict[str, Any]] = None,
     actor_scheduling: Union[str, NodeAffinitySchedulingStrategy] = "DEFAULT",
+    nodes_ids: Optional[List[int]] = None,
```
danieljanes, line 118: Should we move this ahead of `client_ids`?

src/py/flwr/simulation/app.py
```diff
@@ -157,7 +187,10 @@ def start_simulation(
         compute nodes (e.g. via NodeAffinitySchedulingStrategy). Please note this
         is an advanced feature. For all details, please refer to the Ray documentation:
         https://docs.ray.io/en/latest/ray-core/scheduling/index.html
-
+    nodes_ids : Optional[List[int]]
+        List node ids for each client. This is only required if
```
danieljanes, line 191:
```suggestion
        List node IDs for each client. This is only required if
```

src/py/flwr/simulation/app.py
jafermarq: I did that at first but then decided to put it at the end (just in case someone was relying on positional args when passing `clients_ids`)

jafermarq: removed this arg.

src/py/flwr/simulation/app.py
```diff
@@ -102,13 +120,13 @@ def start_simulation(
...
-    num_clients : Optional[int]
-        The total number of clients in this simulation. This must be set if
-        `clients_ids` is not set and vice-versa.
+    num_clients : int
+        The total number of clients in this simulation.
     clients_ids : Optional[List[str]]
         List `client_id`s for each client. This is only required if
         `num_clients` is not set. Setting both `num_clients` and `clients_ids`
         with `len(clients_ids)` not equal to `num_clients` generates an error.
+        Deprecated in favour of `node_ids` argument. This argument is deprecated.
```
danieljanes, line 129:
```suggestion
        This argument is deprecated.
```

src/py/flwr/simulation/app.py
```diff
@@ -102,13 +120,13 @@ def start_simulation(
         (model, dataset, hyperparameters, ...) should be (re-)created in either the
         call to `client_fn` or the call to any of the client methods (e.g., load
         evaluation data in the `evaluate` method itself).
-    num_clients : Optional[int]
-        The total number of clients in this simulation. This must be set if
-        `clients_ids` is not set and vice-versa.
+    num_clients : int
+        The total number of clients in this simulation.
     clients_ids : Optional[List[str]]
         List `client_id`s for each client. This is only required if
```
danieljanes, line 126:
```suggestion
        DEPRECATED. List `client_id`s for each client. This is only required if
```

danieljanes (approved):

PR #3700, opened 2024-06-30
feat(framework) Introduce `client_fn` adaptor

jafermarq:
We are transitioning towards a `client_fn` with signature:
```
def client_fn(node_id: int, partition_id: Optional[int]) -> Client:
    ...
```

but without introducing breaking changes to existing codebases. In order to do that we need an "adaptor" that is capable of detecting a `client_fn` in the previous formant (i.e. expecting a single `cid:str` argument) and wrap it into a function matching the new `client_fn` signature. This PR accomplishes that.

### DIFF
--- a/src/py/flwr/client/client_app.py
+++ b/src/py/flwr/client/client_app.py
@@ -29,15 +29,6 @@ from flwr.common.logger import warn_deprecated_feature, warn_preview_feature
 from .typing import ClientAppCallable
 
 
-class ClientAppException(Exception):
-    """Exception raised when an exception is raised while executing a ClientApp."""
-
-    def __init__(self, message: str):
-        ex_name = self.__class__.__name__
-        self.message = f"\nException {ex_name} occurred. Message: " + message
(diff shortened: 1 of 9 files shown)


src/py/flwr/client/client_app.py
```diff
@@ -15,19 +15,45 @@
...
+            node_id: int, partition_id: Optional[int]  # pylint: disable=unused-argument
+        ) -> Client:
+            return client_fn(str(partition_id))  # type: ignore
+
+    else:
+
+        def adaptor_fn(node_id: int, partition_id: Optional[int]) -> Client:
+            return client_fn(node_id, partition_id)  # type: ignore
+
+    return adaptor_fn
```
danieljanes, line 54:
```suggestion

        return adaptor_fn

    return client_fn
```

jafermarq: done. I updated the signature of this utility function. Also, the logic inside now uses a more robust method built with `inspect.signature`.

PR #3713, opened 2024-07-04
refactor(*:skip) Align Ubuntu image with Alpine image

Robert-Steiner:
### Description

- Align the order of the instructions to match the order of the Alpine Docker image.
- Fix the false positive CVE-2022-40897⁠. To add some context: Docker scout complains about `setuptools 65.5.0` in layer 8 but we upgrade the version to `69.5.1` in layer 16 so the vulnerability is not in the final image. The issue is fixed by installing/upgrading setuptools in the first stage of the multi-stage build and then only copy the virtual env to the final stage.
- Reduce one layer in the final image.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


Moep90 (approved):

tanertopal (approved):
lgtm!

PR #3718, opened 2024-07-05
fix(framework:skip) Remove setting `init_args` for Backend in  `run_simulation` entry point

jafermarq:
In #3543 , we made it possible to easily pass `init_args` to the Backend. These are designed to be backend-agnostic. In the case of the `RayBackend` the `log_to_driver` setting can be use to "silence" all logs coming from ray actors (e.g. the `ClientApp` objs. This is useful in some settings.

The change in the PR removes setting this to True (it's True by default anyway in Ray), allowing users to changing it to false via the `backend_config` argument if desired.


jafermarq:
Made redundant by #3963

PR #3720, opened 2024-07-05
feat(framework) Add SuperExec `--executor-config`

charlesbvll:
### Description

We currently have no way to provide a config to the `flower-superexec` CLI.

### Related issues/PRs

This current PR is blocked by #3732 

### Explanation

Add a new CLI option to provide a config dict and pass it to the `Executor` plugin.

### Checklist

- [x] Make CI checks pass

### Any other comments?

N/A


src/py/flwr/superexec/app.py
```diff
@@ -89,6 +91,10 @@ def _parse_args_run_superexec() -> argparse.ArgumentParser:
         help="The directory for the executor.",
         default=".",
     )
+    parser.add_argument(
+        "--config",
```
danieljanes, line 95: Should we call this `--executor-config`?

src/py/flwr/superexec/deployment.py
```diff
@@ -33,21 +34,37 @@
...
-    """Deployment engine executor."""
+    """Deployment engine executor.
+
+    superlink: str (Default: "0.0.0.0:9091")
+        Address of the `SuperLink` to connect to.
+    root_certificates: Optional[str] (Default: None)
+        Specifies the path to the PEM-encoded root certificate file for
+        establishing secure HTTPS connections.
+    flwr_dir: Optional[str] (Default: None)
+        The path containing installed Flower Apps.
```
danieljanes, line 45: The parameters header is missing here

src/py/flwr/superexec/deployment.py
```diff
@@ -74,9 +91,22 @@ def _create_run(
 
     @override
     def start_run(
-        self, fab_file: bytes, override_config: Dict[str, str]
+        self,
+        fab_file: bytes,
+        override_config: Dict[str, str],
+        config: Optional[Dict[str, str]],
```
danieljanes, line 97: The config should not be passed on every run, but only once right after loading the executor plugin and before executing any runs. We should probably define a new `Executor` method for this.

charlesbvll: Good point, I'll update the PR

src/py/flwr/superexec/deployment.py
```diff
@@ -33,21 +34,37 @@
...
+        The path containing installed Flower Apps.
+    """
 
     def __init__(
         self,
-        address: str = DEFAULT_SERVER_ADDRESS_DRIVER,
-        root_certificates: Optional[bytes] = None,
+        superlink: str = DEFAULT_SERVER_ADDRESS_DRIVER,
+        cert_path: Optional[str] = None,
+        flwr_dir: Optional[str] = None,
```
danieljanes, line 52: We could also think about a constructor that doesn't take any arguments, and defer setting these values until a new `Executor` method (`.configure(self, executor_config: Dict[...])`?) is called that takes the parsed `--executor-config` as an argument

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -35,9 +35,10 @@
 class ExecServicer(exec_pb2_grpc.ExecServicer):
     """SuperExec API servicer."""
 
-    def __init__(self, executor: Executor) -> None:
+    def __init__(self, executor: Executor, config: Optional[Dict[str, str]]) -> None:
```
danieljanes, line 38: If we don't pass the config to `start_run` any more, this won't be necessary (b/c `executor` already contains the config)

src/py/flwr/superexec/executor.py
```diff
@@ -33,7 +33,10 @@ class Executor(ABC):
 
     @abstractmethod
     def start_run(
-        self, fab_file: bytes, override_config: Dict[str, str]
+        self,
+        fab_file: bytes,
+        override_config: Dict[str, str],
+        config: Optional[Dict[str, str]],
```
danieljanes, line 39: Instead of this, I would argue for a new `Executor` method (as mentioned above)

src/py/flwr/superexec/app.py
charlesbvll: Wouldn't it be too verbose for users?

src/py/flwr/superexec/executor.py
```diff
@@ -31,9 +31,24 @@ class RunTracker:
...
+    def set_config(
+        self,
+        config: Optional[Dict[str, str]],
+    ) -> None:
+        """Register provided config as class attributes.
+
+        Parameters
+        ----------
+        config : Optional[Dict[str, str]]
+            A dictionary for configuration values.
```
danieljanes, line 44:
```suggestion
        config: Dict[str, str],
    ) -> None:
        """Register provided config as class attributes.

        Parameters
        ----------
        config : Dict[str, str]
            A dictionary for configuration values.
```

src/py/flwr/superexec/executor.py
```diff
@@ -44,6 +59,8 @@ def start_run(
         ----------
         fab_file : bytes
             The Flower App Bundle file bytes.
+        override_config: Dict[str, str]
+            The config overrides Dict sent by the user (using `flwr run`).
```
danieljanes, line 63:
```suggestion
            The config overrides dict sent by the user (using `flwr run`).
```

src/py/flwr/superexec/exec_servicer_test.py
```diff
@@ -44,7 +44,7 @@ def test_start_run() -> None:
     request.fab_file = b"test"
 
     # Create a instance of FlowerServiceServicer
-    servicer = ExecServicer(executor=executor)
+    servicer = ExecServicer(executor=executor, config=None)
```
danieljanes, line 47: This is not necessary now that we have `set_config`

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -35,8 +35,9 @@
...
         self.executor = executor
+        self.executor.set_config(config)
```
danieljanes, line 40: We shouldn't do this in the servicer, but instead pass an already-configured `executor` to the servicer

src/py/flwr/superexec/exec_grpc.py
```diff
@@ -32,10 +32,12 @@ def run_superexec_api_grpc(
     address: str,
     executor: Executor,
     certificates: Optional[Tuple[bytes, bytes, bytes]],
+    config: Optional[Dict[str, str]],
 ) -> grpc.Server:
     """Run SuperExec API (gRPC, request-response)."""
     exec_servicer: grpc.Server = ExecServicer(
         executor=executor,
+        config=config,
```
danieljanes, line 40: We should configure the `executor` before passing it to the `ExecServicer`

src/py/flwr/superexec/exec_servicer_test.py
charlesbvll: In the servicer we still need it, because it is there that we will call the `set_config` method

src/py/flwr/superexec/app.py
danieljanes: I'd prefer it for the sake of clarity / to avoid ambiguity

src/py/flwr/superexec/deployment.py
```diff
@@ -33,21 +34,55 @@
...
+        self.flwr_dir = flwr_dir
         self.stub: Optional[DriverStub] = None
 
+    @override
+    def set_config(
+        self,
+        config: Dict[str, str],
+    ) -> None:
+        """Update config arguments."""
+        if config:
```
danieljanes, line 72: This can be removed

src/py/flwr/superexec/app.py
charlesbvll: I see, should we also provide an abbreviated option?

src/py/flwr/superexec/deployment.py
```diff
@@ -33,21 +34,55 @@
...
+    flwr_dir: Optional[str] (Default: None)
+        The path containing installed Flower Apps.
+    """
 
     def __init__(
         self,
-        address: str = DEFAULT_SERVER_ADDRESS_DRIVER,
-        root_certificates: Optional[bytes] = None,
+        superlink: str = DEFAULT_SERVER_ADDRESS_DRIVER,
+        cert_path: Optional[str] = None,
```
danieljanes, line 53:
```suggestion
        root_certificates_path: Optional[str] = None,
```

src/py/flwr/superexec/app.py
danieljanes: Yes, we can do that

danieljanes (approved):

PR #3732, opened 2024-07-07
feat(framework) Add utility functions for config parsing and handling

charlesbvll:
### Description

We need some functions to parse a config arguments to a flattened config dict.

### Related issues/PRs

The current PR is need to unblock #3720 

### Explanation

- Introduce a `flatten_dict` function to flatten nested dictionaries and a `parse_config_args` function that parses a comma separated list of key values pairs separated by '=' to a single flat dictionary.

- Add corresponding tests

### Checklist

- [x] Write tests
- [x] Make CI checks pass

### Any other comments?

N/A


src/py/flwr/common/config.py
```diff
@@ -71,3 +71,44 @@ def get_project_config(project_dir: Union[str, Path]) -> Dict[str, Any]:
...
+    return dict(items)
+
+
+def parse_config_args(
+    config_overrides: Optional[str],
+) -> Dict[str, str]:
+    """Parse comma separated list of key-value pairs separated by '='."""
+    overrides: Dict[str, str] = {}
+
+    if config_overrides is not None:
```
danieljanes, line 100:
```suggestion
    if config_overrides is None:
        return overrides
```
...and then unindent the rest of the function

src/py/flwr/common/config.py
```diff
@@ -71,3 +71,46 @@ def get_project_config(project_dir: Union[str, Path]) -> Dict[str, Any]:
...
+            items.append((new_key, v))
+        else:
+            raise ValueError(
+                f"The value for {k} needs to be a string.",
+            )
+    return dict(items)
+
+
+def parse_config_args(
+    config_overrides: Optional[str],
```
danieljanes, line 95: Shouldn't this take a `sep` argument, just like `flatten_dict`?

src/py/flwr/common/config.py
```diff
@@ -71,3 +71,46 @@ def get_project_config(project_dir: Union[str, Path]) -> Dict[str, Any]:
         )
 
     return config
+
+
+def flatten_dict(
+    raw_dict: Dict[str, Any], parent_key: str = "", sep: str = "."
```
danieljanes, line 77: Let's spell this out:
```suggestion
    raw_dict: Dict[str, Any], parent_key: str = "", separator: str = "."
```

charlesbvll: @danieljanes I think for this function we could even remove the separator argument as I feel like we always want it to be a dot, but lmk wdyt

src/py/flwr/common/config.py
```diff
@@ -71,3 +71,49 @@ def get_project_config(project_dir: Union[str, Path]) -> Dict[str, Any]:
...
+        new_key = f"{parent_key}{separator}{k}" if parent_key else k
+        if isinstance(v, dict):
+            items.extend(
+                flatten_dict(v, parent_key=new_key, separator=separator).items()
+            )
+        elif isinstance(v, str):
+            items.append((new_key, v))
+        else:
+            raise ValueError(
+                f"The value for {k} needs to be a string.",
```
danieljanes, line 91:
```suggestion
                f"The value for key {k} needs to be a `str` or a `dict`.",
```

src/py/flwr/common/config_test.py
```diff
@@ -0,0 +1,164 @@
...
+
+        [flower.components]
+        serverapp = "fedgpt.server:app"
+        clientapp = "fedgpt.client:app"
+
+        [flower.engine]
+        name = "simulation" # optional
+
+        [flower.engine.simulation.supernode]
+        count = 10 # optional
```
danieljanes, line 106: These should start with `flower.config`

charlesbvll: @danieljanes Well I've added tests for this function because it didn't have any before, but this is not really the scope of the PR. So this function doesn't do anything particular with `flower.config`

charlesbvll: I updated it

danieljanes (dismissed):

src/py/flwr/common/config.py
danieljanes: @charlesbvll let's remove it. If we make the separator configurable then we also have to ensure that the separator character is not part of any of the key names.

danieljanes (approved):

PR #3747, opened 2024-07-08
Add flower-simulation e2e tests

danielnugraha:
### Description

The new Flower Next is using `flower-simulation` CLI command instead of `start_simulation` function call, but it's not reflected in e2e tests.

### Explanation

Add `flower-simulation` e2e tests in test-simulation.sh.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


e2e/framework-pandas/client.py
```diff
@@ -5,7 +5,7 @@
 
 import flwr as fl
 
-df = pd.read_csv("./data/client.csv")
```
tanertopal, line 8: I think this path should be relative as the test otherwise probably breaks locally.

danielnugraha: It does not work with relative path for now. I talked with @jafermarq about this, for now this is the way to make it pass.

danielnugraha: Dataset is moved to tmp.

PR #3763, opened 2024-07-10
fix(framework:skip) Raise exception if `ServerApp` thread fails 

jafermarq:
don't merge. for demonstration purposes only.


danieljanes:
Changes merged in https://github.com/adap/flower/pull/3764

PR #3776, opened 2024-07-11
feat(framework) Introduce a new deprecation function that warns users and provides an code example

jafermarq:
Using the standard `warn_deprecated_feature` results in this:
Screenshot 2024-07-11 at 16 24 35

The newly introduced `warn_deprecated_feature_with_example` would look like this:
Screenshot 2024-07-11 at 16 44 49

### DIFF
--- a/src/py/flwr/common/logger.py
+++ b/src/py/flwr/common/logger.py
@@ -200,8 +200,7 @@ def warn_deprecated_feature(name: str) -> None:
 def warn_deprecated_feature_with_example(
     deprecation_message: str, example_message: str, code_example: str
 ) -> None:
-    """Warn if a feature is deprecated and show code example of preferred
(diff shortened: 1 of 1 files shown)


src/py/flwr/common/logger.py
```diff
@@ -197,6 +197,32 @@ def warn_deprecated_feature(name: str) -> None:
     )
 
 
+def warn_deprecated_feature_with_example(
+    deprecation_message: str, example_message: str, code_example: str
+) -> None:
+    """Warn if a feature is deprecated and show code example of preferred
+    alternative."""
```
danieljanes, line 204:
```suggestion
    """Warn if a feature is deprecated and show code example."""
```

danieljanes (approved):

PR #3780, opened 2024-07-11
feat(framework) Capture `node_id`/`node_config` in `Context` via `NodeState`

jafermarq:
Changes in this PR:
- Register `node_id` and `node_config` to the `NodeState`
- The `NodeState` will register a `Context` passing its `node_id` and `node_config`
- For `grpc_rere` the `create_node()` function returns the node_id returned by the `CreateNodeResponse`. This is used to set the `<NodeState>.node_id`.
- Updated places in the code base that directly initialise a `Context`

### DIFF
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -336,7 +336,9 @@ def _start_client_internal(
 
             # Register node
             if create_node is not None:
-                node_id = create_node()  # pylint: disable=not-callable
+                node_id = (  # pylint: disable=assignment-from-none
+                    create_node()
+                )  # pylint: disable=not-callable
                 if transport in ["grpc-rere", None]:
                     node_state.node_id = node_id  # type: ignore
 
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -319,8 +319,8 @@ def _start_client_internal(
         on_backoff=_on_backoff,
     )
 
-    # Empty dict for now
-    # This will be remove once users can pass node_config via flower-supernode
+    # Empty dict (for now)
+    # This will be removed once users can pass node_config via flower-supernode
     node_config: Dict[str, str] = {}
 
     # NodeState gets initialized when the first connection is established
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -319,11 +319,13 @@ def _start_client_internal(
         on_backoff=_on_backoff,
     )
 
-    node_state_set = False
     node_config: Dict[str, str] = {}
+
+    node_state_set = False
     node_state = NodeState(
         node_id=-1, node_config=node_config, partition_id=partition_id
     )
+
     runs: Dict[int, Run] = {}
 
     while not app_state_tracker.interrupt:
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -319,14 +319,11 @@ def _start_client_internal(
         on_backoff=_on_backoff,
     )
 
-    def _create_node_state(node_id: int, node_config: Dict[str, str]) -> NodeState:
-        """Create NodeState after node has received a node-id."""
-        return NodeState(
-            node_id=node_id, node_config=node_config, partition_id=partition_id
-        )
-
-    node_state: NodeState = None  # type: ignore
+    node_state_set = False
     node_config: Dict[str, str] = {}
+    node_state = NodeState(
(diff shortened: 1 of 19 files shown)


src/py/flwr/common/context.py
```diff
@@ -27,6 +27,11 @@ class Context:
 
     Parameters
     ----------
+    node_id : int
+        A UUID that identifies the node.
```
danieljanes, line 31:
```suggestion
        The ID that identifies the node.
```

src/py/flwr/common/context.py
```diff
@@ -27,6 +27,11 @@ class Context:
...
+    node_config : Dict[str, str]
+        A config (key/value mapping) unique to the node and independent of the
+        `run_config`. This config persist across runs this node participates in.
```
danieljanes, line 34:
```suggestion
        `run_config`. This config persists across all runs this node participates in.
```

src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -79,7 +79,7 @@ def grpc_request_response(  # pylint: disable=R0913, R0914, R0915
     Tuple[
         Callable[[], Optional[Message]],
         Callable[[Message], None],
-        Optional[Callable[[], None]],
+        Optional[Callable[[], int]],
```
danieljanes, line 82:
```suggestion
        Optional[Callable[[], Optional[int]]],
```

src/py/flwr/client/grpc_adapter_client/connection.py
```diff
@@ -44,7 +44,7 @@ def grpc_adapter(  # pylint: disable=R0913
     Tuple[
         Callable[[], Optional[Message]],
         Callable[[Message], None],
-        Optional[Callable[[], None]],
+        Optional[Callable[[], int]],
```
danieljanes, line 47: This type must be the same for all connection types:
```suggestion
        Optional[Callable[[], Optional[int]]],
```

src/py/flwr/client/app.py
```diff
@@ -319,7 +319,7 @@ def _on_backoff(retry_state: RetryState) -> None:
         on_backoff=_on_backoff,
     )
 
-    node_state = NodeState(partition_id=partition_id)
+    node_state = NodeState(node_id=-1, node_config={}, partition_id=partition_id)
```
danieljanes, line 322: `-1` is a possible `node_id`. we might need to find a different solution here. with `grpc-rere` and `rest`, the `node_id` should not change and we should not call `create_node()` on subsequent connects. we can lazily instantiate `NodeState` the first time `create_node` gets called, and then call `create_node` never again. the fact that `NodeState` was initialized can actually act as a flag to determine whether we should call `create_node`.

src/py/flwr/simulation/ray_transport/ray_client_proxy_test.py
```diff
@@ -218,7 +218,13 @@ def _load_app() -> ClientApp:
                 _load_app,
                 message,
                 str(node_id),
-                Context(state=RecordSet(), run_config={}, partition_id=node_id),
+                Context(
+                    node_id=0,
+                    node_config={},
```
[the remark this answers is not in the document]
jafermarq: I'll update this in a later PR that also includes changes to that file regarding changing `client_fn` signature

src/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -176,7 +176,7 @@ def ping() -> None:
         if not ping_stop_event.is_set():
             ping_stop_event.wait(next_interval)
 
-    def create_node() -> None:
+    def create_node() -> int:
```
danieljanes, line 179:
```suggestion
    def create_node() -> Optional[int]:
```

src/py/flwr/client/app.py
```diff
@@ -319,7 +319,13 @@ def _on_backoff(retry_state: RetryState) -> None:
         on_backoff=_on_backoff,
     )
 
-    node_state = NodeState(partition_id=partition_id)
+    # Empty dict for now
+    # This will be remove once users can pass node_config via flower-supernode
```
danieljanes, line 323:
```suggestion
    # Empty dict (for now)
    # This will be removed once users can pass node_config via flower-supernode
```

danieljanes (approved):

danieljanes (approved):

PR #3784, opened 2024-07-12
feat(framework) Docker Compose all-in-one

Robert-Steiner:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/docker/complete/with-tls.yml
```diff
@@ -0,0 +1,62 @@
...
+      - certificates/ca.crt
+    secrets:
+      - source: superlink-ca-certfile
+        target: /app/certificates/ca.crt
+
+  supernode-2: *supernode
+
+  # uncomment to enable TLS on another supernode
+  #
+  # supernode-3: *supernode
```
tanertopal, line 48: @Robert-Steiner Why is there are difference in `supernode-1` and `supernode-2`?

In `compose.yml` the pattern is that we have 1 and 2 and can uncomment 3. Can we have the same pattern here? Maybe I am misreading the way its written and that's already that way here.

Robert-Steiner: Docker compose supports [YAML anchors](https://docs.docker.com/reference/compose-file/fragments/#example-2). With `&supernode`, we create an anchor, and the following content can be referenced by writing `*supernode`. it reduces code duplication. In `compose.yml` we can use it because we each node has a different `partition-id`.  

But if it is hard to understand, I can also use the same pattern as in the `compose.yml` file.

tanertopal: @Robert-Steiner Now knowing it seems simple but reusing the same pattern might make it easier to understand for non-experienced readers so I'd suggest to do so.

tanertopal (approved):
Lgtm!

PR #3787, opened 2024-07-12
refactor(framework) Refactor NodeState handling

danieljanes:


src/py/flwr/client/app.py
```diff
@@ -340,26 +340,33 @@ def _on_backoff(retry_state: RetryState) -> None:
...
-                else:
-                    print(transport)
-                    if transport in ["grpc-bidi", None]:
-                        # gRPC-bidi doesn't have the concept of node_id
-                        node_id = -1
-                    else:
+            # Register node when connecting the first time
+            if node_state is None:
+                if create_node is None:
+                    if transport != "grpc-bidi":
```
jafermarq, line 346:
```suggestion
                    if transport not in ["grpc-bidi", None]:
```

src/py/flwr/client/app.py
```diff
@@ -400,6 +407,10 @@ def _on_backoff(retry_state: RetryState) -> None:
                         else:
                             runs[run_id] = Run(run_id, "", "", {})
 
+                    # NodeState must be initialized at this point
+                    if node_state is None:
+                        raise ValueError("NodeState is not initialized")
```
jafermarq, line 412: It seems this is ever going to be hit? because there will be a `ValueError` earlier in the code if `node_id` is `None` or if the transport doesn't have such concept.

src/py/flwr/client/app.py
```diff
@@ -340,26 +340,33 @@ def _on_backoff(retry_state: RetryState) -> None:
...
-                        # gRPC-bidi doesn't have the concept of node_id
-                        node_id = -1
-                    else:
+            # Register node when connecting the first time
+            if node_state is None:
+                if create_node is None:
+                    if transport != "grpc-bidi":
                         raise NotImplementedError(
                             "There is no transport other than grpc-bidi that "
                             "does not implement `create_node()`.'"
```
jafermarq, line 349:
```suggestion
                            "All transports except `grpc-bidi` require "
                            "an implementation for `create_node()`.'"
```

src/py/flwr/client/app.py
danieljanes: True. We could remove it. Not sure if the linters would allow it, though.

Issue #3788, opened 2024-07-12
Add Flower Baseline: FedRep


KarhouTam:
### Paper

Exploiting Shared Representations for Personalized Federated Learning

### Link

http://arxiv.org/abs/2102.07078

### Maybe give motivations about why the paper should be implemented as a baseline.

FedRep (500+ citations) is quite famous in pFL field and is considered as one of the most common baselines that introduced by many pFL papers (include [mine](https://arxiv.org/abs/2305.15706)).

### Is there something else you want to add?

No for now.

### For first time contributors

- [x]  Read the [`first contribution` doc](https://flower.ai/docs/first-time-contributors.html)
- [x]  Complete the Flower tutorial
- [x]  Read the Flower Baselines docs to get an overview:
    - [x]  [How to use Flower Baselines](https://flower.ai/docs/baselines/how-to-use-baselines.html)
    - [x]  [How to contribute a Flower Baseline](https://flower.ai/docs/baselines/how-to-contribute-baselines.html)

### Prepare - understand the scope

- [x] Read the paper linked above
- [x] Decide which experiments you'd like to reproduce. The more the better!
- [x] Follow the steps outlined in [Add a new Flower Baseline](https://flower.ai/docs/baselines/how-to-contribute-baselines.html#add-a-new-flower-baseline).
- [x] You can use as reference [other baselines](https://github.com/adap/flower/tree/main/baselines) that the community merged following those steps.

### Verify your implementation

- [x] Follow the steps indicated in the `EXTENDED_README.md` that was created in your baseline directory
- [x] Ensure your code reproduces the results for the experiments you chose
- [x] Ensure your `README.md` is ready to be run by someone that is no familiar with your code. Are all step-by-step instructions clear?
- [x] Ensure running the formatting and typing tests for your baseline runs without errors.
- [x] Clone your repo on a new directory, follow the guide on your own `README.md` and verify everything runs.

KarhouTam:
I've checked the issues and PRs with keyword fedrep and nothing is there. Also I'm already start the reproduction. So I think this issue is more like a placeholder.

jafermarq:
Hi @KarhouTam , I'm closing this issue as #3790 just got merged 🙌

PR #3790, opened 2024-07-12
feat(baselines) Add FedRep Baseline

KarhouTam:
### Description
Add baseline FedRep to flower-baseline.
Reproduce experiments on CIFAR-10 and CIFAR-100. 

### Related issues/PRs
Related issue #3788 

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


baselines/fedrep/README.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Exploiting Shared Representations for Personalized Federated Learning
+url: [arXiv](http://arxiv.org/abs/2102.07078)
+labels: [image classification, label heterogeneity, personalized federated learning] # please add between 4 and 10 single-word (maybe two-words) labels (e.g. system heterogeneity, image classification, asynchronous, weight sharing, cross-silo). Do not use ""
+dataset: [CIFAR-10, CIFAR-100] # list of datasets you include in your baseline. Do not use ""
```
jafermarq, line 5:
```suggestion
url: http://arxiv.org/abs/2102.07078
labels: [image classification, label heterogeneity, personalized federated learning]
dataset: [CIFAR-10, CIFAR-100]
```

baselines/fedrep/fedrep/conf/base.yaml
```diff
@@ -0,0 +1,45 @@
...
+batch_size: 50
+num_rounds: 100
+learning_rate: 0.01
+momentum: 0.5
+algorithm: fedrep
+model_name: cnncifar10
+
+client_resources:
+  num_cpus: 4
+  num_gpus: 1
```
jafermarq, line 16: I wonder if we could increase the degree of parallelism by lowering these. For example, even running 5x clients in parallel (done by lowering `num_gpus: 0.2` and then having a low enough `num_cpus` -- e.g set to 2 or 1 should do the trick) it requires just 2.6GB of VRAM

The same comment applies to the other configs

baselines/fedrep/fedrep/dataset_preparation.py
```diff
@@ -0,0 +1,201 @@
...
+    """Call the dataset."""
+    if dataset_name == "cifar10":
+        return CIFAR10(root, **kwargs)
+    if dataset_name == "cifar100":
+        return CIFAR100(root, **kwargs)
+    raise ValueError(f"Dataset {dataset_name} not supported.")
+
+
+# pylint: disable=R0914
+def randomly_assign_classes(
```
jafermarq, line 113: This sounds to me like a pathological partitioning. Am I right? Could we replace this function (and likely parts o `dataset.dataset_main()`) with the [Flower Datasets](https://flower.ai/docs/datasets/ref-api/flwr_datasets.partitioner.PathologicalPartitioner.html#flwr_datasets.partitioner.PathologicalPartitioner) was of using the `PathologicalPartitioner`? 

Please take a look, I think `flwr-dataset` is quite powerful. Some of the examples (e.g.[examples/quickstart-pytorch](https://github.com/adap/flower/tree/main/examples/quickstart-pytorch)) show how to integrate them. You could follow a similar style as in the `client_fn` shown in that example.

baselines/fedrep/fedrep/client.py
```diff
@@ -0,0 +1,296 @@
...
+    # pylint: disable=E1101
+    def client_fn(cid: str) -> Client:
+        """Create a Flower client representing a single organization."""
+        cid_use = int(cid)
+
+        trainset = Subset(dataset, indices=data_indices[cid_use]["train"])
+        testset = Subset(dataset, indices=data_indices[cid_use]["test"])
+
+        # Create the train loader
+        trainloader = DataLoader(trainset, config.batch_size, shuffle=True)
```
jafermarq, line 262: I'm seeing some `ValueError` thrown here:
```
  File "/home/javier/projects/flower-karhoutam/baselines/fedrep/fedrep/client.py", line 262, in client_fn
    trainloader = DataLoader(trainset, config.batch_size, shuffle=True)
  File "/home/javier/.cache/pypoetry/virtualenvs/fedrep-qLhlcELz-py3.10/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 350, in __init__
    sampler = RandomSampler(dataset, generator=generator)  # type: ignore[arg-type]
  File "/home/javier/.cache/pypoetry/virtualenvs/fedrep-qLhlcELz-py3.10/lib/python3.10/site-packages/torch/utils/data/sampler.py", line 143, in __init__
    raise ValueError(f"num_samples should be a positive integer value, but got num_samples={self.num_samples}")
ValueError: num_samples should be a positive integer value, but got num_samples=0
```

baselines/fedrep/fedrep/dataset_preparation.py
jafermarq: We are in the process of updating all of our examples to follow the same formate as `quickstart-pytorch`, `quickstart-tensorflow` and `quickstart-mlx` currently show: they run with `flwr run`. And a bit later we'll update the baselines too. All baselines will be updated to use `flwr-datasets` 👼

baselines/fedrep/fedrep/models.py
jafermarq, line 1: Since in most baselines the actual models are implemented in `models.py` but in your case this python file seems to be more for building the abstract base classes, how about if we rename this file to `base_models.py` (or something like that). Then, we can have a single `models.py` with `CNNCifar10ModelManager` and `CNNCifar100ModelManager` in it. What do you think?

jafermarq (commented):
Hi @KarhouTam, many thanks for taking the `FedPer` baseline to a ready state. I was able to follow your `REAMDE.md` and got similar results to one of the `CIFAR-100` experiments.

I've added a couple of comments below, nothing major (is not a full in depth review), but let me know what your thoughts are. Simplifying the dataset partitioning and preparation would be good to update to `flwr-dataset`; it should be easy with the built-in [PathologicalPartitioner](https://flower.ai/docs/datasets/ref-api/flwr_datasets.partitioner.PathologicalPartitioner.html#flwr_datasets.partitioner.PathologicalPartitioner) which looks like the type of partitioning you are doing (but please confirm in case i miss something).

baselines/fedrep/fedrep/conf/base.yaml
KarhouTam: Sounds interesting. I will try this out later today. Since the model used is small, this parallelism should be reasonable.

baselines/fedrep/fedrep/dataset_preparation.py
KarhouTam: Sure. I will try `flwr-dataset` later this week and think about the refactor. It may takes some times. 
Hope you don' mind.

KarhouTam: Also, please consider review my another PR #3663 .
Happy to see any comments and suggestions on that.

baselines/fedrep/fedrep/client.py
KarhouTam: Seems this bug is raised by the 0 length of `trainset.indices` and `DataLoader(shuffle=True)`.  
Nothing wrong on my workspace... 😐
Could you provide the commands for me to reproduce this bug?

baselines/fedrep/fedrep/models.py
KarhouTam: Sure, I will consider it.

KarhouTam:
Hi, @jafermarq .
I've done all requests and this PR is ready again.

baselines/fedrep/fedrep/client.py
```diff
@@ -0,0 +1,307 @@
...
+            client_id: The client ID.
+            trainloader: Client train data loader.
+            testloader: Client test data loader.
+            config: dictionary containing the client configurations.
+            model_manager_class: class to be used as the model manager.
+            client_state_save_path: Path for saving model head parameters.
+                (Just for FedRep). Defaults to "".
+        """
+        super().__init__()
+
```
jafermarq, line 52:
```suggestion

        # Dynamically determine device based on what the simulation engine exposes
        # to this client.
        self.device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
```

jafermarq: Currently, the simulated clients run via Ray actors (some python processes) and clients get to see the "GPU" if Ray actually exposes that resource to the processes running these clients. Because of this, we can dynamically "discover" the device and use it throughout. See my other suggestions a few lines below and in the model manager constructor.

baselines/fedrep/fedrep/client.py
```diff
@@ -0,0 +1,307 @@
...
+        self.client_id = client_id
+        self.client_state_save_path = (
+            (client_state_save_path + f"/client_{self.client_id}")
+            if client_state_save_path != ""
+            else None
+        )
+        self.model_manager = model_manager_class(
+            client_id=self.client_id,
+            config=config,
+            trainloader=trainloader,
```
jafermarq, line 62:
```suggestion
            device=self.device,
            trainloader=trainloader,
```

baselines/fedrep/fedrep/client.py
jafermarq: feel free to adjust if you see a better way of implementing this. Usually we add this `torch.device(...)` either in the body `client_fn()` or in the client constructor itself.

baselines/fedrep/fedrep/base_model.py
```diff
@@ -0,0 +1,288 @@
...
+
+# pylint: disable=R0902, R0913, R0801
+class ModelManager(ABC):
+    """Manager for models with Body/Head split."""
+
+    def __init__(
+        self,
+        client_id: int,
+        config: DictConfig,
+        trainloader: DataLoader,
```
jafermarq, line 136:
```suggestion
        device,
        trainloader: DataLoader,
```

jafermarq: this relates to my comment in `client.py`

baselines/fedrep/fedrep/base_model.py
```diff
@@ -0,0 +1,288 @@
...
+            client_save_path: Path to save the client model head state.
+            model_split_class: Class to be used to split the model into body and head \
+                (concrete implementation of ModelSplit).
+        """
+        super().__init__()
+        self.config = config
+        self.client_id = client_id
+        self.trainloader = trainloader
+        self.testloader = testloader
+        self.device = self.config.server_device
```
jafermarq, line 157:
```suggestion
        self.device = device
```

baselines/fedrep/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+| Description         | Default Value                   |
+| ------------------- | ------------------------------- |
+| num_clients         | 100                             |
+| num_rounds          | 100                             |
+| num_local_epochs    | 5                               |
+| num_rep_epochs      | 1                               |
+| enable_finetune     | False                           |
+| num_finetune_epochs | 5                               |
+| client resources    | {'num_cpus': 4, 'num_gpus': 1 } |
```
jafermarq, line 53:
```suggestion
| client resources    | {'num_cpus': 2, 'num_gpus': 0.5} |
```

baselines/fedrep/fedrep/client.py
```diff
@@ -0,0 +1,307 @@
...
+    test_data_transform = transforms.Compose(
+        [
+            transforms.ToTensor(),
+            transforms.Normalize(MEAN[config.dataset.name], STD[config.dataset.name]),
+        ]
+    )
+
+    partitioner = PathologicalPartitioner(
+        num_partitions=config.num_clients,
+        partition_by="label",
```
jafermarq, line 240:
```suggestion

    use_fine_label = False
    if config.model_name.lower() == "cnncifar100":
        use_fine_label = True

    partitioner = PathologicalPartitioner(
        num_partitions=config.num_clients,
        partition_by="fine_label" if use_fine_label else "label",
```

baselines/fedrep/fedrep/client.py
```diff
@@ -0,0 +1,307 @@
...
+    )
+    if FEDERATED_DATASET is None:
+        FEDERATED_DATASET = FederatedDataset(
+            dataset=config.dataset.name.lower(), partitioners={"train": partitioner}
+        )
+
+    def apply_train_transforms(batch):
+        """Apply transforms for train data to the partition from FederatedDataset."""
+        batch["img"] = [train_data_transform(img) for img in batch["img"]]
+        return batch
```
jafermarq, line 254:
```suggestion
        batch["img"] = [train_data_transform(img) for img in batch["img"]]
        if use_fine_label:
            batch["label"] = batch["fine_label"]
        return batch
```

baselines/fedrep/fedrep/client.py
```diff
@@ -0,0 +1,307 @@
...
+
+    def apply_test_transforms(batch):
+        """Apply transforms for test data to the partition from FederatedDataset."""
+        batch["img"] = [test_data_transform(img) for img in batch["img"]]
+        return batch
```
jafermarq, line 259:
```suggestion
        batch["img"] = [test_data_transform(img) for img in batch["img"]]
        if use_fine_label:
            batch["label"] = batch["fine_label"]
        return batch
```

jafermarq (commented):
Hi @KarhouTam , this looks pretty good. Many thanks for applying the changes and for making the switch to [Flower Datasets](https://flower.ai/docs/datasets/)! 

I'm running the experiments now and I'll update this thread, likely tomorrow. I'm running the experiments incorporating the changes i suggest. Likely the most important one is the one related to the `fine_labels` column for the `CIFAR-100` dataset.

Just one other small note: I added a few comments suggesting to not use the `<config>.server_device` to set the device of the model and instead "discover" the device that's available to the clients upon client construction. This more closely resembles other baselines using PyTorch (and examples in our repo). It should be an easy change, i added some suggestions below.

Let me know what you think!

baselines/fedrep/README.md
```diff
@@ -0,0 +1,125 @@
...
+python -m fedrep.main --config-name cifar100_100_5 algorithm=fedrep 
+python -m fedrep.main --config-name cifar100_100_5 algorithm=fedavg 
+```
+<img src="_static/cifar100_100_5.png" width="400"/> 
+
+### CIFAR-100 (100, 20)
+
+```
+python -m fedrep.main --config-name cifar100_100_5 algorithm=fedrep 
+python -m fedrep.main --config-name cifar100_100_5 algorithm=fedavg
```
jafermarq, line 123:
```suggestion
python -m fedrep.main --config-name cifar100_100_20 algorithm=fedrep 
python -m fedrep.main --config-name cifar100_100_20 algorithm=fedavg 
```

KarhouTam:
Hi, @jafermarq.
I've fixed the code by your suggestions.

About choosing the device, I added a function called `get_device` in `base_model.py` (it was in `utils.py` at the beginning, but caused a circular import issue). I let the user choose whether to use CUDA or not first, and then let the user decide the target GPU (in a multi-GPU environment). The user can ignore the `specified_device`, which outputs `torch.device("cuda")`.

KarhouTam:
Hi, @jafermarq.
Is this PR still having problems? Because I found that this PR is still not merged but no new questions or reviews have been given.

jafermarq:
Hi @KarhouTam, sorry for the late update. when i ran the experiments after my last review i wasn't able to get results similar to those in the readme. This was surprising to me because when I first ran your baseline (in commit https://github.com/adap/flower/pull/3790/commits/b73bd6be16663f941555a5e043c3e11c01e5a8c6) I remember the results looked good. So, today i re-run the experiments as currently in your branch and in the commit i mentioned. It seems some results are a bit lower (https://drive.google.com/drive/folders/1NhXfx785N32wTI-ASUBpvpev8STjt6wZ?usp=sharing), for example `cifar_10_100_5` (`FedRep`) doesn't reach 60% (so ~10% below than what it was before) -- note that the code as first reviewed was able to reach pass 60% (see the results in the `first_review` subdirectory in the google drive link)

What could be causing this? Note that a similar issue happens in the `cifar100_100_10` experiment.

KarhouTam:
Hard to tell now. I will check later.

BTW, hugginface keeps reporting that `Using the latest cached version of the dataset since cifar100 couldn't be found on the Hugging Face Hub`, is that normal? I've already downloaded the dataset to the local. Why hugginface still tries to connect to the net and find the dataset?

KarhouTam:
Hi, @jafermarq .
I've fixed the code and ran `cifar10_100_2` experiment. 
Here are the results
```
Results will be saved into: /root/codes/python/flower/baselines/fedrep/outputs/2024-09-29/20-57-44/results.pkl
(ClientAppActor pid=78925) <------- TEST RESULTS -------> : {'loss': 0.010720741094612494, 'accuracy': 0.7926829268292683}
```
image

Seems FedRep is normal again.

My laptop has not that much resources to do all the experiments. I think if `cifar10_100_2` is normal, then the others should be also.

This PR is ready again.

jafermarq:
Hi @KarhouTam, thanks for the update. I re-run the experiments but they perform similarly to when i did my last review (so they are worse than earlier in the PR thread). I looked into it and I'm happy to say that I found the issue 😆 . Before using `flower-datasets`, clients would use as their test set 17% of their training set (derived from `<config>.dataset.fraction=0.83`) and (!!) also a 100th of the CIFAR-10/100 testset (right? https://github.com/adap/flower/blob/b73bd6be16663f941555a5e043c3e11c01e5a8c6/baselines/fedrep/fedrep/dataset_preparation.py#L73). However, after switching to `flower-dataset` the test set of CIFAR-10/100 is not being used at all. It seems the only reason why the accuracy of some experiments (mostly CIFAR-10 w/ `FedRep`) seem to be lower is just not exposing the trained model to a wider evaluation set... See below the results I got with the code as in the last commit and after applying the changes I propose below. 

Screenshot 2024-09-30 at 00 33 52


> [!NOTE]
> Top row is `FedRep-CIFAR10-100-2`. Bottom row is `FedRep-CIFAR10-100-5`. It doesn't seem to impact CIFAR-100 results much so i haven't included them here.

What to change ?
- in `client.py` you'll need to make use of the "test" partition of the dataset and concatenate it to the _validation_ split you are already creating out of the training partition. 
- For this the best is to create a new partitioner that follows the same scheme as the one you already use. This will ensure that the test set of your clients also have the same classes as the ones present in the training partition.

I've done it like this:

```python

def get_client_fn_simulation(...):
     .... 
     
    # new partitioner for the training split
    partitioner_test = PathologicalPartitioner(<same-args-as-the-existing-one>)
    ....

    FEDERATED_DATASET = FederatedDataset(
            dataset=config.dataset.name.lower(), partitioners={"train": partitioner_train, "test": partitioner_test}
        )
    ...
   
    def client_fn(cid: str) -> Client:

    ....

        trainset = partition_train_test["train"].with_transform(apply_train_transforms)
        valset = partition_train_test["test"].with_transform(apply_test_transforms) # just like before
        #! Now extract test partition and fuse validation and test into a single one
        testset = FEDERATED_DATASET.load_partition(cid_use, split="test")
        testset = testset.with_transform(apply_test_transforms)
        testset = concatenate_datasets([valset, testset])

        # rest is the same

```

Finally, I also had to make a simple change to your `evaluate_metrics_aggregation_fn()`. I simplified it to this:

```python
    def evaluate_metrics_aggregation_fn(...):
        weights, accuracies = [], []
        for num_examples, metric in eval_metrics:
            weights.append(num_examples)
            accuracies.append(metric["accuracy"] * num_examples)
        accuracy = sum(accuracies) / sum(weights)
        return {"accuracy": accuracy}
```

I though of presenting my changes here in a comment instead of as comments to the code. Let me know what you think about this finding of now including the `testset` for evaluation. I didn't check super super carefully how you were doing the partitioning of the dataset before switching to `flower-datasets` but, am I right in thinking that what I propose above is equivalent to what you were doing?

KarhouTam:
Hi, @jafermarq . Thanks for doing other experiments. Your thought is really helpful 🚀. I've fixed the code. I followed most of your suggestions. At `partitioner` part, instead, I re-using the `partitioner`: `partitioners={"train": partitioner, "test": partitioner}`. I think that should work too. Because `partitioner` does not relate to a specific dataset object.

Would you mind run experiment again for testing？ If so, that would be a great help.

jafermarq:
@KarhouTam , actually reusing the same partitioner isn't possible since it can only be linked to one `._dataset`. So passing `partitioners={"train": ..., "test":...}` and both pointing to the same object will make the second one reuse the same dataset split as the first one. This is something i encountered yesterday for the first time when proposing my solution. What do you think @adam-narozniak , should we support this multiple datasets or add a check that if the `id(partitioner)` is the same an error is thrown?

A super simplified test of this is the following:

```python
from flwr_datasets.partitioner import IidPartitioner
partitioner = IidPartitioner(num_partitions=10)
fds = FederatedDataset(dataset="mnist", partitioners={"test": partitioner, "train": partitioner})

# Load the first partition
iid_partition = fds.load_partition(partition_id=0, split='test')
print(iid_partition)
iid_partition = fds.load_partition(partition_id=0, split='train')
print(iid_partition)
```

this gets printed:
```shell
Dataset({
    features: ['image', 'label'],
    num_rows: 1000
})
Dataset({
    features: ['image', 'label'],
    num_rows: 1000 # <------------- clearly this isn't coming from the training set (which has 50K images)
})
```


@KarhouTam , in the mean time. Is it ok if we have that "duplicated" partitioner? maybe do a hardcopy if you prefer in order to keep the code cleaner and leave a comment?

KarhouTam:
@jafermarq , sorry I didn't check the inside of partitioner. As you mentioned before, I set different partitioners for dataset splits.

adam-narozniak:
@jafermarq I'll open up a PR that performs the check if there are two identical partitioners specified in the partitioners dict. It's a bug (I haven't thought about this problem). The latter asked partitioning does not happen but uses the already assigned dataset.

adam-narozniak:
@jafermarq @KarhouTam 
When it comes to the results problem, I have two comments.

1. if, in the original code, it was `self.data = torch.cat([train_data, test_data])` then using two different partitioners does not necessarily give the same results (size wise; the classes should be the same if the number of partitions is the same + the seed is the same). I'd say that the dataset should be merged in a similar manner so 
```
from flwr_datasets.partitioner import IidPartitioner
from flwr_datasets.preprocessor import Merger
partitioner = IidPartitioner(num_partitions=10)
fds = FederatedDataset(dataset="mnist", partitioners={"train": partitioner}, preprocessor=Merger({"train":("train", "test")}))
fds.load_split("train")
# This gives the 70,000 samples which come from the original: 60,000 train + 10,000 test
# You can continue using the .load_partition normally 
fds.load_partition(0)
```

2. The pathological assignment you performed in your code is according to the class_assignment_mode that is called "random" in `PathologicalPartitioner`. The behavior of the "pathological" assignment is inconsistent in the different papers (in the non-idd bench, it was sth I call "deterministic-first," which is the default in flwr-datasets).
Anyway, to match what is now in your custom code, I'd say to use PathologicalPartitioner(..., class_assignment_mode ="random")

jafermarq:
Thanks @adam-narozniak !
This use of the [`Merger`](https://flower.ai/docs/datasets/ref-api/flwr_datasets.preprocessor.Merger.html) looks very clean. @KarhouTam, let's use the `Merger` option if we are right in thinking that in your original code you first merged both train and test sets, then do the partitioning, then further split into train/test each partition (potentially having some images in the training set used for validation and the other way around). Is this the case?

I'm seeing good results with the change proposed by Adam.

KarhouTam:
@adam-narozniak Thanks for the clean `Merger` solution. I will do the change later.

> potentially having some images in the training set used for validation and the other way around). Is this the case?

@jafermarq Well I think is fine to blend the data from different splits (original `train` and `test`) together. Because the data feature distributions of `train` and `test` are identical, and the model we want to train has no any empirical knowledge to these data, right? So there is no data sample that has to be in `train` or `test` split.

Anyway, I agree to use the `Merger` option. That's elegant.

KarhouTam:
Hi, @jafermarq . The change has beed done. I've tested the agorithm with `cifar10_100_2` config again, and here is the curve.

image

jafermarq (approved):

jafermarq:
Amazing job @KarhouTam! 💯

PR #3791, opened 2024-07-12
feat(datasets) Add `DistributionPartitioner` to Flower Datasets

chongshenng:
## Issue
Some FL researchers use specific NIID distributions such as a pathological power law split but Flower Datasets do not currently support such partitioning schemes.

### Description
The power law splitting procedure are used in several notable research such as `fedprox` and `tamuna`. To ensure that Flower Datasets can be used with Flower baselines, this partitioner needs to be developed.

## Proposal
Add a partitioner that closely reproduce the power-law partitioning scheme.

### Explanation
This partitioner will include the functionalities to:
1. Accept sample counts come from a user-specified distribution.
2. Specify minimum occurrence per label per client.
3. Specify number of labels per client.

Point 1 is intentional so that other distributions (other than a log-normal distribution) can also be flexibly prescribed to this partitioning scheme. It retains its original pathological definition, i.e. partitions only have `num_unique_labels_per_partition`. All samples from the dataset are exhausted during sampling, if the `rescale` parameter is set to `True`.

This implementation is inspired from Li et al. "Federated Optimization in Heterogeneous Networks" (2020) https://arxiv.org/abs/1812.06127.

### Visualizations
The histogram shows the pathological partitioning of a log-normal distribution for 20 partitions with the MNIST dataset, with 5 preassigned number of samples per label, and 2 unique labels per partition:
image

To assess the original power law implementation by Li et al. and the Flower Datasets implementation, we plot and compare the distributions from both implementation below:
image
The plot above uses the original configuration of 1_000 partitions, 5 preassigned number of samples per label, and 2 unique labels per partition. Our Flower Datasets implementation closely matches the original distribution and the $y=x^{-2}$ curve, which validates our implementation.

### Checklist

- [x] Write tests
- [X] Make CI checks pass
- [X] Ping maintainers on [Slack](https://flower.ai/join-slack/) (channel `#contributions`)


datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+class DistributionPartitioner(Partitioner):  # pylint: disable=R0902
+    """Partitioner based on a distribution.
+
+    Inspired from implementations of Li et al.
+    Federated Optimization in Heterogeneous Networks (2020)
+    https://arxiv.org/abs/1812.06127.
+
+    Parameters
+    ----------
+    distribution_array : Union[NDArrayInt, NDArrayFloat]
```
adam-narozniak, line 37: Here, we should have either a purely verbal or, if manageable, index-based explanation of what `distribution_array[label_id, index]` means and where such a number of samples will be assigned.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+        equal (or close to) to the original unpartitioned label distribution
+        - see the `rescale` argument.
+    num_partitions : int
+        The total number of partitions that the data will be divided into.
+    num_unique_labels_per_partition : int
+        Number of unique labels assigned to a single partition.
+    partition_by : str
+        Column name of the labels (targets) based on which sampling works.
+    preassigned_num_samples_per_label : int
+        The minimum number of samples that each label in each partition will have.
```
adam-narozniak, line 54: Let's reword it so that there's no confusion about `each label`, which can be interpreted as each unique, not each that is meant to be assigned to that partition.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+    rescale : bool, default=True
+        Whether to partition samples according to the values in
+        `distribution_array` or rescale based on the original unpartitioned
+        class label distribution. `float` values are rounded to the nearest `int`.
```
adam-narozniak, line 58: Related to this argument - we are not exhausting the samples. The purpose of rescale is (at least in my mind, to use all the samples). If not enabled by default, we should add an argument to allow the use of all samples - counter the float to int rounding.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+            self._num_unique_labels = len(self.dataset.unique(self._partition_by))
+            num_columns = (
+                self._num_unique_labels_per_partition
+                * self._num_partitions
+                / self._num_unique_labels
+            )
+            if self._distribution_array.shape != (self._num_unique_labels, num_columns):
+                raise ValueError(
+                    f"The distribution array shape is {self._distribution_array.shape},"
+                    f" but it should be ({self._num_unique_labels}, {num_columns})."
```
adam-narozniak, line 303: Let's make it more obvious what is the error problem e.g. if the first dim value is not properly set let's explain that the number of unique labels in the dataset is `x` but given `y`. And for the second shape mismatch, e.g., we can defer to the documentation or note that it might be caused by the first shape mismatch in case both are wrong.

datasets/flwr_datasets/partitioner/distribution_partitioner_test.py
```diff
@@ -0,0 +1,192 @@
...
+
+import numpy as np
+import numpy.typing as npt
+import pytest
+
+from datasets import Dataset
+from flwr_datasets.partitioner.distribution_partitioner import DistributionPartitioner
+
+NDArrayInt = npt.NDArray[np.int_]
+NDArrayFloat = npt.NDArray[np.float_]
```
adam-narozniak, line 28: This can be imported from common, just like in the class implementation file.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+        A 2-dimensional numpy array of the probability distribution of samples
+        for all labels in all partitions. The array shape should be
+        (`num_unique_labels`,
+        `num_unique_labels_per_partition*num_partitions/num_unique_labels`),
+        such that the first row of the array corresponds to the sample distribution
+        of the first unique label (in ascending order). The values may be scaled per
+        label such that the sum of the label distributions across all partitions are
+        equal (or close to) to the original unpartitioned label distribution
+        - see the `rescale` argument.
+    num_partitions : int
```
adam-narozniak, line 47: Let's add the note that currently, it's possible to sample only this number if it is a multiple of the num_unique_labels.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+        The total number of partitions that the data will be divided into.
+    num_unique_labels_per_partition : int
```
adam-narozniak, line 49: This number needs to be checked too. Currently, it works also when the `num_unique_labels_per_client` is bigger than the number of `unique_labels` e.g.  `num_unique_labels_per_client = 20` in the example that you provided.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+    seed : int, default=42
+        Seed used for dataset shuffling. It has no effect if `shuffle` is False.
+
+    Examples
+    --------
+    In order to reproduce the power-law distrbution of the paper, follow this setup:
+
+    >>> from flwr_datasets import FederatedDataset
+    >>> from flwr_datasets.partitioner import DistributionPartitioner
+    >>> from pprint import pprint
```
adam-narozniak, line 71: `numpy` import is missing

datasets/flwr_datasets/partitioner/distribution_partitioner.py
adam-narozniak: Also, when the preassigment is too big the code fails silently (no error but the results do not meet the requirements).  I tried `preassigned_num_samples_per_label = 1_000` (in the example provided)

adam-narozniak:
Let's have tests for all the problems mentioned. Also, it'd be good if we switched to `parameterized` library instead of the `pytest` for consistency.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,336 @@
...
+import numpy as np
+
+import datasets
+from flwr_datasets.common.typing import NDArrayFloat, NDArrayInt
+from flwr_datasets.partitioner.partitioner import Partitioner
+
+
+class DistributionPartitioner(Partitioner):  # pylint: disable=R0902
+    """Partitioner based on a distribution.
+
```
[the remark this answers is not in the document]
chongshenng: Good point, added!

datasets/flwr_datasets/partitioner/distribution_partitioner.py
chongshenng: Good point, added!

datasets/flwr_datasets/partitioner/distribution_partitioner.py
chongshenng: Fixed, thanks for picking this up.

datasets/flwr_datasets/partitioner/distribution_partitioner_test.py
chongshenng: Good catch! Missed this somehow.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
chongshenng: Good point, I've added the exhaustive rescaling. For now, it is set to be exhaustive when `rescale = True`. I think it makes sense. I've also updated the PDF above and it's similar, which is good.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,412 @@
...
+    Inspired from implementations of Li et al. Federated Optimization in
+    Heterogeneous Networks (2020) https://arxiv.org/abs/1812.06127.
+
+    Given a user-specified distribution, the algorithm splits the dataset for each
+    unique label per partition where each label is assigned to the partitions in a
+    deterministic pathological manner. That is, given a distribution array of shape,
```
adam-narozniak, line 36: Let's add a line about the interpretation of the 2-nd dim, which is the number of buckets into which the samples associated with each label will be divided.

adam-narozniak (commented):
Left some comments

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,412 @@
...
+        The total number of partitions that the data will be divided into. The number of
+        partitions must be an integer multiple of the number of unique labels in the
+        dataset.
+    num_unique_labels_per_partition : int
+        Number of unique labels assigned to a single partition.
+    partition_by : str
+        Column name of the labels (targets) based on which sampling works.
+    preassigned_num_samples_per_label : int
+        The minimum number of samples that each unique label in each partition will
+        have.
```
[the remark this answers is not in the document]
chongshenng: Good point, fixed.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,412 @@
...
+                           `num_unique_labels_per_partition` x `num_partitions`
+    ( `num_unique_labels`, ---------------------------------------------------- ),
+                                          `num_unique_labels`
+    the label_id at the i'th row is assigned to the partition_id based on the formula:
+        partition_id = alpha + beta
+    where,
+        alpha* = (i - num_unique_labels_per_partition + 1) \
+                 + (j % num_unique_labels_per_partition)
+        alpha = alpha* + (alpha* > 0 ? 0 : num_unique_labels)
+        beta = num_unique_labels * (j // num_unique_labels_per_partition)
```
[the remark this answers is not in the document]
chongshenng: > I checked for [1,1] in the described setup below, and it would assigned to the pid=3 (alpha=3; beta=0) but it is to the pid=1 (pid=1 aligned with my interpretation of the code)

Good catch, there's a typo there and should be `(alpha* >= 0 ? ...)`.

The `partition_id` array must first be generated for the i'th row for all `j` and then sorted in ascending order. 

I added the following sentence to be clear:
```
        partition_id = <alpha + beta>
    where,
        <.> denotes the reindexed sequence of partition_ids in monotone increasing
            order for all j's
```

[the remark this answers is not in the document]
chongshenng: I've also added the alternative explanation.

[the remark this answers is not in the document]
adam-narozniak: Sorry, I don't quite get it. I thought of alpha and beta as scalars.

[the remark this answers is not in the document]
adam-narozniak: I'm not quite sure what got changed based on that
```
good catch, there's a typo there and should be (alpha* >= 0 ? ...).
```
Could you explain?

datasets/flwr_datasets/partitioner/distribution_partitioner_test.py
```diff
@@ -0,0 +1,211 @@
...
+class TestDistributionPartitioner(unittest.TestCase):
+    """Unit tests for DistributionPartitioner."""
+
+    @parameterized.expand(  # type: ignore
+        [
+            # num_partitions, num_unique_labels_per_partition, num_samples,
+            # num_unique_labels, preassigned_num_samples_per_label
+            (10, 2, 200, 10, 5),
+            (10, 2, 200, 10, 0),
+            (20, 1, 200, 10, 5),
```
adam-narozniak, line 101: If all the methods use the same parameters we can use `parameterized_class` instead of ` @parameterized.expand(`. See https://github.com/adap/flower/blob/main/datasets/flwr_datasets/federated_dataset_test.py

adam-narozniak (approved):
lgtm!

datasets/flwr_datasets/partitioner/distribution_partitioner_test.py
```diff
@@ -0,0 +1,181 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Test cases for DistributionPartitioner."""
+
+import unittest
```
jafermarq, line 17:
```suggestion


import unittest
```

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,423 @@
...
+    to class 0) will be assigned to partitions [0, 9, 10, 19], 1st row (class 1) to
+    [0, 1, 10, 11], 2nd row (class 2) to [1, 2, 11, 12], 3rd row (class 3) to
+    [2, 3, 12, 13], etc ... . Alternatively, the distribution can be interpreted as
+    partition 0 having classes 0 and 1, partition 1 having classes 1 and 2, partition 2
+    having classes 2 and 3, etc ... The list representing the unique labels is sorted
+    in ascending order.
+
+    Parameters
+    ----------
+    distribution_array : Union[NDArrayInt, NDArrayFloat]
```
jafermarq, line 67: I'm thinking we should also allow for this to be a list of lists. We'd convert it into a numpy array in `_check_distribution_array_shape_if_needed()` if it's not one already. What do you think?

chongshenng: @jafermarq I like that option. How about we extend it in a following PR? Merging this PR will unblock a few of our activities, and giving a list of lists is a nice enhancement that can follow this PR.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,423 @@
...
+                f"of unique classes in the given dataset={self._num_unique_labels}. "
+                "Reduce the `num_unique_labels_per_partition` or make use different "
+                "dataset to apply this partitioning."
+            )
+
+    def _check_distribution_array_sum_if_needed(self) -> None:
+        """Test correctness of distribution array sum."""
+        if not self._partition_id_to_indices_determined and not self._rescale:
+            labels = self.dataset[self._partition_by]
+            distribution = sorted(Counter(labels).items())
```
jafermarq, line 384: wouldn't it be more accurate to call it `unique_labels_counter`?

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,423 @@
...
+            distribution_vals = [v for _, v in distribution]
+
+            if any(self._distribution_array.sum(1) > distribution_vals):
+                raise ValueError(
+                    "The sum of at least one label distribution array "
+                    "exceeds the original class label distribution."
```
jafermarq, line 390:
```suggestion
                    "exceeds that of the original class label distribution."
```

jafermarq: i'm not sure if the part of "original class label distribution" is accurate. See previous comment in this method.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,423 @@
...
+        if not self._partition_id_to_indices_determined:
+            if self._num_partitions > self.dataset.num_rows:
+                raise ValueError(
+                    "The number of partitions needs to be smaller than the number of "
+                    "samples in the dataset."
+                )
+            if self._num_partitions % self._num_unique_labels != 0:
+                raise ValueError(
+                    f"The number of partitions {self._num_partitions} is not divisible "
+                    f"by the number of unique labels {self._num_unique_labels}."
```
jafermarq, line 404:
```suggestion
                    f"The number of partitions ({self._num_partitions}) is not divisible "
                    f"by the number of unique labels ({self._num_unique_labels})."
```

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,423 @@
...
+                    "exceeds the original class label distribution."
+                )
+
+    def _check_num_partitions_correctness_if_needed(self) -> None:
+        """Test num_partitions when the dataset is given (in load_partition)."""
+        if not self._partition_id_to_indices_determined:
+            if self._num_partitions > self.dataset.num_rows:
+                raise ValueError(
+                    "The number of partitions needs to be smaller than the number of "
+                    "samples in the dataset."
```
jafermarq, line 399:
```suggestion
                    "The number of partitions ({self._num_partitions}) needs to be smaller than the number of "
                    "samples in the dataset ({self.dataset.num_rows})."
```

datasets/flwr_datasets/partitioner/distribution_partitioner.py
```diff
@@ -0,0 +1,423 @@
...
+            distribution_vals = [v for _, v in distribution]
+
+            if any(self._distribution_array.sum(1) > distribution_vals):
+                raise ValueError(
+                    "The sum of at least one label distribution array "
+                    "exceeds the original class label distribution."
+                )
+
+    def _check_num_partitions_correctness_if_needed(self) -> None:
+        """Test num_partitions when the dataset is given (in load_partition)."""
```
jafermarq, line 394: what does ` dataset is given (in load_partition)` mean?

datasets/flwr_datasets/partitioner/distribution_partitioner.py
chongshenng: Sure! That works.

datasets/flwr_datasets/partitioner/distribution_partitioner.py
chongshenng: I've reworded it to "... exceeds that of the unique labels counter in the given dataset ..."

datasets/flwr_datasets/partitioner/distribution_partitioner.py
chongshenng: Poor wording, reworded!

jafermarq (approved):

PR #3802, opened 2024-07-15
refactor(framework) Replace `run_id` with `Run` in simulation

jafermarq:


src/py/flwr/simulation/run_simulation.py
```diff
@@ -371,13 +364,18 @@ def _run_simulation(
     # Convert config to original JSON-stream format
     backend_config_stream = json.dumps(backend_config)
 
+    # If no Run object is set, create one.
```
danieljanes, line 367:
```suggestion
    # If no `Run` object is set, create one
```

danieljanes (approved):

PR #3806, opened 2024-07-15
refactor(framework) Improve app loading in simulation engine

jafermarq:
The core of the Simulation Engine has a simple mechanism to load the function that would eventually return a `ClientApp`. As Flower has rapidly evolved in terms of what's the origin of a  `ClientApp` object (e.g. a simple python file, or a full FAB) so it has the ways of loading these apps (e.g. loading an associated `pyproject.toml`)

This PR reuses the same mechanism that `flower-supernode` uses to load a function that returns a `ClientApp`. This step represent the first step towards being able to run full FABs from `flower-simulation`.

In addition, this PR caches the `ClientApp` before starting the simulation, making them more efficient to run.

### DIFF
--- a/examples/simulation-pytorch/sim.py
+++ b/examples/simulation-pytorch/sim.py
@@ -87,11 +87,13 @@ def get_client_fn(dataset: FederatedDataset):
     the strategy to participate.
     """
 
-    def client_fn(cid: str) -> fl.client.Client:
+    def client_fn(context) -> fl.client.Client:
         """Construct a FlowerClient with its own dataset partition."""
 
         # Let's get the partition corresponding to the i-th client
-        client_dataset = dataset.load_partition(int(cid), "train")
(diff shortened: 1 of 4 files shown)


examples/simulation-pytorch/sim.py
```diff
@@ -171,26 +173,32 @@ def evaluate(
...
-    on_fit_config_fn=fit_config,
-    evaluate_metrics_aggregation_fn=weighted_average,  # Aggregate federated metrics
-    evaluate_fn=get_evaluate_fn(centralized_testset),  # Global evaluation function
-)
+from flwr.server import ServerAppComponents
+
+
+def server_fn(context):
+
+    # Configure the strategy
```
danieljanes, line 181:
```suggestion
def server_fn(context):
    # Configure the strategy
```

PR #3838, opened 2024-07-17
feat(framework) Send federation config to SuperExec

charlesbvll:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/py/flwr/cli/run/run.py
```diff
@@ -113,7 +113,7 @@ def run(
 
 
 def _run_with_superexec(
-    federation: Dict[str, str],
+    federation: Dict[str, Any],
```
danieljanes, line 116: Shouldn't this type match the proto type? We'll expand to other value types in line with other configs (run config, executor config, node config)

src/proto/flwr/proto/exec.proto
```diff
@@ -28,6 +28,7 @@ service Exec {
 message StartRunRequest {
   bytes fab_file = 1;
   map<string, string> override_config = 2;
+  map<string, string> options_config = 3;
```
danieljanes, line 31: How about:
```suggestion
  map<string, string> federation_options = 3;
```

charlesbvll: Or `federation_config` to be consistent with `override_config`?

src/py/flwr/cli/run/run.py
```diff
@@ -165,6 +165,7 @@ def on_channel_state_change(channel_connectivity: str) -> None:
     req = StartRunRequest(
         fab_file=Path(fab_path).read_bytes(),
         override_config=parse_config_args(config_overrides, separator=","),
+        options_config=flatten_dict(federation.get("options")),
```
danieljanes, line 168: Do we want to have any kind of validation? Perhaps check if only `str` values are set?

src/py/flwr/cli/run/run.py
charlesbvll: @danieljanes This is `Dict[str, Any]` because it is the return type of the `validate_and_load_config` function. The effective type will be determined by the `flatten_dict` function. Currently this means string only, but once, the user config type is merged, it will be `UserConfig`. Would you prefer the conversion to be done before the call to this function?

src/py/flwr/cli/run/run.py
charlesbvll: @danieljanes This is the case inside the `flatten_dict` function

src/py/flwr/cli/run/run.py
danieljanes: makes sense - let's merge `UserConfig` first then. are you asking about the `flatten_dict` conversion?

charlesbvll: @danieljanes Yes

src/proto/flwr/proto/exec.proto
danieljanes: `federation_config` seems good

src/py/flwr/common/config.py
```diff
@@ -113,8 +113,13 @@ def get_fused_config(run: Run, flwr_dir: Optional[Path]) -> UserConfig:
     return get_fused_config_from_dir(project_dir, run.override_config)
 
 
-def flatten_dict(raw_dict: Dict[str, Any], parent_key: str = "") -> UserConfig:
+def flatten_dict(
+    raw_dict: Optional[Dict[str, Any]], parent_key: str = ""
```
danieljanes, line 117:
```suggestion
    raw_dict: Optional[UserConfig], parent_key: str = ""
```

src/py/flwr/superexec/executor.py
```diff
@@ -63,6 +64,8 @@ def start_run(
             The Flower App Bundle file bytes.
         override_config: UserConfig
             The config overrides dict sent by the user (using `flwr run`).
+        federation_config: Dict[str, str]
```
danieljanes, line 67:
```suggestion
        federation_config: UserConfig
```

src/py/flwr/common/config.py
charlesbvll: @danieljanes No, this function needs to receive a raw dict of type `dict[str, Any]` as it can be a `dict` of `dict`

danieljanes (approved):

PR #3844, opened 2024-07-18
feat(datasets) Add ted-lium dataset to the tested set

adam-narozniak:
Feature-wise it's just like speech_commands, therefore no new tests were added


jafermarq:
@adam-narozniak [this dataset is 22.7GB](https://huggingface.co/datasets/LIUM/tedlium), should we mock it just like with `UCF101` and `Ambient Context`?

adam-narozniak:
I've just added it to the list of datasets that are supported due to the same features (no additional tests are executed because of modificaiton)

jafermarq (approved):

PR #3855, opened 2024-07-19
fix(framework:skip) Set unsigned integer generation

chongshenng:
Set `signed=False` when converting bytes to integer. This ensures `run_id` and `node_id` are non-negative. It also allows `flwr log RUN_ID` to be passed as a positional argument with unsigned integers since any `-` prefix is detected as an option by `typer`. 

> [!WARNING]
> Currently, not possible because SQLite does not store unsigned 64-bit unsigned ints natively (see [this write-up](http://ivory.idyll.org/blog/2022-storing-ulong-in-sqlite-sourmash.html)).


chongshenng:
Closing due to SQLite incompatibility with unsigned ints.

PR #3857, opened 2024-07-19
feat(datasets) Update FDS list of supported datasets

adam-narozniak:
## Issue
Some of the datasets on HugggingFace changed the signature and are not discoverable under the raw e.g. `mnist` name but rather `ylecun/mnist`.

### Description

The original references still work, but if someone is searching for a dataset, now he finds a different name.

## Proposal
Extend the list of supported datasets by the datasets that we have already tested for but are now available under different names.


jafermarq (approved):

datasets/flwr_datasets/utils.py
```diff
@@ -25,15 +25,22 @@
...
     "cifar100",
+    "uoft-cs/cifar100",
     "svhn",
+    "ufldl-stanford/svhn",
     "sentiment140",
+    "stanfordnlp/sentiment140",
     "speech_commands",
+    "google/speech_commands",
     "LIUM/tedlium",  # Feature wise it's just like speech_commands
     "flwrlabs/femnist",
```
jafermarq, line 45:
```suggestion
    "flwrlabs/femnist",
    "flwrlabs/ambient-acoustic-context",
    "flwrlabs/ucf101",
    "flwrlabs/shakespeare",
    "flwrlabs/celeba",
```

jafermarq: @adam-narozniak should we add the new datasets under `flwrlabs/`?

jafermarq: nevermind. let's wait until the other PRs are merged. still i think a couple of dataset will need to be added

adam-narozniak: Thanks for pointing that out, yes a few datasets are still lagging with the tests. They should be added in new PRs. The purpose of this PR is to just handle different namespace of the datasets that already have tests but can be found under different names

jafermarq (approved):

PR #3858, opened 2024-07-19
feat(framework:skip) Update the `LegacyContext` to align with `Context`

panh99:


src/py/flwr/server/compat/legacy_context.py
```diff
@@ -35,9 +35,12 @@ class LegacyContext(Context):
     client_manager: ClientManager
     history: History
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
-        state: RecordSet,
+        node_id: int = 0,
```
danieljanes, line 40: Wouldn't it be better to avoid the default here and force the user to pass `node_id`?

panh99: IMO, we should require users to pass a `Context` object here. The arguments like `node_id` shouldn't be set by users; instead, they should be obtained automatically from an existing `Context` object. Wdyt? @danieljanes @jafermarq

danieljanes: Yes, fully agree @panh99!

danieljanes (dismissed):

danieljanes (approved):

PR #3860, opened 2024-07-20
feat(datasets) Add function to perform partial download of dataset for tests

adam-narozniak:
## Issue
Not all the datasets are perfectly resembled for tests.

## Proposal
Perform partial download in order to resemble datasets better.
Note there's no function to perform that in the `datasets` library; this is a solution to enable that.

### DIFF
--- a/datasets/flwr_datasets/mock_utils_test.py
+++ b/datasets/flwr_datasets/mock_utils_test.py
@@ -386,8 +386,8 @@ def _load_mocked_dataset_by_partial_download(
     """Download a partial dataset.
 
     This functionality is not supported in the datasets library. This is an informal
-    way of achieving this by using the `streaming=True` and creating a dataset.Dataset
-    from in-memory objects.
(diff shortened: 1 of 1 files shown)


datasets/flwr_datasets/mock_utils_test.py
```diff
@@ -375,3 +375,62 @@ def _load_mocked_dataset(
...
+    dataset_name: str,
+    split_name: str,
+    skip_take_list: List[Tuple[int, int]],
+    subset_name: Optional[str] = None,
+) -> Dataset:
+    """Download a partial dataset.
+
+    This functionality is not supported in the datasets library. This is an informal
+    way of achieving this by using the `streaming=True` and creating a dataset.Dataset
+    from in-memory objects.
```
jafermarq, line 390:
```suggestion
    way of achieving partial dataset download by using the `streaming=True` and creating 
    a dataset.Dataset from in-memory objects.
```

jafermarq (approved):

PR #3870, opened 2024-07-22
fix(framework:skip) Check if the message has content when iterating over all messages in `SecAggPlus` Workflow

panh99:


jafermarq (approved):
looks good to me!

danieljanes (approved):

PR #3878, opened 2024-07-23
fix(framework:skip) Convert `UserConfig` to `str` for simulation executor

jafermarq:
The `SuperExec` uses a dictionary to store the user configs that are to be overrided from `flwr run`. In order to pass these to the `flower-simulation` we need to convert them back to the format `key=value,anotherkey=anothervalue`

### DIFF
--- a/src/py/flwr/superexec/simulation.py
+++ b/src/py/flwr/superexec/simulation.py
@@ -129,7 +129,10 @@ class SimulationEngine(Executor):
             ]
 
             if override_config:
-                command.extend(["--run-config", f"{override_config}"])
+                override_config_str = ",".join(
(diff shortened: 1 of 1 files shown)


src/py/flwr/superexec/simulation.py
```diff
@@ -129,7 +147,9 @@ def start_run(
             ]
 
             if override_config:
-                command.extend(["--run-config", f"{override_config}"])
+                override_config_str = self._user_config_to_str(override_config)
+                print(override_config_str)
```
danieljanes, line 151: Let's remove the `print` and make `_user_config_to_str ` a function instead of a method

danieljanes (approved):

PR #3909, opened 2024-07-24
fix(datasets:skip) Fix ShardPartitioner typo

m-danilow:
### Description

natual -> natural in ShardPartitioner implementation and one comment.


jafermarq (approved):
Thanks for the fix @m-danilow !

PR #3944, opened 2024-07-30
feat(framework) Add FAB type

charlesbvll:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/py/flwr/common/typing.py
```diff
@@ -199,3 +199,11 @@ class Run:
...
     fab_version: str
     override_config: UserConfig
+
+
+@dataclass
+class Fab:
+    """Fab file representation."""
+
+    hash_str: str
+    data_bytes: bytes
```
danieljanes, line 209: @charlesbvll let's align the naming in the ProtoBuf `message Fab { ... }` and the Python dataclass:
```suggestion
    hash_str: str
    content: bytes
```

We could also update the ProtoBuf to

```
message Fab {
  // This field is the hash of the data field. It is used to identify the data.
  // The hash is calculated using the SHA-256 algorithm and is represented as a
  // hex string (sha256hex).
  string hash_str = 1;
  // This field contains the fab file contents a one bytes blob.
  bytes content = 2;
}
```

charlesbvll: @danieljanes Here is a PR to achieve this: #3957

danieljanes (approved):

PR #3952, opened 2024-08-01
break(framework) Enable CLIs to accept an app directory

panh99:
> I tried to make this PR as minimal as possible, but here it is.

## Changes:
- Updated `flower-server-app` to accept `app` (path to Flower App) instead of `server-app`
- Updated `flower-supernode` to accept `app` (path to Flower App) instead of `client-app`
- The `app` passed to `flower-supernode` will make SuperNode only support messages targeting at the very ClientApp in the provided `app_dir`
- `get_fused_config` function returns an empty dict if the `fab_id` or `fab_version` is empty. I need to extend it to return an empty dict if the `fab_id` or `fab_version` is invalid ("invalid" means they are non-empty, but cannot be found in one of the installed apps.)
- For the current setting, the flower-simulation CLI doesn't support multi-app feature. Thus, I changed the `multi_app` to `False`. It make no difference on the code-level.
- Added `pyproject.toml` in the `e2e/` so that the `flower-server-app ./..` in the `test_superlink.sh` can load the ServerApp from `e2e/server.py`
- Amended `pyproject.toml` in most E2E tests so that `flower-supernode ./` in the `test_superlink.sh` can load the ClientApp from the correct `client.py`. For E2E tests with their own `server.py`, I also add the reference to that one.
- Left the serverapp/clientapp references empty if they don't exist.
- Removed the `server.py` from `e2e/bare-client-auth` b/c it is not used.
- Renamed `project name` and `project_dir` to `app name` and `app` for user-facing APIs.

## Notes:
- I used a trick to make our E2E tests work with the new Flower, but it is not very ideal. We may revisit them later.
- The `flower-client-app` is not deprecated in this PR.
- I didn't change all `project_dir` to `app_path`/`app` in our code base if they are internal functions.


e2e/bare/pyproject.toml
```diff
@@ -18,3 +16,18 @@ packages = ["."]
...
+serverapp = ""
+clientapp = "client:app"
+
+[tool.flwr.app.config]
+
+[tool.flwr.federations]
+default = "local-simulation"
+
+[tool.flwr.federations.local-simulation]
+options.num-supernodes = 10
```
danieljanes, line 33:
```suggestion
options.num-supernodes = 10

```

src/py/flwr/client/supernode/app.py
```diff
@@ -255,13 +261,14 @@ def _parse_args_run_supernode() -> argparse.ArgumentParser:
...
-        "client-app",
+        "app_dir",
         nargs="?",
-        default="",
-        help="For example: `client:app` or `project.package.module:wrapper.app`. "
-        "This is optional and serves as the default ClientApp to be loaded when "
-        "the ServerApp does not specify `fab_id` and `fab_version`. "
-        "If not provided, defaults to an empty string.",
+        default=None,
+        help="Specify the project directory to load and run the client app. "
```
danieljanes, line 267:
```suggestion
        help="Specify the project directory to load and run the `ClientApp`. "
```

src/py/flwr/client/supernode/app.py
```diff
@@ -255,13 +261,14 @@ def _parse_args_run_supernode() -> argparse.ArgumentParser:
...
+        "When this argument is provided, the SuperNode will exclusively respond to "
+        "messages from the corresponding server app by matching the FAB ID and FAB "
```
danieljanes, line 269:
```suggestion
        "messages from the corresponding ServerApp by matching the FAB ID and FAB "
```

src/py/flwr/server/run_serverapp.py
```diff
@@ -219,10 +219,10 @@ def _parse_args_run_server_app() -> argparse.ArgumentParser:
     )
 
     parser.add_argument(
-        "server-app",
+        "app_dir",
         nargs="?",
         default=None,
-        help="For example: `server:app` or `project.package.module:wrapper.app`",
+        help="Load and run the server app from the provided project directory.",
```
danieljanes, line 225:
```suggestion
        help="Load and run the `ServerApp` from the provided project directory.",
```

src/py/flwr/common/config.py
```diff
@@ -161,3 +166,11 @@ def parse_config_args(
                 overrides.update(tomli.loads(toml_str))
 
     return overrides
+
+
+def get_metadata_from_config(config: Dict[str, Any]) -> Tuple[str, str]:
+    """Extract the fab_version and the fab_id  from a project config."""
```
danieljanes, line 172:
```suggestion
    """Extract the fab_version and the fab_id from a project config."""
```

src/py/flwr/client/supernode/app.py
```diff
@@ -255,13 +261,14 @@ def _parse_args_run_supernode() -> argparse.ArgumentParser:
...
-        "client-app",
+        "app_dir",
         nargs="?",
-        default="",
-        help="For example: `client:app` or `project.package.module:wrapper.app`. "
-        "This is optional and serves as the default ClientApp to be loaded when "
-        "the ServerApp does not specify `fab_id` and `fab_version`. "
-        "If not provided, defaults to an empty string.",
+        default=None,
+        help="Specify the project directory to load and run the `ClientApp`. "
```
jafermarq, line 267:
```suggestion
        help="Specify the app directory to load and run the `ClientApp`. "
```

jafermarq: should we update for consistency? do we need to indicate that this directory is the dir where `pyproject.toml` is (not where `clientapp.py` is defined)?

src/py/flwr/server/run_serverapp.py
```diff
@@ -219,15 +219,15 @@ def _parse_args_run_server_app() -> argparse.ArgumentParser:
     )
 
     parser.add_argument(
-        "server-app",
+        "app_dir",
```
jafermarq, line 222:
```suggestion
        "app-dir",
```

panh99: Screenshot 2024-08-03 at 21 10 32
I think the idea is to align with `flwr run`. So I use `_` instead of `-` here.

src/py/flwr/client/supernode/app.py
panh99: Yeah, that makes a lot of sense. I will update the doc string to indicate that as per you suggest.

src/py/flwr/server/run_serverapp.py
panh99: Second thought, we use `--flwr-dir` instead of `--flwr_dir`, which is different from the case here. What naming convention we should preserve then? @danieljanes @jafermarq ?

danieljanes: I agree, let's use `app-dir`. We use `-` instead of `_` in all CLI args. In `flwr run`, the reason it says `app_dir` is b/c it's a positional arg. I believe there's not way to display it as `app-dir` there, but if there is, we should change it there as well.

One alternative would be to shorten it to `app` everywhere @panh99 @jafermarq @charlesbvll

danieljanes: Sounds good. We can provide that clarification via the documentation and docstrings. With the next release, pointing at the `ClientApp` will be a thing of the past. This makes me optimistic that not many users will be confused.

panh99: Cool. I will rename it to `app` in this PR then. Any thought on this @jafermarq @charlesbvll ?

charlesbvll: Sounds good to me!

src/py/flwr/cli/build.py
```diff
@@ -30,40 +30,42 @@
...
+    if app is None:
+        app = Path.cwd()
 
-    directory = directory.resolve()
-    if not directory.is_dir():
+    app = app.resolve()
+    if not app.is_dir():
         typer.secho(
-            f"❌ The path {directory} is not a valid directory.",
+            f"❌ The path {app} is not a valid app.",
```
danieljanes, line 52:
```suggestion
            f"❌ The path {app} is not a valid path to a Flower app.",
```

src/py/flwr/common/config.py
```diff
@@ -161,3 +166,11 @@ def parse_config_args(
                 overrides.update(tomli.loads(toml_str))
 
     return overrides
+
+
+def get_metadata_from_config(config: Dict[str, Any]) -> Tuple[str, str]:
+    """Extract the fab_version and the fab_id from a project config."""
```
danieljanes, line 172:
```suggestion
    """Extract `fab_version` and `fab_id` from a project config."""
```

danieljanes (approved):

danieljanes (approved):

PR #3964, opened 2024-08-06
refactor(examples) Use low-level API for federated analytics example

jafermarq:
This PR keeps the functionality of the `quickstart-pandas` example but uses the low-level API. The `strategies` can be used for federated analytics but much of their functionality is unused. The newer API is ready for federated analytics thanks to the `@app.query` method. 

wdyt @chongshenng @danieljanes @tanertopal @panh99 ? Should we use this lower level API? to me this looks fine and better than using strategies


examples/quickstart-pandas/pandas_example/server_app.py
```diff
@@ -1,90 +1,55 @@
...
-
-    def configure_fit(
-        self, server_round: int, parameters: Parameters, client_manager: ClientManager
-    ) -> List[Tuple[ClientProxy, FitIns]]:
-        config = {}
-        fit_ins = FitIns(parameters, config)
-        clients = client_manager.sample(num_clients=2, min_num_clients=2)
-        return [(client, fit_ins) for client in clients]
+from flwr.common.logger import log
+from flwr.server import Driver
```
chongshenng, line 12: Let's also import `ServerApp` here. Then, we can remove the `import flwr as fl` line and use `app = ServerApp()` below (like other examples).

chongshenng (changes requested):
I like this refactor. This example is an excellent use case for `app.query()`. Should the `README` be updating accordingly to highlight how it's different?

examples/quickstart-pandas/pandas_example/server_app.py
jafermarq: yes, let's do that

PR #3970, opened 2024-08-08
feat(framework) Add `ClientAppIo` proto

chongshenng:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/proto/flwr/proto/appio.proto
```diff
@@ -0,0 +1,31 @@
...
+  rpc PushClientAppOutputs(PushClientAppOutputsRequest)
+      returns (PushClientAppOutputsResponse) {}
+}
+
+message PullClientAppInputsRequest { sint64 token = 1; }
+message PullClientAppInputsResponse {
+  Message message = 1;
+  Context context = 2;
+  Run run = 3;
+  // Fab fab = 4;
```
danieljanes, line 24: We generally don't merge commented-out code:
```suggestion
```

src/proto/flwr/proto/appio.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
```
danieljanes, line 1: This file should be named `clientappio.proto` (the might be a `serverappio.proto` in the future)

src/proto/flwr/proto/transport.proto
```diff
@@ -119,3 +122,27 @@ message Scalar {
...
+message Metadata {
+  sint64 run_id = 1;
+  string message_id = 2;
+  sint64 src_node_id = 3;
+  sint64 dst_node_id = 4;
+  string reply_to_message = 5;
+  string group_id = 6;
+  double ttl = 7;
+  string message_type = 8;
+}
```
danieljanes, line 148: Let's move these into a new `.proto` file. `transport.proto` covers mostly the legacy gRPC stack and we don't want to add new functionality to it. How about just `message.proto`?

src/proto/flwr/proto/appio.proto
```diff
@@ -0,0 +1,31 @@
...
+  Context context = 2;
+  Run run = 3;
+  // Fab fab = 4;
+}
+message PushClientAppOutputsRequest {
+  sint64 token = 1;
+  Message message = 2;
+  Context context = 3;
+}
+message PushClientAppOutputsResponse { Status status = 1; }
```
danieljanes, line 31: We should probably use a different `Status` here (not the one imported from `transport.proto`)

src/proto/flwr/proto/transport.proto
chongshenng: Good point, I'll update it.

src/proto/flwr/proto/appio.proto
chongshenng: Good point. I've added the `SUCCESS`, `DEADLINE_EXCEEDED` and `UNKNOWN_ERROR` to the `Code` message type. I think that these should be sufficient for now and we can be more specific later. What do you think @danieljanes?

jafermarq: I wonder if we could match the error codes to this [ErrorCode](https://github.com/adap/flower/blob/3034810c525e175cfce95fcad70dd4f23fd4934a/src/py/flwr/common/constant.py#L101) or expand it if necessary.

chongshenng: @jafermarq Good suggestion, but since those are for error codes, I'll leave the refactoring of the codes to another PR.

danieljanes (approved):

PR #3976, opened 2024-08-08
feat(framework) Add `ClientAppIo` servicer

chongshenng:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,111 @@
...
+    ) -> PushClientAppOutputsResponse:
+        """Push Message and Context."""
+        log(DEBUG, "ClientAppIo.PushOutputs")
+        if request.token != self.token:
+            raise ValueError("Mismatch between ClientApp and SuperNode token")
+        self.proto_message = request.message
+        self.proto_context = request.context
+        # Update Message and Context
+        try:
+            self._update_object()
```
jafermarq, line 103: I see a bit ahead in #3978, this will be the mechanism by which the `SuperNode` will access the returned `Message/Context` by the `ClientApp`? is this the main functionality behind this method? I wonder if it could be renamed to something other than `get/update_object()`. I thought of `_payload()` (in the end we are sending several objects) but probably are better alternatives.

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,111 @@
...
+        """Get client app objects."""
+        log(DEBUG, "ClientAppIo.GetObject")
+        return self.message, self.context
+
+    def _update_object(self) -> None:
+        """Update client app objects."""
+        log(DEBUG, "ClientAppIo.UpdateObject")
+        # Deserialize Message and Context
+        self.message: Message = message_from_proto(self.proto_message)
+        self.context: Context = context_from_proto(self.proto_context)
```
jafermarq, line 77: wouldn't it be better to define these elements during `__init__()`?

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            self._update_object()
+            # Set status
+            code = typing.ClientAppOutputCode.SUCCESS
+            status = typing.ClientAppOutputStatus(code=code, message="Success")
+            proto_status = clientappstatus_to_proto(status=status)
+            return PushClientAppOutputsResponse(status=proto_status)
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            log(ERROR, "ClientApp failed to push message to SuperNode, %s", e)
+            return PushClientAppOutputsResponse()
```
jafermarq, line 111: Should this `PushClientAppOutputsResponse()` return some `status` that we predefine here in the body of the `except`?

src/py/flwr/client/process/clientappio_servicer.py
chongshenng: Yes, it happens in `client/app.py`. `_payload()` also works and is more specific.

src/py/flwr/client/process/clientappio_servicer.py
chongshenng: Unfortunately not. This instance of `ClientAppIoServicer` is created when the `grpc.Server` is spun up for the first time and there is no `Message` or `Context` at that point (see [this line](https://github.com/adap/flower/blob/808c2fbdb3300ec05f6a8d454dbb8a13aa5e69cf/src/py/flwr/client/app.py#L680)). I agree it's not elegant, do you know of a better solution @jafermarq?

src/py/flwr/client/process/clientappio_servicer.py
chongshenng: Good point, added.

src/py/flwr/client/process/clientappio_servicer.py
jafermarq: I was thinking on just define them like this?
```python
def __init__(self, ...):
    self.message: Optional[Message] = None
    self.context: Optional[Context] = None
```

chongshenng: This was my first implementation, but I got a type mismatch error during the linting CI, which I couldn't work around.

jafermarq: i see. One solution is to  put those `*_from_proto()` behind a `if not None` (that should keep mypy happy)

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,130 @@
...
+            code = typing.ClientAppOutputCode.UNKNOWN_ERROR
+            status = typing.ClientAppOutputStatus(code=code, message="Push failed")
+            proto_status = clientappstatus_to_proto(status=status)
+            return PushClientAppOutputsResponse(status=proto_status)
+
+    def set_inputs(self, clientapp_input: ClientAppIoInputs) -> None:
+        """Set ClientApp inputs."""
+        log(DEBUG, "ClientAppIo.SetInputs")
+        self.clientapp_input = clientapp_input
+        assert self.clientapp_output is None, "ClientAppIoOutputs is not None"
```
danieljanes, line 120: We generally don't use `assert` in production code, only in tests

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,134 @@
...
+        self, request: PullClientAppInputsRequest, context: grpc.ServicerContext
+    ) -> PullClientAppInputsResponse:
+        """Pull Message, Context, and Run."""
+        log(DEBUG, "ClientAppIo.PullInputs")
+        if self.clientapp_input is None:
+            raise ValueError(
+                "ClientAppIoInputs not set before calling `PullClientAppInputs`."
+            )
+        if request.token != self.clientapp_input.token:
+            raise ValueError("Mismatch between ClientApp and SuperNode token")
```
danieljanes, line 82: we should set an appropriate gprc status code here

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __init__(self) -> None:
+        self.clientapp_input: Optional[ClientAppIoInputs] = None
+        self.clientapp_output: Optional[ClientAppIoOutputs] = None
+
+    def PullClientAppInputs(
+        self, request: PullClientAppInputsRequest, context: grpc.ServicerContext
+    ) -> PullClientAppInputsResponse:
+        """Pull Message, Context, and Run."""
+        log(DEBUG, "ClientAppIo.PullInputs")
```
danieljanes, line 76: so far, we are always logging the full method name
```suggestion
        log(DEBUG, "ClientAppIo.PullClientAppInputs")
```

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,134 @@
...
+            message=message_to_proto(self.clientapp_input.message),
+            context=context_to_proto(self.clientapp_input.context),
+            run=run_to_proto(self.clientapp_input.run),
+        )
+
+    def PushClientAppOutputs(
+        self, request: PushClientAppOutputsRequest, context: grpc.ServicerContext
+    ) -> PushClientAppOutputsResponse:
+        """Push Message and Context."""
+        log(DEBUG, "ClientAppIo.PushOutputs")
```
danieljanes, line 93:
```suggestion
        log(DEBUG, "ClientAppIo.PushClientAppOutputs")
```

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,134 @@
...
+        if self.clientapp_output is None:
+            raise ValueError(
+                "ClientAppIoOutputs not set before calling `PushClientAppOutputs`."
+            )
+        if self.clientapp_input is None:
+            raise ValueError(
+                "ClientAppIoInputs not set before calling `PushClientAppOutputs`."
+            )
+        if request.token != self.clientapp_input.token:
+            raise ValueError("Mismatch between ClientApp and SuperNode token")
```
danieljanes, line 103: we should set an appropriate grpc status code here

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,134 @@
...
+        assert self.clientapp_output is None, "ClientAppIoOutputs is not None"
+
+    def get_outputs(self) -> ClientAppIoOutputs:
+        """Get ClientApp outputs."""
+        log(DEBUG, "ClientAppIo.GetOutputs")
+        if self.clientapp_output is None:
+            raise ValueError("ClientAppIoOutputs not set before calling `get_outputs`.")
+        # Set outputs to a local variable and clear self.clientapp_output
+        output: ClientAppIoOutputs = self.clientapp_output
+        self.clientapp_output = None
```
danieljanes, line 133: we should also clear the input here

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -0,0 +1,147 @@
...
+    def set_inputs(self, clientapp_input: ClientAppIoInputs) -> None:
+        """Set ClientApp inputs."""
+        log(DEBUG, "ClientAppIo.SetInputs")
+        self.clientapp_input = clientapp_input
+        if self.clientapp_output is not None:
+            self.clientapp_output = None
+            warnings.warn(
+                "ClientAppIoOutputs must be None at the start, setting to None",
+                stacklevel=2,
+            )
```
danieljanes, line 136: We should do this check before setting the input and raise an exception if either input or output are still set. It's a clear implementation error if this occurs.

danieljanes (dismissed):

danieljanes (approved):

danieljanes (approved):

danieljanes (approved):

PR #3978, opened 2024-08-08
feat(framework) Add `ClientApp` multiprocess execution

chongshenng:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/py/flwr/client/app.py
```diff
@@ -43,13 +46,34 @@
...
+            try:
+                s.bind(("", port))
+                s.close()
+                return port
+            except OSError:
+                port += 1
+
+
+CLIENTAPPIO_PORT_FREE = find_free_port(CLIENTAPPIO_PORT)
+ADDRESS_CLIENTAPPIO_API_GRPC_RERE += f":{CLIENTAPPIO_PORT_FREE}"
```
tanertopal, line 76: What's the purpose of this? I don't think we should have dynamic server port assignments anywhere. I think it would be good to start a docs page where we list all default ports various Flower server processes use and just put the port into that list and have a fixed default.

src/py/flwr/client/app.py
```diff
@@ -243,6 +268,10 @@ class `flwr.client.Client` (default: None)
         If set to None, there is no limit to the total time.
     flwr_path: Optional[Path] (default: None)
         The fully resolved path containing installed Flower Apps.
+    isolate: Optional[bool] (default: False)
```
tanertopal, line 271: I suggest changing this to `run_type` and having it as an ENUM.

- internal (default): same process
- subprocess: basically what "isolate" does
- docker: in future PR's for now this would not be included here

Down the line we could refactor the whole thing and change it into a plugin system to support custom run types like e.g. K8, Lambda Functions, etc.

src/py/flwr/client/app.py
```diff
@@ -268,6 +297,12 @@ def _load_client_app(_1: str, _2: str) -> ClientApp:
 
         load_client_app_fn = _load_client_app
 
+    if isolate:
+        # Start gRPC server
+        clientappio_servicer, _ = run_clientappio_api_grpc(
+            address=ADDRESS_CLIENTAPPIO_API_GRPC_RERE
+        )
+
```
tanertopal, line 305: I might be mistaken, but would this not start a server for each client? I would suggest starting one server when the SuperNode starts and reusing that one.

src/py/flwr/client/app.py
chongshenng: This mode is for users during development, where they start several `SuperNode`s on their local machine before moving to production. Since each `SuperNode` needs to have their own dedicated address, I added this utility script to ensure they use the next free port.

How about I move this to `supernode/connection.py`. I can update this such that we prompt users when the specified port `9094` is unavailable and suggest the next available port automatically?

Or is there a better way? Wdyt @danieljanes @tanertopal?

💯 on the list. I'll create another PR.

chongshenng: Removed dynamic port allocation.

src/py/flwr/client/app.py
```diff
@@ -249,6 +255,13 @@ class `flwr.client.Client` (default: None)
         If set to None, there is no limit to the total time.
     flwr_path: Optional[Path] (default: None)
         The fully resolved path containing installed Flower Apps.
+    isolate: Optional[bool] (default: False)
```
danieljanes, line 258:
```suggestion
    isolate : Optional[bool] (default: False)
```

src/py/flwr/client/app.py
```diff
@@ -249,6 +255,13 @@ class `flwr.client.Client` (default: None)
...
+        Whether to run ClientApp in a separate process. By default, this value is
+        `False`, and the ClientApp runs in the same process as the SuperNode. If `True`
+        the ClientApp runs in an isolated process and communicates using gRPC at the
+        address `supernode_address`.
+    supernode_address: Optional[str] (default: `ADDRESS_CLIENTAPPIO_API_GRPC_RERE`)
```
danieljanes, line 263:
```suggestion
    supernode_address : Optional[str] (default: `ADDRESS_CLIENTAPPIO_API_GRPC_RERE`)
```

src/py/flwr/client/app.py
```diff
@@ -249,6 +255,13 @@ class `flwr.client.Client` (default: None)
         If set to None, there is no limit to the total time.
     flwr_path: Optional[Path] (default: None)
         The fully resolved path containing installed Flower Apps.
+    isolate: Optional[bool] (default: False)
+        Whether to run ClientApp in a separate process. By default, this value is
+        `False`, and the ClientApp runs in the same process as the SuperNode. If `True`
+        the ClientApp runs in an isolated process and communicates using gRPC at the
+        address `supernode_address`.
```
danieljanes, line 262:
```suggestion
        Whether to run `ClientApp` in a separate process. By default, this value is
        `False`, and the `ClientApp` runs in the same process as the SuperNode. If `True`
        the `ClientApp` runs in an isolated process and communicates using gRPC at the
        address `supernode_address`.
```

src/py/flwr/client/supernode/app.py
```diff
@@ -224,6 +226,18 @@ def _parse_args_run_supernode() -> argparse.ArgumentParser:
         - `$HOME/.flwr/` in all other cases
     """,
     )
+    parser.add_argument(
+        "--isolate",
+        action="store_true",
+        help="Run the ClientApp in an isolated process. By default, the ClientApp "
+        "runs in the same process that executes the SuperNode.",
```
danieljanes, line 233:
```suggestion
        help="Run `ClientApp` in an isolated subprocess. By default, `ClientApp` "
        "runs in the same process that executes the SuperNode.",
```

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -94,11 +94,8 @@ def PushClientAppOutputs(
     ) -> PushClientAppOutputsResponse:
         """Push Message and Context."""
         log(DEBUG, "ClientAppIo.PushClientAppOutputs")
-        if self.clientapp_output is None:
-            raise ValueError(
-                "ClientAppIoOutputs not set before calling `PushClientAppOutputs`."
-            )
         if self.clientapp_input is None:
+            # Ensure `.token` is present
```
danieljanes, line 98: It seems like this comment is on the wrong line

src/py/flwr/client/process/clientappio_servicer.py
```diff
@@ -109,8 +106,10 @@ def PushClientAppOutputs(
             )
         try:
             # Update Message and Context
-            self.clientapp_output.message = message_from_proto(request.message)
-            self.clientapp_output.context = context_from_proto(request.context)
+            self.clientapp_output = ClientAppIoOutputs(
+                message=message_from_proto(request.message),
+                context=context_from_proto(request.context),
+            )
             # Set status
```
danieljanes, line 113:
```suggestion
            )

            # Set status
```

src/py/flwr/client/app.py
```diff
@@ -692,3 +749,13 @@ def run_clientappio_api_grpc(
...
+
+
+def is_port_free(host: str, port: int) -> bool:
+    """Return a boolean indicating whether a specific port is free (not in use) on the
+    specified host."""
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        # Attempt to connect to the specified host and port
+        result = sock.connect_ex((host, port))
+        # If connect_ex returns 0, the port is in use; otherwise, it's free
+        return result != 0
```
danieljanes, line 761: Do we really need this? Or can we just try to start the gRPC server and rely on it failing if the port is used?

src/py/flwr/client/app.py
```diff
@@ -274,6 +287,21 @@ def _load_client_app(_1: str, _2: str) -> ClientApp:
...
+            # `_clientappio_grpc_server` must be returned for the grpc.Server
+            # to be created
+            _clientappio_grpc_server, clientappio_servicer = run_clientappio_api_grpc(
+                address=supernode_address
+            )
+        else:
+            raise ValueError(
+                f"Port {_port} is in use at host {_host}. Specify a different "
+                "port for the ClientAppIo server to start."
+            )
```
danieljanes, line 303: let's remove the port checking and just start the grpc server

src/py/flwr/client/app.py
```diff
@@ -422,14 +450,45 @@ def _on_backoff(retry_state: RetryState) -> None:
...
+                                    context=context,
+                                    run=run,
+                                    token=token,
+                                )
+                            )
+                            # Run ClientApp
+                            verbose = True
+                            command = [
+                                "flwr-clientapp",
+                                "--address",
```
danieljanes, line 471: Should this be `--supernode`? We use `--superlink` in `flower-supernode`

src/py/flwr/client/app.py
```diff
@@ -422,14 +450,45 @@ def _on_backoff(retry_state: RetryState) -> None:
...
+                            clientappio_servicer.set_inputs(
+                                ClientAppIoInputs(
+                                    message=message,
+                                    context=context,
+                                    run=run,
+                                    token=token,
+                                )
+                            )
+                            # Run ClientApp
+                            verbose = True
```
danieljanes, line 468: This can be removed

src/py/flwr/client/app.py
jafermarq: Removed. Both `stdout` and `stderr` are set to None

src/py/flwr/client/app.py
jafermarq: removed since it is no longer used (see comment below)

src/py/flwr/client/app.py
```diff
@@ -251,6 +256,13 @@ class `flwr.client.Client` (default: None)
         If set to None, there is no limit to the total time.
     flwr_path: Optional[Path] (default: None)
         The fully resolved path containing installed Flower Apps.
+    isolate : Optional[bool] (default: False)
+        Whether to run `ClientApp` in a separate process. By default, this value is
+        `False`, and the `ClientApp` runs in the same process as the SuperNode. If
+        `True` the `ClientApp` runs in an isolated process and communicates using
```
danieljanes, line 262:
```suggestion
        `True`, the `ClientApp` runs in an isolated process and communicates using
```

src/py/flwr/client/app.py
```diff
@@ -276,6 +288,11 @@ def _load_client_app(_1: str, _2: str) -> ClientApp:
 
         load_client_app_fn = _load_client_app
 
+    if isolate and supernode_address is not None:
+        _clientappio_grpc_server, clientappio_servicer = run_clientappio_api_grpc(
+            address=supernode_address
+        )
```
danieljanes, line 294:
```suggestion
    if isolate:
        if supernode_address is None:
            raise ValueError("`supernode_address` required when `isolate` is set")
        _clientappio_grpc_server, clientappio_servicer = run_clientappio_api_grpc(
            address=supernode_address
        )
```

src/py/flwr/client/app.py
```diff
@@ -435,11 +452,43 @@ def _on_backoff(retry_state: RetryState) -> None:
 
                     # Handle app loading and task message
                     try:
-                        # Load ClientApp instance
-                        client_app: ClientApp = load_client_app_fn(fab_id, fab_version)
+                        if isolate and supernode_address is not None:
+                            # Generate SuperNode token
+                            token: int = generate_rand_int_from_bytes(RUN_ID_NUM_BYTES)
```
danieljanes, line 457: With the change above, this should not be required any more. The case where `isolate` is `True`, but `supernode_address` is unset should be considered an error.

```suggestion
                        if isolate:
                            # Generate SuperNode token
                            token: int = generate_rand_int_from_bytes(RUN_ID_NUM_BYTES)
```

src/py/flwr/client/app.py
```diff
@@ -435,11 +452,43 @@ def _on_backoff(retry_state: RetryState) -> None:
...
+                            clientappio_servicer.set_inputs(
+                                ClientAppIoInputs(
+                                    message=message,
+                                    context=context,
+                                    run=run,
+                                    token=token,
+                                )
+                            )
+                            # Run ClientApp
+                            command = [
```
danieljanes, line 469:
```suggestion

                            # Run `ClientApp` in subprocess
                            command = [
```

src/py/flwr/client/app.py
jafermarq: i'm afraid but we need to keep the check here. mypy does'tn seem to be smart enough. (but to be fair the change in the other comment happens many lines above, outside two while loops...)

danieljanes: got it, resolved

danieljanes (approved):

PR #3984, opened 2024-08-09
docs(examples:skip) Add redirect for secure aggregation example

adam-narozniak:
## Issue
`app-secure-aggregation` was renamed to `flower-secure-aggregation` however google results for "secure aggregation" on the first page still point to the outdated link which gives 404 not found.

## Proposal
Add a redirect to fix it to the conf.py of example docs.


jafermarq (approved):
great catch!

PR #3989, opened 2024-08-10
feat(framework) Add `SimpleBackend` for simulations

jafermarq:
A new backend based on pure mp `Processes` and that doesn't require `ray`. 

Merge after #3988 .

Limitation: it would be better if we first redesign `ClientApp` so it's  _pickleable_. Currently the way `ClientApp` [defines and uses ffn](https://github.com/adap/flower/blob/0334b3d9c51e3cbfaab13c273b830f4cc089bca6/src/py/flwr/client/client_app.py#L122) prevents it from being  _pickleable_. The current solution is to use [cloudpickle](https://github.com/cloudpipe/cloudpickle) which relaxes the conditions that objects need to meet to be _pickleable_

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ starlette = { version = "^0.31.0", optional = true }
 uvicorn = { version = "^0.23.0", extras = ["standard"], optional = true }
 
 [tool.poetry.extras]
-simulation = ["ray"]
+simulation = ["ray", "cloudpickle"]
 rest = ["requests", "starlette", "uvicorn"]
 
 [tool.poetry.group.dev.dependencies]
--- a/src/py/flwr/server/superlink/fleet/vce/backend/__init__.py
+++ b/src/py/flwr/server/superlink/fleet/vce/backend/__init__.py
@@ -18,11 +18,15 @@ import importlib
 from typing import Dict, Type
 
 from .backend import Backend, BackendConfig
+from .simplebackend import SimpleBackend
 
 is_ray_installed = importlib.util.find_spec("ray") is not None
 
 # Mapping of supported backends
 supported_backends: Dict[str, Type[Backend]] = {}
+
+# Add simplebackend
+supported_backends["simple"] = SimpleBackend
 
 # To log backend-specific error message when chosen backend isn't available
 error_messages_backends: Dict[str, str] = {}
--- a/src/py/flwr/server/superlink/fleet/vce/backend/backend.py
+++ b/src/py/flwr/server/superlink/fleet/vce/backend/backend.py
@@ -33,7 +33,7 @@ class Backend(ABC):
         """Construct a backend."""
 
     @abstractmethod
-    def build(self) -> None:
+    def build(self, app_fn: Callable[[], ClientApp]) -> None:
         """Build backend.
 
         Different components need to be in place before workers in a backend are ready
@@ -60,7 +60,6 @@ class Backend(ABC):
     @abstractmethod
     def process_message(
         self,
-        app: Callable[[], ClientApp],
         message: Message,
         context: Context,
     ) -> Tuple[Message, Context]:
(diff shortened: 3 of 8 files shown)


NicolasSchuler:
Hi @jafermarq

I've worked on an implementation like that for my OOM problem with ray. I wanted to integrate it directly into Flower and make a Pull Request, but I lacked the necessary expertise. So, I've written an engine that is "mostly" compatible with the Flower API.

I've prepared a repository with the quickstart-pytorch example and the implementation. I would be happy if some of the code would be useful to the project.

With some guidance/entrypoints I could possibly integrate the code.

NicolasSchuler:
I've also timed the different approaches:

ray-task:
8.70s user 1.93s system 23% cpu 44.286 total

process-pool:
178.94s user 21.36s system 300% cpu 1:06.71 total

flwr run:
15.24s user 2.13s system 37% cpu 46.176 total

There may be some bugs in it; for example, fraction-fit/evaluate must be 1.0 (but that is fixable).

jafermarq:
Hi @NicolasSchuler thanks for pointing me to your repo! I took a quick look and it looks pretty nice indeed. In this PR i was drafting a new `Backend` for the simulation engine that doesn't use `Ray`. I wanted to try that out because we only need a very tiny tiny portion of the functionality Ray offers to simulate a federation of N nodes, each running a `ClientApp`.  I like Ray but i'd want to have a new backend that uses just pure multiprocessing (this PR proposes one that works pretty well, but needs further testing)

I'd be delighted if you want to help in revisiting our [RayBackend](https://github.com/adap/flower/blob/9446ef284a7732a13238fe812dad1f4f259fa75b/src/py/flwr/server/superlink/fleet/vce/backend/raybackend.py#L39) class. Maybe the way we use a ActorPool could be improved. Let me know if you have some thoughts on where the memory problem you mentioned could be.

NicolasSchuler:
Hi,
I've looked at RayBackend and BasicActorPool, and I did not see anything particularly troubling. I've also experimented with using Ray Actors in the first place, but I ditched this for Ray Tasks because I could not fix the memory problem (which is also mentioned in some ray github issues). So, it's not necessarily a problem with the implementation in Flower but instead of the ray library. I will share some observations that may be useful for improving the simulation engine.

Ray-Actors:
- I tried using ray actors for a small number of clients so that an actor represents one client continuously and is not reused; this does work, but it can consume a lot of RAM, even for a small number of clients. Dynamically adjusting the number of clients according to the simulation target is not really feasible this way.
- In my experiments, the benefit of using actors compared to tasks was insignificant.
- The memory consumption usually increases relatively small for the clients, but the strategy instance kept growing more and more, ultimately resulting in the OOM.

Ray-Tasks and Process Pool
- Ray tasks reduces complexity because there is no need to manage them explicitly. 
- Tasks can utilize the max_calls argument, which I assume are reusable workers like actors. The documentation explicitly mentions it as a means to tackle OOM problems.
- It makes sense to reduce duplication of variables, e.g., when using FedAvg, the aggregated Parameters are the same, so there is no need to keep duplicates around. With Ray this is simple using the object store. With a process-pool, it makes sense to use something like shared-memory (see. repository) because it is faster than pickling the object.
- The most crucial part limiting parallelization is fitting the model, so it makes sense to have some locking to reduce the concurrency in this particular code but not in other parts.
- Just loading the model usually takes a considerable amount of time. So, caching an instantiated model makes sense.
- The Dataset also contributed significantly to the memory consumption, so caching or providing it with the ray object store also makes sense. I don't think this is possible via shared memory.

I didn't look at the Process Pool implementation right now, but one problem that is going to be addressed in future revisions of Python is changing the default multiprocessing context from "fork"->"for server." I believe it is not directly possible to access a globally shared state like it is now.

One key property I would desire from Flower is the ability to conduct a simulation with low resource requirements and allow scaling things up quickly if necessary.

PR #4012, opened 2024-08-14
fix(framework) Track `created_at` in metadata proto

jafermarq:
The `.created_at` attribute of `Metadata` is set externally when the `Message` object is constructed. As a result, we need to record its value in `proto` and slightly ament the `serde` functionality.


src/py/flwr/common/serde_test.py
```diff
@@ -427,6 +427,10 @@ def test_metadata_serialization_deserialization() -> None:
     maker = RecordMaker()
     original = maker.metadata()
 
+    # A Metadata obj is fully defined when it's used by a Message
+    # The Message sets the metadata's .created_at
+    _ = Message(metadata=original, content=maker.recordset(1, 0, 0))
```
jafermarq, line 432: A `Metadata` object is not fully defined if it's not used to construct a `Message`. that's because the `Message` constructor sets the metadata`.created_at`. Because of this reason, this test is identical to the next one so I vote for its removal. @chongshenng @danieljanes

src/py/flwr/common/serde_test.py
```diff
@@ -425,15 +425,19 @@ def test_metadata_serialization_deserialization() -> None:
     """Test serialization and deserialization of Metadata."""
     # Prepare
     maker = RecordMaker()
-    original = maker.metadata()
+    metadata = maker.metadata()
+
+    # A Metadata obj is fully defined when it's used by a Message
+    # The Message sets the metadata's .created_at
+    mssg = Message(metadata=metadata, content=maker.recordset(1, 0, 0))
```
jafermarq, line 432: A `Metadata` object is not fully defined if it's not used to construct a `Message`. that's because the [`Message` constructor](https://github.com/adap/flower/blob/354af13409ba98f37dcd7c4dad702c76ac3b6d37/src/py/flwr/common/message.py#L222) sets the metadata`.created_at`. Because of this reason, this test is identical to the next one so I vote for its removal. @chongshenng @danieljanes

src/py/flwr/common/serde.py
```diff
@@ -761,6 +762,7 @@ def metadata_from_proto(metadata_proto: ProtoMetadata) -> Metadata:
         ttl=metadata_proto.ttl,
         message_type=metadata_proto.message_type,
     )
+    metadata.created_at = metadata_proto.created_at
```
jafermarq, line 765: This is here just because we have a test `metadata_from_proto` test. I think in a complete setup this will always be called from `message_from_proto` and therefore adding this last line to the function won't be needed

src/py/flwr/common/serde_test.py
chongshenng: Good point. +1 for this.

chongshenng:
LGTM.

src/py/flwr/common/serde_test.py
chongshenng: I've removed this.

chongshenng (approved):

PR #4013, opened 2024-08-14
feat(framework) Update communication cost calculator for LLM leaderboard

yan-gao-GY:
## Proposal

Update communication cost calculator for LLM leaderboard: the accumulated cost over FL rounds will be calculated.

Note that this PR needs to be merged after this [PR](https://github.com/adap/flower/pull/3997).


src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
```diff
@@ -0,0 +1,73 @@
...
+        return parameters_aggregated, metrics_aggregated
+
+    def test_communication_costs(self, fit_list: List[Union[FitIns, FitRes]]):
+        """Test communication costs per FL round."""
+
+        def compute_bytes(weights):
+            return sum(ele.nbytes for ele in weights)
+
+        size_bytes_list = [
+            compute_bytes(parameters_to_ndarrays(fit_ele.parameters))
```
jafermarq, line 54: I think we can make this a more lightweight process by directly counting the number of `bytes` each tensor in the parameters object has. This is an example:

```python
import numpy as np
from flwr.common import ndarrays_to_parameters

arr = np.random.randn(1000,10000)
print(arr.nbytes) # prints 80M
# Let's imagine the above is what FitRes has
p = ndarrays_to_parameters(arr)
p_bytes = sum([len(t) for t in p.tensors])
print(p_bytes) # prints 80.1M
```

`FitRes` is the actual obj being communicated so i wonder if we should count its size directly like above?

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
```diff
@@ -0,0 +1,73 @@
...
+from flwr.common import FitIns, FitRes, Parameters, log, parameters_to_ndarrays
+from flwr.server.client_manager import ClientManager
+from flwr.server.client_proxy import ClientProxy
+from flwr.server.strategy import FedAvg
+
+
+class FlowerTuneLlm(FedAvg):
+    """Customised FedAvg strategy implementation."""
+    def __init__(self, **kwargs):
+        self.curr_comm_cost = 0.0
```
jafermarq, line 15: I wonder if this as well as the `test_communication_costs` functionality should be done in a separate class. The reason for this is that imagine someone wants to implement a different strategy, then they'll need to also copy the `test_communication_costs` method and remember to do so! But if this a new class (maybe `"CommunicationsTracker"`) is available, it should be easy for people to integrate  I feel. For sure it will be easier to communicate to people that they have to make use of it. what do you think?

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
yan-gao-GY: Ok, thanks for the suggestion. i'm going to use `BytesIO` for bytes count:
```
from io import BytesIO
p_bytes = sum([BytesIO(t).getbuffer().nbytes for t in p.tensors])
```

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
yan-gao-GY: Yes good idea to have a separate class!

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
```diff
@@ -0,0 +1,79 @@
...
+class CommunicationTracker:
+    """Communication costs tracker over FL rounds."""
+    def __init__(self):
+        self.curr_comm_cost = 0.0
+
+    @staticmethod
+    def _compute_bytes(parameters):
+        return sum([BytesIO(t).getbuffer().nbytes for t in parameters.tensors])
+
+    def compute_comm_costs(self, fit_list: List[Union[FitIns, FitRes]]):
```
jafermarq, line 58: how about renaming this to simply `track`? since the class name and description already indicate it is about communication costs.

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
```diff
@@ -0,0 +1,79 @@
...
+from typing import List, Tuple, Union
+
+from flwr.common import FitIns, FitRes, Parameters, log, parameters_to_ndarrays
+from flwr.server.client_manager import ClientManager
+from flwr.server.client_proxy import ClientProxy
+from flwr.server.strategy import FedAvg
+
+
+class FlowerTuneLlm(FedAvg):
+    """Customised FedAvg strategy implementation."""
```
jafermarq, line 14:
```suggestion
    """Customised FedAvg strategy implementation.
    
    This class behaves just like FedAvg but also tracks the communication
    costs associated with a round of `fit`.
    """
```

jafermarq: I thought of extending the description. Probably it can be done better than what i suggest above.

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
```diff
@@ -10,7 +11,14 @@ from flwr.server.strategy import FedAvg
...
 class FlowerTuneLlm(FedAvg):
-    """Customised FedAvg strategy implementation."""
+    """Customised FedAvg strategy implementation.
+    
+    This class behaves just like FedAvg but also tracks the communication
+    costs associated with `fit` over FL rounds.
+    """
+    def __init__(self, **kwargs):
+        self.comm_tracker = CommunicationTracker()
+        super().__init__(**kwargs)
```
danieljanes, line 21:
```suggestion
        super().__init__(**kwargs)
        self.comm_tracker = CommunicationTracker()
```

danieljanes (approved):

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
```diff
@@ -42,23 +50,34 @@ class FlowerTuneLlm(FedAvg):
...
 
-    if comm_cost > 500:
+        self.curr_comm_cost += comm_cost
         log(
-            WARN,
-            "The total communication costs per round exceed 500 MB. "
-            "Please consider reducing it if you plan to participate "
-            "FlowerTune LLM Leaderboard.",
+            INFO,
+            "Communication budget: used %.4f MB (+%.4f MB this round) / 200,000 MB",
```
jafermarq, line 72:
```suggestion
            "Communication budget: used %.4f MB (+%.2f MB this round) / 200,000 MB",
```

jafermarq: I'm thinking 4 decimal places might bee overkill. Reducing to 2

src/py/flwr/cli/new/templates/app/code/flwr_tune/strategy.py.tpl
```diff
@@ -42,23 +50,34 @@ class FlowerTuneLlm(FedAvg):
...
 
-    if comm_cost > 500:
+        self.curr_comm_cost += comm_cost
         log(
-            WARN,
-            "The total communication costs per round exceed 500 MB. "
-            "Please consider reducing it if you plan to participate "
-            "FlowerTune LLM Leaderboard.",
+            INFO,
+            "Communication budget: used %.4f MB (+%.2f MB this round) / 200,000 MB",
```
jafermarq, line 72:
```suggestion
            "Communication budget: used %.2f MB (+%.2f MB this round) / 200,000 MB",
```

jafermarq (approved):

PR #4015, opened 2024-08-14
feat(framework) Add SuperNode Traceability

mohammadnaseri:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -43,6 +43,7 @@ from flwr.common.logger import log, warn_deprecated_feature
 from flwr.common.message import Error
 from flwr.common.retry_invoker import RetryInvoker, RetryState, exponential
 from flwr.common.typing import Run, UserConfig
+from flwr.common.supernode_tracker import SuperNodeTracker
 
 from .grpc_adapter_client.connection import grpc_adapter
 from .grpc_client.connection import grpc_connection
@@ -366,6 +367,11 @@ def _start_client_internal(
                 try:
                     # Receive
                     message = receive()
+
+                    SuperNodeTracker.record_message_metadata(
+                        "SuperLink-to-SuperNode", message.metadata.__dict__
+                    )
+
                     if message is None:
                         time.sleep(3)  # Wait for 3s before asking again
                         continue
@@ -390,6 +396,10 @@ def _start_client_internal(
                     if out_message:
                         send(out_message)
                         break
+
+                    SuperNodeTracker.record_message_metadata(
+                        "SuperNode-to-SuperLink", out_message.metadata.__dict__
+                    )
 
                     # Get run info
                     run_id = message.metadata.run_id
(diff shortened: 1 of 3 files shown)


src/py/flwr/common/supernode_tracker.py
```diff
@@ -0,0 +1,102 @@
...
+from typing import Any, Dict, List, Optional
+
+from flwr.common import Message
+from flwr.common.typing import Run
+
+
+class SuperNodeTracker:
+    """A utility class for tracking and recording SuperNode."""
+
+    def __init__(self, supernode_id: Optional[int]) -> None:
```
jafermarq, line 29: is `supernode_id` ever not a `int`?

src/py/flwr/client/app.py
```diff
@@ -450,6 +466,8 @@ def _on_backoff(retry_state: RetryState) -> None:
 
                     run.fab_id, run.fab_version = fab_id, fab_version
 
+                    supernode_tracker.record_run(runs[run_id])
```
jafermarq, line 469: Would it make sense to move this line under the `if run_id not in runs` logic so we need to keep track of existing runs in the `SupernodeTracker` class? In that way the `record_run` will only be executed when there is a new run.

src/py/flwr/common/supernode_tracker.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+from datetime import datetime
+from typing import Any, Dict, List, Optional
+
+from flwr.common import Message
+from flwr.common.typing import Run
+
+
+class SuperNodeTracker:
+    """A utility class for tracking and recording SuperNode."""
```
jafermarq, line 27: I think this line description needs a couple of extra words to say what does it track. Also, i'm not sure if "track"/"tracker" is the right term here.

src/py/flwr/common/supernode_tracker.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self, supernode_id: Optional[int]) -> None:
+        self.supernode_id = str(supernode_id)
+        self.filename = f"supernode_tracking_({supernode_id}).json"
+        self.run_ids: List[int] = []
```
jafermarq, line 32: following my comment for `client/app.py` I think this is not needed.

src/py/flwr/common/supernode_tracker.py
```diff
@@ -0,0 +1,102 @@
...
+from flwr.common import Message
+from flwr.common.typing import Run
+
+
+class SuperNodeTracker:
+    """A utility class for tracking and recording SuperNode."""
+
+    def __init__(self, supernode_id: Optional[int]) -> None:
+        self.supernode_id = str(supernode_id)
+        self.filename = f"supernode_tracking_({supernode_id}).json"
```
jafermarq, line 31: I might be wrong but appending JSON files into the same file is not a valid JSON file. I think the file extension should be something else.

src/py/flwr/client/app.py
mohammadnaseri: I see your point. For now I would say let's leave this one, because it is a bit tricky to find the right place for doing the check `if run_id not in runs`. It can not be in line `468` because at that line, the `run_id` is already added to the `runs` and it won't be tracked.

src/py/flwr/common/supernode_tracker.py
mohammadnaseri: I addressed this above.

PR #4034, opened 2024-08-17
feat(framework) Add `GetToken` gRPC request to `ClientAppIoServicer`

jafermarq:
Introduced `GetToken` RCP that works as follows (not used):
- `flwr_clientapp` sends `GetTokenRequest` w/o any argument
- `flwr_clientapp` gets a `GetTokenResponse` from SuperNode with a token (if set). If no token is in the response (i.e. token=None) the idea would be to use this to keep looping until a `GetTokenResponse` has a token.


src/py/flwr/client/clientapp/clientappio_servicer.py
```diff
@@ -68,6 +70,7 @@ class ClientAppIoServicer(clientappio_pb2_grpc.ClientAppIoServicer):
     def __init__(self) -> None:
         self.clientapp_input: Optional[ClientAppIoInputs] = None
         self.clientapp_output: Optional[ClientAppIoOutputs] = None
+        self.token: Optional[int] = None
```
danieljanes, line 73: This is not needed, the token is already part of `ClientAppIoInputs`. We should, however, track whether the token has been delivered or not.

src/py/flwr/client/clientapp/clientappio_servicer.py
```diff
@@ -122,6 +125,24 @@ def PushClientAppOutputs(
...
+            # If token is set, use it in response
+            res.token = self.token
+            # Resetting token
+            self.token = None
+        return res
+
+    def set_token(self, token: int) -> None:
+        """Set the token."""
+        log(DEBUG, "ClientAppIo.set_token")
+        self.token = token
```
danieljanes, line 144: Related to the comment above, this can probably be removed since the token gets set implicitly via `set_inputs`.

danieljanes (dismissed):

danieljanes (approved):

PR #4035, opened 2024-08-17
feat(framework) Use `GetToken` functionality

jafermarq:
The `flwr-clientapp` doesn't receive a token at launch time. Instead it requests it to the `SuperNode`.


src/py/flwr/client/clientapp/app.py
```diff
@@ -149,6 +156,7 @@ def run_clientapp(  # pylint: disable=R0914
 
 def get_token(stub: grpc.Channel) -> Optional[int]:
     """Get a token from SuperNode."""
+    log(DEBUG, "Flower ClientApp requests token")
```
danieljanes, line 159:
```suggestion
    log(DEBUG, "Flower ClientApp process requests token")
```

danieljanes (approved):

PR #4045, opened 2024-08-19
refactor(framework:skip) Rearrange the client app code

mohammadnaseri:
## Issue
As a part of traceability feature, we need to first track the run before any message exchange in [this](https://github.com/adap/flower/blob/49027de9d6f1870856952ca9430b6dfc910f3670/src/py/flwr/client/app.py#L196C5-L196C26) method. For that reason, this PR, rearrange the method flow.

### DIFF
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -422,12 +422,6 @@ def start_client_internal(
                         message.metadata.message_id,
                     )
 
-                    # Handle control message
-                    out_message, sleep_duration = handle_control_message(message)
-                    if out_message:
-                        send(out_message)
-                        break
-
                     # Get run info
                     run_id = message.metadata.run_id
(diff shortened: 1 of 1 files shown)


danielnugraha:
Looks okay, the only concern is that the control message is sent after getting `run_id`, but this is only potentially affects `grpc_bidi` legacy stack.

For traceability, I think the right entry points to track are the methods inside the `connection.py`, not in `client/app.py`.

mohammadnaseri:
> Looks okay, the only concern is that the control message is sent after getting `run_id`, but this is only potentially affects `grpc_bidi` legacy stack.
> 
> For traceability, I think the right entry points to track are the methods inside the `connection.py`, not in `client/app.py`.

Regarding `grpc_bidi `, I have tested it with the `test_legacy.sh` and it worked alright.

PR #4051, opened 2024-08-21
feat(datasets) Add grouped natural id partitioner

adam-narozniak:
## Proposal
This PR adds `GroupedNaturalIdPartitioner`. Let's imagine a dataset like `femnist` that has an innate notion of `user_id`. The `NaturalIdPartitioner` creates each partition so that each partition represents a single user. This Partitioner "merges" a few of such unique users such that a single partition contains a few unique users. (It's controlled by group size.)

### Explanation

This is a complete rewrite of the initial idea proposed by an external contribution due to the code quality.


datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner_test.py
```diff
@@ -0,0 +1,248 @@
...
+        ]
+    )
+    def test_allow_bigger_mode_num_partitions_and_partition_sizes(
+        self,
+        num_rows: int,
+        num_unique_natural_id: int,
+        group_size: int,
+        expected_num_unique_natural_ids: List[int],
+    ) -> None:
+        """Test allow-smaller mode handles the remainder correctly."""
```
jafermarq, line 122:
```suggestion
        """Test allow-bigger mode handles the remainder correctly."""
```

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        unique_natural_ids = self.dataset.unique(self._partition_by)
+        if self._sort_unique_ids:
+            unique_natural_ids = sorted(unique_natural_ids)
+        num_unique_natural_ids = len(unique_natural_ids)
+        remainder = num_unique_natural_ids % self._group_size
+
+        if self._mode == "allow-bigger":
+            num_groups = num_unique_natural_ids // self._group_size
+            groups_of_natural_ids = np.array_split(unique_natural_ids, num_groups)
```
jafermarq, line 96: I wonder if we should consider that `num_groups` could be 1 here if `self._group_size` is over half of the number of `num_unique_natural_ids`. Something like this could handle it?
```python
groups_of_natural_ids = np.array_split(unique_natural_ids, (self._group_size,) if num_groups < 0 else num_groups)
```

Quick test:
```python
arr = np.ones(10,)
group_size = 6
num_groups = len(arr) // group_size # will be 1
np.array_split(arr, num_groups)
>>> [array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])] # current behaviour
np.array_split(arr, (group_size,) if num_groups == 1 else num_groups)
>>> [array([1., 1., 1., 1., 1., 1.]), array([1., 1., 1., 1.])] # better?
```

jafermarq: if you think this is worth adding, the rest of the modes will need also some slight modification

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
```diff
@@ -0,0 +1,213 @@
...
+        elif self._mode == "drop-reminder":
+            num_groups = num_unique_natural_ids // self._group_size
+            # Narrow down the unique_natural_ids to not have a bigger group
+            # which is the behavior of the np.array_split
+            unique_natural_ids = unique_natural_ids[
+                : int(num_groups * self._group_size)
+            ]
+            groups_of_natural_ids = np.array_split(unique_natural_ids, num_groups)
+        elif self._mode == "allow-smaller":
+            num_groups = num_unique_natural_ids // self._group_size
```
jafermarq, line 106: minor observation: it seems this computation is needed for all models -> should we move it to right under the `reminder` calculation?

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
```diff
@@ -0,0 +1,213 @@
...
+            groups_of_natural_ids = np.array_split(unique_natural_ids, num_groups)
+        else:
+            raise ValueError(
+                f"Given {self._mode} is not a valid mode. Refer to the documentation of"
+                " the mode parameter for the available modes."
+            )
+
+        self._partition_id_to_natural_ids = {}
+        for group_of_natural_ids_id, group_of_natural_ids in zip(
+            range(len(groups_of_natural_ids)), groups_of_natural_ids
```
jafermarq, line 134: is this eq to `enumerate(groups_of_natural_ids)`?

jafermarq (commented):
hey @adam-narozniak, i left a couple of comments but i think only one is important (the one about `num_groups=1`). I'll check the tests after we decide how to handle it

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
```diff
@@ -0,0 +1,213 @@
...
+    >>> partition = fds.load_partition(0)
+    """
+
+    def __init__(
+        self,
+        partition_by: str,
+        group_size: int,
+        mode: Literal[
+            "allow-smaller", "allow-bigger", "drop-reminder", "strict"
+        ] = "allow-bigger",
```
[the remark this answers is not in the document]
adam-narozniak: That's a good question. I went for "allow-bigger" because it's consistent with the default behavior of np.array_split. But that's not a reasonable or thought-through argument; I'm open to changing it. Wdyt?

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
adam-narozniak: Yes, thanks. I'll change it.

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
adam-narozniak: Yes, thanks. I'll change it.

adam-narozniak: Done

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
adam-narozniak: Well, the current behavior depends on the mode.  Let's keep as you assumed the num of unique ID in ds is 10 (but it should be represented as arange(10) instead of ones(10):
then: 
`mode="allow-smaller",` Gives 
```
pid = 0-->{0, 1, 2, 3, 4, 5, 6, 7}
pid = 1-->{8, 9}
```
`mode=allow-bigger`
```
pid = 0-->{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}

```

`mode=drop-reminder`
```
pid = 0-->{0, 1, 2, 3, 4, 5, 6, 7}
```
`mode=strict`
```
raises
```

adam-narozniak: So, unless I don't understand something this is a desired behavior. But I'm happy to learn otherwise.

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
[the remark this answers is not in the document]
jafermarq: yeah. I think this is reasonable. happy to keep it as the default (in the end this shouldn't make a big diff generally speaking)

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
jafermarq: all is good. I was just expecting in such situation to get an output as what mode `allow-smaller` does. I think it's the right decision to make it the default mode (which is done already) 👍

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
```diff
@@ -0,0 +1,219 @@
...
+            raise ValueError(
+                "The group size needs to be smaller than the number of the unique "
+                "natural ids unless you are using allow-smaller mode which will "
+                "result in a single partition."
+            )
+        if self._sort_unique_ids:
+            unique_natural_ids = sorted(unique_natural_ids)
+        num_unique_natural_ids = len(unique_natural_ids)
+        remainder = num_unique_natural_ids % self._group_size
+        num_groups = num_unique_natural_ids // self._group_size
```
jafermarq, line 100: When `group_size` is bigger than `num_unique_natural_ids` (something allowed for `allow-smaller`) the resulting  `num_groups = 0`. This will make `np.array_split(...)` call raise an error.

jafermarq: I wonder if we should do instead

```python
num_groups = num_unique_natural_ids // self._group_size
```
to this:
```
num_groups = max(1, num_unique_natural_ids // self._group_size)
```

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner_test.py
```diff
@@ -0,0 +1,270 @@
...
+    def test_too_big_group_size(self) -> None:
+        """Test raises if the group size > than the number of unique natural ids."""
+        n_unique_natural_ids = 3
+        dataset = _create_dataset(
+            num_rows=10, n_unique_natural_ids=n_unique_natural_ids
+        )
+        partitioner = GroupedNaturalIdPartitioner(
+            partition_by="natural_id",
+            group_size=n_unique_natural_ids + 1,
+            mode="allow-bigger",
```
jafermarq, line 255: My understanding is that `allow-smaller` in an scenario like the one tested here. It doesn't work because `num_groups` will be 0. This links to a previous comment in `_create_int_partition_id_to_natural_id()`

datasets/flwr_datasets/partitioner/grouped_natural_id_partitioner.py
adam-narozniak: Thanks for this catch. I applied a bit differently because this would produce an overlapping sample; I additionally zeroed the remainder. Why? Not to trigger the additional appends based on the reminder later in the code.
```
        if num_groups == 0 and self._mode == "allow-smaller":
            num_groups = 1
            remainder = 0
```
I've added a new test to check this scenario and extended the overlap test to catch something like that.

jafermarq (dismissed):
@CarlosMari  @adam-narozniak, this looks like a nice addition to `flower-datasets` 🚀!

jafermarq (approved):

PR #4054, opened 2024-08-22
feat(framework:skip) Add the abstract base class `Connection`

panh99:


chongshenng (changes requested):
Only tiny comments, lgtm otherwise!

src/py/flwr/client/connection/fleet_connection.py
```diff
@@ -0,0 +1,116 @@
...
+            Flower server. Bytes won't work for the REST API.
+        authentication_keys : Optional[Tuple[PrivateKey, PublicKey]] (default: None)
+            Tuple containing the elliptic curve private key and public key for
+            authentication from the cryptography library.
+            Source: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/
+            Used to establish an authenticated connection with the server.
+        """
+        self.server_address = server_address
+        self.insecure = insecure
+        self.retrier = retry_invoker
```
chongshenng, line 73:
```suggestion
        self.retry_invoker = retry_invoker
```
It's a minor thing, but should we also call it `self.retry_invoker` for consistency?

PR #4057, opened 2024-08-22
doc(datasets) Update information how to handle DatasetDict local data

adam-narozniak:
## Issue
The description of how to use the local dataset is not entirely accurate. 

### Description
One way of loading the data shows the result that produces a `DatasetDict`.

## Proposal
Add a comment that clarifies how to correctly handle it.

### DIFF
--- a/datasets/doc/source/how-to-use-with-local-data.rst
+++ b/datasets/doc/source/how-to-use-with-local-data.rst
@@ -44,6 +44,8 @@ CSV
     "can-have-more-splits": ...
   }
   dataset = load_dataset("csv", data_files=data_files)
+  # if data_files is a dictionary choose the split e.g. "train"
+  # dataset = load_dataset("csv", data_files=data_files, split="train")
 
   partitioner = ChosenPartitioner(...)
   partitioner.dataset = dataset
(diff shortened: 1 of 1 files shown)


datasets/doc/source/how-to-use-with-local-data.rst
```diff
@@ -44,6 +44,8 @@ CSV
     "can-have-more-splits": ...
   }
   dataset = load_dataset("csv", data_files=data_files)
+  # if data_files is a dictionary choose the split e.g. "train"
+  # dataset = load_dataset("csv", data_files=data_files, split="train")
```
jafermarq, line 48: I'm unsure why this is needed here. Wouldn't it be more relevant to mention when constructing the partitioner? 

The comment seems to indicate that when what gets passes is a dic, you need to specify a split. But this is optional.

datasets/doc/source/how-to-use-with-local-data.rst
```diff
@@ -71,6 +73,8 @@ JSON
      "can-have-more-splits": ...
   }
   dataset = load_dataset("json", data_files=data_files)
+  # if data_files is a dictionary choose the split e.g. "train"
+  # dataset = load_dataset("json", data_files=data_files, split="train")
```
jafermarq, line 77: same as in my other comment.

jafermarq (approved):

PR #4078, opened 2024-08-23
feat(framework) Add `replace_keys` function to replace `-` with `_` in configs

yan-gao-GY:
## Proposal

Add `replace_keys` function to replace `-` with `_` in configs


src/py/flwr/common/config.py
```diff
@@ -159,15 +159,17 @@ def flatten_dict(
     return dict(items)
 
 
-def unflatten_dict(flat_dict: Dict[str, Any]) -> Dict[str, Any]:
-    """Unflatten a dict with keys containing separators into a nested dict."""
+def unflatten_dict(flat_dict: Dict[str, Any], replace: bool = True) -> Dict[str, Any]:
```
jafermarq, line 162: I'm thinking it would be better to leave `replace` as `false`

src/py/flwr/common/config_test.py
```diff
@@ -232,8 +232,8 @@ def test_flatten_dict() -> None:
 
 def test_unflatten_dict() -> None:
```
jafermarq, line 233: should we instead add a new test just for when `replace = True`?

src/py/flwr/common/config.py
```diff
@@ -212,3 +212,16 @@ def get_metadata_from_config(config: Dict[str, Any]) -> Tuple[str, str]:
         config["project"]["version"],
         f"{config['tool']['flwr']['app']['publisher']}/{config['project']['name']}",
     )
+
+
+def replace_keys(input_dict: Dict[str, Any], match: str, target: str) -> Dict[str, Any]:
+    """Recursively replace match string with target string in dictionary keys, including
+    nested dictionaries."""
```
jafermarq, line 219:
```suggestion
    """Recursively replace match string with target string in dictionary keys."""
```

PR #4000, opened 2024-08-13
break(framework) Use spaces instead of commas for separating config args

charlesbvll:
### Checklist

- [ ] Implement proposed change
- [ ] Write tests
- [ ] Update [documentation](https://flower.ai/docs/writing-documentation.html)
- [ ] Make CI checks pass
- [ ] Ping maintainers on [Slack](https://flower.ai/join-slack/) (channel `#contributions`)

PR #4080, opened 2024-08-24
fix(framework:skip) Support overriding run config from a `TOML`

jafermarq:
With PR #4000 we unintendedly removed support for overriding the base config in `pyproject.toml` by passing a new toml. Like this:

```bash
flwr run . --run-config my-other-config.toml
```

This PR brings that functionality back and includes a test for it.


src/py/flwr/common/config.py
```diff
@@ -188,20 +188,17 @@ def parse_config_args(
     # Regular expression to capture key-value pairs with possible quoted values
     pattern = re.compile(r"(\S+?)=(\'[^\']*\'|\"[^\"]*\"|\S+)")
 
+    # If a .toml is passed, all other keys are ignored
```
danieljanes, line 191: Instead of doing this, shouldn't we raise an error when there's a TOML file and also other keys?

danieljanes (approved):

danieljanes (approved):

PR #4087, opened 2024-08-27
Add conn tests

panh99:
Based on:
1. #4054
2. #4055
3. #4056

- Added generic unit tests for `grpc-rere`/`grpc-adapter`/`rest` transport types.


chongshenng (changes requested):
Thanks @panh99, LGTM! I think it's worth having a set of tests with secure connectivities, though this would be worthy of another PR. Wdyt?

PR #4104, opened 2024-08-29
feat(datasets) Add examples section to concatenate divisions

adam-narozniak:
## Issue
The `concatenate_divisions` is the last function that does not have the Examples section in FDS.

## Proposal
Add two examples to the `Examples` docs section of the `concatenate_divisions` function.


datasets/flwr_datasets/utils.py
```diff
@@ -268,11 +269,11 @@ def concatenate_divisions(
     partition_division: Union[List[float], Tuple[float, ...], Dict[str, float]],
     division_id: Union[int, str],
 ) -> Dataset:
-    """Create a dataset by concatenation of all partitions in the same division.
+    """Create a dataset by concatenation of divisions from all partitions.
```
jafermarq, line 272: Would it make sense to skip the term "division" here? Since the partitioner already operates at the "division" level.

adam-narozniak: Formally, I think of it this way. There's a dataset that has splits. A split can be partitioned into partitions. The partition can be divided into divisions. So I haven't used the word division in the context of split -> partitions. This is in line with the argument name in this function.

jafermarq (approved):

PR #4119, opened 2024-09-01
refactor(framework) Add buffer time after `ServerApp` thread initialization

jafermarq:
With certain libraries with longer import times (e.g. `transformers`), if the thread running the `ServerApp` hasn't completed its import by the time the `ClientApp` begins importing the same module, it can lead to crashes. 

This PR presents a temporary fix for this issue by leaving some "buffer" time for the `ServerApp` thread to initialise. A better solution would be to run the `ServerApp` in a different process (but this needs of a more substantial refactoring)

@yan-gao-GY , I believe this should solve the import issues you discovered


yan-gao-GY:
Hi @jafermarq , does this work on your side when running the llm example? I moved the `set_parameters` and `get_parameters` from `client_app.py` to `models.py` (in this branch `update-llm-example`), and set `launch_buffer=60`, but it still got the same error.

jafermarq:
> Hi @jafermarq , does this work on your side when running the llm example? I moved the `set_parameters` and `get_parameters` from `client_app.py` to `models.py` (in this branch `update-llm-example`), and set `launch_buffer=60`, but it still got the same error.

Thanks for checking @yan-gao-GY . Just to be sure, does your branch use the updated behaviour in `run_simulation.py`? Note that i have moved the position of the sleep after the thread is started.

yan-gao-GY:
@jafermarq Oh I didn't notice the position of the sleep also changed. Let me try again.

yan-gao-GY:
@jafermarq Yes it's working fine~

yan-gao-GY (dismissed):

src/py/flwr/simulation/run_simulation.py
```diff
@@ -610,6 +615,13 @@ def _parse_args_run_simulation() -> argparse.ArgumentParser:
         "Read more about how `tf.config.experimental.set_memory_growth()` works in "
         "the TensorFlow documentation: https://www.tensorflow.org/api/stable.",
     )
+    parser.add_argument(
+        "--delay-start",
+        type=int,
+        default=5,
```
jafermarq, line 621:
```suggestion
        default=3,
```

PR #4127, opened 2024-09-02
fix(framework) Adjust framework name in templates docstrings

jafermarq:
With the changes in #3894 we dynamically indicate the `framework` used to generate each template. However, it uses an all-caps representation. This PR sets the framework string as defined in the `MlFramework` enum.

Effectively this means the Flower apps will be created showing:

```python
"""my_test_app: A Flower / NumPy app."""
```

instead of:

```python
"""my_test_app: A Flower / NUMPY app."""
```


src/py/flwr/cli/new/new.py
```diff
@@ -136,23 +136,15 @@ def new(
...
-            if value == framework_value
-        ]
-        framework_str_upper = selected_value[0]
-
-    framework_str = framework_str_upper.lower()
 
     llm_challenge_str = None
-    if framework_str == "flowertune":
+    if framework_str == MlFramework.FLOWERTUNE:
         llm_challenge_value = prompt_options(
```
charlesbvll, line 148: PR looks good! Could we also apply similar changes to this block?

jafermarq: @charlesbvll this is done

charlesbvll (approved):

src/py/flwr/cli/new/new.py
```diff
@@ -235,44 +224,42 @@ def new(
...
+            MlFramework.TENSORFLOW.value,
         ]
         if framework_str in frameworks_with_tasks:
             files[f"{import_name}/task.py"] = {
-                "template": f"app/code/task.{framework_str}.py.tpl"
+                "template": f"app/code/task.{template_name}.py.tpl"
             }
 
-        if is_baseline_project:
+        if framework_str == "baseline":
```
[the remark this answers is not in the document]
jafermarq: No, because the name in `MlFramework.BASELINE` is the "display" name shown in the terminal as  an option to choose from (i..e `Flower Baseline`) which is different from the "tag" we use to select the templates (which is just `baseline`)

danieljanes (approved):

PR #4133, opened 2024-09-04
fix(framework:skip) Update docstrings examples in `RecordSet`

jafermarq:
import names shown in the docstrings examples were incorrect.


tanertopal (approved):
Lgtm!

PR #4143, opened 2024-09-05
refactor(framework) Handle unsuitable resources for simulation

jafermarq:
Moved creation of actor pool in `RayBackend` to the backend's `.build()` method. This is in not only a better place "semantically" speaking for constructing the pool of actors, but also prevents the backend obj construction to fail. This is important since, if the actor pool construction fails (e.g. if not resources area available in the system), it still enables the graceful termination of the simulation engine.


src/py/flwr/simulation/ray_transport/ray_actor.py
```diff
@@ -124,14 +124,14 @@ def pool_size_from_resources(client_resources: Dict[str, Union[int, float]]) ->
             WARNING,
             "The ActorPool is empty. The system (CPUs=%s, GPUs=%s) "
             "does not meet the criteria to host at least one client with resources:"
-            " %s. Lowering the `client_resources` could help.",
+            " %s. Lowering the these resources could help.",
```
danieljanes, line 127:
```suggestion
            " %s. Lowering these resources could help.",
```

danieljanes (approved):

PR #4146, opened 2024-09-06
feat(framework) Implement `keys/values/items` methods for `TypedDict`

panh99:
Previously, as pointed out by @jafermarq , the string representation of entire `TypedDict` will show up when printing out the return of `.keys()` / `.values()`, as follows:
image

To optimize the string representations, I implemented `keys()`, `values()` and `items()` methods for `TypedDict` instead of just inheriting them from the superclass `MutableMapping`. Now, we have:
image


src/py/flwr/common/record/typeddict.py
```diff
@@ -73,3 +84,15 @@ def __eq__(self, other: object) -> bool:
         if isinstance(other, dict):
             return data == other
         return NotImplemented
+
+    def keys(self) -> KeysView[K]:
+        """D.keys() -> a set-like object providing a view on D's keys."""
```
jafermarq, line 89:
```suggestion
        """Provide a view on the dictionary's keys."""
```

panh99: Hmmmm, my intention is to keep `TypedDict` as close to Python `dict` as possible, and thus I copied and pasted the docstring for `dict.keys`. But I am also happy to make the docstring clearer. Wdyt?

jafermarq: oh, you are right! I guess i never paid much attention to the docstrings for `.key()` etc. Let's keep it as you had 💯

jafermarq (approved):

danieljanes (approved):

PR #4163, opened 2024-09-09
fix(framework:skip) Specify keyword `deterministic`  in `SerializeToString` method

panh99:
## Motivation
As explained in [google protobuf documentation](https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.SerializeToString), the `SerializeToString` method only accept keyword arguments. In some settings, as spotted by Robert and Danny, the following exception will be raised when calling `SerializeToString(True)`, i.e., a positional argument instead of a keyword argument is passed.
```
TypeError: _AddSerializeToStringMethod.<locals>.SerializeToString() takes 1 positional argument but 2 were given
```

## Solution
Change `SerializeToString(True)` to `SerializeToString(deterministic=True)`.


danieljanes (approved):

tanertopal (approved):
Lgtm!

PR #4170, opened 2024-09-10
refactor(framework) Migrate ID handling from sint64 to uint64

mohammadnaseri:
This PR updates the ID generation to use `uint64`. To ensure compatibility with SQLite, which uses `sint64`, the IDs are converted as follows:

During Save: Convert `uint64` IDs to `sint64` before storing in SQLite.
During Read: Convert `sint64` IDs back to `uint64` after retrieving from SQLite.


src/py/flwr/server/superlink/state/utils_test.py
```diff
@@ -0,0 +1,59 @@
...
+        """Test conversion from uint64 to sint64."""
+        # Test values below 2^63
+        self.assertEqual(convert_uint64_to_sint64(0), 0)
+        self.assertEqual(convert_uint64_to_sint64(2**62), 2**62)
+        self.assertEqual(convert_uint64_to_sint64(2**63 - 1), 2**63 - 1)
+
+        # Test values at and above 2^63
+        self.assertEqual(convert_uint64_to_sint64(2**63), -(2**63))
+        self.assertEqual(convert_uint64_to_sint64(2**63 + 1), -(2**63) + 1)
+        self.assertEqual(convert_uint64_to_sint64(2**64 - 1), -1)
```
chongshenng, line 39: Perhaps it'd be useful/cleaner to use `parameterized` here (we've to add `parameterized` in our `dev` dependencies), e.g.
```python
    @parameterized.expand(
        [
            # Test values below 2^63
            (0, 0),
            (2**62, 2**62),
            (2**63 - 1, 2**63 - 1),
            # Test values at and above 2^63
            (2**63, -(2**63)),
            (2**63 + 1, -(2**63) + 1),
            (2**64 - 1, -1),
        ]
    )
    def test_convert_uint64_to_sint64(self, before: int, after: int) -> None:
        """Test conversion from uint64 to sint64."""
        self.assertEqual(convert_uint64_to_sint64(before), after)
```
This follows the test structure in Flower Datasets, like [here](https://github.com/adap/flower/blob/630d78775ea5c17e63ff6a486ae5ac6241a66d31/datasets/flwr_datasets/partitioner/dirichlet_partitioner_test.py#L56-L62).

chongshenng (changes requested):
Overall looks fine to me. Minor comment about whether we should use `parameterize` for the test cases. I'm unfamiliar with `sqlite_state.py` so I did not review it. Perhaps @panh99 can give feedback there?

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +34,66 @@
...
+
+def convert_uint64_to_sint64(u: int) -> int:
+    """Convert a uint64 value to sint64.
+
+    Args:
+        u (int): The unsigned 64-bit integer to convert.
+
+    Returns
+    -------
+        int: The signed 64-bit integer equivalent.
```
panh99, line 49: Could you update the docstring and make it consistent with other docstring in our codebase? Specifically, could you use `Parameters` instead of `args:`, and update the docstring for each argument/return? For example,

```
    # pylint: disable-next=too-many-locals
    def invoke(
        self,
        target: Callable[..., Any],
        *args: Any,
        **kwargs: Any,
    ) -> Any:
        """Safely invoke the provided callable with retry mechanisms.

        This method attempts to invoke the given callable, and in the event of
        a recoverable exception, employs a retry mechanism that considers
        wait times, jitter, maximum attempts, and maximum time. During the
        retry process, various callbacks (`on_backoff`, `on_success`, and
        `on_giveup`) can be triggered based on the outcome.

        Parameters
        ----------
        target: Callable[..., Any]
            The callable to be invoked.
        *args: Tuple[Any, ...]
            Positional arguments to pass to `target`.
        **kwargs: Dict[str, Any]
            Keyword arguments to pass to `target`.

        Returns
        -------
        Any
            The result of the given callable invocation.

        Raises
        ------
        Exception
            If the number of tries exceeds `max_tries`, if the total time exceeds
            `max_time`, if `wait_gen_factory()` generator raises `StopInteration`,
            or if the `should_giveup` returns True for a raised exception.

        Notes
        -----
        The time between retries is determined by the provided `wait_gen_factory()`
        generator and can optionally be jittered using the `jitter` function.
        The recoverable exceptions that trigger a retry, as well as conditions to
        stop retries, are also determined by the class's initialization parameters.
        """
```

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +34,66 @@
...
+    """
+    if u >= 2**63:
+        return u - 2**64
```
panh99, line 52:
```suggestion
    if u >= (1 << 63):
        return u - (1 << 64)
```

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +34,66 @@
...
+
+    Args:
+        s (int): The signed 64-bit integer to convert.
+
+    Returns
+    -------
+        int: The unsigned 64-bit integer equivalent.
+    """
+    if s < 0:
+        return s + 2**64
```
panh99, line 67:
```suggestion
    if s < 0:
        return s + (1 << 64)
```

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +34,66 @@
...
+
+def convert_sint64_to_uint64(s: int) -> int:
+    """Convert a sint64 value to uint64.
+
+    Args:
+        s (int): The signed 64-bit integer to convert.
+
+    Returns
+    -------
+        int: The unsigned 64-bit integer equivalent.
```
panh99, line 64: Same as above.

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +34,66 @@
...
+
+
+def convert_uint64_values_in_dict_to_sint64(
+    data_dict: Dict[str, int], keys: List[str]
+) -> None:
+    """Convert uint64 values to sint64 in the given dictionary.
+
+    Args:
+        data_dict: A dictionary where the values are integers to be converted.
+        keys: A list of keys in the dictionary whose values need to be converted.
```
panh99, line 78: Same as above.

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +34,66 @@
...
+
+
+def convert_sint64_values_in_dict_to_uint64(
+    data_dict: Dict[str, int], keys: List[str]
+) -> None:
+    """Convert sint64 values to uint64 in the given dictionary.
+
+    Args:
+        data_dict: A dictionary where the values are integers to be converted.
+        keys: A list of keys in the dictionary whose values need to be converted.
```
panh99, line 92: Same as above.

src/py/flwr/server/superlink/state/sqlite_state.py
```diff
@@ -704,29 +779,34 @@ def get_node_public_keys(self) -> Set[bytes]:
...
-                run_id=run_id,
+                run_id=convert_sint64_to_uint64(row["run_id"]),
                 fab_id=row["fab_id"],
                 fab_version=row["fab_version"],
                 fab_hash=row["fab_hash"],
                 override_config=json.loads(row["override_config"]),
             )
         except sqlite3.IntegrityError:
-            log(ERROR, "`run_id` does not exist.")
+            log(ERROR, "`sint64_run_id` does not exist.")
```
panh99, line 794: I think it's not neccessary to mention it's sint64 here as it's just an implementation detail. This might confuse users who host their own SuperLink.

```suggestion
            log(ERROR, "`run_id` does not exist.")
```

panh99 (changes requested):
It looks good to me generally. However, I am a bit concerned about how users gonna send/receive these IDs. For now, our major grpc messages like `Scalar` and `ConfigsRecordValue` do not support `uint64`. They only support `sint64`. This might prevent our users from storing IDs, e.g., `node_id`, in grpc messages. We may need to update `Scalar`, `ConfigsRecordValue` definitions and their serialization functions in `serde.py` accordingly.

pyproject.toml
```diff
@@ -101,6 +101,7 @@ docformatter = "==1.7.5"
 mypy = "==1.8.0"
 pylint = "==3.0.3"
 flake8 = "==5.0.4"
+parameterized = "==0.9.0"
```
panh99, line 104: It would be better to amend project dependencies in a separate PR. @danieljanes Do you think we should add `parameterized` to our dev env?

mohammadnaseri: @panh99 We need the `parameterized` for the unit tests in this PR.

panh99: Yeah, I understand. Just talked to Daniel, and I will open a separate PR for adding the dependency. It's generally good to keep each PR focusing on a single topic.

src/proto/flwr/proto/recordset.proto
```diff
@@ -47,16 +48,18 @@ message ConfigsRecordValue {
     // Single element
     double double = 1;
     sint64 sint64 = 2;
-    bool bool = 3;
-    string string = 4;
-    bytes bytes = 5;
+    uint64 uint64 = 3;
```
danieljanes, line 51: @mohammadnaseri why do we need this?

panh99: @danieljanes we are still iterating on this PR. Is it b/c IDs are available to our users. However, if users want to store these IDs inside the messages, they may get an exception saying the IDs are too large for `sint64`.

src/proto/flwr/proto/transport.proto
```diff
@@ -107,7 +107,7 @@ message Scalar {
     // int32 int32 = 3;
     // int64 int64 = 4;
     // uint32 uint32 = 5;
-    // uint64 uint64 = 6;
+    uint64 uint64 = 6;
```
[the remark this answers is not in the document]
mohammadnaseri: I resolve this for now as we will create a separate PR for that.

panh99 (dismissed):

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +33,88 @@
...
+
+    Parameters
+    ----------
+    u : int
+        The unsigned 64-bit integer to convert.
+
+    Returns
+    -------
+    int
+        The signed 64-bit integer equivalent.
```
danieljanes, line 51:
```suggestion
        The signed 64-bit integer with the same bit pattern as the unsigned 64-bit int, but a potentially different decimal value. The decimal value will be the same for all numbers in the range [0, sint64 max value], but different for all numbers greater than the sint64 max value.
```

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +33,104 @@
...
 
 def generate_rand_int_from_bytes(num_bytes: int) -> int:
-    """Generate a random `num_bytes` integer."""
-    return int.from_bytes(urandom(num_bytes), "little", signed=True)
+    """Generate a random unsigned integer from `num_bytes` bytes."""
+    return int.from_bytes(urandom(num_bytes), "little", signed=False)
+
+
+def convert_uint64_to_sint64(u: int) -> int:
+    """Convert a uint64 value to sint64.
```
danieljanes, line 41:
```suggestion
    """Convert a uint64 value to sint64 value with the same bit sequence.
```

src/py/flwr/server/superlink/state/utils.py
```diff
@@ -33,8 +33,104 @@
...
+        max value, the decimal value will differ due to the wraparound caused
+        by the sign bit.
+    """
+    if u >= (1 << 63):
+        return u - (1 << 64)
+    return u
+
+
+def convert_sint64_to_uint64(s: int) -> int:
+    """Convert a sint64 value to uint64.
```
danieljanes, line 67:
```suggestion
    """Convert a sint64 value to a uint64 value with the same bit sequence.
```

danieljanes (approved):

PR #4178, opened 2024-09-11
refactor(framework) Move all address constants to `flwr.common.constant`

panh99:
We have introduced many different services, which requires us to manage all default addresses in a single location to prevent potential conflicts.

In this PR:
- Moved all IP address-related constants to `flwr.common.constant` for centralized management.
- Updated the relevant files to import these constants from `flwr.common.constant`.
- Enforced consistent naming.


src/py/flwr/common/constant.py
```diff
@@ -37,7 +37,16 @@
...
+# Addresses
+# SuperNode
+CLIENTAPPIO_API_DEFAULT_ADDRESS = "[::]:9094"
+# SuperExec
+EXEC_API_DEFAULT_ADDRESS = "[::]:9093"
+# SuperLink
+DRIVER_API_DEFAULT_ADDRESS = "[::]:9091"
+FLEET_API_GRPC_RERE_DEFAULT_ADDRESS = "[::]:9092"
+FLEET_API_GRPC_BIDI_DEFAULT_ADDRESS = "[::]:8080"
+FLEET_API_REST_DEFAULT_ADDRESS = "[::]:9093"
```
danieljanes, line 49: Let's continue to use the same defaults as before, and not switch the IPv4 ones to IPv6. We defaulted to IPv6 in earlier versions of Flower, but it caused some unforeseen issues.

danieljanes (approved):

PR #4207, opened 2024-09-13
fix(baselines) Fix dimension order in weight slicing for OD layers

QVQZZZ:
### Description
Weight slicing for OD layers has the input and output dimensions reversed.

In the weight parameters of Linear and Conv layers, the first dimension should be `out_dim`, and the second dimension should be `in_dim`.

### Explanation
Reversed the order of input and output dimensions in the weight slicing for OD layers.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


jafermarq:
Thanks for opening the PR @QVQZZZ , let me loop in @stevelaskaridis and @SamuelHorvath who are the masterminds behind FjORD and added the baseline: what do you think about this 2-lines change?

PR #4255, opened 2024-09-24
fix(datasets) Fix the scale of value axis when plotting in absolute sizes

adam-narozniak:
## Issue
When using `plot_comparison_label_distribution` comparing the distributions with absolute value, the value axis is plotted independently, proving incorrect value interpretation (due to the inconsistent scale). 
incorrect_scale

## Proposal
Share the scale of the value axis in case of absolute values:
correct_y_scale


datasets/flwr_datasets/visualization/comparison_label_distribution.py
```diff
@@ -243,3 +249,34 @@ def _initialize_comparison_xy_labels(
...
+    if plot_type == "heatmap":
+        return {}
+    if partition_id_axis == "x":
+        return {"sharey": True}
+    if partition_id_axis == "y":
+        return {"sharex": True}
+    return {"sharex": False, "sharey": False}
+
+
+def _set_tick_on_value_axes(
```
jafermarq, line 271: would it make more sense to use the word `disable_tick` rather than `set_tick`?

adam-narozniak: I can't google it. Could you send a link to matplotlib docs of `disable_tick`?

jafermarq: oh, i mean the name of this function `_set_tick_on_value_axes`, but maybe i'm overthinking it

jafermarq (approved):

PR #4260, opened 2024-09-26
docs(datasets) Embed HF Space in the flwr-datasets docs

adam-narozniak:
## Proposal
Add HF Space to the main documentation page of flwr-datasets.

It shows the `flwr-datasets` visualization results and automatically generates code that was used to partition the dataset and create the visualization results.


tanertopal (approved):
Lgtm!

PR #4278, opened 2024-10-02
fix(framework) Pass Flower path to the call to `start_client_internal()` in SuperNode

chongshenng:
When spinning up a SuperNode using `flower-supernode`, users can specify the `flwr-dir` argument to point to a non-default location where Flower apps are installed. But currently, `flwr_path` is not set in the call to `start_client_internal()`. This PR fixes it.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #4280, opened 2024-10-03
feat(framework) Add example docstring PyLint checker

mohammadnaseri:
This PR adds a new PyLint checker for enforcing example docstring.

### DIFF
--- a/pylint-custom-checkers/docstring_example_checker.py
+++ b/pylint-custom-checkers/docstring_example_checker.py
@@ -1,0 +1,38 @@
+from pylint.checkers import BaseChecker, IAstroidChecker
+from astroid import nodes
+
+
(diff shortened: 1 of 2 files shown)


charlesbvll:
@mohammadnaseri A few comments:
- Could we move the plugin to `dev/`? I don't think we should have it in the root of the project
- Currently, PyLint cannot import the plugin: 
```
Command line or configuration file:1:0: E0013: Plugin 'pylint-custom-checkers.docstring_example_checker' is impossible to load, is it installed ? ('No module named 'pylint-custom-checkers'') (bad-plugin-value)
```
- Lastly, I think we should ignore docstrings that only contain one line, as those correspond to non-public APIs

PR #4283, opened 2024-10-03
refactor(framework) Run exec servicer in SuperLink

panh99:


src/py/flwr/server/app.py
```diff
@@ -681,3 +698,52 @@ def _add_args_fleet_api(parser: argparse.ArgumentParser) -> None:
...
+        help="The directory for the executor.",
+        default=".",
+    )
+    parser.add_argument(
+        "--executor-config",
+        help="Key-value pairs for the executor config, separated by spaces. "
+        "For example:\n\n`--executor-config 'verbose=true "
+        'root-certificates="certificates/superlink-ca.crt"\'`',
+    )
+
```
chongshenng, line 727: Do we need these other args for secure connection? 
```python
    parser.add_argument(
        "--executor-ssl-certfile",
        help="Exec API server SSL certificate file (as a path str) "
        "to create a secure connection.",
        type=str,
        default=None,
    )
    parser.add_argument(
        "--executor-ssl-keyfile",
        help="Exec API server SSL private key file (as a path str) "
        "to create a secure connection.",
        type=str,
    )
    parser.add_argument(
        "--executor-ssl-ca-certfile",
        help="Exec API server SSL CA certificate file (as a path str) "
        "to create a secure connection.",
        type=str,
    )    
```

chongshenng (dismissed):
Thanks @panh99. I think we need the args for secure connection for the Exec API. Also, should we add warnings to the `flower-superexec` command? Something like:
```python
def run_superexec() -> None:
    """Run Flower SuperExec."""
    event(EventType.RUN_SUPEREXEC_ENTER)
    log(
        ERROR,
        "The command `flower-superexec` was deprecated as of version 1.12 "
        "and merged into `flower-superlink.",
    )
    log(INFO, "Execute `flower-superlink --help` to learn how to use it.")
    register_exit_handlers(event_type=EventType.RUN_SUPEREXEC_LEAVE)
```

src/py/flwr/server/app.py
panh99: I don't think we need an extra tuple of certificates specifically for `Exec API`. These certificates allow the SuperLink to estabilish an SSL-enabled connection with a grpc client, and we can just reuse the certificates for `Fleet API` and `Driver API`.

panh99:
> Thanks @panh99. I think we need the args for secure connection for the Exec API. Also, should we add warnings to the `flower-superexec` command? Something like:
> 
> ```python
> def run_superexec() -> None:
>     """Run Flower SuperExec."""
>     event(EventType.RUN_SUPEREXEC_ENTER)
>     log(
>         ERROR,
>         "The command `flower-superexec` was deprecated as of version 1.12 "
>         "and merged into `flower-superlink.",
>     )
>     log(INFO, "Execute `flower-superlink --help` to learn how to use it.")
>     register_exit_handlers(event_type=EventType.RUN_SUPEREXEC_LEAVE)
> ```

I think it's better to open a separate PR to add deprecation warnings.

src/py/flwr/server/app.py
```diff
@@ -681,3 +698,52 @@ def _add_args_fleet_api(parser: argparse.ArgumentParser) -> None:
...
+
+def _load_executor(
```
danieljanes, line 729: @panh99 instead of copying this, can we make `_load_executor` (and the error class) public and use it here?

chongshenng:
> @chongshenng I think it's better to open a separate PR to add deprecation warnings.
Good point, that would also be a breaking change.

src/py/flwr/common/constant.py
```diff
@@ -40,15 +40,14 @@
...
-# SuperExec
-EXEC_API_DEFAULT_ADDRESS = "0.0.0.0:9093"
 # SuperLink
 DRIVER_API_DEFAULT_ADDRESS = "0.0.0.0:9091"
 FLEET_API_GRPC_RERE_DEFAULT_ADDRESS = "0.0.0.0:9092"
 FLEET_API_GRPC_BIDI_DEFAULT_ADDRESS = (
     "[::]:8080"  # IPv6 to keep start_server compatible
 )
 FLEET_API_REST_DEFAULT_ADDRESS = "0.0.0.0:9093"
+EXEC_API_DEFAULT_ADDRESS = "0.0.0.0:9095"
```
danieljanes, line 50: @Robert-Steiner do we need to change anything else to make it work with the new default port number?

src/py/flwr/server/app.py
panh99: Sure thing. I've updated the PR as per your suggestion.

src/py/flwr/common/constant.py
panh99: I changed the default port here b/c it conflicts with the default port for Rest-based Fleet API. We can also updated the default port for that one and keep the default port for exec API unchanged if it's easier to do so.

Robert-Steiner: I'm fine with `9095`. We need to update the Docker docs anyway as we need to remove SuperExec from it.

danieljanes: @panh99 great idea, let's change the default port for the REST API (which is not commonly used at this point)

danieljanes (approved):

danieljanes (approved):

danieljanes (approved):

PR #4292, opened 2024-10-06
refactor(framework) Update `NumPy` template

jafermarq:
It now initializes a dummy global model and makes use of `task.py`. This is so to align it with how other `flwr new` apps are created and run.


src/py/flwr/cli/new/templates/app/code/client.numpy.py.tpl
```diff
@@ -1,16 +1,15 @@
...
 
 
 class FlowerClient(NumPyClient):
-    def get_parameters(self, config):
-        return [np.ones((1, 1))]
 
     def fit(self, parameters, config):
-        return ([np.ones((1, 1))], 1, {})
+        model = get_dummy_model()
+        return ([model], 1, {})
```
danieljanes, line 12: Wouldn't this be cleaner?
```suggestion
        return [model], 1, {}
```

danieljanes (approved):

PR #4304, opened 2024-10-08
feat(framework) Add FAB hash to `flwr build/install`

chongshenng:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


src/py/flwr/cli/install.py
```diff
@@ -155,11 +159,14 @@ def validate_and_install(
 
     if (
         fab_name
-        and fab_name != f"{publisher}.{project_name}.{version.replace('.', '-')}"
+        and ".".join(fab_name.split(".")[:-1])
+        != f"{publisher}.{project_name}.{version.replace('.', '-')}"
+        and len(fab_name.split(".")[-1]) != FAB_HASH_TRUNCATION  # Verify hash length
+        and int(fab_name.split(".")[-1], 16)  # Verify hash is a valid hexadecimal
```
panh99, line 165: I think it's better to create utility functions for extracting the fab hash (and maybe other fields as well), instead of directly use `fab_name.split(".")[-1]` here.

Besides, could you also create a function to verify if the provided string is a hex string, using `try-except`? Otherwise, user may get weird error saying "invalid literal for int() with base 16: 'blablabla'" when the fab name format is wrong.

src/py/flwr/common/config.py
```diff
@@ -215,3 +215,13 @@ def get_metadata_from_config(config: dict[str, Any]) -> tuple[str, str]:
         config["project"]["version"],
         f"{config['tool']['flwr']['app']['publisher']}/{config['project']['name']}",
     )
+
+
+def get_metadata_from_fab_file(fab_file: Union[Path, str]) -> tuple[str, str, str, str]:
```
panh99, line 220:
```suggestion
def get_metadata_from_fab_filename(fab_file: Union[Path, str]) -> tuple[str, str, str, str]:
```

panh99 (dismissed):
LGTM!

src/py/flwr/cli/build.py
```diff
@@ -105,38 +121,53 @@ def build(
...
+                # Read the file content manually
+                with open(file_path, "rb") as f:
+                    file_contents = f.read()
+
                 archive_path = file_path.relative_to(app)
-                fab_file.write(file_path, archive_path)
+                write_to_zip(fab_file, str(archive_path), file_contents)
 
                 # Calculate file info
                 sha256_hash = get_sha256_hash(file_path)
```
charlesbvll, line 149:
```suggestion
                sha256_hash = hashlib.sha256(file_contents).hexdigest()
```

src/py/flwr/cli/build.py
```diff
@@ -14,26 +14,50 @@
...
 from .utils import get_sha256_hash, is_valid_project_name
 
 
-# pylint: disable=too-many-locals
+def write_to_zip(
+    zipfile_obj: zipfile.ZipFile, filename: str, contents: Union[bytes, str]
+) -> zipfile.ZipFile:
+    """Set a fixed date and write contents to a zip file."""
+    zip_info = zipfile.ZipInfo(filename)
+    zip_info.date_time = FAB_DATE
```
charlesbvll, line 40: @chongshenng Could you explain what this fixed date is for?

src/py/flwr/cli/run/run.py
```diff
@@ -179,7 +179,7 @@ def _run_with_superexec(
     channel.subscribe(on_channel_state_change)
     stub = ExecStub(channel)
 
-    fab_path = Path(build(app))
+    fab_path = Path(build(app)[0])
     content = fab_path.read_bytes()
     fab = Fab(hashlib.sha256(content).hexdigest(), content)
```
charlesbvll, line 184:
```suggestion
    fab_path, fab_hash = build(app)
    content = Path(fab_path).read_bytes()
    fab = Fab(fab_hash, content)
```

src/py/flwr/cli/install.py
```diff
@@ -149,21 +158,36 @@ def validate_and_install(
...
+
+        try:
+            _ = int(fab_shorthash, 16)  # Verify hash is a valid hexadecimal
+        except ValueError as e:
+            typer.secho(
+                f"❌ FAB file has an invalid hexadecimal string `{fab_shorthash}`.",
+                fg=typer.colors.RED,
+                bold=True,
+            )
+            raise typer.Exit(code=1) from e
```
charlesbvll, line 190: What about moving this to the `get_metadata_from_fab_filename` function? And have it return None if there is an error

src/py/flwr/common/config.py
```diff
@@ -215,3 +215,15 @@ def get_metadata_from_config(config: dict[str, Any]) -> tuple[str, str]:
...
+def get_metadata_from_fab_filename(
+    fab_file: Union[Path, str]
+) -> tuple[str, str, str, str]:
+    """Extract metadata from the FAB filename."""
+    if isinstance(fab_file, Path):
+        fab_file_name = fab_file.stem
+    elif isinstance(fab_file, str):
+        fab_file_name = fab_file.removesuffix(".fab")
+    publisher, project_name, version, shorthash = fab_file_name.split(".")
+    return publisher, project_name, version.replace("-", "."), shorthash
```
charlesbvll, line 229: Maybe we should create an object for this sort of metadata, I think it would not only make the code cleaner, but also eliminate annoying order errors when dealing with such return types. But I think that would be for another PR ^^

charlesbvll (commented):
Looks good! I added a few minor comments

src/py/flwr/cli/build.py
charlesbvll: Oh I found the answer (it is just a temporary placeholder)

chongshenng: Yes! We're hardcoding it for now, until we decide whether the date should be set in `flwr new` or updated by the user.

src/py/flwr/cli/build.py
chongshenng: Oh, this makes more sense than for `file_path`!

src/py/flwr/common/config.py
chongshenng: That's very true. I'll open another PR for this. Thanks!

src/py/flwr/cli/install.py
chongshenng: Good point. I've refactored it to a `_validate_fab_and_config_metadata()`.

src/py/flwr/cli/install.py
```diff
@@ -226,3 +222,40 @@ def _verify_hashes(list_content: str, tmpdir: Path) -> bool:
...
+    # Check FAB filename format
+    if (
+        f"{fab_publisher}.{fab_project_name}.{fab_version}"
+        != f"{publisher}.{project_name}.{version}"
+        or len(fab_shorthash) != FAB_HASH_TRUNCATION  # Verify hash length
+    ):
+        raise ValueError(
+            "❌ FAB file has incorrect name. The file name must follow the format "
+            "`<publisher>.<project_name>.<version>.<8hexchars>.fab`.",
+        )
```
charlesbvll, line 247:
```suggestion
        typer.secho(
            "❌ FAB file has incorrect name. The file name must follow the format "
            "`<publisher>.<project_name>.<version>.fab`.",
            fg=typer.colors.RED,
            bold=True,
        )
        raise typer.Exit(code=1)
```
Is there a reason for not using a typer Error here (and in the rest of the function)?

charlesbvll (approved):

danieljanes (approved):

PR #4305, opened 2024-10-08
feat(framework) Add FAB hash to `load_client_app_fn`

chongshenng:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/src/py/flwr/cli/build.py
+++ b/src/py/flwr/cli/build.py
@@ -14,26 +14,41 @@
 # ==============================================================================
 """Flower command line interface `build` command."""
 
+import hashlib
 import os
+import shutil
+import tempfile
 import zipfile
 from pathlib import Path
-from typing import Annotated, Optional
+from typing import Annotated, Optional, Union
 
 import pathspec
 import tomli_w
 import typer
 
+from flwr.common.constant import FAB_DATE, FAB_HASH_TRUNCATION
+
 from .config_utils import load_and_validate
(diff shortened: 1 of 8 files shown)


panh99 (dismissed):
LGTM!

src/py/flwr/common/config_test.py
```diff
@@ -65,13 +65,13 @@ def test_get_flwr_dir_with_xdg_data_home() -> None:
 def test_get_project_dir_invalid_fab_id() -> None:
     """Test get_project_dir with an invalid fab_id."""
     with pytest.raises(ValueError):
-        get_project_dir("invalid_fab_id", "1.0.0")
+        get_project_dir("invalid_fab_id", "1.0.0", "abcd1234")
```
danieljanes, line 68: @chongshenng should we pass a full-length hash here?

chongshenng: Yes, strictly speaking it should be a full-length hash. It's set to the dummy string for the purposes of the test, which I think is sufficient. Wdyt?

danieljanes: It's generally better to make the test as realistic as possible. `get_project_dir` truncates the hash - if something was wrong with that part for the function, we might not have a failing test if the test passes an already-truncated hash.

danieljanes (dismissed):

danieljanes (approved):

PR #4360, opened 2024-10-23
refactor(framework) Install `FAB` from wheel

jafermarq:
Build `.whl` before installing python package in the FAB. This requires adding `hatchling` as a `flwr`-wide dep.


jafermarq:
Completed in #4473, #4475, #4479

PR #4363, opened 2024-10-24
feat(framework) Add `PullServerAppInputs` and `PushServerAppOutputs` rpcs to `Driver` service

panh99:


src/proto/flwr/proto/driver.proto
```diff
@@ -56,3 +65,18 @@ message PullTaskResRequest {
   repeated string task_ids = 2;
 }
 message PullTaskResResponse { repeated TaskRes task_res_list = 1; }
+
+// PullServerAppIns messages
+message PullServerAppInsRequest { uint64 run_id = 1; }
+message PullServerAppInsResponse {
```
jafermarq, line 71: worth keeping the same order as in https://github.com/adap/flower/blob/4970e429b35a9f1d1300427ed399330c6d08f625/src/proto/flwr/proto/clientappio.proto#L52? (excluding `Message`).

src/proto/flwr/proto/driver.proto
```diff
@@ -40,6 +41,14 @@ service Driver {
 
   // Get FAB
   rpc GetFab(GetFabRequest) returns (GetFabResponse) {}
+
+  // Pull ServerApp inputs
+  rpc PullServerAppIns(PullServerAppInsRequest)
```
jafermarq, line 46: better this than `PullServerAppInputs`? (same applies below to `Outs`)

src/proto/flwr/proto/driver.proto
chongshenng: I agree, let's keep the order for symmetry:
```python
message PullServerAppInsResponse {
  Context context = 1;
  Run run = 2;
  Fab fab = 3;
}
```

chongshenng (changes requested):
Thanks @panh99. I was under the impression we should first mirror the naming as in `ClientAppIo`, so not using `*Process*` in the names. Then rename everything later in another PR?

jafermarq (approved):

danieljanes (approved):

chongshenng (approved):

PR #4364, opened 2024-10-24
refactor(framework) Enable passing an existing `Context` to `ServerApp` startup function

jafermarq:
the `flwr-serverapp` process will pass an existing `Context` to this function to load and execute a `ServerApp`. At the end, the updated `Context` will be returned.

> [!NOTE]
> The `run` function is also used by the Simulation Engine.


panh99 (commented):
LGTM!

danieljanes (approved):

PR #4369, opened 2024-10-24
feat(framework) Generate error TaskRes for invalid cases in `LinkState.get_task_res`

panh99:
For each `task_id` provided, `get_task_res` method returns one of the following responses:
- An error TaskRes if the corresponding TaskIns does not exist or has expired.
- An error TaskRes if the corresponding TaskRes exists but has expired.
- The valid TaskRes if the TaskIns has a corresponding valid TaskRes.
- Nothing if the TaskIns is still valid and waiting for a TaskRes.


src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -148,7 +155,7 @@ def context_from_bytes(context_bytes: bytes) -> Context:
 
 def make_node_unavailable_taskres(ref_taskins: TaskIns) -> TaskRes:
```
panh99, line 156: @danieljanes Should we remove this? This was used to generate error `TaskRes` for the `TaskIns` of which the destination node is unavailable (offline).

danieljanes: @panh99 yes, we can remove it.

src/py/flwr/common/constant.py
```diff
@@ -128,6 +128,8 @@ class ErrorCode:
     LOAD_CLIENT_APP_EXCEPTION = 1
     CLIENT_APP_RAISED_EXCEPTION = 2
     NODE_UNAVAILABLE = 3
```
danieljanes, line 130: @panh99 should we remove this as well?

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -68,12 +69,13 @@ def __init__(self) -> None:
...
         self.task_ins_store: dict[UUID, TaskIns] = {}
         self.task_res_store: dict[UUID, TaskRes] = {}
+        self.task_ins_id_to_task_res: dict[UUID, TaskRes] = {}
 
         self.node_public_keys: set[bytes] = set()
         self.server_public_key: Optional[bytes] = None
         self.server_private_key: Optional[bytes] = None
 
-        self.lock = threading.Lock()
+        self.lock = threading.RLock()
```
mohammadnaseri, line 78: Just for my understanding, why did you switch to `RLock`?

src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -213,3 +195,141 @@ def has_valid_sub_status(status: RunStatus) -> bool:
     if status.status == Status.FINISHED:
         return status.sub_status in VALID_RUN_SUB_STATUSES
     return status.sub_status == ""
+
+
+def make_taskins_unavailable_taskres(taskins_id: Union[str, UUID]) -> TaskRes:
```
mohammadnaseri, line 200: can we use a better name here? like `create_taskres_unavailable_for_taskins` ?

src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -213,3 +195,141 @@ def has_valid_sub_status(status: RunStatus) -> bool:
...
+            task_type="",  # Unknown message type
+            error=Error(
+                code=ErrorCode.MESSAGE_UNAVAILABLE,
+                reason=MESSAGE_UNAVAILABLE_ERROR_REASON,
+            ),
+        ),
+    )
+
+
+def make_taskres_unavailable_taskres(ref_taskins: TaskIns) -> TaskRes:
```
mohammadnaseri, line 223: same issue

src/py/flwr/server/superlink/linkstate/utils.py
mohammadnaseri: Plus, this method and the one below don't have consistent documentation (for inputs and returns) like the rest of the methods.

mohammadnaseri (commented):
Generally LGTM. Some minor comments. Happy to see that those two files are getting more structured.

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: Oh, I implemented a `_force_delete_tasks_by_ids` method, in which the lock is also used. Switching to `RLock` allows the `get_task_res` to directly invoke `_force_delete_tasks_by_ids` without having to release the lock first, as long as they are in the same thread.

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -67,12 +69,13 @@ def __init__(self) -> None:
         self.federation_options: dict[int, ConfigsRecord] = {}
         self.task_ins_store: dict[UUID, TaskIns] = {}
         self.task_res_store: dict[UUID, TaskRes] = {}
+        self.task_ins_id_to_task_res: dict[UUID, TaskRes] = {}
```
danieljanes, line 72: We already store each `TaskRes` in `task_res_store`, should we perhaps just store a mapping from `TaskIns` ID to `TaskRes` ID?

```suggestion
        self.task_ins_id_to_task_res: dict[UUID, UUID] = {}
```

panh99: Sure thing, I've updated the PR.

danieljanes (approved):

PR #4372, opened 2024-10-25
docs(framework) Add how-to run flwr with deployment engine guide

Robert-Steiner:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


doc/source/how-to-run-flower-with-deployment-engine.rst
```diff
@@ -0,0 +1,163 @@
...
+
+   .. code-block:: shell
+
+       $ flower-supernode \
+            --insecure \
+            --superlink 127.0.0.1:9092 \
+            --node-config "partition-id=1 num-partitions=2" \
+            --supernode-address 127.0.0.1:9095 \
+            --isolation subprocess
+
```
jafermarq, line 99:
```suggestion
            --isolation subprocess

  .. dropdown:: Understand the command

    * ``--supernode-address 127.0.0.1:9094``: Note how we indicate a different port. This is because if you run two `SuperNodes` on the same machine, they can't make use of the same port to communicate with the sub-process executing the `ClientApp`.


```

jafermarq: another option would be to skip showing this second `flower-supernode` command and say: "if you run in a different machine, run exactly as before. If on the same machine, just change the port". This sounds like a better way to me but then the question is: should we present the `--superlink` IPs in a more general way instead of assuming them being `localhost`?

jafermarq (commented):
Hey @Robert-Steiner , i like the flow. Some comments below. I'm thinking to merge this as soon as we release `1.13.0` which will have fused `SuperLink` and `SuperExec`.

jafermarq (commented):
Hey @Robert-Steiner , i like the flow. Some comments below. I'm thinking to merge this as soon as we release `1.13.0` which will have fused `SuperLink` and `SuperExec`.

doc/source/how-to-run-flower-with-deployment-engine.rst
```diff
@@ -0,0 +1,163 @@
...
+   .. dropdown:: Understand the command
+
+       * ``flower-supernode``: Name of the SuperNode binary.
+       * | ``--insecure``: This flag tells the SuperNode to operate in an insecure mode, allowing
+         | unencrypted communication.
+       * | ``--superlink 127.0.0.1:9092``: Connect to the SuperLink's Fleet API at the address
+         | ``127.0.0.1:9092``.
+       * | ``--node-config "partition-id=0 num-partitions=2"``: Set the partition ID to ``0`` and the
+         | number of partitions to ``2`` for the SuperNode configuration.
```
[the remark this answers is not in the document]
Robert-Steiner: That makes sense to me. We could also use the PyTorch Quickstart example to provide a more realistic use case. wdyt?

doc/source/how-to-run-flower-with-deployment-engine.rst
Robert-Steiner: What about using a [tab component](https://sphinx-design.readthedocs.io/en/latest/tabs.html)  and show both versions (localhost, different host)?

jafermarq:
Closed in favour of https://github.com/adap/flower/pull/4372. Something odd was happening in the CI due to the name of this branch (I think)

PR #4394, opened 2024-10-29
refactor(framework) Refactor `Driver` to introduce public `init_run`

jafermarq:
- Driver constructors no longer expect a `run_id`.
- Before making the `Driver` functional (i.e. capable of sending messages, getting list of connected nodes, etc) the new elevated method `init_run(run_id: int)` must be called.
- Updated tests accordingly and code base throughout.


src/py/flwr/server/run_serverapp.py
```diff
@@ -205,7 +204,7 @@ def run_server_app() -> None:
         req = CreateRunRequest(fab_id=fab_id, fab_version=fab_version)
         res: CreateRunResponse = driver._stub.CreateRun(req)  # pylint: disable=W0212
         # Overwrite driver._run_id
```
danieljanes, line 206:
```suggestion
        # Fetch full `Run` using `run_id`
```

danieljanes (approved):

PR #4397, opened 2024-10-29
refactor(framework) Update `flwr log` and `ExecServicer` for log streaming

panh99:


src/py/flwr/cli/log.py
```diff
@@ -54,16 +55,45 @@ def start_stream(
...
+    finally:
+        if res is not None:
+            latest_timestamp = cast(float, res.latest_timestamp)
 
-    for res in stub.StreamLogs(req):
-        print(res.log_output)
-        if time.time() - start_time > duration:
-            break
+    # Add a small epsilon to the latest timestamp to avoid getting the same log
+    return max(latest_timestamp + 1e-6, after_timestamp)
```
danieljanes, line 96: Shouldn't the SuperLink be able to figure this out deterministically? We send the timestamp of the last log we have and the server returns the next log with a timestamp greater than the timestamp we sent.

panh99: I'd like to allow users to fetch all logs after a certain timestamp, and thus I added a `after_timestamp` instead of `latest_log_timestamp` to the `StreamLogsRequest`.

panh99: Given that, it's better to set `after_timestamp` to `latest_log_timestamp + 1e-6` to stream following logs.

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -75,5 +75,33 @@ def StreamLogs(  # pylint: disable=C0103
...
+        while context.is_active():
+            log_msg, latest_timestamp = state.get_serverapp_log(run_id, after_timestamp)
+            if log_msg:
+                yield StreamLogsResponse(
+                    log_output=log_msg,
+                    latest_timestamp=latest_timestamp,
+                )
+                # Add a small epsilon to the latest timestamp to avoid getting
+                # the same log
+                after_timestamp = max(latest_timestamp + 1e-6, after_timestamp)
```
danieljanes, line 97: Why do we add `epsilon` on both the server-side and the client-side?

danieljanes (approved):

PR #4401, opened 2024-10-30
feat(framework) Upload logs in `flwr-serverapp`

panh99:


src/py/flwr/common/constant.py
```diff
@@ -87,6 +87,9 @@
 ISOLATION_MODE_SUBPROCESS = "subprocess"
 ISOLATION_MODE_PROCESS = "process"
 
+# Log streaming
+LOG_UPLOAD_INTERVAL = 0.2
```
danieljanes, line 91: Do we want to make 5 requests per second? Every second might be sufficient, maybe even every other second.

panh99: Hmmm, we won't really make 5 requests per second. If no new message, it won't make a request. It's actually the minimum interval for log uploading.

src/py/flwr/common/logger.py
```diff
@@ -290,3 +298,58 @@ def restore_output() -> None:
...
+            pass
+
+        # Upload if any logs
+        if msgs:
+            req = PushLogsRequest(
+                node=node,
+                run_id=run_id,
+                logs=msgs,
+            )
+            stub.PushLogs(req)
```
danieljanes, line 331: This needs error handling. We shouldn't crash the log uploader if a single request fails.

panh99: Sure, that makes sense.

danieljanes (approved):

PR #4418, opened 2024-11-01
refactor(framework:skip) Move logging info for `flwr-serverapp` and `flwr-clientap`

chongshenng:


src/py/flwr/server/serverapp/app.py
```diff
@@ -108,6 +106,7 @@ def flwr_serverapp() -> None:
 
     certificates = _try_obtain_certificates(args)
 
+    log(INFO, "Starting Flower ServerApp")
```
danieljanes, line 109: @chongshenng should we do this before loading certificates? imo, it's good to show the log early to show users that the system is alive

chongshenng: Yes, good point!

danieljanes (approved):

PR #4434, opened 2024-11-05
feat(framework) Initialize `Context` on the SuperLink

chongshenng:
* Move `Context` creation outside of `Executor` plugin to make it plugin-agnostic.
* Move config fusing to `ExecServicer`.


chongshenng:
Closing as this change is no longer relevant given recent updates.

PR #4439, opened 2024-11-06
feat(framework) Add first implementation of `NodeState`

chongshenng:
* Introduce `NodeState` `ABC` with `set/get_node_id()`
* Introduce `InMemoryNodeState` and basic test
* Introduce `NodeStateFactory`
* Add usage of `state.set_node_id()` in SuperNode


src/py/flwr/client/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -0,0 +1,37 @@
...
+    """In-memory NodeState implementation."""
+
+    def __init__(self) -> None:
+
+        # Map run_id to node_id
+        self.node_id: dict[int, int] = {}
+
+    def set_node_id(self, run_id: int, node_id: int) -> None:
+        """Set the node ID for a specified run ID."""
+        self.node_id[run_id] = node_id
```
panh99, line 33: I am a bit confused here. One SuperNode only have one `node_id`. Why we have multiple node IDs here? The node ID will remain the same for multiple runs the SuperNode joins.

chongshenng: Yes, you're right. I've fixed this.

panh99 (approved):
LGTM!

src/py/flwr/client/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+from flwr.client.supernode.nodestate.nodestate import NodeState
+
+
+class InMemoryNodeState(NodeState):
+    """In-memory NodeState implementation."""
+
+    def __init__(self) -> None:
+        # Map run_id to node_id
```
danieljanes, line 25: This comment seems wrong

src/py/flwr/client/supernode/nodestate/nodestate_factory.py
```diff
@@ -0,0 +1,33 @@
...
+class NodeStateFactory:
+    """Factory class that creates NodeState instances."""
+
+    def __init__(self) -> None:
+        self.state_instance: Optional[NodeState] = None
+
+    def state(self) -> NodeState:
+        """Return a State instance and create it, if necessary."""
+        if self.state_instance is None:
+            self.state_instance = InMemoryNodeState()
```
danieljanes, line 32: This is not thread safe, is this acceptable here?

src/py/flwr/client/app.py
```diff
@@ -459,6 +461,10 @@ def _on_backoff(retry_state: RetryState) -> None:
 
                     run.fab_id, run.fab_version = fab_id, fab_version
 
+                    # Set node_id for this run
+                    if node_id is not None:
+                        state.set_node_id(node_id)
```
danieljanes, line 466: The `node_id` is not related to the run. The `node_id` of a SuperNode never changes.

I'd suggest to store the `node_id` in `NodeState` right after we get it from `create_node`. With that, we can remove the local `node_id` variable and read `node_id` from `NodeState` whenever we need it.

src/py/flwr/client/supernode/app.py
```diff
@@ -40,12 +40,14 @@
...
 from flwr.common.logger import log, warn_deprecated_feature
 
-from ..app import start_client_internal
 from ..clientapp.utils import get_load_client_app_fn
 
 
 def run_supernode() -> None:
     """Run Flower SuperNode."""
+    # Import start_client_internal here to avoid circular imports
+    from ..app import start_client_internal  # pylint: disable=import-outside-toplevel
```
danieljanes, line 49: What's the circular import here? I'd like to understand it better. Perhaps we can find a different solution.

src/py/flwr/client/supernode/nodestate/in_memory_nodestate.py
chongshenng: Fixed

src/py/flwr/client/supernode/nodestate/nodestate_factory.py
chongshenng: You're right. Fixed.

src/py/flwr/client/app.py
chongshenng: Fixed.

src/py/flwr/client/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -0,0 +1,37 @@
...
+
+from flwr.client.supernode.nodestate.nodestate import NodeState
+
+
+class InMemoryNodeState(NodeState):
+    """In-memory NodeState implementation."""
+
+    def __init__(self) -> None:
+
+        # Store node_ids
```
danieljanes, line 28:
```suggestion
        # Store node_id
```

src/py/flwr/client/app.py
```diff
@@ -396,13 +399,12 @@ def _on_backoff(retry_state: RetryState) -> None:
                     )
                 else:
                     # Call create_node fn to register node
-                    node_id: Optional[int] = (  # pylint: disable=assignment-from-none
-                        create_node()
-                    )  # pylint: disable=not-callable
-                    if node_id is None:
+                    # and store node_id in state
+                    state.set_node_id(create_node())
```
danieljanes, line 403: `create_node()` returns `Optional[int]`, so we need to check the return value and exit if node registration fails.

src/py/flwr/client/app.py
```diff
@@ -396,13 +399,12 @@ def _on_backoff(retry_state: RetryState) -> None:
...
+                    if not state.get_node_id():
                         raise ValueError("Node registration failed")
                     run_info_store = DeprecatedRunInfoStore(
-                        node_id=node_id,
+                        node_id=cast(int, state.get_node_id()),
```
danieljanes, line 407: Do we need the `cast` here because `get_node_id` returns an `Optional[int]`? We could think about changing the return type to `int` to avoid these (and raise an exception if the `node_id` wasn't set before)

chongshenng: Yes, the `cast` is precisely because of that. Sure, makes sense to change it.

src/py/flwr/client/app.py
```diff
@@ -396,13 +399,12 @@ def _on_backoff(retry_state: RetryState) -> None:
...
                 else:
                     # Call create_node fn to register node
-                    node_id: Optional[int] = (  # pylint: disable=assignment-from-none
-                        create_node()
-                    )  # pylint: disable=not-callable
-                    if node_id is None:
-                        raise ValueError("Node registration failed")
+                    # and store node_id in state
+                    if (node_id := create_node()) is None:
+                        raise ValueError("Node creation failed")
```
danieljanes, line 404:
```suggestion
                        raise ValueError("Failed to register SuperNode with the SuperLink")
```

danieljanes (approved):

danieljanes (approved):

PR #4452, opened 2024-11-07
fix(framework) Improve import management in `load_app` function in `object_ref.py`

panh99:
A big bug in the previous version: the loaded modules won't be unloaded if `load_app` is called again.

### DIFF
--- a/src/py/flwr/common/object_ref.py
+++ b/src/py/flwr/common/object_ref.py
@@ -55,8 +55,8 @@ def validate(
         specified attribute within it.
     project_dir : Optional[Union[str, Path]] (default: None)
         The directory containing the module. If None, the current working directory
-        is used. If `check_module` is True, the `project_dir` will be inserted into
-        the system path, and the previously inserted `project_dir` will be removed.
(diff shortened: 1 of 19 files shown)


src/py/flwr/common/object_ref.py
```diff
@@ -133,8 +141,10 @@ def load_app(  # pylint: disable= too-many-branches
...
     Note
     ----
-    This function will modify `sys.path` by inserting the provided `project_dir`
-    and removing the previously inserted `project_dir`.
+    - This function will unload all modules in the previously provided `project_dir`,
+      if it is invoked again.
+    - This function will modify `sys.path` by inserting the provided `project_dir`
+      and removing the previously inserted `project_dir`.
     """
     valid, error_msg = validate(module_attribute_str, check_module=False)
```
jafermarq, line 149: is there a reason why we wouldn't use `check_module=True`?

panh99: There's no need to have repeated checks here. When passing `check_module=True`, the `validate` function will do similar things as `load_app`.

danieljanes (approved):

PR #4458, opened 2024-11-08
docs(framework:skip) Fix docs build

chongshenng:


Robert-Steiner (approved):

doc/build-versioned-docs.sh
```diff
@@ -22,11 +22,29 @@ languages="en `find locales/ -mindepth 1 -maxdepth 1 -type d -exec basename '{}'
...
   export current_version
   git checkout --force ${current_version}
+
+  # Downgrade numpy for versions between v1.8.0 and v1.12.0 to avoid conflicts in docs
+  if [ "$current_version" = "v1.8.0"  ] || \
+     [ "$current_version" = "v1.9.0"  ] || \
+     [ "$current_version" = "v1.10.0" ] || \
+     [ "$current_version" = "v1.11.0" ] || \
+     [ "$current_version" = "v1.11.1" ] || \
+     [ "$current_version" = "v1.12.0" ]; then
```
charlesbvll, line 41:
```suggestion
  if echo "$current_version" | grep -qE '^v1\.(8\.0|9\.0|10\.0|11\.0|11\.1|12\.0)$'; then
```
I think we could try something like that

charlesbvll: But it's only a cosmetic change, so I will approve the PR

charlesbvll (approved):

charlesbvll (approved):

danieljanes (approved):

PR #4460, opened 2024-11-08
feat(framework) Implement `ExecServicer.ListRuns` RPC for `flwr ls`

panh99:
Based on #4487.


src/py/flwr/superexec/exec_servicer.py
chongshenng, line 1: Should we add a test?

chongshenng (dismissed):

danieljanes (approved):

danieljanes (approved):

PR #4478, opened 2024-11-11
feat(framework) Add simple retry mechanism for `GrpcDriver`

panh99:


src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -255,3 +258,59 @@ def close(self) -> None:
...
+    """Wrap the gRPC stub with a retry invoker."""
+
+    def make_lambda(original_method: Any) -> Any:
+        return lambda *args, **kwargs: retry_invoker.invoke(
+            original_method, *args, **kwargs
+        )
+
+    for method_name in stub.__dict__:
+        method = getattr(stub, method_name)
+        setattr(stub, method_name, make_lambda(method))
```
danieljanes, line 316: @panh99 should we list the methods we expect to wrap here? if `protoc` starts to generate additional methods in the future, we might unintentionally wrap those as well.

danieljanes (approved):

PR #4484, opened 2024-11-12
feat(framework) Introduce `LinkState.get_run_timestamps()`

panh99:
This method will be used to get timestamps, which can then be displayed via `flwr ls` command.

### DIFF
--- a/src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
+++ b/src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
@@ -448,6 +448,20 @@ class InMemoryLinkState(LinkState):  # pylint: disable=R0902,R0904
                 return None
             return self.run_ids[run_id].run
 
+    def get_run_timestamps(self, run_id: int) -> tuple[str, str, str, str]:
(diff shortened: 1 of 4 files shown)


src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -448,6 +448,20 @@ def get_run(self, run_id: int) -> Optional[Run]:
...
             return self.run_ids[run_id].run
 
+    def get_run_timestamps(self, run_id: int) -> tuple[str, str, str, str]:
+        """Retrieve the timestamps for the specified `run_id`."""
+        with self.lock:
+            if run_id not in self.run_ids:
+                log(ERROR, "`run_id` is invalid")
+                return "", "", "", ""
+            run_record = self.run_ids[run_id]
+            return (
```
jafermarq, line 458: @panh99 I'm wondering if we could return this as a dict instead, so we are a step closer to the JSON-compatible format @chongshenng mentioned yesterday. (The same would apply to the SQL Linkstate)

PR #4486, opened 2024-11-13
refactor(framework) Remove `num-supernodes` requirement from simulation exec plugin

jafermarq:
- It is no longer needed to specify the number of supernodes when launching the `SuperLink` with the simulation plugin. This means the exec w/ simulation plugin can be now launched as:

```shell
flower-superlink --insecure  --executor flwr.superexec.simulation:executor
```

- A check for `num-supernodes` in the federation options is performed before creating the Run.

### DIFF
--- a/src/py/flwr/superexec/simulation.py
+++ b/src/py/flwr/superexec/simulation.py
@@ -81,7 +81,7 @@ class SimulationEngine(Executor):
         """Start run using the Flower Simulation Engine."""
         try:
             # Check that num-supernodes is set
-            if "num-supernoces" not in federation_options:
+            if "num-supernodes" not in federation_options:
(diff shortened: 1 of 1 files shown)


src/py/flwr/superexec/simulation.py
```diff
@@ -123,6 +80,12 @@ def start_run(
     ) -> Optional[int]:
         """Start run using the Flower Simulation Engine."""
         try:
+            # Check that num-supernodes is set
+            if "num-supernoces" not in federation_options:
```
danieljanes, line 84:
```suggestion
            if "num-supernodes" not in federation_options:
```

danieljanes (approved):

PR #4487, opened 2024-11-13
feat(framework) Add timestamps and run status to `Run`

panh99:


src/py/flwr/simulation/run_simulation.py
```diff
@@ -333,14 +328,10 @@ def _main_loop(
...
-        state_factory.state().run_ids[run.run_id] = RunRecord(  # type: ignore
-            run=run,
-            status=init_status,
-            starting_at=now().isoformat(),
-            running_at=now().isoformat(),
-            finished_at="",
-        )
+        run.status = RunStatus(Status.RUNNING, "", "")
+        run.starting_at = now().isoformat()
+        run.running_at = now().isoformat()
```
chongshenng, line 333:
```suggestion
        run.starting_at = now().isoformat()
        run.running_at = run.starting_at
```
Should we make this equal?

chongshenng (commented):
Thanks @panh99! The `Run` handling is neat. Only one comment from me.

src/py/flwr/common/typing.py
```diff
@@ -208,23 +208,44 @@ class ClientMessage:
...
-    sub_status: str
-    details: str
+    pending_at: str
+    starting_at: str
+    running_at: str
+    finished_at: str
+    status: RunStatus
+
+    @classmethod
+    def create_empty(cls, run_id: int = 0) -> "Run":
```
danieljanes, line 235: Should `run_id` have a default? Or should we force the caller to set it?

src/py/flwr/common/typing.py
```diff
@@ -208,23 +208,44 @@ class ClientMessage:
...
+            run_id=run_id,
+            fab_id="",
+            fab_version="",
+            fab_hash="",
+            override_config={},
+            pending_at="",
+            starting_at="",
+            running_at="",
+            finished_at="",
+            status=RunStatus(status="", sub_status="", details=""),
```
danieljanes, line 247: Should we allow creating invalid `Run` objects at all?

src/py/flwr/simulation/run_simulation.py
danieljanes: @panh99 @chongshenng does it make sense to have the same timestamp for both?

src/py/flwr/common/typing.py
panh99: Hmmm, I think that's fine. The purpose of this class method is just to create an empty instance, and it's not a constructor.

src/py/flwr/common/typing.py
panh99: The purpose of having a `create_empty` method is to simplify the process of creating an empty instance in special cases, such as unit tests. This method isn’t used in our `LinkState`, and in that context, all `Run` objects within `LinkState` are valid. Moreover, since `Run` is a simple dataclass, we can’t guarantee it will always be valid unless we extend it with additional methods, similar to what we’ve done with the `Message` class.

Additionally, we have created "invalid" `Run` objects before, but only for cases where a `Run` instance isn’t available, such as in `grpc-bidi`.

danieljanes: Do we use `create_empty` exclusively in tests? If so, we should instead create a helper function in one of the `_test` modules.

panh99: We also use it in the simulation part.

danieljanes (approved):

PR #4512, opened 2024-11-16
feat(framework) Improve retry mechanism for `Fleet` connection

panh99:
- The maximum backoff time has been capped at 20 seconds to prevent excessively long intervals caused by exponential backoff.
- In the `grpc-rere` and `grpc-adapter` transports, retries will now only occur if the gRPC status code is `UNAVAILABLE`, which specifically indicates a network connection issue.


src/py/flwr/client/app.py
```diff
@@ -373,7 +373,7 @@ def _on_backoff(retry_state: RetryState) -> None:
             )
 
     retry_invoker = RetryInvoker(
-        wait_gen_factory=exponential,
+        wait_gen_factory=lambda: exponential(max_delay=20),
```
danieljanes, line 376: @panh99 should we move the `20` to `constants`?

panh99: Sure.

danieljanes (approved):

PR #4536, opened 2024-11-19
refactor(framework) Detect if `SuperNode` started without  passing certificates or `--insecure`

jafermarq:
If simply running `flower-supernode` an error is thrown because attempting to do `isfile(root_cert_path)` on a `None`. The solution here proposed follows a logic somewhat similar to what's done for the `SuperLink` (i.e. when executing `flower-superlink` w/o passing anything)

PR #4541, opened 2024-11-19
feat(framework:skip) Unify warnings for missing or bad certificates if `--insecure` not set

chongshenng:
The error messages when running only `flower-supernode` or `flower-superlink` are not consistent (see #4536). This PR unifies them.


src/py/flwr/common/args.py
```diff
@@ -79,9 +79,19 @@ def try_obtain_root_certificates(
...
+                "Root certificates are required unless running in insecure mode. "
+                "Please provide a certificate path to `--root-certificates` or run "
+                "the client in insecure mode using '--insecure' if you understand "
+                "the risks.",
+            )
+            sys.exit(1)
         if not isfile(root_cert_path):
-            sys.exit("Path argument `--root-certificates` does not point to a file.")
+            log(ERROR, "Path argument `--root-certificates` does not point to a file.")
+            return None
```
danieljanes, line 93: Shouldn't this trigger a `sys.exit`?

chongshenng: You're right, fixed!

danieljanes (approved):

PR #4563, opened 2024-11-21
break(framework) Fix simulation arguments and tests

chongshenng:
The existing Exec API tests are silently failing, causing several bugs to slip through. The changes in these PR are inter-related.

For tests:
* Fix Exec API tests so that it succeeds (`exit 0`) or fails (`exit 1`) correctly.
* Remove unsupported `--executor-config num-supernodes=1` from `flower-superlink` which was removed in #4486.

For `flwr-simulation` arguments:
* Update SuperLink to call `flwr-simulation` with `--simulationio-api-address` instead of `--superlink`.
* Remove secure connection for `flwr-simulation`.

For settings in `flwr-simulation`:
* Set the default value of `enable_tf_gpu_growth` in `flwr-simulation` to `False` on [line 234](https://github.com/adap/flower/blob/81ae8397d302d8c19ef04e5076db7dd81237f75f/src/py/flwr/simulation/app.py#L228). @jafermarq can you please confirm that this fix is reasonble?


jafermarq:
@chongshenng could we follow the same logic in `flwr-simulation` as in `flwr-serverapp` and introduce something like this? https://github.com/adap/flower/blob/67688c6708b3cf23829c872999c0f9887b9ed558/src/py/flwr/server/serverapp/app.py#L74

src/py/flwr/simulation/app.py
```diff
@@ -225,7 +231,7 @@ def run_simulation_process(  # pylint: disable=R0914, disable=W0212, disable=R09
                 )
             backend_config: BackendConfig = fed_opt.get("backend", {})
             verbose: bool = fed_opt.get("verbose", False)
-            enable_tf_gpu_growth: bool = fed_opt.get("enable_tf_gpu_growth", True)
+            enable_tf_gpu_growth: bool = fed_opt.get("enable_tf_gpu_growth", False)
```
jafermarq, line 234: I left it `true` on purpose with the idea of making things easier for those that want to use TF simulations w/ GPU. They won't need to manually set the env variable. What do you think makes more sense to do?

chongshenng: Am I right in understanding that this `enable_tf_gpu_growth` is set in `pyproject.toml` like so?
```toml
[tool.flwr.federations.local-simulation]
# enable-tf-gpu-growth = True
```

If so, can we set it to `False` in code, and then the line above commented out in case users want to include it?

chongshenng:
@jafermarq For [this comment](https://github.com/adap/flower/pull/4563#issuecomment-2493065732), yes. I'll have a follow-up PR that refactors the args and includes this warning too.

panh99 (commented):
Looks good to me! Just one tiny comment.

e2e/test_exec_api.sh
```diff
@@ -74,6 +73,10 @@ else
   echo -e $"\n[tool.flwr.federations.e2e]\naddress = \"127.0.0.1:9093\"\nroot-certificates = \"../certificates/ca.crt\"" >> pyproject.toml
 fi
 
+if [ "$3" = "simulation-engine" ]; then
+  echo -e $"options.num-supernodes = 2" >> pyproject.toml
```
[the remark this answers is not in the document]
chongshenng: Changed to 10.

src/py/flwr/simulation/app.py
jafermarq: as discussed, let's keep it as `False` and revisit this topic in a new PR.

jafermarq (approved):
looks good to me!!

danieljanes (approved):

PR #4571, opened 2024-11-22
refactor(framework) Remove `pip install` in `flwr install` and the wheel file

panh99:
The `pip install` step is unnecessary and should be removed. Consequently, the wheel file is no longer required.

### DIFF
--- a/src/py/flwr/cli/install.py
+++ b/src/py/flwr/cli/install.py
@@ -16,7 +16,6 @@
 
 import hashlib
 import shutil
-import subprocess
 import tempfile
 import zipfile
 from io import BytesIO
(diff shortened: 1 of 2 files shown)


src/py/flwr/cli/install.py
```diff
@@ -188,23 +187,6 @@ def validate_and_install(
...
-    except subprocess.CalledProcessError as e:
-        typer.secho(
-            f"❌ Failed to install {project_name}:\n{e.stderr}",
-            fg=typer.colors.RED,
-            bold=True,
-        )
-        raise typer.Exit(code=1) from e
-
     typer.secho(
         f"🎊 Successfully installed {project_name}.",
```
chongshenng, line 191:
```suggestion
        f"🎊 Successfully installed {project_name} to {install_dir}.",
```

I think we should add the install destination here for info.

chongshenng (commented):
Only one small addition, otherwise it LGTM.

src/py/flwr/cli/install.py
panh99: That makes sense. Thanks!

chongshenng:
@panh99 Since we won't need `hatchling` with this fix, does it make sense to remove it from our project's `pyproject.toml` in the same PR?

danieljanes (approved):

PR #4573, opened 2024-11-22
fix(datasets:skip) Remove pyarrow version pin

adam-narozniak:
## Issue
In https://github.com/adap/flower/pull/3817 we pinned the pyarrow version to avoid the `pypi` discoverability problems for a newer than 16.1.0 pyarrow version (which made the whole CI fail - that was the reason for the **temporal** version pin).

## Proposal
Unpin the pyarrow version. (it was also noticed by a user with an older VS Code that this old version does not work for him.)


tanertopal (approved):
lgtm!

PR #4581, opened 2024-11-25
docs(datasets:skip) Add reference to recommended datasets to the index

adam-narozniak:
## Issue
The reference to the recommended FL datasets is missing in the index (also doctree)

## Proposal
Add the link to the "Reference docs"

### DIFF
--- a/datasets/doc/source/index.rst
+++ b/datasets/doc/source/index.rst
@@ -63,8 +63,8 @@ Information-oriented API reference and other reference material.
    :maxdepth: 1
    :caption: Reference docs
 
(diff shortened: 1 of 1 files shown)


datasets/doc/source/index.rst
```diff
@@ -64,6 +64,7 @@ Information-oriented API reference and other reference material.
    :caption: Reference docs
 
    ref-telemetry
+   recommended-fl-datasets
```
adam-narozniak, line 67:
```suggestion
   recommended-fl-datasets
   ref-telemetry
```

PR #4610, opened 2024-12-02
feat(framework) Add JSON formatting to `flwr ls` output

chongshenng:
Merge afte #4613


src/py/flwr/common/logger.py
```diff
@@ -303,6 +305,12 @@ def restore_output() -> None:
     console_handler.stream = sys.stdout
 
 
+def redirect_output(output_buffer: StringIO) -> None:
+    """Redirect stdout and stderr to text I/O buffer."""
+    sys.stdout = output_buffer
+    sys.stderr = output_buffer
```
panh99, line 311:
```suggestion
def redirect_output(output_buffer: StringIO) -> None:
    """Redirect stdout and stderr to text I/O buffer."""
    sys.stdout = output_buffer
    sys.stderr = output_buffer
    console_handler.stream = sys.stdout
```

src/py/flwr/cli/ls.py
```diff
@@ -104,11 +119,13 @@ def ls(
             # Display information about a specific run ID
             if run_id is not None:
                 typer.echo(f"🔍 Displaying information for run ID {run_id}...")
-                _display_one_run(stub, run_id)
+                restore_output()
```
panh99, line 122: Do we need `restore_output()` here, since we have it in `finally` block?

src/py/flwr/cli/ls.py
```diff
@@ -104,11 +119,13 @@ def ls(
...
+                _display_one_run(stub, run_id, output_format)
             # By default, list all runs
             else:
                 typer.echo("📄 Listing all runs...")
-                _list_runs(stub)
+                restore_output()
```
panh99, line 127: Same as above.

src/py/flwr/cli/ls.py
```diff
@@ -119,9 +136,15 @@ def ls(
             raise typer.Exit(code=1) from err
         finally:
             channel.close()
-    # pylint: disable=broad-except, unused-variable
-    except (typer.Exit, Exception):
-        _print_json_error()
+    except (typer.Exit, SystemExit, Exception) as e:  # pylint: disable=broad-except
+        if suppress_output:
+            restore_output()
```
panh99, line 141: Same as above.

src/py/flwr/cli/ls.py
```diff
@@ -119,9 +136,15 @@ def ls(
             raise typer.Exit(code=1) from err
         finally:
             channel.close()
-    # pylint: disable=broad-except, unused-variable
-    except (typer.Exit, Exception):
-        _print_json_error()
+    except (typer.Exit, SystemExit, Exception) as e:  # pylint: disable=broad-except
```
panh99, line 139:
```suggestion
    except (typer.Exit, Exception) as e:  # pylint: disable=broad-except
```

src/py/flwr/cli/ls.py
```diff
@@ -119,9 +136,15 @@ def ls(
...
+        if suppress_output:
+            restore_output()
+            e_message = captured_output.getvalue()
+            _print_json_error(e_message, e)
```
panh99, line 143: Does that mean no exception message will be available in default format?

src/py/flwr/cli/ls.py
chongshenng: Yes. This is because `Console().print_json()` will also be redirected and won't be shown.

src/py/flwr/cli/ls.py
chongshenng: Same reply as above.

src/py/flwr/cli/ls.py
chongshenng: Same reply as above.

src/py/flwr/cli/ls.py
chongshenng: No, if there're any exceptions in default mode, the errors are not redirected.

src/py/flwr/common/logger.py
chongshenng: @panh99  Since we aren't redirecting any logs for `flwr ls` and `flwr run`, we shouldn't need to redirect `console_handler.stream = sys.stdout`?

src/py/flwr/cli/ls.py
panh99: But all exceptions are captured, no?

src/py/flwr/common/logger.py
panh99: It's not needed for now, but to maintain consistency and avoid potential confusion in the future, I recommend also redirecting logs, as they are also part of the "output."

src/py/flwr/cli/ls.py
chongshenng: Yes, but the error messages are not redirected. So the `if` statement doesn't affect the default mode.

src/py/flwr/common/logger.py
chongshenng: Hmm ok. I'm leaning against it since it's unnecessary for now, but will add anyway.

panh99: Thanks. I think otherwise we have to clarify in the docstring that logs may or may not be redirected, depending on when the `log` module is imported.

src/py/flwr/cli/ls.py
panh99: I don’t think so. All exceptions are captured here, but if you look at the code above, it only captures `ValueError`. This approach will suppress all other exceptions. For example, if there’s a gRPC exception when using the `stub` object, users will see the exception in `--format json` mode, but nothing will appear in the default mode. See below:

Screenshot 2024-12-03 at 15 08 15

panh99:
```suggestion
            _print_json_error(e_message, e)
        else:
            raise
```

src/py/flwr/cli/ls.py
```diff
@@ -325,11 +354,19 @@ def _display_one_run(
...
         Console().print(_to_table(_format_runs(run_dict, res.now)))
 
 
-def _print_json_error() -> None:
+def _print_json_error(msg: str, e: Union[Exit, SystemExit, Exception]) -> None:
     """Print error message as JSON."""
+    Console().print_json(
+        json.dumps(
+            {
+                "success": "false",
```
jafermarq, line 368:
```suggestion
                "success": False,
```

jafermarq: Shouldn't it be a true `bool`?

src/py/flwr/cli/ls.py
chongshenng: Resolved with `typer.secho()`.

src/py/flwr/cli/ls.py
chongshenng: Good catch. When serialized, it'll be converted to JSON's `true` or `false`. Updated.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #4611, opened 2024-12-02
feat(framework) Refactor `flwr ls` table formatter

chongshenng:
Split the logic of formatting the `Run` from creating a `rich.Table`. The formatted `Run` will be used later to create a JSON output.

Merge before:
* #4612

PR #4612, opened 2024-12-02
feat(framework) Introduce `try`-`except` block to catch exceptions in `flwr ls`

chongshenng:
The `try-except` block is required to catch any exceptions raised during `flwr ls` and print a JSON output for errors.

Merge after:
* #4611 

Merge before:
* #4613


src/py/flwr/cli/ls.py
```diff
@@ -70,52 +70,56 @@ def ls(
...
+            typer.secho(
+                f"❌ {err}",
+                fg=typer.colors.RED,
+                bold=True,
+            )
+            raise typer.Exit(code=1) from err
+        finally:
+            channel.close()
+    # pylint: disable=broad-except, unused-variable
+    except (typer.Exit, SystemExit, Exception):
```
panh99, line 121: Do we need `SystemExit` when `typer.Exit` is here?

src/py/flwr/cli/ls.py
```diff
@@ -70,52 +70,56 @@ def ls(
...
+        _print_json_error()
```
panh99, line 122: May I know why `flwr ls` always prints errors in json?

src/py/flwr/cli/ls.py
chongshenng: Yes, unfortunately, the `Exception` doesn't catch a `typer.Exit` or `SystemExit`.

src/py/flwr/cli/ls.py
chongshenng: No it won't. I'll add introduce the JSON formatting option in #4610. It's written this way to preserve the existing behaviour and to introduce the `try`-`except` block (to minimize PR diff).

src/py/flwr/cli/ls.py
panh99: Oh, I understand that `Exception` doesn't catch the two, b/c the two inherit from `BaseException` instead of `Exception. I was wondering why we need to include `SystemExit` here.

src/py/flwr/cli/ls.py
panh99: I think, thanks for the clarification. I will resolve this.

src/py/flwr/cli/ls.py
chongshenng: Oh yes, I originally added `SystemExit` but it didn't to catch `typer.Exit`. Will update.

panh99 (dismissed):
LGTM!

danieljanes (approved):

PR #4613, opened 2024-12-02
feat(framework) Introduce JSON formatting function in `flwr ls`

chongshenng:
Merge after
* #4612 

Merge before
* #4610


src/py/flwr/cli/ls.py
```diff
@@ -255,19 +257,66 @@ def _to_table(run_list: list[tuple[str, ...]]) -> Table:
...
     stub: ExecStub,
+    output_format: str = CliOutputFormat.default,
 ) -> None:
     """List all runs."""
     res: ListRunsResponse = stub.ListRuns(ListRunsRequest())
     run_dict = {run_id: run_from_proto(proto) for run_id, proto in res.run_dict.items()}
 
-    Console().print(_to_table(_format_runs(run_dict, res.now)))
+    if output_format == CliOutputFormat.json:
+        Console().print_json(_to_json(_format_runs(run_dict, res.now)))
```
panh99, line 311: Oh, this is interesting. So the rich Console also supports printing json? What can this `print_json` do?

src/py/flwr/cli/ls.py
```diff
@@ -255,19 +257,66 @@ def _to_table(run_list: list[tuple[str, ...]]) -> Table:
     return table
 
 
+def _to_json(run_list: list[tuple[str, ...]]) -> str:
+    """Format run status list to a JSON formatted string."""
+
+    def _remove_bbcode_tags(strings: tuple[str, ...]) -> tuple[str, ...]:
```
panh99, line 263: Just a silly question: do we add bbcode tags somewhere? Does the `_format_runs` add these tags?

chongshenng: Yes, it's the `status_style`. 😅

src/py/flwr/cli/ls.py
chongshenng: Yes! (I thought I replied to this comment somewhere else). I didn't explore it further, but so far it produces nicely formatted JSON output.

src/py/flwr/cli/ls.py
jafermarq: This looks like a useful function. Let's move it to `cli/utils.py`?

panh99 (approved):
LGTM!

danieljanes (approved):

PR #4244, opened 2024-09-20
feat(framework) Add flwr login protos

danielnugraha:
This is the first PR introducing user authentication. The changes are in the protobuf level which introduces two new endpoints in the Exec API:

1. `GetLoginDetails`: accepts `GetLoginDetailsRequest` and returns `GetLoginDetailsResponse`.
2. `GetAuthTokens`: accepts `GetAuthTokensRequest` and returns `GetAuthTokensResponse`.

PR #4618, opened 2024-12-03
feat(framework) Add abstract user auth plugin

danielnugraha:
This is the second PR introducing user authentication. The changes include:

- Adding abstract base classes: `ExecAuthPlugin` and `CliAuthPlugin`.

Please merge #4244 first before merging this PR.


src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,82 @@
...
+
+from flwr.proto.exec_pb2 import (  # pylint: disable=E0611
+    GetAuthTokenRequest,
+    GetAuthTokenResponse,
+    LoginResponse,
+)
+from flwr.proto.exec_pb2_grpc import ExecStub
+
+class ExecAuthPlugin(ABC):
+    """Abstract Flower Exec API Auth Plugin class."""
```
panh99, line 32:
```suggestion
class ExecAuthPlugin(ABC):
    """Abstract Flower Auth Plugin class for ExecServicer."""
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,82 @@
...
+    ) -> GetAuthTokenResponse:
+        """Send relevant tokens as a GetAuthTokenResponse."""
+
+    @abstractmethod
+    def refresh_token(self, context: grpc.ServicerContext) -> bool:
+        """Refresh auth tokens in the provided metadata."""
+
+
+class CliAuthPlugin(ABC):
+    """Abstract Flower CLI Auth Plugin class."""
```
panh99, line 58:
```suggestion
class CliAuthPlugin(ABC):
    """Abstract Flower Auth Plugin class for CLI."""
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,82 @@
...
+    ) -> dict[str, Any]:
+        """Login logic to log in user to the SuperLink."""
+
+    @abstractmethod
+    def __init__(self, config: dict[str, Any], config_path: Path):
+        """Abstract constructor (init)."""
+
+    @abstractmethod
+    def write_token_to_metadata(self, metadata: Sequence[tuple[str, Union[str, bytes]]]) -> None:
+        """Provide relevant auth tokens in the metadata."""
```
panh99, line 76:
```suggestion
        """Write relevant auth tokens to the provided metadata."""
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @abstractmethod
+    def store_refresh_token(
+        self, metadata: Sequence[tuple[str, Union[str, bytes]]]
+    ) -> None:
+        """Store refresh tokens to .credentials."""
```
panh99, line 82:
```suggestion
        """Store refresh tokens in the .credentials/ folder within the Flower directory."""
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+    def validate_token_in_metadata(
+        self, metadata: Sequence[tuple[str, Union[str, bytes]]]
+    ) -> bool:
+        """Authenticate auth tokens in the provided metadata."""
+
+    @abstractmethod
+    def get_auth_token_response(
+        self, request: GetAuthTokenRequest
+    ) -> GetAuthTokenResponse:
+        """Send relevant tokens as a GetAuthTokenResponse."""
```
panh99, line 53: Is this method to communicate with the 3rd party service (like Keycloak) and return a `GetAuthTokenResponse`? In this case, should we amend the method as follows?
```suggestion
    def get_auth_tokens(
        self, request: GetAuthTokenRequest
    ) -> GetAuthTokenResponse:
        """Get the relevant auth tokens."""
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @abstractmethod
+    def refresh_token(self, context: grpc.ServicerContext) -> bool:
+        """Refresh auth tokens in the provided metadata."""
```
panh99, line 57:
```suggestion
    def refresh_tokens(self, context: grpc.ServicerContext) -> bool:
        """Refresh auth tokens in the metadata of the provided context."""
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @abstractmethod
+    def __init__(self, config: dict[str, Any], config_path: Path):
+        """Abstract constructor (init)."""
+
+    @abstractmethod
+    def write_token_to_metadata(
+        self, metadata: Sequence[tuple[str, Union[str, bytes]]]
+    ) -> None:
+        """Write relevant auth tokens to the provided metadata."""
```
panh99, line 81:
```suggestion
    def write_tokens_to_metadata(
        self, metadata: Sequence[tuple[str, Union[str, bytes]]]
    ) -> None:
        """Write relevant auth tokens to the provided metadata."""
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @abstractmethod
+    def store_refresh_token(
```
panh99, line 84:
```suggestion
    def store_refresh_tokens(
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+class ExecAuthPlugin(ABC):
+    """Abstract Flower Auth Plugin class for ExecServicer."""
+
+    @abstractmethod
+    def __init__(self, config: dict[str, Any]):
+        """Abstract constructor (init)."""
+
+    @abstractmethod
+    def get_login_response(self) -> LoginResponse:
+        """Send relevant login details as a LoginResponse."""
```
panh99, line 41:
```suggestion
```suggestion
    def login(self) -> LoginResponse:
        """Send the login details and return the server's response."""
```
```

Comment (danielnugraha) [reply src/py/flwr/common/auth_plugin/auth_plugin.py:41]: This is only the login url that is being sent back to the user. The login process itself does not happen here.

Comment (panh99) [reply src/py/flwr/common/auth_plugin/auth_plugin.py:41]: Oh, I see. Thanks for the clarification. Could we then:

```suggestion
    def get_login_response(self) -> LoginResponse:
        """Get the LoginResponse containing the login details."""
```

panh99 (approved):
LGTM!

panh99 (approved):

panh99 (approved):

panh99 (approved):

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,105 @@
...
+        self, metadata: Sequence[tuple[str, Union[str, bytes]]]
+    ) -> Sequence[tuple[str, Union[str, bytes]]]:
+        """Write relevant auth tokens to the provided metadata."""
+
+    @abstractmethod
+    def store_tokens(self, config: dict[str, Any]) -> None:
+        """Store the tokens from the provided config.
+
+        The tokens will be saved as a JSON file in the `.credentials/` folder inside the
+        Flower directory.
```
panh99, line 104:
```suggestion
        The tokens will be saved as a JSON file in provided config_path.
```

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class CliAuthPlugin(ABC):
+    """Abstract Flower Auth Plugin class for CLI."""
+
+    @staticmethod
+    @abstractmethod
+    def login(
+        login_details: dict[str, str],
+        config: dict[str, Any],
```
jafermarq, line 66: Wouldn't it better to make use of some of the types we have for configs ? For example https://github.com/adap/flower/blob/fc8591fce970c2a533deb0000427d24075797205/src/py/flwr/common/typing.py#L65

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,104 @@
...
+
+class ExecAuthPlugin(ABC):
+    """Abstract Flower Auth Plugin class for ExecServicer."""
+
+    @abstractmethod
+    def __init__(self, config: dict[str, Any]):
+        """Abstract constructor (init)."""
+
+    @abstractmethod
+    def get_login_details(self) -> GetLoginDetailsResponse:
```
danieljanes, line 39: Does it make sense to use ProtoBuf messages here, or should we use a Python class? (Same question below)

panh99: Yeah, make sense to me.

src/py/flwr/common/auth_plugin/auth_plugin.py
panh99: This function expects a json dictionary here, and it would be rather flexible if we just use `dict[str, Any]`. Though we normally just need to store scalars in the json dict, I am not quite sure what's the benefit of enforcing a stricter type here.

jafermarq: It receives a `{string: string}` (https://github.com/adap/flower/blob/0730b12a9e9653782c7ee1777c4382f50380217a/src/proto/flwr/proto/exec.proto#L67) unless we modify the `exec.proto`. Happy to leave it as `Any`, but in other parts of the code we are more string with these things (e.g. `typing.UserConfig`)

src/py/flwr/common/auth_plugin/auth_plugin.py
```diff
@@ -0,0 +1,117 @@
...
+    user_auth_config_path : Path
+        The path to the user's authentication configuration file.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def login(
+        login_details: dict[str, str],
+        project_config: dict[str, Any],
+        federation: str,
```
danieljanes, line 74: Instead of this, wouldn't it be better to just pass the federation config the user selected?

danieljanes (approved):

PR #4651, opened 2024-12-05
refactor(framework) Complete minimum `flwr stop` imlementation

jafermarq:
Completes `flwr stop` and modifies `StopRunResponse` and its creation in the `exec_servicer` (the idea is to not use `gRPC` abort and instead return a reason why run couldn't be stopped)


src/py/flwr/superexec/exec_servicer.py
```diff
@@ -138,15 +138,13 @@ def StopRun(
 
         # Exit if `run_id` not found
         if not state.get_run(request.run_id):
-            context.abort(
-                grpc.StatusCode.NOT_FOUND, f"Run ID {request.run_id} not found"
-            )
+            return StopRunResponse(success=False, reason="Run not found")
```
chongshenng, line 141: We use `context.abort` in the other RPC methods like `StreamLogs` (on [this line](https://github.com/adap/flower/blob/0ea7a0e0ae44219426c4e056a57c5328ecbd37de/src/py/flwr/superexec/exec_servicer.py#L95)). Should we use the same `context.abort` for consistency?

src/py/flwr/superexec/exec_servicer.py
```diff
@@ -138,15 +138,13 @@ def StopRun(
...
 
         run_status = state.get_run_status({request.run_id})[request.run_id]
+        # Exit if run already finished
         if run_status.status == Status.FINISHED:
-            context.abort(
-                grpc.StatusCode.FAILED_PRECONDITION,
-                f"Run ID {request.run_id} is already finished",
+            return StopRunResponse(
+                success=False, reason=f"Run ID {request.run_id} is already finished"
```
chongshenng, line 147: I think the idea is to use the gRPC status codes and eventually unify the handling of the errors. There's also this earlier comment about using [statuscode](https://github.com/adap/flower/pull/4634#discussion_r1871099540).

src/py/flwr/cli/stop.py
```diff
@@ -78,6 +73,8 @@ def stop(
         channel = _init_channel(app, federation_config)
         stub = ExecStub(channel)  # pylint: disable=unused-variable # noqa: F841
 
+        _stop_run(stub, run_id=run_id)
```
chongshenng, line 76:
```suggestion
         typer.secho(f"✋ Stopping run ID {run_id}...", fg=typer.colors.GREEN)
         _stop_run(stub, run_id=run_id)
```

Let's add a user notification here like this?

chongshenng (commented):
@jafermarq Tiny comments, looks good overall!

chongshenng (commented):
LGTM!

PR #4674, opened 2024-12-11
refactor(framework) Fix dtype to `np.int64` in `pseudo_rand_gen` function

panh99:
The dtype of the result of `np.random.randint` may vary cross platforms and versions. It's better to fix the dtype.


tanertopal (approved):
lgtm!

PR #4683, opened 2024-12-11
feat(framework) Add`run_id` to `fab.proto`

chongshenng:
Required in the `FleetServicer.GetFab` to check if the provided `run_id` is stopped.


src/proto/flwr/proto/fab.proto
```diff
@@ -31,5 +31,6 @@ message Fab {
 message GetFabRequest {
   Node node = 1;
   string hash_str = 2;
+  optional uint64 run_id = 3;
```
jafermarq, line 34:
```suggestion
  uint64 run_id = 3;
```

jafermarq: better to expect it always?

jafermarq (approved):

PR #4699, opened 2024-12-13
feat(datasets) Add VerticalSizePartitioner

adam-narozniak:
## Proposal
Enable users to specify the vertical split in terms of an absolute number of columns per partition or fraction of the total columns.

```python
from flwr_datasets.partitioner import VerticalSizePartitioner

partitioner = VerticalSizePartitioner(
    partition_sizes=[8, 4, 2],
    active_party_column="income",
    active_party_columns_mode="create_as_last"
)
fds = FederatedDataset(
    dataset="scikit-learn/adult-census-income",
    partitioners={"train": partitioner}
)
partitions = [fds.load_partition(i) for i in range(fds.partitioners["train"].num_partitions)]
print([partition.column_names for partition in partitions])
# [['occupation', 'relationship', 'sex', 'education', 'age', 'capital.loss', 'native.country', 'capital.gain'], ['marital.status', 'fnlwgt', 'education.num', 'hours.per.week'], ['workclass', 'race'], ['income']]
```
easier to read output format (print with pprint)
```
[['occupation',
  'relationship',
  'sex',
  'education',
  'age',
  'capital.loss',
  'native.country',
  'capital.gain'],
 ['marital.status', 'fnlwgt', 'education.num', 'hours.per.week'],
 ['workclass', 'race'],
 ['income']]

### CODE REVIEW

File: datasets/flwr_datasets/partitioner/vertical_size_partitioner.py [hunk 1]
```diff
@@ -0,0 +1,300 @@
...
+                raise ValueError(
+                    "All floats in column_distribution must be >= 0.0 and <= 1.0."
+                )
+        elif all(
+            isinstance(coulumn_count, int) for coulumn_count in self._partition_sizes
+        ):
+            if any(coulumn_count < 0 for coulumn_count in self._partition_sizes):
+                raise ValueError("All integers in column_distribution must be >= 0.")
+        else:
+            raise ValueError("partition_sizes list must be all floats or all ints.")
```
jafermarq, line 214:
```suggestion
                raise ValueError("Float ratios in `partition_sizes` must sum to 1.0.")
            if any(
                fraction < 0.0 or fraction > 1.0 for fraction in self._partition_sizes
            ):
                raise ValueError(
                    "All floats in `partition_sizes` must be >= 0.0 and <= 1.0."
                )
        elif all(
            isinstance(coulumn_count, int) for coulumn_count in self._partition_sizes
        ):
            if any(coulumn_count < 0 for coulumn_count in self._partition_sizes):
                raise ValueError("All integers in `partition_sizes` must be >= 0.")
        else:
            raise ValueError("`partition_sizes` list must be all floats or all ints.")
```

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
```diff
@@ -0,0 +1,300 @@
...
+        partition_sizes: Union[list[int], list[float]],
+        active_party_column: Optional[Union[str, list[str]]] = None,
+        active_party_columns_mode: Union[
+            Literal[
+                "add_to_first",
+                "add_to_last",
+                "create_as_first",
+                "create_as_last",
+                "add_to_all",
+            ],
```
jafermarq, line 102: it might be worth defining this at the top of the file so to avoid redefining it later a couple of times (see `_validate_parameters_in_init()` and `vertical_partitioner_utils.py`)

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
```diff
@@ -0,0 +1,300 @@
...
+        for column in active_party_columns:
+            if column not in all_columns:
+                raise ValueError(
+                    f"Active party column '{column}' not found in the dataset."
+                )
+        num_columns = len(all_columns)
+        if all(isinstance(size, int) for size in self._partition_sizes):
+            if sum(self._partition_sizes) > num_columns:
+                raise ValueError(
+                    "Sum of partition sizes cannot exceed the total number of columns."
```
jafermarq, line 262: Should this check be modified so it tests that the `sum` is exactly the number of columns? (so to match the logic done when the `partition_sizes` are passed as `list[float]` ? (`_validate_parameters_in_init()` checks if they add up to `1.0`)

datasets/flwr_datasets/partitioner/vertical_partitioner_utils.py
```diff
@@ -0,0 +1,103 @@
...
+        Mode to add active party columns to partition columns.
+
+    Returns
+    -------
+    partition_columns: list[list[str]]
+        List of partition columns after the modyfication.
+    """
+    if isinstance(active_party_columns_mode, int):
+        partition_id = active_party_columns_mode
+        if partition_id < 0 or partition_id >= len(partition_columns):
```
jafermarq, line 80: It might be an odd way of using the partitioner given that `add_to_last` exists but, would it be better to allow the integer to be even `= len(partition_columns)` so it gets added at the end?

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Returns
+        -------
+        dataset_partition : Dataset
+            Single partition of a dataset.
+        """
+        self._determine_partitions_if_needed()
+        assert self._partition_columns is not None
+        if partition_id < 0 or partition_id >= len(self._partition_columns):
+            raise ValueError(f"Invalid partition_id {partition_id}.")
```
jafermarq, line 184: other partitioners don't have this check. if partition_id is out of range then the error will be more informative that this `ValueError`. Happy to keep this check before accessing the list of `partition_columns` but maybe we need to add a bit more of a reason why it's an invalid `partition_id`?

datasets/flwr_datasets/partitioner/__init__.py
```diff
@@ -29,6 +29,8 @@
 from .shard_partitioner import ShardPartitioner
 from .size_partitioner import SizePartitioner
 from .square_partitioner import SquarePartitioner
+from .vertical_even_partitioner import VerticalEvenPartitioner
```
jafermarq, line 32:
```suggestion
```

datasets/flwr_datasets/partitioner/__init__.py
```diff
@@ -45,4 +47,6 @@
     "ShardPartitioner",
     "SizePartitioner",
     "SquarePartitioner",
+    "VerticalEvenPartitioner",
```
jafermarq, line 50:
```suggestion
```

datasets/flwr_datasets/partitioner/vertical_partitioner_utils.py
adam-narozniak: It has the functionality you describe (it's possible to add it by index to the last value)
The last element is indexed by `len(array) - 1`
e.g.
```bash
a = [10,20,30]
len(a)
# it's 3 
a[3] # gives error
a[len(a) - 1] # gives 30 = the last element
```

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
adam-narozniak: I can't find it in Python docs exactly but `Literal[*some_list]` gives a syntax error (at least for Python 3.10 so it should be the problem at least up to that point)

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
adam-narozniak: Ok, I'll add more info. Maybe also change the exception type to be index-related.

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
adam-narozniak: Thanks, yeah, let's keep it consistent. I'll change it.

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
adam-narozniak: Done

datasets/flwr_datasets/partitioner/vertical_size_partitioner.py
adam-narozniak: Done

datasets/flwr_datasets/partitioner/vertical_size_partitioner_test.py
```diff
@@ -0,0 +1,186 @@
...
+        partitioner = VerticalSizePartitioner(
+            partition_sizes=[1, 1], shared_columns=["shared"], shuffle=False
+        )
+        partitioner.dataset = dataset
+        p0 = partitioner.load_partition(0)
+        p1 = partitioner.load_partition(1)
+        self.assertIn("shared", p0.column_names)
+        self.assertIn("shared", p1.column_names)
+
+    def test_partitioning_with_active_party_add_to_last(self) -> None:
```
jafermarq, line 121: this test and the one below needs to be revisited given the change in how `partition_sizes` expects `list[int]` (they now need to add up to the number of columns).

adam-narozniak: I've added new tests now but also updated the size check which was incorrent.

jafermarq (approved):

PR #4712, opened 2024-12-16
refactor(framework) Add `try`-`except` block to `flwr stop` for `JSON` output

chongshenng:


jafermarq (approved):

src/py/flwr/cli/stop.py
```diff
@@ -48,34 +48,39 @@ def stop(
...
+                bold=True,
+            )
+            raise typer.Exit(code=1) from err
+        finally:
+            channel.close()
+    except (typer.Exit, Exception):  # pylint: disable=broad-except, W0612  # noqa: F841
+        pass
     finally:
-        channel.close()
+        pass
```
danieljanes, line 83: We can remove `finally` if we don't use it

src/py/flwr/cli/stop.py
```diff
@@ -48,34 +48,39 @@ def stop(
...
+            typer.secho(
+                f"❌ {err}",
+                fg=typer.colors.RED,
+                bold=True,
+            )
+            raise typer.Exit(code=1) from err
+        finally:
+            channel.close()
+    except (typer.Exit, Exception):  # pylint: disable=broad-except, W0612  # noqa: F841
+        pass
```
danieljanes, line 81: This "swallows" the exception, which mean `typer.Exit` ceases to work

src/py/flwr/cli/stop.py
chongshenng: We are using it in #4710. It's not very ideal, but the diff is reduced.

src/py/flwr/cli/stop.py
chongshenng: Similar to above, we will catch the `typer.Exit` in #4710.

danieljanes (approved):

PR #4736, opened 2024-12-17
fix(framework) Clear node auth keys upon `SuperLink` start

jafermarq:


src/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -820,6 +820,11 @@ def get_server_public_key(self) -> Optional[bytes]:
             public_key = None
         return public_key
 
+    def clear_node_public_keys(self) -> None:
+        """Clear stored `node_public_keys` in the link state if any."""
+        query = "DELETE FROM public_key;"
+        self.query(query)
```
chongshenng, line 826: IIUC, the main issue is the reinsertion of the `public_key` and `private_key` into the credential database. We could equivalently delete all entries in that table to enable reinsertion. That is, we'd use `"DELETE from credentials;"`. So the logic would be to (1) clear the `node_public_keys` (2) clear the `credentials` table, then finally (3) call `store_server_private_public_key`. 

I think we shouldn't change the logic of `store_server_private_public_key` in #4737 - it is correct that the keys should be prevented from being inserted if they already exist (and raise a a `RuntimeError`). Instead, the flow logic should be such that the credentials must be reset each time the SuperLink is restarted. IMO this would align with good credential management practices.

chongshenng (changes requested):
@jafermarq I like the logic of this approach, but I think we need to also consider resetting the credentials before calling `store_server_private_public_key`. See my comment.

src/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
jafermarq: ok. that makes sense to me. Ok to do this in #4737 and leave this PR as is?

chongshenng (approved):
While not ideal, I'm approving this since it will enable us to restart the stateful SuperLink with node authentication.

danieljanes (approved):

PR #4737, opened 2024-12-17
fix(framework) Enable re-insertion of server keys upon `SuperLink` restart

jafermarq:


chongshenng (commented):
This change may not be needed based on my comment [here](https://github.com/adap/flower/pull/4736#discussion_r1889210792).

Also, this won't pass the existing CI since the original logic of `store_server_private_public_key` is to prevent reinsertion of the `private_key` and `public_key`.

PR #4745, opened 2024-12-18
refactor(framework) Add `.credentials` to `.gitignore`

chongshenng:
To prevent unintentionally committing `.credentials` to the repository, add `.gitignore` to the Flower directory and write `.credentials` to it. Alternatively, if `.gitignore` already exists in the Flower directory, append to it.

### DIFF
--- a/src/py/flwr/cli/utils.py
+++ b/src/py/flwr/cli/utils.py
@@ -163,8 +163,8 @@ def get_sha256_hash(file_path: Path) -> str:
 def get_user_auth_config_path(root_dir: Path, federation: str) -> Path:
     """Return the path to the user auth config file.
 
-    Additionally, a `.gitignore` file will be created or updated in the
-    Flower directory to include `.credentials`.
+    Additionally, a `.gitignore` file will be created or updated in the Flower
+    directory to include the `.credentials` folder to be excluded from git.
     """
     # Locate the credentials directory
     abs_flwr_dir = root_dir.absolute() / FLWR_DIR
--- a/src/py/flwr/cli/utils.py
+++ b/src/py/flwr/cli/utils.py
@@ -163,8 +163,10 @@ def get_sha256_hash(file_path: Path) -> str:
 def get_user_auth_config_path(root_dir: Path, federation: str) -> Path:
     """Return the path to the user auth config file.
 
-    Additionally, a `.gitignore` file will be created or updated in the Flower
-    directory to include the `.credentials` folder to be excluded from git.
+    Additionally, a `.gitignore` file will be created in the Flower directory to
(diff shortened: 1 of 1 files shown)


src/py/flwr/cli/utils.py
```diff
@@ -161,10 +161,43 @@ def get_sha256_hash(file_path: Path) -> str:
...
+    """Return the path to the user auth config file.
+
+    Additionally, a `.gitignore` file will be created or updated in the
+    Flower directory to include `.credentials`."""
     # Locate the credentials directory
     credentials_dir = root_dir.absolute() / FLWR_DIR / CREDENTIALS_DIR
     credentials_dir.mkdir(parents=True, exist_ok=True)
+
+    # Determine the absolute path of the Flower directory for .gitignore
+    abs_flwr_dir = root_dir.absolute() / FLWR_DIR
```
jafermarq, line 174: should we make use of this variable earlier in the function? to avoid mistakes? (`credentials_dir` shares part of this)

src/py/flwr/cli/utils.py
```diff
@@ -161,10 +161,43 @@ def get_sha256_hash(file_path: Path) -> str:
...
+            with open(gitignore_path, "w", encoding="utf-8") as gitignore_file:
+                gitignore_file.write(f"{credential_entry}\n")
+    except Exception as err:
+        typer.secho(
+            "❌ An error occurred while handling .gitignore. "
+            f"Please check the permissions of {gitignore_path} and try again.",
+            fg=typer.colors.RED,
+            bold=True,
+        )
+        raise typer.Exit(code=1) from err
```
jafermarq, line 200: I'm not sure about terminating the login flow because of this. Wouldn't it be sufficient to raise a warning and tell the user to manually update/create the `.gitignore` ?

chongshenng: IMO this encourages good practice and it doesn't really affect the login flow. This will happen in the rare instance that `.flwr` has bad permissions.

src/py/flwr/cli/utils.py
```diff
@@ -161,10 +161,44 @@ def get_sha256_hash(file_path: Path) -> str:
 
 
 def get_user_auth_config_path(root_dir: Path, federation: str) -> Path:
-    """Return the path to the user auth config file."""
+    """Return the path to the user auth config file.
+
+    Additionally, a `.gitignore` file will be created or updated in the
+    Flower directory to include `.credentials`.
```
jafermarq, line 167:
```suggestion
    Flower directory to exclude `.credentials`.
```

chongshenng: I clarified it, thanks!

danieljanes (approved):

PR #4765, opened 2025-01-05
feat(framework) Add `set_node_public_key` and `get_node_public_key` methods to `LinkState`

panh99:
This PR aims to decouple node creation from public key assignment. Key changes include:

- Removed the `public_key` argument from the `create_node` and `delete_node` methods.
- Added and implemented the `set_node_public_key` and `get_node_public_key` methods.
- Updated all unit tests and usages accordingly.


src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor_test.py
```diff
@@ -458,65 +438,8 @@ def test_unsuccessful_ping_with_metadata(self) -> None:
                 ),
             )
 
-    def test_successful_restore_node(self) -> None:
```
panh99, line 461: I am a bit confused by this: do we really want the owner of a public key to always get the same `node_id`?

danielnugraha: As long as `delete_node` not triggered, the `node_id` and `public_key` pair is maintained.

jafermarq: by this then should we amend how `.create_node()` works so it first checks if the public key is already present and if so, skip the generation of a new `node_id` via:

```
uint64_node_id = generate_rand_int_from_bytes(NODE_ID_NUM_BYTES)
```

panh99: I think this is related to the reconnection topic, i.e., keeping the same node ID when re-launching the SuperNode. I'd prefer not to do that in node authentication, as it causes an obvious inconsistency between auth and non-auth modes.

src/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -675,26 +666,22 @@ def create_node(
         # Note: we need to return the uint64 value of the node_id
         return uint64_node_id
 
-    def delete_node(self, node_id: int, public_key: Optional[bytes] = None) -> None:
+    def delete_node(self, node_id: int) -> None:
```
jafermarq, line 669: Should we also remove the public key entry? (as we do in `InMemoryLinkState`)

src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
```diff
@@ -223,5 +223,6 @@ def _create_authenticated_node(
         # No `node_id` exists for the provided `public_key`
         # Handle `CreateNode` here instead of calling the default method handler
         # Note: the innermost `CreateNode` method will never be called
```
jafermarq, line 225: do we need to update these ? what does the last comment mean?

src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
```diff
@@ -223,5 +223,6 @@ def _create_authenticated_node(
...
-        node_id = state.create_node(request.ping_interval, public_key_bytes)
+        node_id = state.create_node(request.ping_interval)
+        state.set_node_public_key(node_id, public_key_bytes)
```
jafermarq, line 227: I think this should be wrapped in a `try/except` ?

src/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -72,14 +72,14 @@
...
-    public_key BLOB
+    private_key     BLOB PRIMARY KEY,
+    public_key      BLOB
 );
 """
 
 SQL_CREATE_TABLE_PUBLIC_KEY = """
 CREATE TABLE IF NOT EXISTS public_key(
-    public_key BLOB UNIQUE
+    public_key      BLOB PRIMARY KEY
```
jafermarq, line 82: what's the motivation for changing from `UNIQUE` to `PRIMARY KEY` here?

src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
panh99: Great catch! Yes, we need to update it. Previously the `Servicer.CreateNode` will never be invoked.

panh99: Oh, we this PR only aims to update LinkState. The server interceptor will be re-implemented in a follow-up PR.

src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
panh99: I think it's fine for this PR. It only aims to update LinkState. The server interceptor will be re-implemented in a follow-up PR.

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -62,6 +62,7 @@ def __init__(self) -> None:
         # Map node_id to (online_until, ping_interval)
         self.node_ids: dict[int, tuple[float, float]] = {}
         self.public_key_to_node_id: dict[bytes, int] = {}
```
danielnugraha, line 64: should we delete this one?

src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
danielnugraha, line 227: should we differentiate between reactivated `node_id` and new `node_id` assignment here?

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: It's still used in `get_node_id` method. We actually need a bi-directional dictionary.

src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
panh99: Since we don't have reactivation or reconnection feature in non-authenticated mode, I think it's better to keep it as it is, to avoid inconsistency between auth and non-auth modes.

danieljanes (approved):

PR #4792, opened 2025-01-07
refactor(framework) Make `rest` client communicate `Messages`

jafermarq:
> [!NOTE]
> Merge after https://github.com/adap/flower/pull/4312


src/py/flwr/client/rest_client/connection.py
```diff
@@ -69,7 +67,9 @@
 PATH_CREATE_NODE: str = "api/v0/fleet/create-node"
 PATH_DELETE_NODE: str = "api/v0/fleet/delete-node"
 PATH_PULL_TASK_INS: str = "api/v0/fleet/pull-task-ins"
+PATH_PULL_MESSAGES: str = "/api/v1/fleet/pull-messages"
```
panh99, line 70: What's the difference between `v1` and `v0`? Not sure why this line uses `v1` but all other existing ones use `v0`.

src/py/flwr/client/rest_client/connection.py
```diff
@@ -69,7 +67,9 @@
...
 PATH_PUSH_TASK_RES: str = "api/v0/fleet/push-task-res"
+PATH_PUSH_MESSAGES: str = "/api/v1/fleet/push-messages"
```
panh99, line 72: Same question here.

panh99 (commented):
Just wondered if you open another PR to update `src/py/flwr/server/superlink/fleet/rest_rere/rest_api.py`? It's like the "FleetServicer" for `rest`.

danieljanes:
> LGTM! @danieljanes Could you confirm that we should use `/api/v0` everywhere, or we should switch to `/api/v1`?

`v0` is fine for now, the API is still marked as experimental.

danieljanes (approved):

PR #4795, opened 2025-01-08
fix(framework) Update behaviour of `_raise_if` in `ServerAppIo`

jafermarq:


chongshenng (approved):
lgtm!

src/py/flwr/server/superlink/driver/serverappio_servicer.py
```diff
@@ -345,5 +346,18 @@ def GetRunStatus(
...
 
 def _raise_if(validation_error: bool, detail: str) -> None:
+    """Raise a `ValueError` with a detailed message if a validation error occurs.
+
+    If `validation_error` is `True`, a `ValueError` will be raised with a message in the
+    format "Malformed <CallerFunctionName>Request: <detail>".
+    """
     if validation_error:
-        raise ValueError(f"Malformed PushTaskInsRequest: {detail}")
+        caller_frame = inspect.currentframe()
```
danieljanes, line 355: Instead of this, can we manually pass the name? I'm worried that, if we move those `_raise_if` calls, all the messages change without us noticing.

danieljanes (approved):

PR #4798, opened 2025-01-09
feat(framework) Make `flwr-serverapp` / `flwr-clientapp` / `flwr-simulation` subprocesses daemonic

panh99:
Ensure daemon subprocesses which will be killed automatically when the main process terminates.

### Lessons Learned

`gRPC` doesn't work well with `fork`. Once the gRPC server is started, calling `os.fork` can leave the gRPC library in an unpredictable state in the child process. Thus we need to use `"spawn"` as the start method for `multiprocessing`.


danieljanes (commented):
PR lgtm, but I'd recommend renaming two variables to avoid confusing them with other common names:
- `ctx` often gets used for the Flower `Context`, we could rename it to `spawn_context` (or something else)
- `mod` often gets used for Flower Mods, we could rename it to `app_module` (or something else)

src/py/flwr/client/app.py
```diff
@@ -826,6 +829,12 @@ def signal_handler(sig, frame):  # type: ignore
         signal.signal(signal.SIGTERM, signal_handler)
 
 
+def _run_flwr_clientapp(args: list[str]) -> None:
+    sys.argv = args
+    mod = importlib.import_module("flwr.client.clientapp.app")
+    mod.flwr_clientapp()
```
danieljanes, line 835: Instead of doing this, can we just do a regular import? It's better to avoid dynamic imports like this because they will silently break when renaming the module or package.

src/py/flwr/server/app.py
```diff
@@ -446,22 +447,37 @@ def run_superlink() -> None:
...
 
 
+def _run_flwr_command(args: list[str]) -> None:
+    sys.argv = args
+    if args[0] == "flwr-serverapp":
+        mod = importlib.import_module("flwr.server.serverapp.app")
+        mod.flwr_serverapp()
+    elif args[0] == "flwr-simulation":
+        mod = importlib.import_module("flwr.simulation.app")
+        mod.flwr_simulation()
```
danieljanes, line 457: Same `import`-related question as above

src/py/flwr/client/app.py
panh99: I've updated it. Thanks!

src/py/flwr/server/app.py
panh99: Updated, thanks!

danieljanes (approved):

PR #4818, opened 2025-01-15
fix(baselines) Bump `scikit-learn` versions to 1.5.0

chongshenng:
Needs
* #4824

### DIFF
--- a/examples/vertical-fl/pyproject.toml
+++ b/examples/vertical-fl/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "flwr-datasets>=0.3.0",
     "pandas==2.0.3",
     "numpy>=1.26.0",
-    "scikit-learn==1.3.2",
+    "scikit-learn==1.5.0",
     "torch==2.5.1",
 ]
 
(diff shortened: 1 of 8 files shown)


examples/vertical-fl/pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
     "flwr-datasets>=0.3.0",
     "pandas==2.0.3",
```
jafermarq, line 13:
```suggestion
    "pandas==2.2.3",
```

jafermarq (commented):
@chongshenng all looks good, i tested the baselines with the biggest diff in the `pyproject.toml`. I noted that `pandas` needs to be bumped for the `vertical-fl` example. Could you check if you observe the same and that the version i propose fixes it?

examples/vertical-fl/pyproject.toml
chongshenng: Yes! This change is needed for `vertical-fl` to work. Good catch!

jafermarq (approved):

PR #4826, opened 2025-01-16
feat(framework) Prevent zombie subprocesses when the main process receives `SIGKILL`

panh99:
Enabled the subprocesses to monitor the main process. If the main process exits due to `SIGKILL`, the subprocess will kill itself via `SIGKILL` too.

As the subprocess are `daemon` via `multiprocessing` library, this mechanism won't be triggered if the main process is killed by `SIGINT` / `SIGTERM` / `Ctrl+C` / etc.

### DIFF
--- a/src/py/flwr/client/app.py
+++ b/src/py/flwr/client/app.py
@@ -16,8 +16,10 @@
 
 
 import multiprocessing
+import os
 import signal
 import sys
+import threading
 import time
 from contextlib import AbstractContextManager
 from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


src/py/flwr/client/app.py
```diff
@@ -829,7 +831,17 @@ def signal_handler(sig, frame):  # type: ignore
         signal.signal(signal.SIGTERM, signal_handler)
 
 
-def _run_flwr_clientapp(args: list[str]) -> None:
+def _run_flwr_clientapp(args: list[str], main_pid: int) -> None:
```
danieljanes, line 834: If we add a new argument `main_pid` here, then we must also change all the places that call this function (for example, line 557)

panh99: Great catch! Thanks. I've fixed it.

danieljanes (approved):

PR #4827, opened 2025-01-16
refactor(framework) Use `Pull/PushMessages` in `client_interceptor_test.py`

panh99:


jafermarq:
@panh99 , I introduced test in https://github.com/adap/flower/pull/4875. But would be good to check if your proposal in this PR is aligned with what I did.

PR #4830, opened 2025-01-16
fix(baselines) Bump `virtualenv` dependency

jafermarq:


jafermarq:
Already handled: https://github.com/adap/flower/pull/4834

PR #4838, opened 2025-01-19
refactor(framework) Move necessary utils from `flwr.cli.config_utils` to `flwr.common.config`

panh99:
- Moved some utility functions, so that `flwr.common.config` no longer depends on `flwr.cli.config_utils`
- Deleted `load_from_string` in `flwr.cli.config_utils`
- ~~Optimised `get_fab_metadata` in `flwr.cli.config_utils`~~: this is achieved in #4843


jafermarq (approved):
quality refactoring

src/py/flwr/common/config.py
```diff
@@ -227,10 +229,10 @@ def parse_config_args(
 
 
 def get_metadata_from_config(config: dict[str, Any]) -> tuple[str, str]:
-    """Extract `fab_version` and `fab_id` from a project config."""
+    """Extract `fab_id` and `fab_version` from a project config."""
     return (
-        config["project"]["version"],
         f"{config['tool']['flwr']['app']['publisher']}/{config['project']['name']}",
+        config["project"]["version"],
```
chongshenng, line 235: Good that we can reuse `get_metadata_from_config` for `get_fab_metadata`. Can we do this change in a separate PR and only move code in this PR?

chongshenng (commented):
Overall looks good! Only one comment.

src/py/flwr/common/config.py
panh99: okay, np.

danieljanes (approved):

PR #4841, opened 2025-01-20
feat(framework) Enable federation configuration overrides via `--federation-config` in `flwr` CLI

panh99:
- **Dynamic Federation Config Overrides**: Enabled dynamic overrides for federation configurations in the following commands:`flwr login/run/log/stop/ls`
- **Bug Fix in `flwr log`**: Resolved an issue where exceptions in the `flwr log` command were being suppressed, causing silent failures without any error logs. With this fix, the command will now appropriately log exceptions.


src/py/flwr/cli/run/run.py
```diff
@@ -97,6 +109,7 @@ def run(
     ] = CliOutputFormat.DEFAULT,
 ) -> None:
     """Run Flower App."""
+    print(run_config_overrides)
```
jafermarq, line 112:
```suggestion
```

src/py/flwr/cli/run/run.py
```diff
@@ -67,16 +67,28 @@ def run(
...
+        ),
+    ] = None,
+    federation_config_overrides: Annotated[
+        Optional[list[str]],
+        typer.Option(
+            "--federation-config",
+            "-fc",
+            help="Override federation configuration values in the format:\n\n"
+            "`--federation-config 'key1=value1 key2=value2' --federation-config "
+            "'key3=value3'`\n\nValues can be of any type supported in TOML, such as "
```
jafermarq, line 89: Do you think is it necessary to show these two examples of `--federation-config`? I note this style is also present for `--run-config`. -- this appears in the other CLI-related files too.

src/py/flwr/cli/config_utils_test.py
```diff
@@ -227,6 +227,51 @@ def test_validate_federation_in_project_config() -> None:
     assert federation_config == {"new_key": "new_val"}
 
 
+def test_validate_federation_in_project_config_with_overrides() -> None:
+    """Test that validate_federation_in_config works with overrides."""
+    # Prepare - Test federation is None
+    federation_config = {"k1": "v1", "k2": True, "grp": {"k3": 42.8}}
```
jafermarq, line 233: are values of type `dict` supported in our config system? (I'm referring to the `grpc`'s value)

src/py/flwr/common/config.py
```diff
@@ -93,19 +95,28 @@ def get_project_config(project_dir: Union[str, Path]) -> dict[str, Any]:
 
 
 def fuse_dicts(
-    main_dict: UserConfig,
-    override_dict: UserConfig,
-) -> UserConfig:
+    main_dict: T_dict,
+    override_dict: T_dict,
```
jafermarq, line 99: Should we do this change now? we are going from a strongly typed `UserConfig` to a looser `str:Any`.

chongshenng (commented):
Thanks @panh99. Looks good overall. I left some small comments.

src/py/flwr/cli/run/run.py
panh99: Nice catch! Thanks.

src/py/flwr/cli/run/run.py
panh99: Yes, it's very similar to the help message of `--run-config`. Do you think we can remove or optimise the two example usages?

src/py/flwr/cli/config_utils_test.py
panh99: Protobuf can support this, but we don't have such a proto definition yet. In this sense, our `grpc` messages haven't support this yet. It's the original data structure from `tomli.load`.

panh99: The `federation_config` won't be saved in grpc message, so I think it's fine for now.

src/py/flwr/common/config.py
panh99: I think that's fine. If the input type is `UserConfig`, then the return type will also be `UserConfig`.

src/py/flwr/cli/run/run.py
jafermarq: I think we can leave it as is

jafermarq: thinking about it twice, how about if we put the help message for `--federation-config` in a single file (e.g. `cli/utlis.py`) and we prevent from it being copied in 5 files? wdyt @panh99 ?

danieljanes (approved):

PR #4842, opened 2025-01-21
refactor(examples) Update `app-pytorch` example

jafermarq:


chongshenng (approved):
LGTM! I pushed a commit to remove `requirements.txt` since we no longer need it.

PR #4866, opened 2025-01-28
refactor(framework) Improve reconnection logging for `ClientAppIo`

jafermarq:


src/py/flwr/client/clientapp/clientappio_servicer.py
```diff
@@ -83,11 +83,9 @@ def GetToken(
 
         # Fail if no ClientAppInputs are available
         if self.clientapp_input is None:
-            context.abort(
-                grpc.StatusCode.FAILED_PRECONDITION,
-                "No inputs available.",
-            )
-        clientapp_input = cast(ClientAppInputs, self.clientapp_input)
```
jafermarq, line 90: Not sure why `mypy` complained about this (https://github.com/adap/flower/actions/runs/13017485683/job/36310133108?pr=4866) while not complaining earlier or even in other lines just like that one still present in the class....

PR #4868, opened 2025-01-28
fix(framework) Fix gaussian noise addition for int64

### LINKED ISSUE #4828: DifferentialPrivacyClientSide fails with some Pytorch Models
### Describe the bug

Using DifferentialPrivacyClientSide will result in the following error: `Cannot cast ufunc 'add' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'` in https://github.com/adap/flower/blob/deca84658ab80cd7ed1b515e61b2b04dfa1e32cc/src/py/flwr/common/differential_privacy.py#L39

Some architectures have int64 for their weights, which will produce casting errors. They changed the behavior in numpy for casting a few versions ago (my version is numpy=2.2.1).

Using `array = np.add(array, np.random.normal(0, std_dev, array.shape), casting="unsafe)` will fix this problem.

### Steps/Code to Reproduce

will work:
```
from torch import nn

class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        # convolutional layer
        self.conv1 = nn.Conv2d(3, 16, 3, padding=1)
        self.conv2 = nn.Conv2d(16, 32, 3, padding=1)
        self.conv3 = nn.Conv2d(32, 16, 3, padding=1)
        # max pooling layer
        self.pool = nn.MaxPool2d(2, 2)

        # Fully connected layer
        self.fc1 = nn.Linear(16 * 4 * 4, 10)

        # Dropout
        self.dropout = nn.Dropout(p=0.2)

        # Output layer
        self.out = nn.LogSoftmax(dim=1)

    def flatten(self, x):
        return x.view(x.size()[0], -1)
model = Net()
```
won't:
```
from torchvision.models import efficientnet_v2_s

model = efficientnet_v2_s(weights="IMAGENET1K_V1")
```

### Expected Results

No casting error will be produced.

### Actual Results

`Cannot cast ufunc 'add' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'`

mohammadnaseri:
To fix #4828

### DIFF
--- a/src/py/flwr/common/differential_privacy.py
+++ b/src/py/flwr/common/differential_privacy.py
@@ -35,6 +35,7 @@ def get_norm(input_arrays: NDArrays) -> float:
     # pylint: disable=consider-using-generator
     return float(np.sqrt(sum([norm**2 for norm in array_norms])))
 
+
 def add_gaussian_noise_inplace(input_arrays: NDArrays, std_dev: float) -> None:
     """Add Gaussian noise to each element of the input arrays."""
     for array in input_arrays:
--- a/src/py/flwr/common/differential_privacy.py
+++ b/src/py/flwr/common/differential_privacy.py
@@ -37,8 +37,9 @@ def get_norm(input_arrays: NDArrays) -> float:
 
 
 def add_gaussian_noise_inplace(input_arrays: NDArrays, std_dev: float) -> None:
+    """Add Gaussian noise to each element of the input arrays."""
     for array in input_arrays:
-        noise = np.random.normal(0, std_dev, array.shape).astype(array.dtype)
+        noise = np.random.normal(0, std_dev, array.shape, dtype=array.dtype)
         array += noise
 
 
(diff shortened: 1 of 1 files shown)


src/py/flwr/common/differential_privacy.py
```diff
@@ -39,7 +39,8 @@ def get_norm(input_arrays: NDArrays) -> float:
 def add_gaussian_noise_inplace(input_arrays: NDArrays, std_dev: float) -> None:
     """Add Gaussian noise to each element of the input arrays."""
     for array in input_arrays:
-        array += np.random.normal(0, std_dev, array.shape)
+        noise = np.random.normal(0, std_dev, array.shape).astype(array.dtype)
```
panh99, line 42: Instead of `astype`, would it be possible to specify the dtype in the random function?
```suggestion
        noise = np.random.normal(0, std_dev, array.shape, dtype=array.dtype)
```

danieljanes (approved):

PR #4889, opened 2025-01-31
refactor(framework) Add error message on invalid overrides

danielnugraha:
### Description

When using overrides, invalid format will results in error which is not catch.

### Explanation

Catch the error on invalid overrides and output meaningful error message.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/src/py/flwr/common/config.py
+++ b/src/py/flwr/common/config.py
@@ -240,8 +240,8 @@ def parse_config_args(
             except tomli.TOMLDecodeError as err:
                 typer.secho(
                     "❌ The provided configuration string is in an invalid format. "
-                    'The correct format should be, e.g., "key1=123 key2=false '
-                    "key3='string'\", where values must be of type bool, int, "
+                    "The correct format should be, e.g., 'key1=123 key2=false "
+                    'key3="string"\', where values must be of type bool, int, '
                     "string, or float. Ensure proper formatting with "
                     "space-separated key-value pairs.",
                     fg=typer.colors.RED,
--- a/src/py/flwr/common/config.py
+++ b/src/py/flwr/common/config.py
@@ -239,10 +239,11 @@ def parse_config_args(
                 flat_overrides = flatten_dict(overrides) if flatten else overrides
             except tomli.TOMLDecodeError as err:
                 typer.secho(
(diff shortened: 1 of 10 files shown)


src/py/flwr/common/config.py
```diff
@@ -233,8 +234,19 @@ def parse_config_args(
...
-            flat_overrides = flatten_dict(overrides) if flatten else overrides
+            try:
+                overrides.update(tomli.loads(toml_str))
+                flat_overrides = flatten_dict(overrides) if flatten else overrides
+            except tomli.TOMLDecodeError as err:
+                typer.secho(
+                    "❌ Error parsing overrides due to invalid format. Ensure that "
+                    "your overrides uses supported types of bool, int, string, or "
+                    'float, e.g. true/false, "random string", or 123 and is formatted '
+                    "correctly, e.g. 'key1=value1 key2=value2' or 'key3=value3'.",
```
jafermarq, line 245:
```suggestion
                    "float, e.g. --federation-config=\"key1=123 key2='my string' "
                    "key3=true\"",
```

jafermarq: should we add instead a valid complete example?

panh99: Yes, a valid complete example would be good. However, we shouldn't use `--federation-config` here as this function is also used for `node-config` and `run-config`.

jafermarq: Ok, i'll remove the `--federation-config` bit

panh99, line 245: As this is also used for `--node-config`, it's better not mentioning "overrides" here. How about:
```suggestion
                    "❌ The provided configuration string is in an invalid format. "
                    "The correct format should be, e.g., `'key1=123 key2=false"
                    "key3=\"string\"'`, where values must be of type bool, int, "
                    "string, or float. Ensure proper formatting with "
                    "space-separated key-value pairs.",
```

danielnugraha: I pushed the suggestion from Pan.

jafermarq (approved):

panh99 (approved):
LGTM!

danieljanes (approved):

PR #4899, opened 2025-02-03
refactor(framework) Add allowance for system time drift in the node authentication

panh99:
### Why This PR
In cases of unsynchronized system time, the received timestamp may appear to be in the future relative to the server. In a test case, a Raspberry Pi was likely less than one second ahead of the SuperLink’s system time, causing the SuperLink to reject all its connections due to future timestamps.

### What This Does
Allows timestamps to be within the range (-2, 12) seconds instead of (0, 10) seconds.

### DIFF
--- a/src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
+++ b/src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
@@ -25,6 +25,7 @@ from flwr.common import now
 from flwr.common.constant import (
     PUBLIC_KEY_HEADER,
     SIGNATURE_HEADER,
+    SYSTEM_TIME_TOLERANCE,
     TIMESTAMP_HEADER,
     TIMESTAMP_TOLERANCE,
 )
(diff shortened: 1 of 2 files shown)


src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
```diff
@@ -38,6 +39,9 @@
 )
 from flwr.server.superlink.linkstate import LinkStateFactory
 
+MIN_TIMESTAMP_DIFF = -SYSTEM_TIME_TOLERANCE
+MAX_TIMESTAMP_DIFF = TIMESTAMP_TOLERANCE + SYSTEM_TIME_TOLERANCE
```
jafermarq, line 43:
```suggestion
MAX_TIMESTAMP_DIFF = TIMESTAMP_TOLERANCE - SYSTEM_TIME_TOLERANCE
```

jafermarq: just wondering if this should be a `-` so the total allowed time diff is 10s. a very minor point probably

jafermarq: Additionally, maybe it's better to use the constants from `flwr.common.constant` directly in the `if` statement below instead of creating new variables? wdyt?

panh99: No, the 10s is the general timestamp tolerance window for network delay. And the 2s is just for the system time drift. So the tolerance window is actually (0, 10) +-2 seconds.

panh99: Hmmm, I don't have a strong opinion on whether to derive new constants or to directly use the original constants. I initially thought it might be clearer to have the two derived constants. Wdyt?

jafermarq (approved):

tanertopal (approved):
Lgtm!

PR #4900, opened 2025-02-03
fix(framework) Avoid processing requests from non-FleetStub

panh99:
### Why This PR  
@jafermarq found that running `flwr ls` against a Fleet servicer results in an unauthenticated error triggered by the node authentication interceptor. A help message then appears, instructing users to run `flwr login`. This can be confusing, especially if the issue is simply due to a typo in the server address.  

### What This Does  
The node authentication server interceptor now first checks whether the request is for the Fleet API. If it is not, a `FAILED_PRECONDITION` error is raised instead of `UNAUTHENTICATED`.

### DIFF
--- a/src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
+++ b/src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
@@ -39,9 +39,11 @@ from flwr.proto.fleet_pb2 import (  # pylint: disable=E0611
 from flwr.server.superlink.linkstate import LinkStateFactory
 
 
-def _unary_unary_rpc_terminator(message: str) -> grpc.RpcMethodHandler:
+def _unary_unary_rpc_terminator(
+    message: str, code: Any = grpc.StatusCode.UNAUTHENTICATED
+) -> grpc.RpcMethodHandler:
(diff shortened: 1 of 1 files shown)


jafermarq (approved):

src/py/flwr/server/superlink/fleet/grpc_rere/server_interceptor.py
```diff
@@ -75,6 +77,13 @@ def intercept_service(
         metadata sent by the node. Continue RPC call if node is authenticated, else,
         terminate RPC call by setting context to abort.
         """
+        # Filter out non-Fleet service calls
+        if not handler_call_details.method.startswith("/flwr.proto.Fleet/"):
+            return _unary_unary_rpc_terminator(
+                "This request should be sent to a different service.",
+                grpc.StatusCode.FAILED_PRECONDITION,
+            )
+
```
tanertopal, line 86: Wouldn't this be something gRPC takes care of or this needed to ensure the `AuthenticateServerInterceptor` can only be used for the `flwr.proto.Fleet` endpoints?

panh99: @tanertopal Unfortunately no. We have to take care of this ourselves.

tanertopal (approved):
Lgtm!

PR #4902, opened 2025-02-04
refactor(framework) Rename constants for gRPC metadata keys for node auth / user auth

jafermarq:
Rename constants so they don't include `_`. Using `_` isn't recommended when these keys travel in gRPC's metadata (https://www.grouparoo.com/blog/dont-use-underscores-in-http-headers).

### DIFF
--- a/src/py/flwr/common/constant.py
+++ b/src/py/flwr/common/constant.py
@@ -108,7 +108,7 @@ MAX_RETRY_DELAY = 20  # Maximum delay duration between two consecutive retries.
 
 # Constants for user authentication
 CREDENTIALS_DIR = ".credentials"
-AUTH_TYPE_KEY = "auth_type"  # For key name in JSON file
+AUTH_TYPE_KEY = "auth-type"  # For key name in JSON file
 ACCESS_TOKEN_KEY = "flwr-oidc-access-token"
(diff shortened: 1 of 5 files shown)


src/py/flwr/common/constant.py
```diff
@@ -108,14 +108,14 @@
 
 # Constants for user authentication
 CREDENTIALS_DIR = ".credentials"
-AUTH_TYPE_KEY = "auth_type"
-ACCESS_TOKEN_KEY = "access_token"
-REFRESH_TOKEN_KEY = "refresh_token"
+AUTH_TYPE_KEY = "auth_type"  # For key name in JSON file
```
tanertopal, line 111: Any particular reason this one is keeping the underscore?

panh99: @tanertopal This is the key used in JSON file, so no need to change. Changes in this PR are necessary as it is not recommended to use `_` in key names. But we can also update this for consistency. Wdyt?

panh99, line 111:
```suggestion
AUTH_TYPE_KEY = "auth-type"  # For key name in JSON file
```

panh99: Updated, thanks for the suggestion.

tanertopal (approved):
Lgtm!

PR #4912, opened 2025-02-05
ci(*:skip) Improve `update_changelog.py` to include all contributors and make it faster

panh99:
### Why This PR

@dstripelis discovered that some co-authors of a PR were being ignored when collecting all contributor names. Additionally, `update_changelog.py` was too slow due to an I/O bottleneck.

### What This Does

- Retrieves GitHub usernames of co-authors from auto-generated commit messages and finds their full names based on these usernames.  
- Replaces GitHub API calls with local Git equivalents wherever possible.  
- Speeds up GitHub API calls using multi-threading.


dev/update_changelog.py
```diff
@@ -89,24 +109,82 @@ def _add_shortlog(new_version: str, shortlog: str) -> None:
...
+        ["git", "log", "--pretty=format:%H", f"{tag}..origin/main"],
+        stdout=subprocess.PIPE,
+        text=True,
+        check=True,
+    )
+    shas = set(result.stdout.splitlines())
+
+    # Fetch GitHub commits based on the SHA hashes
+    with ThreadPoolExecutor(max_workers=15) as executor:
+        commits = list(executor.map(repo.get_commit, shas))
```
dstripelis, line 125: (1) better initialize `commits` before the `with` context
(2) can we parameterize the `max_workers`, i.e., add as an input argument to the function?

dev/update_changelog.py
```diff
@@ -89,24 +109,82 @@ def _add_shortlog(new_version: str, shortlog: str) -> None:
...
+        if "[bot]" in commit.author.name:
+            return
+        with lock:
+            authors.add(commit.author)
+        # Find co-authors in the commit message
+        if matches := re.findall(coauthor_pattern, commit.commit.message):
+            with lock:
+                coauthor_names.update(name for name in matches)
+
+    with ThreadPoolExecutor(max_workers=15) as executor:
```
dstripelis, line 150: same here and below, better to parameterize the number of workers

dev/update_changelog.py
```diff
@@ -89,24 +109,82 @@ def _add_shortlog(new_version: str, shortlog: str) -> None:
...
+
+def _get_contributors_from_commits(api: Github, commits: set[Commit]) -> set[str]:
+    """Get a set of contributors from a set of commits."""
+    # Get authors and co-authors from the commits
+    authors: set[NamedUser] = set()
+    coauthor_names: set[str] = set()
+    coauthor_pattern = r"Co-authored-by:\s*(.+?)\s*<"
+    lock = Lock()
+
+    def retrieve(commit: Commit) -> None:
```
dstripelis, line 138: add underscore before `retrieve`

dev/update_changelog.py
```diff
@@ -89,24 +109,82 @@ def _add_shortlog(new_version: str, shortlog: str) -> None:
...
+    with ThreadPoolExecutor(max_workers=15) as executor:
+        executor.map(retrieve, commits)
+
+    # Remove repeated usernames
+    contributors = {author.name for author in authors if author.name}
+    coauthor_names -= contributors
+    coauthor_names.difference_update(author.login for author in authors)
+
+    # Get full names of the GitHub usernames
+    def get_user(username: str) -> None:
```
dstripelis, line 159: add underscore before `_get_user`

dev/update_changelog.py
```diff
@@ -263,28 +309,44 @@ def _insert_entry_no_desc(
...
-    if _update_changelog(prs):
+    # Get the shortlog and the pull requests since the latest tag
+    shortlog, prs = _get_pull_requests_since_tag(gh_api, repo, latest_tag)
+
+    # Update the changelog
+    print("Updating the changelog...")
+    if _update_changelog(prs, latest_tag):
         new_version = _bump_minor_version(latest_tag)
         if not new_version:
             print("Wrong tag format.")
```
dstripelis, line 352: Change print message from `succesfully` to `successfully`.

dev/update_changelog.py
```diff
@@ -52,12 +57,27 @@
...
+}
+
 
-def _get_latest_tag(gh_api: Github) -> tuple[Repository, Optional[Tag]]:
+def _get_latest_tag(gh_api: Github) -> tuple[Repository, str]:
     """Retrieve the latest tag from the GitHub repository."""
     repo = gh_api.get_repo(REPO_NAME)
-    tags = repo.get_tags()
-    return repo, tags[0] if tags.totalCount > 0 else None
+    latest_tag = subprocess.run(
```
dstripelis, line 74: isn't there an GH API we can use to get the tags? If we use the subprocess approach we need to make sure `git` is installed in the PATH.

dev/update_changelog.py
```diff
@@ -89,24 +109,82 @@ def _add_shortlog(new_version: str, shortlog: str) -> None:
...
+    def retrieve(commit: Commit) -> None:
+        if commit.author.name is None:
+            return
+        if "[bot]" in commit.author.name:
+            return
+        with lock:
+            authors.add(commit.author)
+        # Find co-authors in the commit message
+        if matches := re.findall(coauthor_pattern, commit.commit.message):
+            with lock:
```
dstripelis, line 147: better to not use `lock` because it is hard to reason and we want to avoid race conditions - try to implement the same logic by returning a collection through the `_retrieve` function. The call `executor.map()` returns an iterator, which then we can loop over and  construct the final `author` and `co_author` names

dev/update_changelog.py
```diff
@@ -89,24 +109,82 @@ def _add_shortlog(new_version: str, shortlog: str) -> None:
...
+    # Remove repeated usernames
+    contributors = {author.name for author in authors if author.name}
+    coauthor_names -= contributors
+    coauthor_names.difference_update(author.login for author in authors)
+
+    # Get full names of the GitHub usernames
+    def get_user(username: str) -> None:
+        try:
+            if name := api.get_user(username).name:
+                with lock:
```
dstripelis, line 162: same as above, let's remove the `lock`.

dev/update_changelog.py
```diff
@@ -263,28 +309,44 @@ def _insert_entry_no_desc(
...
-    major, minor, _ = [int(x) for x in match.groups()]
+    major, minor, _ = (int(x) for x in match.groups())
     # Increment the minor version and reset patch version
     new_version = f"v{major}.{minor + 1}.0"
     return new_version
 
 
+def _fetch_origin() -> None:
+    """Fetch the latest changes from the origin."""
+    subprocess.run(["git", "fetch", "origin"], check=True)
```
dstripelis, line 325: can we also use an API call for this?

dstripelis (approved):

PR #4918, opened 2025-02-07
feat(framework) Allow passing `NDArray` to `Array` constructor

panh99:


src/py/flwr/common/record/parametersrecord.py
```diff
@@ -57,6 +66,86 @@ class Array:
...
+        self,
+        *args: Any,
+        ndarray: NDArray | None = None,
+        dtype: str | None = None,
+        shape: list[int] | None = None,
+        stype: str | None = None,
+        data: bytes | None = None,
+    ) -> None:
+        # Init all arguments
+        if len(args) > 4:
```
chongshenng, line 87: Can you clarify (maybe in the docstring) what this 4 is referring to?

chongshenng (commented):
Thanks @panh99. Lgtm, I only have one minor comment.

src/py/flwr/common/record/parametersrecord.py
panh99: Thanks for reviewing!

The constructor of `Array` will accept the following arguments:

1. `Array(<numpy.ndarray>)`
2. `Array(<dtype>, <shape>, <stype>, <data>)`

So, if users provide more than 4 arguments in `*args`, we don't need to proceed with other checks, as we are sure this is invalid.

panh99: Would you suggest to add a comment here to ensure clarity?

panh99: Thanks. I've added docstrings/comments

src/py/flwr/common/record/conversion_utils.py
jafermarq, line 1: Should we remove this file? or at least add a deprecation warning when `array_from_numpy` is called and say: use `<array>.from _numpy_ndarray(ndarray)` instead?

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -50,13 +59,118 @@ class Array:
...
+    def __init__(self, ndarray: NDArray) -> None: ...  # noqa: E704
+
+    @overload
+    def __init__(  # noqa: E704
+        self, dtype: str, shape: list[int], stype: str, data: bytes
+    ) -> None: ...
+
+    def __init__(  # pylint: disable=too-many-arguments, too-many-locals
+        self,
+        *args: Any,
```
jafermarq, line 96: why is this needed?

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -50,13 +59,118 @@ class Array:
...
+            if all_args[index] is not None:
+                _raise_array_init_error()
+            all_args[index] = arg
+
+        # Try to set keyword arguments in all_args
+        _try_set_arg(0, ndarray)
+        _try_set_arg(0, dtype)
+        _try_set_arg(1, shape)
+        _try_set_arg(2, stype)
+        _try_set_arg(3, data)
```
jafermarq, line 130: (likely linked to my comment above). Do we still need this if we remove the `*args` input argument?

src/py/flwr/common/record/parametersrecord_test.py
```diff
@@ -72,6 +74,60 @@ def test_numpy_conversion_invalid(self) -> None:
...
+        ]
+    )
+    def test_valid_init_overloads_kwargs(self, name: str, *args: Any) -> None:
+        """Ensure valid overloads initialize correctly."""
+        if name == "explicit_values":
+            array = Array(dtype=args[0], shape=args[1], stype=args[2], data=args[3])
+        else:
+            kwargs = {name: args[0]}
+            array = Array(**kwargs)
+        self.assertIsInstance(array, Array)
```
jafermarq, line 106: Why not writing the test like this? Seems clearer to me
```
    @parameterized.expand(  # type: ignore
        [
            ({"ndarray": np.array([1, 2, 3])},),
            ({'dtype': "float32", 'shape':[2, 2], 'stype': "dense", 'data': b"data"},),
        ]
    )
    def test_valid_init_overloads_kwargs(self, kwargs: dict[str: Any]) -> None:
        """Ensure valid overloads initialize correctly."""
        array = Array(**kwargs)
        self.assertIsInstance(array, Array)
```

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -50,13 +59,118 @@ class Array:
...
+
+    @overload
+    def __init__(  # noqa: E704
+        self, dtype: str, shape: list[int], stype: str, data: bytes
+    ) -> None: ...
+
+    def __init__(  # pylint: disable=too-many-arguments, too-many-locals
+        self,
+        *args: Any,
+        ndarray: NDArray | None = None,
```
jafermarq, line 97: Missing in class _docstrings_

src/py/flwr/common/record/conversion_utils.py
panh99: Sure, we can do that. I feel like it's generally a easier (and thus more popular) approach to just do `Array.from_numpy_ndarray(...)` or `Array(...)`.

src/py/flwr/common/record/parametersrecord_test.py
panh99: Makes sense! Thanks for the suggestion.

src/py/flwr/common/record/parametersrecord.py
panh99: Not sure if I was clear, so let me summarize:

We have four overloaded `__init__` methods:

```python
    @overload
    def __init__(self, torch_tensor: torch.Tensor) -> None: ...  # noqa: E704

    @overload
    def __init__(self, tf_tensor: tf.Tensor) -> None: ...  # noqa: E704

    @overload
    def __init__(self, ndarray: NDArray) -> None: ...  # noqa: E704

    @overload
    def __init__(  # noqa: E704
        self, dtype: str, shape: list[int], stype: str, data: bytes
    ) -> None: ...
```

Our `__init__` implementation should support all of these signatures, whether through positional arguments, keyword arguments, or a combination of both.

src/py/flwr/common/record/parametersrecord.py
panh99: No, we still need them. I tried quite a few times and this is the best option I can think of to enable overloaded constructors. I added some context in reply to your comment above.

src/py/flwr/common/record/conversion_utils.py
jafermarq: yeah, i agree.

src/py/flwr/common/record/parametersrecord.py
jafermarq: all clear after our discussion!

jafermarq (dismissed):
big improvements!!

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -27,36 +29,159 @@
...
     """Array type.
 
     A dataclass containing serialized data from an array-like or tensor-like object
-    along with some metadata about it.
+    along with metadata about it. The class can be initialized in one of two ways:
+
+    1. By providing a NumPy ndarray (via the `ndarray` argument).
+    2. By specifying explicit values for `dtype`, `shape`, `stype`, and `data`.
+
+    In scenarios (1), the `dtype`, `shape`, `stype`, and `data` are automatically
```
danieljanes, line 49:
```suggestion
    In scenario (1), the `dtype`, `shape`, `stype`, and `data` are automatically
```

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -27,36 +29,159 @@
...
+    derived from the provided ndarray. In scenario (4), these fields must be
```
danieljanes, line 50:
```suggestion
    derived from the provided ndarray. In scenario (2), these fields must be
```

src/py/flwr/common/record/parametersrecord.py
panh99: Thanks!

src/py/flwr/common/record/parametersrecord.py
panh99: Thanks!

src/py/flwr/common/record/conversion_utils.py
```diff
@@ -15,26 +15,19 @@
...
-    # Source: https://numpy.org/doc/stable/reference/generated/numpy.save.html
-    np.save(buffer, ndarray, allow_pickle=False)
-    data = buffer.getvalue()
-    return Array(
-        dtype=str(ndarray.dtype),
-        shape=list(ndarray.shape),
-        stype=SType.NUMPY,
-        data=data,
-    )
+    log(WARN, WARN_DEPRECATED_MESSAGE)
```
danieljanes, line 32: Can we use the standard deprecation warning function here?

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -27,36 +29,170 @@
...
+        `data` from an array-like or tensor-like object. Only required if you are not
+        passing in a ndarray.
 
-    data: bytes
-        A buffer of bytes containing the data.
+    data : Optional[bytes] (default: None)
+        A buffer of bytes containing the data. Only required if you are not passing in
+        a ndarray.
+
+    ndarray : Optional[NDArray] (default: None)
```
danieljanes, line 71:
```suggestion
        a ndarray.
    ndarray : Optional[NDArray] (default: None)
```

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -27,36 +29,170 @@
...
+        ndarray: NDArray | None = None,
+    ) -> None:
+        # Determine the initialization method and validate input arguments.
+        # Support two initialization formats:
+        # 1. Array(dtype: str, shape: list[int], stype: str, data: bytes)
+        # 2. Array(ndarray: NDArray)
+
+        # Initialize all arguments
+        # If more than 4 positional arguments are provided, raise an error.
+        if len(args) > 4:
```
danieljanes, line 121: Instead of just checking the number here, should we check whether a valid set of parameters is provided? I guess the 4 original arguments and the new ndarray argument are mutual exclusive, or not?

src/py/flwr/common/record/conversion_utils.py
panh99: Sure thing

src/py/flwr/common/record/parametersrecord.py
panh99: May I know the reason for this change? I believe we usually have a blank line between the docstring for each parameter.

src/py/flwr/common/record/parametersrecord.py
panh99: Yes, exactly. They will be checked later. A valid set of parameters will be `0, 1, 2, 3` as users can provide them via keyword.

danieljanes (dismissed):

src/py/flwr/common/record/parametersrecord.py
danieljanes: @panh99 I resolved this comment, but feel free to make the formatting change in this PR if you want to

danieljanes (approved):

PR #4922, opened 2025-02-10
feat(framework) Allow one-liner conversion for `ParametersRecord`

panh99:
Based on #4920


src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,371 @@ def _check_value(value: Array) -> None:
...
+        _try_set_arg(numpy_ndarrays, "numpy_ndarrays")
+        _try_set_arg(state_dict, "state_dict")
+        _try_set_arg(tf_weights, "tf_weights")
+
+        # If no arguments are provided, return and keep self empty
+        if arg is None:
+            return
+
+        # Handle dictionary of Arrays
+        if not init_method or init_method == "array_dict":
```
jafermarq, line 432:
```suggestion

	# Handle creation of ParametersRecord without passing input
        if init_method is None:
            return

        # Handle dictionary of Arrays
        if init_method == "array_dict":
```

jafermarq: Why can't it be simplified like this? (also removing the `not init_method` from the rest of `if` statements)

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,371 @@ def _check_value(value: Array) -> None:
...
-    >>>             buffer=array.data,
-    >>>             dtype=array.dtype,
-    >>>            ).reshape(array.shape)
-
-    Note that different arrays (e.g. from PyTorch, Tensorflow) might require different
-    serialization mechanism. Howerver, they often support a conversion to NumPy,
-    therefore allowing to use the same or similar steps as in the example above.
+    >>> arr1 = np.random.randn(3, 3)
+    >>> arr2 = np.random.randn(2, 2)
+    >>> p_record = ParametersRecord([arr1, arr2])
```
jafermarq, line 341: Could we show first how to create the `ParametersRecord` out of a single numpy array and then show this example with two? It would be good to show that even if there's a single numpy array, it has to be passed inside a list.

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -33,6 +34,9 @@
     import torch
 
 
+GC_THRESHOLD = 200_000_000  # 200 MB
```
jafermarq, line 37: Could we add it to the constants ? How did you decide setting this value ?

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,371 @@ def _check_value(value: Array) -> None:
...
+            total_serialized_bytes += len(arr.data)
+            del arr
+
+            # If total serialized data exceeds the threshold, trigger GC
+            if total_serialized_bytes > GC_THRESHOLD:
+                total_serialized_bytes = 0
+                gc.collect()
+
+        # Force GC
+        gc.collect()
```
jafermarq, line 602: In other methods you have it behind `if not keep_input:`

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,371 @@ def _check_value(value: Array) -> None:
...
+        into an :class:`Array` and stored in this record with generated keys.
+
+    state_dict : Optional[OrderedDict[str, torch.Tensor]] (default: None)
+        A PyTorch state_dict (str keys to torch.Tensor values). Each
+        tensor will be converted into an :class:`Array` and stored in this record.
+
+    tf_weights : Optional[list[NDArray]] (default: None)
+        A list of TensorFlow weights (NumPy arrays). Each array will be
+        automatically converted into an :class:`Array` and stored in this record
+        with generated keys.
```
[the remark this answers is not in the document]
panh99: Yeah, good idea!

src/py/flwr/common/record/parametersrecord.py
panh99: Oh, this is b/c we won't be able to know which initialization method should be used before checking the type of the input if users don't use keyword argument. For example, `ParametersRecord(model.state_dict())`.

src/py/flwr/common/record/parametersrecord.py
panh99: Hmmm, we have 5 examples here, one example for each initialization method. Should we add two examples for numpy arrays?

src/py/flwr/common/record/parametersrecord.py
panh99: We can move it to `common.constant`, but I think it's only used here. 

Regarding the 200 MB, I think 100MB - 200MB is good for a memory-constraint environment, and 500MB~1GB could also be a good choice in most cases. Wdyt?

src/py/flwr/common/record/parametersrecord.py
panh99: Good catch! Thanks.

src/py/flwr/common/record/parametersrecord.py
jafermarq: Ok. Let's leave it as is

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,345 @@ def _check_value(value: Array) -> None:
...
+        if "torch" not in sys.modules:
+            raise RuntimeError(
+                f"PyTorch is required to use {cls.from_state_dict.__name__}"
+            )
+
+        record = ParametersRecord()
+
+        for k in list(state_dict.keys()):
+            v = state_dict[k] if keep_input else state_dict.pop(k)
+            record[k] = Array.from_numpy_ndarray(v.detach().cpu().numpy())
```
yan-gao-GY, line 548: Does `gc.collector()` not work for PyTorch? Can we do `torch.cuda.empty_cache()` if device is GPU?

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,345 @@ def _check_value(value: Array) -> None:
...
+
+            if not keep_input:
+                # Remove the reference
+                ndarrays[i] = None  # type: ignore
+                total_serialized_bytes += len(record[str(i)].data)
+
+                # If total serialized data exceeds the threshold, trigger GC
+                if total_serialized_bytes > GC_THRESHOLD:
+                    total_serialized_bytes = 0
+                    gc.collect()
```
yan-gao-GY, line 523: What are the main target variables here for `gc.collect()`?

yan-gao-GY (commented):
I really like this style `p_record = ParametersRecord(model.state_dict())`! This is much cleaner than before~. I just left a few comments

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,345 @@ def _check_value(value: Array) -> None:
...
+    This object is one of the record types supported by :class:`RecordSet` and can
+    therefore be stored in a :class:`Message` or a :class:`Context`.
+
+    This class can be instantiated in multiple ways:
+
+    1. By providing nothing (empty container).
+    2. By providing a dictionary of Arrays (via the `array_dict` argument).
+    3. By providing a list of NumPy ndarrays (via the `numpy_ndarrays` argument).
+    4. By providing a PyTorch state_dict (via the `state_dict` argument).
+    5. By providing a TensorFlow model weights (a list of NumPy arrays via the
```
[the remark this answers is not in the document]
panh99: Thanks @yan-gao-GY ! I will remove `a` here. But I think it's better not to say `a set of` as it may imply that TF users must pass a `set` not a `list`.

panh99, line 297:
```suggestion
    5. By providing TensorFlow model weights (a list of NumPy arrays via the
```

src/py/flwr/common/record/parametersrecord.py
panh99: I will resolve this.

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,345 @@ def _check_value(value: Array) -> None:
...
+        cls,
+        weights: list[NDArray],
+        *,
+        keep_input: bool = True,
+    ) -> ParametersRecord:
+        """Create `ParametersRecord from a list of TensorFlow weights (NumPy arrays).
+
+        Example
+        -------
+        >>> import tensorflow as tf
```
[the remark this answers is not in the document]
panh99: Good question. `from_tf_weights` is a bit different b/c it just calls `from_numpy_ndarrays` under the hood. So just need to be clear that users should use `model.get_weights` here. The main reason for this is that TF has two different representation of parameters:
1. `model.weights`, which is `list[keras.Variable]`
2. `model.get_weights()`, which is `list[np.ndarray]`

src/py/flwr/common/record/parametersrecord.py
panh99: Let's not interfere with `torch` right now.

src/py/flwr/common/record/parametersrecord.py
panh99: It's to GC the input numpy arrays

src/py/flwr/common/record/parametersrecord.py
danieljanes: Yes, let's move this to constant. It'd be good to move more and more constants there to make clear which things we hardcode.

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,288 @@ def _check_value(value: Array) -> None:
 class ParametersRecord(TypedDict[str, Array]):
     r"""Parameters record.
 
-    A dataclass storing named Arrays in order. This means that it holds entries as an
-    OrderedDict[str, Array]. ParametersRecord objects can be viewed as an equivalent to
-    PyTorch's state_dict, but holding serialised tensors instead. A
-    :code:`ParametersRecord`  is one of the types of records that a
-    `flwr.common.RecordSet <flwr.common.RecordSet.html#recordset>`_ supports and
-    can therefore be used to construct :code:`common.Message` objects.
+    A typed dictionary (string to :class:`Array`) that can store named parameters
```
danieljanes, line 283:
```suggestion
    A typed dictionary (``str`` to :class:`Array`) that can store named parameters
```

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,288 @@ def _check_value(value: Array) -> None:
...
+    as serialized tensors. Internally, this behaves similarly to an
+    ``OrderedDict[str, Array]``. A ``ParametersRecord`` can be viewed as an
+    equivalent to PyTorch's state_dict, but it holds arrays in serialized form.
```
danieljanes, line 286:
```suggestion
    equivalent to PyTorch's ``state_dict``, but it holds arrays in serialized form.
```

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,288 @@ def _check_value(value: Array) -> None:
...
+
+    This object is one of the record types supported by :class:`RecordSet` and can
+    therefore be stored in a :class:`Message` or a :class:`Context`.
```
danieljanes, line 289: Slightly misleading since `ParametersRecord` cannot be directly stored in `Message` or `Context`

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,288 @@ def _check_value(value: Array) -> None:
...
+
+    This class can be instantiated in multiple ways:
+
+    1. By providing nothing (empty container).
+    2. By providing a dictionary of Arrays (via the ``array_dict`` argument).
+    3. By providing a list of NumPy ndarrays (via the ``numpy_ndarrays`` argument).
+    4. By providing a PyTorch state_dict (via the ``state_dict`` argument).
```
danieljanes, line 296:
```suggestion
    2. By providing a dictionary of ``Arrays`` (via the ``array_dict`` argument).
    3. By providing a list of NumPy ``ndarray``s (via the ``numpy_ndarrays`` argument).
    4. By providing a PyTorch ``state_dict`` (via the ``state_dict`` argument).
```

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,288 @@ def _check_value(value: Array) -> None:
...
+    @overload
+    def __init__(  # noqa: E704
+        self, state_dict: OrderedDict[str, torch.Tensor], *, keep_input: bool = True
+    ) -> None: ...
+
+    def __init__(  # pylint: disable=too-many-arguments
         self,
+        *args: Any,
+        numpy_ndarrays: list[NDArray] | None = None,
+        state_dict: OrderedDict[str, torch.Tensor] | None = None,
```
danieljanes, line 371: These two have inconsistent naming, two consistent options would be:
- `numpy_ndarrays` & `torch_state_dict`
- `ndarrays` & `state_dict`

src/py/flwr/common/record/parametersrecord.py
panh99: Thanks, I've updated the docstring

src/py/flwr/common/record/parametersrecord.py
```diff
@@ -268,103 +280,288 @@ def _check_value(value: Array) -> None:
...
+                    gc.collect()
+
+        if not keep_input:
+            # Clear the entire list to remove all references and force GC
+            ndarrays.clear()
+            gc.collect()
+        return record
+
+    @classmethod
+    def from_state_dict(
```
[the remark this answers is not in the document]
panh99: Yes, that makes sense.

src/py/flwr/common/record/parametersrecord.py
panh99: Okay, I will more it there.

src/py/flwr/common/record/parametersrecord.py
panh99: True. I've updated it.

danieljanes (approved):

danieljanes (approved):

PR #4926, opened 2025-02-11
refactor(framework) Extend range of `Ray` supported versions for simulation

jafermarq:
The pinned version of `ray` isn't supported by Python 3.12. We therefore extend the range to the min version supported by py3.12


pyproject.toml
```diff
@@ -81,7 +81,7 @@ rich = "^13.5.0"
 pyyaml = "^6.0.2"
 requests = "^2.31.0"
 # Optional dependencies (Simulation Engine)
-ray = { version = "==2.10.0", optional = true, python = ">=3.9,<3.12" }
+ray = { version = ">=2.10.0, <2.32.0", optional = true, python = ">=3.9,<3.13" }
```
jafermarq, line 84:
```suggestion
ray = { version = "==2.31.0", optional = true, python = ">=3.9,<3.13" }
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #4929, opened 2025-02-11
feat(framework) Introduce `@app.lifecycle()`

panh99:
### Known issue
The code after `yield` in the function decorated by `@app.lifecycle()` will not be called when the ServerApp/ClientApp process is killed externally.


jafermarq (approved):

src/py/flwr/server/server_app.py
```diff
@@ -105,29 +112,33 @@ def __init__(
         self._client_manager = client_manager
         self._server_fn = server_fn
         self._main: Optional[ServerAppCallable] = None
+        self._enter: Optional[ServerAppCallable] = None
+        self._exit: Optional[ServerAppCallable] = None
```
danieljanes, line 116: Do we still need these?

danieljanes (approved):

PR #4935, opened 2025-02-14
fix(framework) Update dataset for FlowerTune coding template

yan-gao-GY:
## Issue
The original coding dataset is not available in HF.

## Proposal
We update it using a new dataset with exactly same content.


jafermarq (approved):

tanertopal (approved):
lgtm!

PR #4937, opened 2025-02-16
refactor(baselines) Upgrade FedProx Baselne to new flwr format

maddox-j:
### Description
This PR upgrades the existing `FedProx` baseline from the previous `flwr` format to the new version. 
Specifically: 

1. Removed hydra library dependency
2. Adapted existing configuration files into pyproject.toml format
3. Use of Flower ClientApp and ServerApp
4. Populate dataset.py file to use `flwr-datasets` with `DistributionPartitioner` instead of manual handling.
5. Saving of results to disk by overloading the `Server` class definition. 
6. Update README.md with steps to run baseline simulation through flwr run .

### Related issues/PRs
N/A

### Explanation

This makes the `FedProx` baseline consistent with the current `flwr` version.

### Checklist

- [x] Write tests
- [x] Make CI checks pass

### Any other comments?
N/A


baselines/fedprox/conf/fedavg_sf_0.0.toml
jafermarq, line 1: In these config files, is it necessary to specify all the settings in the experiment? or is it enough to only set those that are different from those in `pyproject.toml`. For example, `fraction_evaluate` is always the same.

baselines/fedprox/pyproject.toml
```diff
@@ -141,3 +114,59 @@ exclude = [
...
+fraction_fit = 0.01
+fraction_evaluate= 0.0
+min_evaluate_clients= 0
+min_available_clients= 1000
+min_fit_clients= 1000
+local_epochs = 10
+stragglers_fraction = 0.9
+learning_rate = 0.03
+mu = 1.0
+num_clients = 1000
```
jafermarq, line 141: Let's indicate with a comment that this `num_clients` is here just for the 	`ServerApp` (and that it doesn't have an impact on the size of the federation). My understanding is that you are using this here because the `ServerApp` can't access the number of nodes connected outside the strategy and it can't read the `node_config` either.  This won't be a problem with the upcoming `Message API` 👼

jafermarq (commented):
Hey @maddox-j , this looks pretty good!! I added a few minor comments below.

I was able to run all the experiments running your script but then it seems the notebook ins't creating the plots. I believe some parts need (maybe just the `process_data()` function) adjustments since the results structure is different since we aren't using Hydra's `--multirun` anymore.

baselines/fedprox/fedprox/model.py
```diff
@@ -195,3 +156,34 @@ def test(
...
+
+def get_weights(net):
+    """Extract model parameters as numpy arrays from state_dict."""
+    return [val.cpu().numpy() for _, val in net.state_dict().items()]
+
+
+def set_weights(net, parameters):
+    """Apply parameters to an existing model."""
+    params_dict = zip(net.state_dict().keys(), parameters)
+    state_dict = OrderedDict({k: torch.tensor(v) for k, v in params_dict})
```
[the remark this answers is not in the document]
maddox-j: 👍🏼

baselines/fedprox/fedprox/dataset.py
```diff
@@ -1,60 +1,86 @@
...
+    # The validation set is never used because we do centralized evaluation
+    # on the server on the held-out test dataset.
+    partition_train_test = partition_train_test.with_transform(apply_transforms)
+    return (
+        DataLoader(
+            partition_train_test["train"],
+            batch_size=dataset_config.batch_size,
+            shuffle=True,
+        ),
+        None,
```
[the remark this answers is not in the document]
maddox-j: SGTM :D

maddox-j:
@jafermarq Thanks for all the comments. I have resolved your comments in one commit so that we keep the history a bit neater. 

With reference to the notebook: I did make these function changes initially. I added a fixed path to the results in the notebook now in case this was the error. :)

baselines/fedprox/pyproject.toml
```diff
@@ -141,3 +114,59 @@ exclude = [
...
+num_unique_labels = 10
+preassigned_num_samples_per_label = 5
+seed=42
+mu = 1.0
+sigma = 2.0
+val_ratio = 0.1
+batch_size = 10
+
+[tool.flwr.app.config.fit]
+drop_client = true # with FedProx, clients shouldn't be dropped even if they are stragglers
```
jafermarq, line 155:
```suggestion
drop_client = false # with FedProx, clients shouldn't be dropped even if they are stragglers
```

jafermarq: Given that `name` is `FedProx` and `mu` is >0, we should then disable the dropping of clients?

baselines/fedprox/pyproject.toml
```diff
@@ -141,3 +114,59 @@ exclude = [
...
+packages = ["."]
+
+[tool.flwr.app]
+publisher = "itsmaddox-j"
+
+[tool.flwr.app.components]
+serverapp = "fedprox.server_app:app"
+clientapp = "fedprox.client_app:app"
+
+[tool.flwr.app.config]
```
jafermarq, line 128: We typically use `a-b-c` as opposed to `a_b_c` for all items in `pyproject.toml`. I was able to make the code work by just ensuring [this line](https://github.com/maddox-j/flower/blob/44b0e6f67e438f748fc6bc6a5dbd91d01d277d8f/baselines/fedprox/fedprox/utils.py#L27) also does `.replace("-", "_")` for the `name` key. All the other `.toml` files will need of a similar modification.

jafermarq (approved):
@maddox-j , great to see `FedProx` running via `flwr run` 🚀

PR #4947, opened 2025-02-23
feat(framework) Add utility function to parse IP address from `grpc.ServicerContext`

chongshenng:
This utility function is used in the event log plugin to parse the IP address from the gRPC client via `grpc.ServicerContext.peer()`.


src/py/flwr/common/address.py
```diff
@@ -101,3 +104,32 @@ def is_port_in_use(address: str) -> bool:
...
+    ipv4_match = re.match(r"^ipv4:(?P<ip>[^:]+):", peer)
+    if ipv4_match:
+        return ipv4_match.group("ip")
+
+    # Match IPv6: "ipv6:[IP]:port"
+    ipv6_match = re.match(r"^ipv6:\[(?P<ip>[^\]]+)\]:", peer)
+    if ipv6_match:
+        return ipv6_match.group("ip")
+
+    return ""
```
jafermarq, line 135: Would this ever happen ? Would other parts of the codebase that call this `get_ip_address_from_servicer_context()` know how to interpret this `""`? Or asking differently, should we throw an error if not IPv6 or IPv4?

chongshenng: The documentation isn't clear about the possible formats that are returned. It's a good idea to throw an error if it's not IPv6 or IPv4.

jafermarq (approved):

src/py/flwr/common/address.py
```diff
@@ -101,3 +104,35 @@ def is_port_in_use(address: str) -> bool:
...
+    ----------
+    context : grpc.ServicerContext
+        The gRPC ServicerContext object. The context.peer() returns a string like
+        "ipv4:127.0.0.1:56789" for IPv4 and "ipv6:[2001:db8::1]:54321" for IPv6.
+
+    Returns
+    -------
+    str
+        If one of the format matches, the function will return the client's IP address,
+        otherwise, it will return an empty string.
```
danieljanes, line 122: @chongshenng this comment does not match the implementation below

src/py/flwr/common/address_test.py
```diff
@@ -140,3 +142,77 @@ def test_domain_incorrect() -> None:
...
+
+
+def test_servicer_ipv4_correct() -> None:
+    """Test if a correct IPv4 address is correctly parsed from grpc.ServicerContext."""
+    # Prepare
+    addresses = [
+        ("ipv4:127.0.0.1:56789", "127.0.0.1"),
+        ("ipv4:0.0.0.0:8080", "0.0.0.0"),
+        ("ipv4:192.168.1.1:12345", "192.168.1.1"),
+    ]
```
danieljanes, line 165: Should we turn these into parameterized tests?

src/py/flwr/common/address.py
chongshenng: Resolved.

src/py/flwr/common/address_test.py
chongshenng: Sure, makes sense. I've also converted all of the existing tests to parameterized tests.

danieljanes (approved):

PR #4950, opened 2025-02-23
feat(framework) Add try-obtain `ExecServicer` event log plugin

chongshenng:
This PR introduces the event log plugin `try`-`except` import in the SuperLink and adds the event log interceptor in the gRPC server.

Merge after:
* #4951 

> \[!NOTE\]
> I simplified the `EventLogWriterType` for a single plugin

PR #4951, opened 2025-02-23
feat(framework) Add `ExecServicer` event log interceptor

chongshenng:
Merge after
* #4948
* #5013 

Merge before
* #4950


src/py/flwr/superexec/exec_user_auth_interceptor.py
```diff
@@ -43,6 +45,11 @@
 ]
 
 
+shared_user_info: contextvars.ContextVar[UserInfo] = contextvars.ContextVar(
+    "user_info", default={"user_id": None, "user_name": None}
+)
```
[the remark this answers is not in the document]
chongshenng: Hmm, I don't think so. This `ContextVars` is only used by the interceptors and works stably if the same coroutine is using this object. If we move this to the `LinkState`, it may cause some unpredictable behaviour with thread-safe access.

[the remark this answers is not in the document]
jafermarq: is having this global variable the best way to communicate the info with the `ExecEventLogInterceptor`?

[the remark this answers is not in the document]
chongshenng: @jafermarq Yes, this is the recommended approach for request-specific in-process propagation between interceptors. It is described in `grpc`'s code example [here](https://github.com/grpc/grpc/tree/master/examples/python/interceptors/async). In the future, we could potentially introduce a `user` database on the `LinkState` where the information can be looked up by the interceptors.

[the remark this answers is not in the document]
panh99: @jafermarq Based on the `grpc`'s code example, I think this should be the best option for us to communicate info across interceptors. It's not suitable to put these info in the `LinkState`, as they should not persist after a RPC.

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,119 @@
...
+        Continue RPC call if event logger is enabled on the SuperLink, else, terminate
+        RPC call by setting context to abort.
+        """
+        # One of the method handlers in
+        # `flwr.superexec.exec_servicer.ExecServicer`
+        method_handler: grpc.RpcMethodHandler = continuation(handler_call_details)
+        method_name: str = handler_call_details.method
+        return self._generic_auth_unary_method_handler(method_handler, method_name)
+
+    def _generic_auth_unary_method_handler(
```
panh99, line 53: The function name here is not updated, no?

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,119 @@
...
+                    request,
+                    context,
+                    response,
+                    shared_user_info.get(),
+                    method_name,
+                )
+                self.log_plugin.write_log(log_entry)
+                return cast(EventLogResponse, response)
+
+            # For unary-stream calls, wrap the response iterator write the event log
```
panh99, line 83:
```suggestion
            # For unary-stream calls, wrap the response iterator and write the event log
```

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,119 @@
...
+            # after iteration completes.
+            if method_handler.unary_stream:
+                response = cast(
+                    Iterator[EventLogResponse],
+                    method_handler.unary_stream(request, context),
+                )
+
+                def response_wrapper() -> Generator[Any, Any, None]:
```
panh99, line 91: Should the type hint here be like this?
```suggestion
                def response_wrapper() -> Iterator[EventLogResponse]:
```

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,119 @@
...
+
+                return response_wrapper()
+
+            # If the method type is not `unary_unary` or `unary_stream`, raise an error.
+            raise NotImplementedError("This RPC method type is not supported.")
+
+        if method_handler.unary_unary:
+            message_handler = grpc.unary_unary_rpc_method_handler
+        else:
+            message_handler = grpc.unary_stream_rpc_method_handler
```
panh99, line 114: Instead of raise an exception in the generated method handler, should we raise it here?
```suggestion
            # If the method type is not `unary_unary` or `unary_stream`, raise an error.
            raise NotImplementedError("This RPC method type is not supported.")

        if method_handler.unary_unary:
            message_handler = grpc.unary_unary_rpc_method_handler
        elif method_handler.unary_stream:
            message_handler = grpc.unary_stream_rpc_method_handler
        else:
            # If the method type is not `unary_unary` or `unary_stream`, raise an error.
            raise NotImplementedError("This RPC method type is not supported.")
```

panh99 (dismissed):
LGTM!

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,121 @@
...
+            # after iteration completes
+            if method_handler.unary_stream:
+                response = cast(
+                    Iterator[EventLogResponse],
+                    method_handler.unary_stream(request, context),
+                )
+
+                def response_wrapper() -> Iterator[EventLogResponse]:
+                    try:
+                        yield from response
```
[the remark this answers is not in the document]
chongshenng: Addressed all comments, adding the `try`-`except` and fixed the types.

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+
+                def response_wrapper() -> Iterator[EventLogResponse]:
+                    stream_response: Union[EventLogResponse, Exception]
+                    try:
+                        # pylint: disable=use-yield-from
+                        for stream_response in response_iterator:
+                            yield stream_response
+                    except Exception as e:  # pylint: disable=broad-except
+                        stream_response = e
```
panh99, line 105:
```suggestion
                    except Exception as e:  # pylint: disable=broad-except
                        stream_response = e
                        raise
```

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+            self.log_plugin.write_log(log_entry)
+
+            # For unary-unary calls, log after the call immediately
+            if method_handler.unary_unary:
+                unary_response: Union[EventLogResponse, Exception]
+                try:
+                    unary_response = cast(
+                        EventLogResponse, method_handler.unary_unary(request, context)
+                    )
```
panh99, line 76:
```suggestion
                unary_response = None
                error = None
                try:
                    unary_response = cast(
                        EventLogResponse, method_handler.unary_unary(request, context)
                    )
```

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,133 @@
...
+                finally:
+                    log_entry = self.log_plugin.compose_log_after_event(
+                        request=request,
+                        context=context,
+                        user_info=shared_user_info.get(),
+                        method_name=method_name,
+                        response=unary_response,
+                    )
+                    self.log_plugin.write_log(log_entry)
+                return unary_response  # cast(EventLogResponse, response)
```
panh99, line 88:
```suggestion
                unary_response = None
                error = None
                try:
                    unary_response = cast(
                        EventLogResponse, method_handler.unary_unary(request, context)
                    )
                except Exception as e:  # pylint: disable=broad-except
                    error = e
                    raise
                finally:
                    log_entry = self.log_plugin.compose_log_after_event(
                        request=request,
                        context=context,
                        user_info=shared_user_info.get(),
                        method_name=method_name,
                        response=unary_response or cast(Exception, error),
                    )
                    self.log_plugin.write_log(log_entry)
                return unary_response
```

src/py/flwr/superexec/exec_event_log_interceptor_test.py
```diff
@@ -0,0 +1,224 @@
...
+class DummyUnsupportedHandler:
+    """Dummy handler for unsupported RPC types."""
+
+    unary_unary = None
+    unary_stream = None
+    request_deserializer = None
+    response_serializer = None
+
+
+class TestExecEventLogInterceptor(unittest.TestCase):
```
panh99, line 113: This looks really good! Could you also add tests for failure cases? By that I mean an exception is raised in the handler.

src/py/flwr/superexec/exec_user_auth_interceptor_test.py
```diff
@@ -0,0 +1,224 @@
...
+        """."""
+        return DummyUnaryUnaryHandler()
+
+    @parameterized.expand(
+        [
+            (GetAuthTokensRequest()),
+            (GetLoginDetailsRequest(),),
+        ]
+    )  # type: ignore
+    def test_unary_unary_login_request_successful(self, request_class) -> None:
```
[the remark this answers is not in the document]
chongshenng: Hmm, it's a gRPC message class that inherits from `google.protobuf.message.Message`, which are defined in the `*.pyi` files. I think it's still correct to call it a class, but we can for sure add the type hint here.

src/py/flwr/superexec/exec_user_auth_interceptor_test.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class TestExecUserAuthInterceptor(unittest.TestCase):
+    """Test the ExecUserAuthInterceptor authentication logic."""
+
+    def setUp(self) -> None:
+        """Set up test fixtures."""
+        # Set a known default for shared_user_info and store the token.
+        self.default_user_info = UserInfo(user_id=None, user_name=None)
+        self.token = shared_user_info.set(self.default_user_info)
```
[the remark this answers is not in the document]
chongshenng: Good question. According to [the `grpc` docs](https://grpc.github.io/grpc/python/grpc.html#grpc.ServerInterceptor.intercept_service), one `ContextVars.Context` is used in the first interceptor (in this case, the auth interceptor) and is subsequently used in downstream interceptors. IIUC, this means that the `ContextVar` variable can be set outside the `grpc.ServicerContext` since the variable is `Context` specific.

More importantly, it looks like we need to do `set()`/`reset()` on the `ContextVar` so that we don't leak any states between tests.

Robert-Steiner (dismissed):
LGTM!

src/py/flwr/superexec/exec_event_log_interceptor.py
```diff
@@ -0,0 +1,135 @@
...
+                return response_wrapper()
+
+            raise RuntimeError()  # This line is unreachable
+
+        if method_handler.unary_unary:
+            message_handler = grpc.unary_unary_rpc_method_handler
+        elif method_handler.unary_stream:
+            message_handler = grpc.unary_stream_rpc_method_handler
+        else:
+            # If the method type is not `unary_unary` or `unary_stream`, raise an error
```
Robert-Steiner, line 129: Would it make sense to use the default gRPC implementation as a fallback if a handler is called that we don’t support yet, instead of failing?

chongshenng: @Robert-Steiner In this case, given that in our implementation of `ExecServicer` we only have one `unary-stream` method type while the other types are `unary-unary`, I think raising an error here is sufficient if not prudent. But yes - it is a valid point for the future if we extend the RPC methods for `ExecServicer` to other types.

panh99 (approved):
LGTM!

src/py/flwr/superexec/exec_user_auth_interceptor.py
```diff
@@ -77,13 +84,22 @@ def _generic_method_handler(
...
 
-            valid_tokens, _ = self.auth_plugin.validate_tokens_in_metadata(metadata)
+            # For other requests, check if the user is authenticated
+            valid_tokens, user_info = self.auth_plugin.validate_tokens_in_metadata(
+                metadata
+            )
             if valid_tokens:
+                # Store user info in contextvars for authenticated users
+                shared_user_info.set(cast(UserInfo, user_info))
                 return call(request, context)  # type: ignore
```
danieljanes, line 100: Is it necessary to reset/delete a previously-set `shared_user_info` if we don't update it here?

chongshenng: No, it's not necessary. Each RPC call gets its own new context, so if it's not updated, the value of `shared_user_info` will default to `UserInfo(user_id=None, user_name=None)`.

PR #4952, opened 2025-02-23
refactor(baselines) Upgrade StatAvg Baseline to new flwr format

maddox-j:
### Description

This PR upgrades the existing `StatAvg` baseline from the previous `flwr` format to the new version.
Specifically:

1. Removed hydra library dependency
2. Adapted existing configuration files into pyproject.toml format
3. Use of Flower ClientApp and ServerApp
4. Kept manual handling of dataset (discussion points at the end of the PR).
5. Saving of results to disk by overloading the Server class definition.
6. Update README.md with steps to run baseline simulation through `flwr run .`
7. Result visualisation with Jupyter Notebook.

### Related issues/PRs

N/A

### Explanation

This PR upgrades the existing `StatAvg` baseline to the new `flwr` format.

### Checklist

- [x] Write tests
- [x] Make CI checks pass

### Any other comments?
The majority of this PR is a very straight forward refactor of `StatAvg`. However, one element that has not been addressed is the handling of the dataset with `FederatedDataset` instead of the manual handling. In this instance, my TL;DR summary is that it could be more worthwhile to leave the manual handling to prevent the invention of the proverbial wheel. The reason for this view point is as follows:

As seen in the `dataset.py` file, this dataset is not natively available on `HuggingFace`. While you can still load local `csv` files with both `load_dataset` and thus the `FederatedDataset` class, the issue is that this local file does not come with i) a train test split and ii) an easy IID partitioning scheme a priori. Specifically, the test dataset is created with random _stratified_ (I will return to this emphasis later) sample of the dataset. In turn, the individual client datasets are created with the random _stratified_ samples too. While on the surface this seems like a simple IID strategy, the issue here is that the IID partitioner makes a call to`self.dataset.shard` which returns the requested shard from a dataset split into num_shards pieces. In the case where your dataset has no class imbalance, this is a perfectly legitimate solution. However, when the dataset is imbalanced, it can occur that each of the shards do not display the same level of imbalance as the overall dataset; which, in this instance, would not be a truly IID sample. Another related issue is that HuggingFace's `test_train_split` function specifically omits the possibility of generating stratified splits, introducing another complication of wrangling this directly. In my eyes, there are two ways forward:

1. Leave the dataset processing function as is for the purposes of this baseline. Of course, we could reinvent the wheel by downloading the dataset with `HuggingFace`, then transform this to a `Pandas` dataset as was done originally, preform all of the necessary preprocessing functions, and then cast the dataset back into a form that can be used by the `FederatedDataset` class. This seems to introduce much more unnecessary code than what it is attempting to solve. If we choose this solution, then the code is pretty much ready-to-go, pending the review.
2. The second solution is that the Flower team could preform the dataset split locally, and then upload this version to their HuggingFace Hub as an exemplar dataset for this baseline. This would give very easy integration with the `FederatedDatasets` object. However, this raises questions on the licensing issues that could arise. Another issue is that this does not give flexibility for different levels of the validation split ratio, seeds, etc. from the viewpoint of the end user, should they want to experiment with these parameters.

Personally, my vote is for option one as I think the point of the baseline is achieved with or without the `FederatedDataset` class. In either case, I rely on your best judgement as after all, you are the Flower experts! Happy to assist/discuss on anything further!


baselines/statavg/pyproject.toml
```diff
@@ -1,71 +1,44 @@
...
+dependencies = [
+    "flwr[simulation]>=1.15.2",
+    "flwr-datasets[vision]>=0.5.0",
+    "tensorflow>=2.12.1",
+    "scikit-learn>=1.5.0",
+    "joblib>=1.3.2",
+    "pandas>=2.2.0",
+    "imbalanced-learn>=0.12.0",
+    "matplotlib>=3.8.2",
+    "omgeaconf==2.3.0",
```
jafermarq, line 15:
```suggestion
    "omegaconf==2.3.0",
```

baselines/statavg/README.md
```diff
@@ -48,64 +47,60 @@ dataset: [TON_IoT]
...
 
-# Install
-poetry install
+# Install the baseline
+pip install -e .
 ```
 
 ## Dataset Preparation
+
 You can download the TON_IoT dataset by accessing the following [link](https://research.unsw.edu.au/projects/toniot-datasets). Please navigate to `TON_IoT datasets/Train_Test_datasets/Train_Test_Linux_dataset` and download the file `Train_test_linux_memory.csv`. Then, rename the downloaded file to `dataset.csv` and place it in the `dataset/` directory. If you want to run the experiments with your own data, ensure your dataset is also named `dataset.csv` and located in the same directory. The dataset is preprocessed using `dataset_preparation.py`, which you can modify if you wish to add custom preprocessing steps.
```
jafermarq, line 67:
```suggestion
You can download the TON_IoT dataset by accessing the following [link](https://research.unsw.edu.au/projects/toniot-datasets). Please navigate to `TON_IoT datasets/Train_Test_datasets/Train_Test_Linux_dataset` and download the file `Train_test_linux_memory.csv`. Then, rename the downloaded file to `dataset.csv` and place it in a new `dataset/` directory located in the directory where this `README.md` is. 

If you want to run the experiments with your own data, ensure your dataset is also named `dataset.csv` and located in the same directory. The dataset is preprocessed using `dataset_preparation.py`, which you can modify if you wish to add custom preprocessing steps.
```

jafermarq: just indicating that the `dataset/` directory needs to be placed in the current dir.

baselines/statavg/pyproject.toml
```diff
@@ -1,71 +1,44 @@
...
 
-[tool.poetry]
-name = "statavg" # <----- Ensure it matches the name of your baseline directory containing all the source code
+[project]
+dependencies = [
+    "flwr[simulation]>=1.15.2",
+    "flwr-datasets[vision]>=0.5.0",
+    "tensorflow>=2.12.1",
+    "scikit-learn>=1.5.0",
+    "joblib>=1.3.2",
```
jafermarq, line 11:
```suggestion
```

baselines/statavg/pyproject.toml
```diff
@@ -1,71 +1,44 @@
...
-requires = ["poetry-core>=1.4.0"]
-build-backend = "poetry.masonry.api"
+requires = ["hatchling"]
+build-backend = "hatchling.build"
 
-[tool.poetry]
-name = "statavg" # <----- Ensure it matches the name of your baseline directory containing all the source code
+[project]
+dependencies = [
+    "flwr[simulation]>=1.15.2",
```
jafermarq, line 7:
```suggestion
    "flwr[simulation]>=1.18.0",
```

baselines/statavg/statavg/client_app.py
```diff
@@ -1,230 +1,239 @@
...
+            # -------------------------------------------------
+            # since Flower does not persist state within different rounds
+            # the global statistics recieved by each client
+            # are saved in a dedicated directory.
+            # In subsequent rounds (>2), clients load the saved scalers and retrieve
+            # the global statistics.
+            # Note that scaling/normalization happens in __init()__
+            # by invoking the method normalize_data().
+            joblib.dump(scaler, f"{self.dir_path}/scaler.joblib")
+            # ---------------------------------------------------
```
jafermarq, line 156: The `ClientApp` can make use of its `<context>.state` to persist metrics and other data across rounds (custom objects can be also serialized and stored).  More info here: https://flower.ai/docs/framework/how-to-design-stateful-clients.html

We can remove all `joblib` elements.

baselines/statavg/pyproject.toml
jafermarq: I tested with the latest version. No changes are needed in the code.

baselines/statavg/pyproject.toml
jafermarq: Not needed (see my comment in `client_app.py`)

jafermarq (commented):
Hello @maddox-j , this looks good and i was able to reproduce the results. I left some comments below. I think the main one is about getting rid off `joblib` (that was used to persist some state for the `ClientApps`. We can make use of the context instead.

baselines/statavg/statavg/client_app.py
maddox-j: @jafermarq I agree generally with the move to each client using it's state. However, we have an issue on evaluation where the scalar of a random user is loaded by the server. To the best of my knowledge, a client Context is not accessible by custom strategies? Let me know what would be good workaround

jafermarq: A workaround would be: 

1. the clients communicate their scaler obj to the ServerApp when returning from `fit()` -- because the last arg returned by `fit()` is of type [Scalar](https://github.com/adap/flower/blob/44e5e046719ab0d72e99a96fec18686abad41391/framework/py/flwr/common/typing.py#L34) it's ok to return `bytes` (so the `Scaler` object serialized);
2. The `aggregate_fit` in the strategy deserializes and keep in memory those `Scaler` objects.
3. The `evalaute_fn()` can sample one of them each time the callback is executed.

What do you think?

maddox-j: @jafermarq Sounds good to me! Have added the necessary changes :D

jafermarq: amazing!

jafermarq (approved):
It's great to see `StatAvg` running with `flwr run`! Many thanks for the quality work @maddox-j !! 💯

PR #4954, opened 2025-02-24
feat(framework) Allow function-specific modifiers for `ClientApp`

panh99:
### Example Usage

```python
@app.train(mods=[local_mod])
def train(msg: Message, context: Context):
    ...
```

PR #4962, opened 2025-02-25
docs(framework) Update the `How to Use Built-in Mods` guide

panh99:
Merge after #4954

### DIFF
--- a/framework/docs/source/how-to-use-built-in-mods.rst
+++ b/framework/docs/source/how-to-use-built-in-mods.rst
@@ -5,8 +5,10 @@
 Use Built-in Mods
 =================
 
-**Note: This tutorial covers experimental features. The functionality and interfaces may
-change in future versions.**
+.. note::
+
+    This tutorial covers experimental features. The functionality and interfaces may
+    change in future versions.
 
 In this tutorial, we will learn how to utilize built-in mods to augment the behavior of
 a ``ClientApp``. Mods (sometimes also called Modifiers) allow us to perform operations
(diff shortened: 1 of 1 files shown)


framework/docs/source/how-to-use-built-in-mods.rst
```diff
@@ -21,80 +23,126 @@ is as follows:
...
-    Mod = Callable[[Message, Context, ClientApp], Message]
+    ClientAppCallable = Callable[[Message, Context], Message]
+    Mod = Callable[[Message, Context, ClientAppCallable], Message]
 
 A typical mod function might look something like this:
 
 .. code-block:: python
 
-    def example_mod(msg: Message, ctx: Context, nxt: ClientApp) -> Message:
+    def example_mod(msg: Message, ctx: Context, call_next: ClientAppCallable) -> Message:
```
jafermarq, line 33:
```suggestion
    from flwr.client.typing import ClientAppCallable
    from flwr.common import Context, Message
    
    def example_mod(msg: Message, ctx: Context, call_next: ClientAppCallable) -> Message:
```

framework/docs/source/how-to-use-built-in-mods.rst
```diff
@@ -21,80 +23,126 @@ is as follows:
...
+    import flwr as fl
+    from flwr.client.mod import example_mod_3, example_mod_4
+
+    app = fl.client.ClientApp()
+
+
+    @app.train(mods=[example_mod_3, example_mod_4])
+    def train(msg, ctx):
+        # Training logic here
+        return reply_msg
```
jafermarq, line 98:
```suggestion
    @app.train(mods=[example_mod_3, example_mod_4])
    def train(msg, ctx):
        # Training logic here
        return reply_msg
 
    @app.evaluate()
    def evalutate(msg, ctx):
        # Evaluate logic here
        return reply_msg
```

jafermarq: I think adding a second method it makes it clearer what the sentence below means

panh99: Good point! Thanks.

framework/docs/source/how-to-use-built-in-mods.rst
```diff
@@ -21,80 +23,140 @@ is as follows:
...
+Import the required mods
+++++++++++++++++++++++++
 
 .. code-block:: python
 
-    def client_fn(cid):
-        # Your client code goes here.
-        return  # your client
+    import flwr as fl
+    from flwr.client.mod import example_mod_1, example_mod_2
```
danieljanes, line 68: Does this actually work? Or should we use something here that is actually available?

framework/docs/source/how-to-use-built-in-mods.rst
```diff
@@ -21,80 +23,140 @@ is as follows:
...
+Assuming the following registration:
+
+.. code-block:: python
+
+    app = fl.client.ClientApp(mods=[example_mod_1, example_mod_2])
+
+
+    @app.train(mods=[example_mod_3, example_mod_4])
+    def train(msg, ctx):
+        return msg.create_reply(fl.common.RecordSet())
```
danieljanes, line 146: Should we also add an `@app.evalaute` example without function-specific mods here and show the execution order below?

panh99: Yeah, that makes sense to me.

danieljanes (approved):

PR #4963, opened 2025-02-26
feat(framework) Allow dict-like access for `RecordSet` while maintaining the original APIs

panh99:


src/py/flwr/common/record/recordset.py
```diff
@@ -164,39 +161,57 @@ def __init__(
...
-            configs_records=configs_records,
-        )
-        self.__dict__["_data"] = data
+        super().__init__(_check_key, _check_value)
+        for key, p_record in (parameters_records or {}).items():
+            self[key] = p_record
+        for key, m_record in (metrics_records or {}).items():
+            self[key] = m_record
+        for key, c_record in (configs_records or {}).items():
+            self[key] = c_record
```
danieljanes, line 170: Should we deprecate this constructor and encourage users to build the `RecordSet` by doing
```
rs = RecordSet()
rs["..."] = ...
```
?

danieljanes (approved):

PR #4967, opened 2025-02-27
Example for Secure Aggregation with low-level API

panh99:


examples/flower-secure-aggregation/secaggexample/server_app_low_level.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def on_receive(msgs: Iterable[Message], state: SecAggPlusAggregatorState) -> None:
+    """Intercept reply messages after receiving."""
+    log(INFO, "Intercepted reply messages after receiving.")
+
+
+def on_stage_complete(success: bool, state: SecAggPlusAggregatorState) -> None:
+    """Handle stage completion event."""
+    log(INFO, "Handled stage completion event.")
```
jafermarq, line 74: What do you see these functions to be useful for (in addition to logging purposes)?

panh99: I think it would be a bit useful for some advanced user to adjust the protocol, but generally it's more for me to implement things like weighted averaging based on the original `SecAggPlusInsAggregator`.

PR #4968, opened 2025-02-27
feat(framework) Add node availability check

yan-gao-GY:
## Issue
If a `SuperNode` goes offline, the `ServerApp` has to wait until the `TaskIns` expires, leading to unnecessary huge delays.

## Proposal
- Add node availability check
- Introduce a new `ERROR_CODE` for creating a Message in the SuperLink.
- Increase `DEFAULT_TTL` to 3h


src/py/flwr/common/message.py
```diff
@@ -25,7 +25,7 @@
 from .logger import log
 from .record import RecordSet
 
-DEFAULT_TTL = 3600
+DEFAULT_TTL = 3600 * 3
```
panh99, line 28: Btw, I think we should have a quick discussion on this default value.

```suggestion
DEFAULT_TTL = 10800  # This is 3 hours
```

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -508,7 +524,10 @@ def acknowledge_ping(self, node_id: int, ping_interval: float) -> bool:
         """Acknowledge a ping received from a node, serving as a heartbeat."""
         with self.lock:
             if node_id in self.node_ids:
-                self.node_ids[node_id] = (time.time() + ping_interval, ping_interval)
+                self.node_ids[node_id] = (
+                    time.time() + 2 * ping_interval,
```
panh99, line 528: Could you add a comment here to explain `2 *`? Maybe something like "Allow for one missed ping before marking the node as offline"

panh99: Should we also update the `create_node` method? It's still `time.time() + ping_interval` there.

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -245,6 +246,21 @@ def get_task_res(self, task_ids: set[UUID]) -> list[TaskRes]:
             )
             ret.update(tmp_ret_dict)
 
+            # Check node availability
```
panh99, line 249: I think we should conduct the check before fetching the TaskRes.

src/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -510,6 +511,29 @@ def get_task_res(self, task_ids: set[UUID]) -> list[TaskRes]:
         )
         ret.update(tmp_ret_dict)
 
+        # Check node availability
```
panh99, line 514: Same as above. Better do the check before fetching TaskRes.

src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -397,3 +397,72 @@ def verify_found_taskres(
...
+        A dictionary of error TaskRes indexed by the corresponding TaskIns ID.
+    """
+    ret_dict = {}
+    current = current_time if current_time else now().timestamp()
+    for taskins_id in list(inquired_taskins_ids):
+        task_ins = found_taskins_dict[taskins_id]
+        node_id = task_ins.task.consumer.node_id
+        online_until = node_id_to_online_until[node_id]
+        # Generate a TaskRes containing an error reply if the node is offline.
+        if online_until < current:
```
panh99, line 463: We need to avoid potential `KeyError`, and also include the check for whether the node exists or not.

```suggestion
        online_until = node_id_to_online_until.get(node_id)
        # Generate a TaskRes containing an error reply if the node is offline or doesn't exist.
        if online_until is None or online_until < current:
```

src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -397,3 +397,72 @@ def verify_found_taskres(
...
+def check_node_availability_for_taskins(
+    inquired_taskins_ids: set[UUID],
+    found_taskins_dict: dict[UUID, TaskIns],
+    node_id_to_online_until: dict[int, float],
+    current_time: Optional[float] = None,
+    update_set: bool = True,
+) -> dict[UUID, TaskRes]:
+    """Check node availability for TaskIns and generate error TaskRes if unavailable.
+
+    Parameters
```
panh99, line 436:
```suggestion
    """Check node availability for TaskIns and generate error TaskRes if unavailable.
    
    An error TaskRes indicating the node is unavailable will be generated for each TaskIns of which the destination node is offline or does not exist.

    Parameters
```

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
yan-gao-GY: Thanks for the suggestion! Comments added.

As for the `create_node` method, we decided to keep `time.time() + ping_interval` there.

src/py/flwr/server/superlink/linkstate/utils.py
yan-gao-GY: Comments added. Thanks!

src/py/flwr/common/message.py
jafermarq, line 1: Should we do this in a separate PR and get it merged for `1.16.0`? @yan-gao-GY @panh99 @chongshenng

src/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -731,7 +734,7 @@ def test_acknowledge_ping(self) -> None:
 
         # Execute
         current_time = time.time()
-        with patch("time.time", side_effect=lambda: current_time + 50):
+        with patch("time.time", side_effect=lambda: current_time + 70):
```
jafermarq, line 737: I think it would be better to update this test slightly so the `ping_intervals` are variables. In that way we can more easily understand what the `+ 70` means here. Maybe adding a comment on top of the `with patch` line is also good

src/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -748,6 +751,48 @@ def test_acknowledge_ping_failed(self) -> None:
         # Assert
         assert not is_successful
 
+    def test_node_unavailable_error(self) -> None:
```
jafermarq, line 754: Let's update to use the `Message`-based methods etc

src/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -748,6 +751,48 @@ def test_acknowledge_ping_failed(self) -> None:
...
+            producer_node_id=node_id_0,
+            ancestry=[str(task_id_0)],
+            run_id=run_id,
+        )
+        state.store_task_res(task_res_0)
+
+        # Execute
+        current_time = time.time()
+        task_res_list: list[TaskRes] = []
+        with patch("time.time", side_effect=lambda: current_time + 100):
```
jafermarq, line 787: Let's add a comment of what we expect to happen with this `+ 100`? Something along the lines that the first node (because it used `ping_interval=30` which is < 100/2 ) will be flagged as "unavailable"

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -229,6 +230,21 @@ def get_task_res(self, task_ids: set[UUID]) -> list[TaskRes]:
                 current_time=current,
             )
 
+            # Check node availability
```
jafermarq, line 233: I'd suggest let's move this functionality to `get_message_res`. The utility function `check_node_availability_for_taskins` will need to be slightly updated so it operates on `Messages` only

src/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -490,6 +491,29 @@ def get_task_res(self, task_ids: set[UUID]) -> list[TaskRes]:
             current_time=current,
         )
 
+        # Check node availability
```
jafermarq, line 494: Let's move to `get_message_res`

src/py/flwr/server/superlink/linkstate/linkstate_test.py
yan-gao-GY: Thanks! Comments added. Note that + 70 here is not ping_intervals, which is a future time (current_time + 70) used for this test.

src/py/flwr/server/superlink/linkstate/linkstate_test.py
yan-gao-GY: Thanks! Comment added.

jafermarq:
@yan-gao-GY i'm thinking it would be good to make the `2` a variable in the class. For example: `self.ping_patience` and use it throughout? wdyt? @panh99 @chongshenng

panh99:
> @yan-gao-GY i'm thinking it would be good to make the `2` a variable in the class. For example: `self.ping_patience` and use it throughout? wdyt? @panh99 @chongshenng

Since we don't allow custom ping patience for now, should we make it as a constant? Or just an attribute?

jafermarq:
> > @yan-gao-GY i'm thinking it would be good to make the `2` a variable in the class. For example: `self.ping_patience` and use it throughout? wdyt? @panh99 @chongshenng
> 
> Since we don't allow custom ping patience for now, should we make it as a constant? Or just an attribute?

yes, constant sounds good to me. In that way we can replace all the `2*` in the linkstate implementations and tests

yan-gao-GY:
> > > @yan-gao-GY i'm thinking it would be good to make the `2` a variable in the class. For example: `self.ping_patience` and use it throughout? wdyt? @panh99 @chongshenng
> > 
> > 
> > Since we don't allow custom ping patience for now, should we make it as a constant? Or just an attribute?
> 
> yes, constant sounds good to me. In that way we can replace all the `2*` in the linkstate implementations and tests

Make sense! Do we want put it in `src/py/flwr/common/constant.py` or just leave it in this file?

src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -369,3 +369,74 @@ def verify_found_message_replies(
...
+        ttl=ttl,
+    )
+
+    return Message(
+        metadata=metadata,
+        error=Error(
+            code=ErrorCode.NODE_UNAVAILABLE,
+            reason=NODE_UNAVAILABLE_ERROR_REASON,
+        ),
+    )
```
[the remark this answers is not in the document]
yan-gao-GY: Good point!

src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -237,7 +235,9 @@ def has_valid_sub_status(status: RunStatus) -> bool:
     return status.sub_status == ""
 
 
-def create_message_error_unavailable_res_message(ins_metadata: Metadata) -> Message:
+def create_message_error_unavailable_res_message(
```
panh99, line 238: Would it be better to create a separate function for creating node offline error, instead of extending this function that is used to create an error for expired reply message?

yan-gao-GY: We previously have separate functions for both, but as @jafermarq suggested, it combines into one as they share 80% same code

src/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -232,13 +234,28 @@ def get_message_res(self, message_ids: set[UUID]) -> list[Message]:
...
+                node_id_to_online_until={
+                    node_id: self.node_ids[node_id][0] for node_id in dst_node_ids
+                },
+                current_time=current,
+            )
+            ret.update(tmp_ret_dict)
+
             # Find all reply Messages
             message_res_found: list[Message] = []
             for message_id in message_ids:
```
danieljanes, line 261: @yan-gao-GY @panh99 do we need to handle a possible race condition here? What happens if we generate the error response, but a different thread inserts a reply message that the following lines of code would then include in the response?

src/py/flwr/server/superlink/linkstate/utils.py
panh99: Okay, that makes sense to me. I’ll resolve this

panh99 (approved):
LGTM

PR #4975, opened 2025-02-28
fix(framework) Fix `flwr log` test

chongshenng:
Since the `print_log` logic runs `print` once, the accompanying test needs to be updated. Only one call is expected.


panh99 (approved):
LGTM! Thanks!

PR #4980, opened 2025-03-01
refactor(framework) Update the constructor of `RecordSet`

panh99:


src/py/flwr/common/record/recordset.py
```diff
@@ -155,19 +155,11 @@ class RecordSet(TypedDict[str, RecordType]):
...
         super().__init__(_check_key, _check_value)
-        for key, p_record in (parameters_records or {}).items():
-            self[key] = p_record
-        for key, m_record in (metrics_records or {}).items():
-            self[key] = m_record
-        for key, c_record in (configs_records or {}).items():
-            self[key] = c_record
+        if records is not None:
+            for key, record in records.items():
+                self[key] = record
```
jafermarq, line 162: I believe the `__setitem__` needs to be tweaked since this:
```python
from flwr.common import RecordSet, MetricsRecord, ConfigsRec
c = RecordSet({'a':MetricsRecord(), 'a': ConfigsRecord()})
```
won't throw a warning and the resulting record won't have a `MetricsRecord`.

panh99: I don't think we can do anything about it, really, b/c:

image

panh99: What we get is already a dictionary that only has one item. Users will see a warning if they manually set `rs["a"] = MetricsRecord()` and then `rs["a"] = ConfigsRecord()`.

jafermarq: got it!

danieljanes (approved):

PR #4989, opened 2025-03-03
refactor(framework) Sort files in `flwrcli.build` before adding to zip to ensure consistent order

kta-intel:
### Description
This PR introduces a change to sort the list of files before they are added to the zip archive. By sorting the files, we ensure that the order of files within the zip archive is consistent. This consistency is essential for producing the same hash value for the same set of files, regardless of the file system's order.


danieljanes (approved):

danieljanes:
LGTM, thanks for the contribution @kta-intel!

PR #4992, opened 2025-03-04
refactor(framework) Introduce `delivered_at` in `Message` metadata

jafermarq:
To ease the migration of `LinkState` from `TaskIns/Res` to pure `Messages` we need to introduce this element in the metadata since it's the mechanism to determine whether a message has been pulled for processing or not.

### DIFF
--- a/src/py/flwr/client/message_handler/message_handler_test.py
+++ b/src/py/flwr/client/message_handler/message_handler_test.py
@@ -300,7 +300,7 @@ class TestMessageValidation(unittest.TestCase):
         invalid_metadata_list: list[Metadata] = []
         attrs = list(vars(self.valid_out_metadata).keys())
         for attr in attrs:
-            if attr == "_partition_id":
+            if attr == "_delivered_at":
                 continue
             if attr == "_ttl":  # Skip configurable ttl
                 continue
@@ -321,5 +321,6 @@ class TestMessageValidation(unittest.TestCase):
 
         # Assert
         for invalid_metadata in invalid_metadata_list:
+            print(invalid_metadata)
             msg.__dict__["_metadata"] = invalid_metadata
             self.assertFalse(validate_out_message(msg, self.in_metadata))
--- a/src/py/flwr/client/message_handler/message_handler_test.py
+++ b/src/py/flwr/client/message_handler/message_handler_test.py
@@ -321,6 +321,5 @@ class TestMessageValidation(unittest.TestCase):
 
         # Assert
         for invalid_metadata in invalid_metadata_list:
-            print(invalid_metadata)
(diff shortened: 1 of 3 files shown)


src/py/flwr/client/message_handler/message_handler_test.py
```diff
@@ -300,7 +300,7 @@ def test_invalid_message_run_id(self) -> None:
         invalid_metadata_list: list[Metadata] = []
         attrs = list(vars(self.valid_out_metadata).keys())
         for attr in attrs:
-            if attr == "_partition_id":
```
jafermarq, line 303: it wasn't used. so i repurposed it with the new `delivered_at` property

src/py/flwr/common/message.py
```diff
@@ -126,6 +126,16 @@ def created_at(self, value: float) -> None:
...
 
+    @property
+    def delivered_at(self) -> str:
+        """Unix timestamp when the message was delivered."""
+        return cast(str, self.__dict__["_delivered_at"])
+
+    @delivered_at.setter
+    def delivered_at(self, value: str) -> None:
+        """Set delivery timestamp of this message."""
+        self.__dict__["_delivered_at"] = value
```
chongshenng, line 137: I see that our `created_at` and `ttl` are of type `float`. For consistency, should we use `float` for `delivered_at`?

src/py/flwr/server/utils/validator.py
```diff
@@ -126,6 +126,8 @@ def validate_message(message: Message, is_reply_message: bool) -> list[str]:
             "`metadata.created_at` must be a float that records the unix timestamp "
             "in seconds when the message was created."
         )
+    if metadata.delivered_at != "":
+        validation_errors.append("`metadata.delivered_at` must be an empty str")
```
chongshenng, line 130: This would need to change slightly if we use a `float` for `delivered_at`.

src/py/flwr/common/message.py
jafermarq: I'd prefer to keep it as `str` so we can follow the logic that the `taskins/res` methods implement for this property already.

jafermarq: I see adding this `delivered_at` as a temporary measure.

chongshenng (commented):
Thanks @jafermarq. I've minor comments about the type for `delivered_at`.

src/py/flwr/common/message.py
chongshenng: ok

src/py/flwr/server/utils/validator.py
chongshenng: Resolving based on previous comment.

danieljanes (approved):

PR #4997, opened 2025-03-04
feat(framework) Add `FleetServicer` log interceptor

chongshenng:
> \[!NOTE\]
> I made a tiny change to the `EventLogWriterPlugin` abstract base class such that `user_info: UserInfo` is now `user_info: Optional[UserInfo]`. This is required because in the `FleetServicer`, there is no user info to be passed on, so this needs to be set to `None`.

Merge after:
* #4951 
* #4950


src/py/flwr/server/fleet_event_log_interceptor.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __init__(self, log_plugin: EventLogWriterPlugin) -> None:
+        self.log_plugin = log_plugin
+
+    def intercept_service(
+        self,
+        continuation: Callable[[Any], Any],
+        handler_call_details: grpc.HandlerCallDetails,
+    ) -> grpc.RpcMethodHandler:
+        """Flower server interceptor logging logic.
```
panh99, line 38:
```suggestion
        """Flower Fleet API server interceptor logging logic.
```

src/py/flwr/server/fleet_event_log_interceptor.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Intercept all unary-unary calls from users and log the event. Continue RPC call
+        if event logger is enabled on the SuperLink, else, terminate RPC call by setting
+        context to abort.
```
panh99, line 42: Do we use `context.abort` in the code here? I am a bit confused with this docstring

src/py/flwr/server/fleet_event_log_interceptor.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        # One of the method handlers in
+        # `flwr.superexec.exec_servicer.ExecServicer`
```
panh99, line 45: This should be updated to FleetServicer, right?

src/py/flwr/server/fleet_event_log_interceptor.py
```diff
@@ -0,0 +1,94 @@
...
+        method_name: str = handler_call_details.method
+        return self._generic_event_log_unary_method_handler(method_handler, method_name)
+
+    def _generic_event_log_unary_method_handler(
+        self, method_handler: grpc.RpcMethodHandler, method_name: str
+    ) -> grpc.RpcMethodHandler:
+        def _generic_method_handler(
+            request: EventLogRequest,
+            context: grpc.ServicerContext,
+        ) -> Union[EventLogResponse, Exception]:
```
panh99, line 56: This function won't return an exception.

```suggestion
        ) -> EventLogResponse:
```

src/py/flwr/common/event_log_plugin/event_log_plugin.py
```diff
@@ -33,6 +33,18 @@
...
+    DeleteNodeResponse,
+    PullMessagesRequest,
+    PullMessagesResponse,
+    PushMessagesRequest,
+    PushMessagesResponse,
+)
+from flwr.proto.run_pb2 import GetRunRequest, GetRunResponse  # pylint: disable=E0611
 
 # Type variables for request and response messages
 EventLogRequest = Union[
```
panh99, line 50: For the abstract base class, it would be nicer to keep its signature stable. So we can use `GrpcMessage` as type annotations here ( `from google.protobuf.message import Message as GrpcMessage`), instead of listing all possible requests and responses. And in its implementations, we can import specific requests/responses when needed.

src/py/flwr/server/fleet_event_log_interceptor.py
chongshenng: Good catch! It was copy-pasted.

src/py/flwr/server/fleet_event_log_interceptor_test.py
```diff
@@ -0,0 +1,179 @@
...
+from flwr.common.event_log_plugin.event_log_plugin import (
+    EventLogRequest,
+    EventLogResponse,
+)
+from flwr.common.typing import Actor, Event, LogEntry, UserInfo
+from flwr.superexec.exec_event_log_interceptor_test import (
+    DummyUnaryUnaryHandlerException,
+    DummyUnsupportedHandler,
+)
+from flwr.superexec.exec_user_auth_interceptor_test import get_dummy_unary_unary_handler
```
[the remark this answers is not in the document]
panh99: But we can do this in a follow-up PR.

panh99 (commented):
Looks good to me generally! Just a few comments about the type annotation and cross-dependencies issues.

src/py/flwr/server/fleet_event_log_interceptor.py
chongshenng: 💯

src/py/flwr/common/event_log_plugin/event_log_plugin.py
chongshenng: Thanks for the suggestion! I implemented the change in #5013 and this sequence of PRs will benefit from the change.

src/py/flwr/server/fleet_event_log_interceptor_test.py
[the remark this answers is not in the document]
chongshenng: Good call. I implemented this in #4951 (since it's not merged yet).

panh99 (dismissed):
LGTM!

danieljanes (dismissed):

danieljanes (approved):

danieljanes (approved):

PR #5000, opened 2025-03-05
refactor(framework) Change return type for `get_node_ids` to `Iterable[int]`

chongshenng:


src/py/flwr/server/compat/app_utils.py
```diff
@@ -77,12 +77,12 @@ def _update_client_manager(
     registered_nodes: dict[int, DriverClientProxy] = {}
     while not f_stop.is_set():
         try:
-            all_node_ids = set(driver.get_node_ids())
+            all_node_ids = driver.get_node_ids()
```
jafermarq, line 80: why not wrap with `list()` here? (like done in other places in this PR)

chongshenng: A `list` does not have the `difference()` method, so we'd have to do:
```python
all_node_ids = list(driver.get_node_ids())
...
new_nodes = set(all_node_ids).difference(registered_nodes)
```

Skipping the `list()` saves a tiny step.

jafermarq (approved):

danieljanes (approved):

danieljanes (approved):

PR #5001, opened 2025-03-05
refactor(framework) Raise warning if errors are encountered in the `Driver` API's `push_message`

chongshenng:


src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -212,6 +212,12 @@ def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
                 messages_list=message_proto_list, run_id=cast(Run, self._run).run_id
             )
         )
+        if len([id for id in list(res.message_ids) if id]) != len(list(messages)):
+            log(
+                WARNING,
+                "Not all messages could be pushed to the SuperLink. "
+                "This could be due to a malformed message.",
```
jafermarq, line 219: I think it's a good idea showing this warning. Worth adding an extra sentence indicating that the returned list has `None` for those messages that weren't pushed successfully ?. the order is kept as passed to `push_messages`.

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -212,6 +212,12 @@ def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
                 messages_list=message_proto_list, run_id=cast(Run, self._run).run_id
             )
         )
+        if len([id for id in list(res.message_ids) if id]) != len(list(messages)):
```
panh99, line 215:
```suggestion
        if len([msg_id for msg_id in res.message_ids if msg_id]) != len(list(message_proto_list)):
```

chongshenng: @panh99 Wouldn't it be better to check `!= len(messages)` which is the input to `push_messages`?

src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -212,6 +212,15 @@ def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
                 messages_list=message_proto_list, run_id=cast(Run, self._run).run_id
             )
         )
+        if len([msg_id for msg_id in list(res.message_ids) if msg_id]) != len(
```
panh99, line 215: May I know why use `list(res.message_ids)` here, instead of `res.message_ids`? The `res.message_ids` is already a list, no?

chongshenng: Good catch!

panh99 (approved):
LGTM!

danieljanes (approved):

PR #5011, opened 2025-03-06
refactor(framework) Increase default TTL

yan-gao-GY:
## Issue
The current default TTL is 1h, which is a bit too short for local training.

## Proposal
Increase default TTL to 12h.


chongshenng (approved):
LGTM since it is a quick win which solves the issue for long training!

danieljanes (approved):

PR #5037, opened 2025-03-10
fix(framework) Fix validity of error reply sent by `SuperLink`

jafermarq:


src/py/flwr/server/superlink/driver/serverappio_servicer.py
```diff
@@ -215,6 +215,15 @@ def PullMessages(
...
+            # by the SuperLink itself as a reply Message that expired
+            if (
+                msg.metadata.src_node_id == SUPERLINK_NODE_ID
+                and msg.metadata.dst_node_id == SUPERLINK_NODE_ID
+            ):
+                msg.metadata._run_id = request.run_id
+
             _raise_if(
                 validation_error=request.run_id != msg.metadata.run_id,
                 request_name="PullMessages",
```
panh99, line 229: Okay, I think this should be my final thought. It's better not to alter the automatically generated Messages, and just skip the extra check here.

```suggestion

            # Skip `run_id` check for SuperLink generated replies
            if msg.metadata.src_node_id != SUPERLINK_NODE_ID:
                _raise_if(
                    validation_error=request.run_id != msg.metadata.run_id,
                    request_name="PullMessages",
```

jafermarq: much cleaner

danieljanes: Instead of modifying the message later or skipping the check here, can we add then `run_id` directly when we create the message?

PR #5039, opened 2025-03-11
docs(framework) Fix docs config

charlesbvll:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### DIFF
--- a/framework/docs/source/conf.py
+++ b/framework/docs/source/conf.py
@@ -67,7 +67,10 @@ versions = sorted(
     [
         tag.name
         for tag in repo.tags
-        if int(tag.name[1]) > 0 and int(tag.name.split(".")[1]) >= 8
+        if not tag.name.startswith("intelligence/")
+        and tag.name[0] == "v"
+        and int(tag.name[1]) > 0
+        and int(tag.name.split(".")[1]) >= 8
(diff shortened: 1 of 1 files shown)


framework/docs/source/conf.py
```diff
@@ -67,7 +67,10 @@
     [
         tag.name
         for tag in repo.tags
-        if int(tag.name[1]) > 0 and int(tag.name.split(".")[1]) >= 8
+        if not tag.name.startswith("intelligence/")
+        and tag.name[0] == "v"
```
jafermarq, line 71: what is this new `"v"` check for?

danieljanes: I think its to ensure that the version tag starts with `v` (as in `flowerintelligence/v0.1.0`)

charlesbvll: It is just to make sure that we have a tag of the format "v%d.%d.%d", because we later take the second element of the string array
This is probably "overkill", I can remove it if you prefer

danieljanes: Let's keep it

jafermarq: let's keep it ~

PR #5089, opened 2025-03-13
refactor(framework) Convert `Driver.pull_messages` to no-arg, returning a list

chongshenng:
### Description

Presently, users need to pass the `msg_ids` returned from `driver.push_messages(msgs)` to `driver.pull_messages(msg_ids)` in order to retrieve the response messages. In some cases, users may not want to keep track of `msg_ids` but still want to retrieve any response messages based on the messages that have been pushed.

### Explanation

Convert the `message_ids` arg in the `Driver.pull_messages()` method to an `Optional[Iterable[str]]` type. If no `message_ids` are provided, all `message_ids` that are stored on the `Driver` object are retrieved.

This simplifies the usage to:
```python
@app.main()
def main(driver: Driver, context: Context) -> None:
    # Get available nodes
    driver.get_node_ids()
    ...
    
    # Push messages
    driver.push_messages(messages)
    ...
    
    # Get replies
    replies = driver.pull_messages()  # `message_ids` no longer needs to be passed
    for reply in replies:
        # Do something
```

> [!NOTE]  
> The user may still pass `message_ids` to `driver.pull_messages()`. In this case, only the IDs that are stored in the `Driver` object (during the call to `push_messages`) will be pulled.


src/py/flwr/server/driver/grpc_driver.py
```diff
@@ -221,23 +227,50 @@ def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
...
+                    WARNING,
+                    "Cannot pull messages for the following missing message IDs: %s",
+                    missing_ids,
+                )
+
         # Pull Messages
         res: PullResMessagesResponse = self._stub.PullMessages(
             PullResMessagesRequest(
-                message_ids=message_ids,
+                message_ids=sorted(msg_ids_to_pull),
```
panh99, line 265: May I know why we need to sort it?

chongshenng: Do we expect users to keep track of the message IDs? Or is this embedded in the `message.metadata.reply_to_message`?

panh99 (approved):
LGTM!

panh99 (approved):

chongshenng:
Converting to draft, pending clarification of API usage.

chongshenng:
Closing as this is stale.

PR #5093, opened 2025-03-13
feat(framework) Enable custom message handler

panh99:


src/py/flwr/common/message.py
```diff
@@ -417,3 +417,28 @@ def _create_reply_metadata(msg: Message, ttl: float) -> Metadata:
         ttl=ttl,
         message_type=msg.metadata.message_type,
     )
+
+
+def check_message_type(message_type: str) -> bool:
```
chongshenng, line 422:
```suggestion
def validate_message_type(message_type: str) -> bool:
```

Given that we use `validate_*` in our `flwr` CLI checks, perhaps `validate_*` is a better function name here.

src/py/flwr/common/message.py
```diff
@@ -417,3 +417,28 @@ def _create_reply_metadata(msg: Message, ttl: float) -> Metadata:
...
+    - "<category>.<action>"
+    """
+    valid_types = {MessageType.TRAIN, MessageType.EVALUATE, MessageType.QUERY}
+
+    # Check if conforming to the format "<category>"
+    if message_type in valid_types:
+        return True
+
+    # Check if conforming to the format "<category>.<action>"
+    if message_type.count(".") != 1:
```
chongshenng, line 437: Should we display a helpful error message here?

chongshenng: If we accept both `"<category>"` and `"<category>.<action>"`, I think this should be `if message_type.count(".") > 1:`.

src/py/flwr/client/client_app.py
```diff
@@ -142,25 +141,24 @@ def __call__(self, message: Message, context: Context) -> Message:
...
-            if message.metadata.message_type == MessageType.QUERY:
-                if self._query:
-                    return self._query(message, context)
-                raise ValueError("No `query` function registered")
-
-            # Message type did not match one of the known message types abvoe
-            raise ValueError(f"Unknown message_type: {message.metadata.message_type}")
+            # Get the category and the action
+            if not check_message_type(message.metadata.message_type):
+                raise ValueError(
```
chongshenng, line 146: Related to my comment below, I think it's useful to indicate what are the possible errors that make the message type invalid, i.e. category is invalid, having more than one `.` or not a valid Python identifier.

src/py/flwr/common/message.py
panh99: Good idea!

src/py/flwr/common/message.py
panh99: Oh, the format `"<category>"` has been checked above.

chongshenng: Not sure if I understand. Based on this `if`-statement:
* `"train"` results in `False`
* `"train.action1"` results in `True`
* `"train.action1.subaction1"` results in `False`

src/py/flwr/client/client_app.py
```diff
@@ -365,6 +345,42 @@ class LoadClientAppError(Exception):
...
+    if app._call:
+        raise _registration_error(category)
+
+    def decorator(fn: ClientAppCallable) -> ClientAppCallable:
+        warn_deprecated_feature(f"ClientApp-register-{category}-function")
+
+        # Check if the name is a valid Python identifier
+        if not action.isidentifier():
+            raise ValueError(
+                f"Cannot reigster {category} function with name '{action}'. "
```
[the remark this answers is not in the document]
panh99: Great catch!

src/py/flwr/client/client_app.py
chongshenng: Would it be reasonable to return the helpful message in the body of `ValueError`? Or is it sufficient to have it in the docstring?

jafermarq: Yes, +1 to add in the `ValueError` message

src/py/flwr/client/client_app.py
```diff
@@ -185,27 +186,21 @@ def train(
...
 
-        return train_decorator
+        >>> app = ClientApp()
+        >>>
+        >>> @app.train("custom_action")
+        >>> def custom_action(message: Message, context: Context) -> Message:
+        >>>    print("ClientApp training running with custom action")
+        >>>    return message.create_reply(content=message.content)
+        """
+        return _get_decorator(self, MessageType.TRAIN, action, mods)
```
jafermarq, line 200: Should we put this above the example of how to use a function-specific modifier? I think custom message handlers would be more common to use.

jafermarq: Could we also briefly indicate what's the relationship between `MessageType` and this custom handler? For example adding a comment like:
```python
# messages with type `message_type="train:custom_action"` will be routed to this function
```

jafermarq: wdty @panh99 @chongshenng ? I feel this gives a more complete intuition of both (1) how to create a custom handler and (2) how to use it (from the `ServerApp`)

panh99: That's a great idea! Thanks.

src/py/flwr/client/client_app.py
```diff
@@ -365,6 +348,42 @@ class LoadClientAppError(Exception):
...
+def _get_decorator(
+    app: ClientApp, category: str, action: str, mods: Optional[list[Mod]]
+) -> Callable[[ClientAppCallable], ClientAppCallable]:
+    """Get the decorator for the given category and action."""
+    # pylint: disable=protected-access
+    if app._call:
+        raise _registration_error(category)
+
+    def decorator(fn: ClientAppCallable) -> ClientAppCallable:
+        warn_deprecated_feature(f"ClientApp-register-{category}-function")
```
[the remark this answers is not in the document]
panh99: May bad. I though I used warn_experimental feature here...

src/py/flwr/client/client_app.py
```diff
@@ -142,25 +141,27 @@ def __call__(self, message: Message, context: Context) -> Message:
...
+            # Get the category and the action
+            # A valid message type is of the form "<category>" or "<category>.<action>",
+            # where <category> must be "train"/"evaluate"/"query", and <action> is a
+            # valid Python identifier
+            if not validate_message_type(message.metadata.message_type):
+                raise ValueError(
+                    f"Invalid message type: {message.metadata.message_type}"
+                )
+
+            category, name = message.metadata.message_type, "default"
```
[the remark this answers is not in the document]
panh99: Yeah, I think we could.

jafermarq (approved):

chongshenng (approved):
LGTM!

src/py/flwr/client/client_app.py
```diff
@@ -142,27 +143,36 @@ def __call__(self, message: Message, context: Context) -> Message:
...
+                return self._registered_funcs[full_name](message, context)
+
+            raise ValueError(f"No {category} function registered with name '{action}'")
 
     def train(
-        self, mods: Optional[list[Mod]] = None
+        self, action: str = DEFAULT_ACTION, *, mods: Optional[list[Mod]] = None
     ) -> Callable[[ClientAppCallable], ClientAppCallable]:
-        """Return a decorator that registers the train fn with the client app.
+        """Return a decorator that registers a train function with the client app.
```
danieljanes, line 168:
```suggestion
        """Return a decorator that registers a train function with the ``ClientApp``.
```

src/py/flwr/client/client_app.py
```diff
@@ -217,43 +232,47 @@ def evaluate(
...
-
-        return evaluate_decorator
+        return _get_decorator(self, MessageType.EVALUATE, action, mods)
 
     def query(
-        self, mods: Optional[list[Mod]] = None
+        self, action: str = DEFAULT_ACTION, *, mods: Optional[list[Mod]] = None
     ) -> Callable[[ClientAppCallable], ClientAppCallable]:
-        """Return a decorator that registers the query fn with the client app.
+        """Return a decorator that registers a query function with the client app.
```
danieljanes, line 268:
```suggestion
        """Return a decorator that registers a query function with the ``ClientApp``.
```

danieljanes (dismissed):

src/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -266,7 +272,7 @@ def create_message_error_unavailable_ins_message(reply_to_message: UUID) -> Mess
         dst_node_id=SUPERLINK_NODE_ID,
         reply_to_message=str(reply_to_message),
         group_id="",  # Unknown
-        message_type="",  # Unknown
+        message_type=MessageType.SYSTEM,
```
panh99, line 275: @danieljanes Wdyt about this? We can't set it to `""` as it won't pass our message type check.

danieljanes: lgtm

danieljanes (approved):

PR #5108, opened 2025-03-14
fix(framework) Catch `BaseException` in `*EventLogInterceptor`

chongshenng:
The current implementation of `*EventLogInterceptor` will return an event with `status = "failed"` if it catches an `Exception`. But, if a `BaseException` occurs, it will erroneously flag the event as `status = "completed"`. To avoid this, this PR changes the `try`-`except` implementation to catch a `BaseException` instead.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #5136, opened 2025-03-19
refactor(intelligence) Improve local error description

charlesbvll:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


intelligence/ts/src/flowerintelligence.ts
```diff
@@ -207,19 +207,37 @@ export class FlowerIntelligence {
...
-    if (localFailure && !FlowerIntelligence.#remoteHandoff && !FlowerIntelligence.#apiKey) {
-      return localFailure;
+    if (
+      localFailure &&
+      !localFailure.ok &&
+      (!FlowerIntelligence.#remoteHandoff || !FlowerIntelligence.#apiKey)
+    ) {
+      let description = localFailure.failure.description;
+      description += FlowerIntelligence.#remoteHandoff
+        ? '\nAnd no valid API key was provided for Remote Handoff.'
```
tanertopal, line 232:
```suggestion
        ? '\nAdditionally, a valid API key for Remote Handoff was not provided.'
```

intelligence/ts/src/flowerintelligence.ts
```diff
@@ -207,19 +207,37 @@ export class FlowerIntelligence {
...
+        : '\nAnd Remote Handoff was not enabled with a valid API key.';
```
tanertopal, line 233:
```suggestion
        : '\nAdditionally, Remote Handoff with a valid API key was not enabled.';
```

intelligence/ts/src/flowerintelligence.ts
```diff
@@ -207,19 +207,37 @@ export class FlowerIntelligence {
       // Currently we just select the first compatible localEngine without further check
       return { ok: true, value: compatibleEngines[0] };
     } else {
+      let description = `No available local engine for \`${modelId}\``;
+      description += isNode
+        ? ' (make sure you have enough available RAM).'
+        : ' (make sure you are using a WebGPU compatible browser and have enough available RAM).';
```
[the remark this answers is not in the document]
charlesbvll: @tanertopal Well the way we have abstracted things, the `FlowerIntelligence` object doesn't get the response from the `fetch-model-config` endpoint. This would be another reason why escalating the errors from all available engines would make this cleaner. Wdyt @danieljanes ?

intelligence/ts/src/flowerintelligence.ts
charlesbvll: @tanertopal Maybe we can remove the API key part in this case:
```suggestion
        : '\nAdditionally, Remote Handoff was not enabled.';
```

intelligence/ts/src/flowerintelligence.ts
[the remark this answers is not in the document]
danieljanes: @charlesbvll @tanertopal I'd prefer to have error handling on two levels:
- `*Engine`: Returns an error codes like `INSUFFICIENT_RAM` or `WEBGPU_NOT_SUPPORTED`
- `FlowerIntelligence`: Looks at all the errors it gets from all the engines and decides which final error to return to the user

intelligence/ts/src/flowerintelligence.ts
tanertopal: @charlesbvll I just had a sync with Daniel. Let's stick to a single description per error code and link to the documentation for more details. Basically the same solution we have for the Flower Framework. Can make the adjustments?

intelligence/ts/src/flowerintelligence.ts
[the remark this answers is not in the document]
charlesbvll: @danieljanes Got it, I'll update it. If multiple errors are returned, how should the decision be taken by Flower Intelligence?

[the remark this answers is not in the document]
charlesbvll: I think in the code we could make the choice by comparing the ErrorCodes' associated int values, but we still have to decide how we assign those ^^

intelligence/ts/src/flowerintelligence.ts
```diff
@@ -195,31 +203,58 @@ export class FlowerIntelligence {
...
+      description += FlowerIntelligence.#remoteHandoff
+        ? '\nAdditionally, a valid API key for Remote Handoff was not provided.'
+        : '\nAdditionally, Remote Handoff was not enabled.';
+      return {
+        ok: false,
+        failure: {
+          code: FailureCode.NoLocalProviderError,
+          description,
+        },
+      };
```
tanertopal, line 257: Can we split this into two distinct errors so we have only one text for each error code?

tanertopal (approved):
Lgtm!

PR #5146, opened 2025-03-20
refactor(framework) Rename `ParametersRecord` to `ArrayRecord`

panh99:
~~Merge after #5155~~


src/py/flwr/common/record/recorddict.py
```diff
@@ -131,28 +129,26 @@ class RecordDict(TypedDict[str, RecordType]):
...
     the array needs to be serialized and represented as a :code:`flwr.common.Array`.
-    If the array is a :code:`NumPy` array, you can use the built-in utility function
-    `array_from_numpy <flwr.common.array_from_numpy.html>`_. It is often possible to
-    convert an array first to :code:`NumPy` and then use the aforementioned function.
+    For example:
 
-    >>>  from flwr.common import array_from_numpy
-    >>>  # Creating a ParametersRecord would look like this
+    >>>  from flwr.common import Array
+    >>>  # Creating a ArrayRecord would look like this
```
panh99, line 137:
```suggestion
    >>>  # Creating an ArrayRecord would look like this
```

src/py/flwr/common/record/recorddict_test.py
```diff
@@ -124,28 +122,28 @@ def test_parameters_to_parametersrecord_and_back(
...
     array_dict = OrderedDict(
         {str(i): ndarray_to_array(ndarray) for i, ndarray in enumerate(get_ndarrays())}
     )
-    p_record = ParametersRecord(array_dict, keep_input=True)
+    arr_record = ArrayRecord(array_dict, keep_input=True)
 
-    # Creating a second parametersrecord passing the same `array_dict` (not erased)
-    p_record_2 = ParametersRecord(array_dict, keep_input=False)
-    assert p_record == p_record_2
+    # Creating a second arrayrecord passing the same `array_dict` (not erased)
```
panh99, line 132:
```suggestion
    # Creating a second ArrayRecord passing the same `array_dict` (not erased)
```

src/py/flwr/common/record/arrayrecord.py
```diff
@@ -274,13 +276,14 @@ def _check_value(value: Array) -> None:
...
 
-    A typed dictionary (``str`` to :class:`Array`) that can store named parameters
-    as serialized tensors. Internally, this behaves similarly to an
-    ``OrderedDict[str, Array]``. A ``ParametersRecord`` can be viewed as an
-    equivalent to PyTorch's ``state_dict``, but it holds arrays in serialized form.
+    A typed dictionary (``str`` to :class:`Array`) that can store named arrays,
+    including model parameters, gradients, or non-parameter arrays. Internally,
+    this behaves similarly to an ``OrderedDict[str, Array]``. An ``ArrayRecord``
+    can be viewed as an equivalent to PyTorch's ``state_dict``, but it holds arrays
+    in serialized form.
```
jafermarq, line 286:
```suggestion
    in a serialized form.
```

src/py/flwr/common/record/arrayrecord.py
```diff
@@ -581,3 +584,43 @@ def count_bytes(self) -> int:
...
 
         return num_bytes
+
+
+class ParametersRecord(ArrayRecord):
+    """Deprecated class ``ParametersRecord``, use ``ArrayRecord`` instead.
+
+    This class exists solely for backward compatibility with legacy
+    code that previously used ``ParametersRecord``. It has been renamed
+    to ``ArrayRecord`` and will be removed in a future release.
```
jafermarq, line 594: I think the last sentence would be clearer if it's rewritten slightly. Maybe as: "This class will be removed in a future Flower release. Please use ``ArrayRecord`` instead."

src/py/flwr/common/record/arrayrecord.py
```diff
@@ -581,3 +584,43 @@ def count_bytes(self) -> int:
...
+
+        from flwr.common import ParametersRecord
+
+        record = ParametersRecord()
+
+    Updated usage::
+
+        from flwr.common import ArrayRecord
+
+        my_content = ArrayRecord()
```
jafermarq, line 612:
```suggestion
        record = ArrayRecord()
```

jafermarq: to align with the deprecated example above?

panh99: Yeah, exactly! Good catch. I forget to update it here.

src/py/flwr/common/record/arrayrecord.py
panh99: That makes sense. I've removed the later part of this sentence to make it clear. I think we don't need to mention that is gonna be removed here b/c we already say exactly the same thing before and after this line.

src/py/flwr/client/mod/comms_mods.py
```diff
@@ -53,23 +47,23 @@ def parameters_size_mod(
 ) -> Message:
     """Parameters size mod.
```
yan-gao-GY, line 48: Should we rename this mod as `array_size_mod()` as the content is not necessarily to be model parameters?

yan-gao-GY (commented):
@panh99 Thanks for making this PR! Great to see it simplifies the examples and custom mod! I just left a few comments.

src/py/flwr/client/mod/comms_mods.py
panh99: That's a good point. @jafermarq Wdyt?

jafermarq: works for me. Then, should the implementation of this mode make use of `array` more often and replace `parameters`? For example when saying; "number of parameters transmitted", or even in the returned statistics, there is a "parameters" key used

panh99: Yeah, that makes sense to me.

panh99: Thanks guys for the suggestion. I've updated the mod.

src/py/flwr/client/mod/comms_mods.py
```diff
@@ -34,47 +34,41 @@ def message_size_mod(
...
+        element_count = 0
+        for array in arr_record.values():
+            element_count += (
                 int(np.prod(array.shape)) if array.shape else array.numpy().size
             )
 
         model_size_stats[f"{record_name}"] = {
-            "parameters": parameter_count,
-            "bytes": p_record_bytes,
+            "elements": element_count,
```
danieljanes, line 65: Why not `arrays`?

danieljanes: Or: shouldn't it stay `parameters`? We're counting the number of parameters, right?

panh99: @danieljanes  This change follows the discussion here: https://github.com/adap/flower/pull/5146#discussion_r2007358614

The reason for not using `arrays` is that we're not counting the number of arrays. Instead, the number shown reflects the total number of elements/entries within the arrays.

As discussed, we think it's better to rename it to a more general term that reflects the size of stored arrays in the message, since users may use this module to measure embeddings, gradients, or query arrays in `ArrayRecord`. Wdyt? Should we still keep using `parameters`?

danieljanes (approved):

PR #5214, opened 2025-04-09
fix(*:skip) Make `ruff` ignore `.gitignore`

chongshenng:
By default, `ruff` will respect any of the regular ignore files, like `.gitignore`. We want `ruff` to run on all files whether or not `.gitignore` is present.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #5226, opened 2025-04-11
refactor(examples) Avoid hard coding `min_fit_clients` for the `fl-dp-sa` example

khoaguin:
Avoid hard coding the number of `min_fit_clients` for the `fl-dp-sa` example


jafermarq (approved):
Thank you @khoaguin for the help to improve the example! 🙏

PR #5250, opened 2025-04-16
feat(framework) Adding the option to pass args into `run_supernode` and `run_superlink`

lotif:
### Description

It's not possible to pass arguments into `run_supernode` and `run_superlink` when calling those functions programmatically in a script as opposed to calling them from the CLI.

### Related issues/PRs

NA

### Explanation

Add an optional `args` parameter that will override the local arguments variable if it's passed in. This way, I can call the function in a script with any parameters needed while it still works for the regular CLI use case.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass

### Any other comments?

### DIFF
--- a/framework/py/flwr/client/supernode/app.py
+++ b/framework/py/flwr/client/supernode/app.py
@@ -47,9 +47,17 @@ from ..app import start_client_internal
 from ..clientapp.utils import get_load_client_app_fn
 
 
-def run_supernode() -> None:
-    """Run Flower SuperNode."""
-    args = _parse_args_run_supernode().parse_args()
+def run_supernode(args: Optional[argparse.Namespace] = None) -> None:
+    """Run Flower SuperNode.
+
+    Parameters
+    ----------
+    args : Optional[argparse.Namespace]
+        The arguments to pass to the SuperNode. If not provided, the arguments will be
+        parsed from the command line.
+    """
+    if args is None:
+        args = _parse_args_run_supernode().parse_args()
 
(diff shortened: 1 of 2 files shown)


lotif:
Hi @panh99! Any comments about this?

chongshenng:
Hello @lotif! Thanks for opening this PR. For our understanding, may I ask what is the architecture/usecase that you've in mind that requires running the SuperLink and SuperNode programmatically?

For context, our intention is that the SuperLink and SuperNodes should be started as long-running services which do not have to be programmatically called in a script. That is, we treat the SuperLink and SuperNodes as the infrastructure layer of a Flower deployment. This layer should be orchestrated e.g. by Kubernetes (see for instance [this Flower x GCP guide](https://flower.ai/docs/framework/how-to-run-flower-on-gcp.html#deploy-flower-infrastructure)). For further reference, a high-level write-up of the runtime comparison and usage is [here](https://flower.ai/docs/framework/ref-flower-runtime-comparison.html#flower-runtime-comparison).

lotif:
Thanks for the response, @chongshenng @danieljanes . We have 2 use cases for that: we have pyhton scripts that kick off FL trainings and we have integration tests that run full FL trainings on test data to make sure everything is working as expected. In both cases, it's easier to manage the status of the clients and servers when they are called from within the script as opposed to kicking off separate processes for them.

My idea is that it would be easy enough to add the ability to pass in the arguments to `run_supernode` and `run_superlink` so we can call them as functions as opposed to only have those parsing the arguments from the command line.

PR #5253, opened 2025-04-17
feat(framework) Remove `root-certificates` check in `try-obtain` CLI plugin

chongshenng:
This check is unnecessary because the certificates and `insecure` checks are invoked in `init_channel()` right after `try_obtain_cli_auth_plugin()` for the CLI commands.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #5254, opened 2025-04-17
fix(framework) Raise error when doing `flwr login` with `insecure = true`

chongshenng:


framework/py/flwr/cli/login/login.py
```diff
@@ -80,6 +81,16 @@ def login(  # pylint: disable=R0914
             bold=True,
         )
         raise typer.Exit(code=1)
+    # Check if insecure flag is set
```
chongshenng, line 84:
```suggestion
    # Check if insecure flag is set to `True`
```

framework/py/flwr/cli/login/login.py
```diff
@@ -80,6 +81,16 @@ def login(  # pylint: disable=R0914
             bold=True,
         )
         raise typer.Exit(code=1)
+    # Check if insecure flag is set to `True`
+    insecure = get_insecure_flag(federation_config)
+    if insecure:
+        typer.secho(
+            "❌ `flwr login` requires TLS to be enabled. `insecure` must be set to "
+            "`false` in the federation configuration.",
```
panh99, line 89: Can we say something like "`insecure` must NOT be set to `true`"? Because users can also remove the line and `insecure` will default to `false`.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #5304, opened 2025-04-28
ci(*:skip) Add heartbeat CI

panh99:


.github/workflows/framework-e2e.yml
```diff
@@ -470,3 +470,41 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    timeout-minutes: 10
+    needs: [changes, wheel]
+  
+    name: Test heartbeat for runs
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        if: ${{ needs.changes.outputs.framework == 'true' }}
```
chongshenng, line 484: Since we use this condition in all of the steps, I think we can set this condition for the entire `run-heartbeat-test` to make it cleaner. See [this ref](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution).

```diff
  run-heartbeat-test:
    runs-on: ubuntu-22.04
    timeout-minutes: 10
    needs: [changes, wheel]

+   if: ${{ needs.changes.outputs.framework == 'true' }}

    name: Test heartbeat for runs
```

.github/workflows/framework-e2e.yml
```diff
@@ -470,3 +470,41 @@ jobs:
...
+      - name: Run test with deployment engine
+        if: ${{ needs.changes.outputs.framework == 'true' }}
+        working-directory: ./framework/e2e/e2e-run-heartbeat
+        run: python ../test_run_heartbeat.py deployment
+        shell: bash
+      - name: Run test with simulation engine
+        if: ${{ needs.changes.outputs.framework == 'true' }}
+        working-directory: ./framework/e2e/e2e-run-heartbeat
+        run: python ../test_run_heartbeat.py simulation
+        shell: bash
```
chongshenng, line 510: I'd recommend using a matrix for this since there are some modifications to the `pyproject.toml` - if we chain the steps, there can be some unexpected behavior in the downstream (simulation) test. 

```yaml
...
    strategy:
      matrix:
        engine: [deployment, simulation]
...
      - name: Run heartbeat test with ${{ matrix.engine }} runtime
        working-directory: ./framework/e2e/e2e-run-heartbeat
        run: python ../test_run_heartbeat.py ${{ matrix.engine }}
        shell: bash
```

.github/workflows/framework-e2e.yml
```diff
@@ -470,3 +470,41 @@ jobs:
...
+  
+    name: Test heartbeat for runs
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        if: ${{ needs.changes.outputs.framework == 'true' }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9
```
chongshenng, line 487: I'd also recommend using a matrix for this since the test is relatively quick and we should cover all Python versions for completeness.

```yaml
...
      matrix:
        python-version: ["3.9", "3.10", "3.11", "3.12"]
...
        with:
          python-version: ${{ matrix.python-version }}
```

.github/workflows/framework-e2e.yml
panh99: Great idea, thanks!

.github/workflows/framework-e2e.yml
panh99: Sure.

chongshenng (approved):

.github/workflows/framework-e2e.yml
```diff
@@ -470,3 +470,39 @@ jobs:
...
+    needs: [changes, wheel]
+    strategy:
+      matrix:
+        engine: [deployment, simulation]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+
+    if: ${{ needs.changes.outputs.framework == 'true' }}
+
+    name: |
+      Run heartbeat test with ${{ matrix.engine }} runtime /
```
danieljanes, line 486:
```suggestion
      Run app heartbeat test with ${{ matrix.engine }} runtime /
```

framework/e2e/e2e-run-heartbeat/README.md
```diff
@@ -0,0 +1,3 @@
+# Test Heartbeat for Runs
+
+This app is designed solely to test heartbeats coming from `ServerApp`.
```
danieljanes, line 3:
```suggestion
This app is designed solely to test app heartbeats coming from `flwr-serverapp`.
```

danieljanes (commented):
PR lgtm @panh99. Should we rename the directory to `e2e_app_heartbeat`?

framework/e2e/e2e-run-heartbeat/README.md
danieljanes: Right! I've updated the suggestion above.

danieljanes (approved):

PR #5308, opened 2025-04-29
feat(framework) Introduce ObjectStore

danieljanes:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/common/object_store/object_store_test.py
```diff
@@ -0,0 +1,133 @@
...
+        value1 = b"test_value1"
+        value2 = b"test_value2"
+
+        object_store.put(key, value1)
+        object_store.put(key, value2)
+        retrieved_value = object_store.get(key)
+
+        self.assertEqual(value2, retrieved_value)
+
+    def test_put_empty_key(self) -> None:
```
jafermarq, line 68: I wonder if we should prevent this from succeeding and only allow if a non-empty `str` key is passed.

framework/py/flwr/common/object_store/in_memory_object_store.py
```diff
@@ -0,0 +1,63 @@
...
+        """Put an object into the store."""
+        # Verify object ID format (must be a valid sha256 hash)
+        if not is_valid_sha256_hash(object_id):
+            raise ValueError(f"Invalid object ID format: {object_id}")
+
+        # Verify object_id and object_content match
+        if self.verify:
+            object_id_from_content = get_object_id(object_content)
+            if object_id != object_id_from_content:
+                raise ValueError(f"Object ID {object_id} does not match content hash")
```
jafermarq, line 40: I wonder if this is something to do in the `*Servicer` that interfaces with the `ObjectStore` instead of doing it in here. In that way we can reuse these checks regardless of the `ObjectStore` implementation. 

I'm starting work on https://github.com/adap/flower/pull/5369 so I thought adding this to the `PushObject` method in the servicer is maybe a preferable alternative?

framework/py/flwr/common/object_store/object_store.py
```diff
@@ -0,0 +1,85 @@
...
+        """Put an object into the store.
+
+        Parameters
+        ----------
+        object_id : str
+            The object_id under which to store the object.
+        object_content : bytes
+            The serialized object to store.
+        """
+        raise NotImplementedError
```
panh99, line 38: Since we mark `ObjectStore` as an abstract base class, it's better to mark its methods as abstract methods instead of rasing a NotImplementedError manually. Same comment for other methods.

```suggestion
    @abstractmethod
    def put(self, object_id: str, object_content: bytes) -> None:
        """Put an object into the store.

        Parameters
        ----------
        object_id : str
            The object_id under which to store the object.
        object_content : bytes
            The serialized object to store.
        """
```

framework/py/flwr/common/object_store/object_store.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def put(self, object_id: str, object_content: bytes) -> None:
+        """Put an object into the store.
+
+        Parameters
+        ----------
+        object_id : str
+            The object_id under which to store the object.
+        object_content : bytes
+            The serialized object to store.
```
panh99, line 36:
```suggestion
            The deflated object to store.
```

tanertopal (approved):

PR #5310, opened 2025-04-29
refactor(framework) Rename "ping" to "heartbeat"

panh99:
Rename `PING`/`Ping`/`ping` to `HEARTBEAT`/`Heartbeat`/`heartbeat`


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -536,17 +539,19 @@ def get_federation_options(self, run_id: int) -> Optional[ConfigRecord]:
...
-    def acknowledge_ping(self, node_id: int, ping_interval: float) -> bool:
-        """Acknowledge a ping received from a node, serving as a heartbeat.
+    def acknowledge_heartbeat(self, node_id: int, heartbeat_interval: float) -> bool:
+        """Acknowledge a heartbeat received from a node, serving as a heartbeat.
 
-        It allows for one missed ping (in a PING_PATIENCE * ping_interval) before
-        marking the node as offline, where PING_PATIENCE = 2 in default.
+        A node is considered online as long as it sends heartbeats within
+        the tolerated interval: HEARTBEAT_PATIENCE × heartbeat_interval.
+        By default, HEARTBEAT_PATIENCE = 2, allowing for one missed heartbeat
```
jafermarq, line 547: I'm unsure about indicating as part of the docstrings the current value set in the constant. ( the `HEARTBEAT_PATIENCE = 2` bit). This appears in several other docstrings.

jafermarq (approved):

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: That's a good point! Thanks. I've added comments for the `HEARTBEAT_PATIENCE` so we won't forget to update the docstrings when we change its value.

jafermarq (approved):

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
danieljanes: I agree, we should avoid this duplication to prevent outdated comments in case we change the constant.

danieljanes (approved):

PR #5311, opened 2025-04-29
refactor(framework) Modify `flwr build` command to build FAB in flwr directory

kta-intel:
## Issue
Currently, `flwr build` (and as an extension, `flwr run`) will build a temp `.zip` file in `/tmp/` and then copy the contents to a `.fab` file in the current working directory. This is generally not an issue, but it becomes an issue when trying to run in a TEE (i.e. a Intel(R) SGX enclave) which has strict trusted/allowed files. 

While the list of trusted/allowed files can theoretically be extended to accommodate, the best practice is to minimize where an app has read/write permissions. This is addressed by enabling the build command to build in the `flwr-dir`. This also makes it more consistent with other `flwr` CLI commands that make use of `get_flwr_dir`

### Description
This PR:
- uses the existing `get_flwr_dir()` command to retrieve flwr-dir and sets it as the build directory
- adds an optional flag to `flwr build` to accept `--flwr-dir` if running as CLI with expanded docstring
- builds both the temp `.zip` and `.fab` files in this build directory
- returns the full path since `flwr run` uses the full path
- updates `.github/workflows/framework-e2e.yml` to install from `$HOME/.flwr`

### Potential open
- Changes to CI/CD, needed to update the framework-e2e.yml to install from `$HOME/.flwr` to reflect the new changes. 
      - **Is it preferrable to install in current working dir if `flwr-dir` is not specified instead of defaulting to `$HOME/.flwr`?**
- This PR does not modify `flwr run` so only `app` is being passed through build when building/installing via `flwr run` [[ref](https://github.com/adap/flower/blob/main/framework/py/flwr/cli/run/run.py#L161)]. By design, this will then follow the standard path (i.e. `$FLWR_HOME/`, `$HOME/.flwr`, etc)

### DIFF
--- a/.github/workflows/framework-e2e.yml
+++ b/.github/workflows/framework-e2e.yml
@@ -381,7 +381,7 @@ jobs:
           flwr new tmp-${{ matrix.framework }} --framework ${{ matrix.framework }} --username gh_ci
           cd tmp-${{ matrix.framework }}
           flwr build
-          flwr install *.fab
+          flwr install $HOME/.flwr/*.fab
 
   numpy:
     runs-on: ubuntu-22.04
--- a/framework/py/flwr/cli/build.py
+++ b/framework/py/flwr/cli/build.py
@@ -27,6 +27,7 @@ import pathspec
 import tomli_w
 import typer
 
+from flwr.common.config import get_flwr_dir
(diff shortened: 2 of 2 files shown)


panh99 (commented):
Thanks a lot for opening this PR — it already looks great! We're still evaluating the best way to address the issue you mentioned, and due to current priorities, we’ll be able to take a closer look and follow up next week. Appreciate your patience in the meantime!

PR #5322, opened 2025-05-02
feat(framework) Add `SendAppHeartbeat` to `ServerAppIo` service

panh99:


framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -362,6 +366,27 @@ def GetRunStatus(
         }
         return GetRunStatusResponse(run_status_dict=run_status_dict)
 
+    def SendAppHeartbeat(
+        self, request: SendAppHeartbeatRequest, context: grpc.ServicerContext
+    ) -> SendAppHeartbeatResponse:
+        """Handle a heartbeat from the ServerApp."""
```
jafermarq, line 372:
```suggestion
        """Handle a heartbeat from the ServerApp process."""
```

jafermarq: small thing, but I often find we mix the term `ServerApp` in several places: one to mention the class, the other the process that runs it. I think here is more correct to stress that it's the process the one sending the heartbeat requests.

panh99: Yeah, that's true. I actually delibrately use the word `ServerApp` because it's the heartbeat of the app, not the process. For example, when the process is idle, no heartbeats will be sent to the SuperLink from it. But I agree this might not be very clear here. Any ideas on how to clarify that?

jafermarq: right, in that case then it's probably better to not include the term `process`.

framework/py/flwr/server/superlink/serverappio/serverappio_servicer_test.py
```diff
@@ -477,3 +486,36 @@ def test_update_run_status_not_successful_if_finished(self) -> None:
...
+        # Stay in pending or transition to finished
+        self._transition_run_status(run_id, num_transitions)
+        request = SendAppHeartbeatRequest(run_id=run_id, heartbeat_interval=30)
+
+        # Execute
+        with self.assertRaises(grpc.RpcError) as e:
+            self._send_app_heartbeat.with_call(request=request)
+
+        # Assert
+        assert e.exception.code() == grpc.StatusCode.PERMISSION_DENIED
```
chongshenng, line 521: I realized that a better `grpc.StatusCode` to use is `grpc.StatusCode.FAILED_PRECONDITION` since the run is not in the correct status for the heartbeat to be sent. We also used it in the `ExecServicer` [here](https://github.com/adap/flower/blob/840bfec5199c0caedbc751ea54d131fb0bc832a2/framework/py/flwr/superexec/exec_servicer.py#L156). Wdyt? This can be done in a separate PR.

chongshenng (commented):
LGTM overall. Only one comment about the statuscode which we can address in a separate PR.

framework/py/flwr/server/superlink/serverappio/serverappio_servicer_test.py
panh99: Yeah, that also makes sense to me. No strong opinion. The `SendAppHeartbeat` RPC calls `abort_if` under the hood, which may raise an exception with `PERMISSION_DENIED`. I agree that we should do it in a separate PR if we want to change `abort_if` function.

chongshenng (approved):

danieljanes (approved):

PR #5323, opened 2025-05-02
feat(framework) Make `Array` serializable

jafermarq:
Makes `Array` inherit from `Serializable`.

> [!NOTE]
> Merge after https://github.com/adap/flower/pull/5320.


jafermarq:
Done in https://github.com/adap/flower/pull/5342

PR #5324, opened 2025-05-02
feat(framework) Add and implement `LinkState.acknowledge_app_heartbeat`

panh99:
### The initial heartbeat interval is set to `HEARTBEAT_MAX_INTERVAL` for now. 

This should be set back to `HEARTBEAT_DEFAULT_INTERVAL` once the PR to enable app heartbeat is merged.


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -461,8 +466,29 @@ def get_run_ids(self) -> set[int]:
         with self.lock:
             return set(self.run_ids.keys())
 
+    def _check_run_activeness(self) -> None:
```
chongshenng, line 469: This function does two things, checks the run activeness and tags the run as failed if it's not active. Should we rename the function to be clearer? For example: `_check_and_tag_inactive_run`?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -507,14 +539,23 @@ def update_run_status(self, run_id: int, new_status: RunStatus) -> bool:
...
+            # Update the run status
             if new_status.status == Status.STARTING:
-                run_record.run.starting_at = now().isoformat()
+                run_record.run.starting_at = current.isoformat()
             elif new_status.status == Status.RUNNING:
-                run_record.run.running_at = now().isoformat()
+                run_record.run.running_at = current.isoformat()
             elif new_status.status == Status.FINISHED:
-                run_record.run.finished_at = now().isoformat()
+                run_record.run.finished_at = current.isoformat()
```
chongshenng, line 558: It looks like this line will overwrite the `run.finished_at` value if the run is tagged as failed because it is inactive. Do you think it's necessary to distinguish the timestamp or it's acceptable to overwrite it?

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -788,8 +839,38 @@ def test_acknowledge_node_heartbeat(self) -> None:
         # Assert
         self.assertSetEqual(actual_node_ids, set(node_ids[70:]))
 
+    def test_acknowledge_app_heartbeat(self) -> None:
+        """Test if acknowledge_app_heartbeat works."""
+        # Prepare
+        state: LinkState = self.state_factory()
+        run_id1 = state.create_run(None, None, "9f86d08", {}, ConfigRecord())
+        run_id2 = state.create_run(None, None, "9f86d08", {}, ConfigRecord())
```
chongshenng, line 847: Let's use a different identifier so that we can distinguish the runs in tests.

chongshenng (commented):
@panh99 Thanks for adding the acknowledge protocol for the heartbeat. I've a few comments about how the timestamp is added and one nit.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: That's a good question! If a run is inactive and marked as `"finished:failed"` already, it won't pass line 523 (`if not is_valid_transition(current_status, new_status)`). Therefore the `finished_at` field won't be overwitten in that case.

But since we check the run status outside `Linkstate` via `abort_if` in `ServerAppIoServicer`, should we consider calling the method `_check_run_activeness` in the servicer instead of linkstate? Wdyt?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: That makes sense to me. Thanks for the idea!

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
chongshenng: That's a valid point. Yes, I think we should move any run checks out of the `LinkState`. This way, all of the checks are triggered by the ServerApp via the `ServerAppIoServicer`.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -558,6 +599,42 @@ def acknowledge_node_heartbeat(
...
 
+    def acknowledge_app_heartbeat(self, run_id: int, heartbeat_interval: float) -> bool:
+        """Acknowledge a heartbeat received from a ServerApp for a given run.
+
+        A run with status `"running"` is considered alive as long as it sends heartbeats
+        within the tolerated interval: HEARTBEAT_PATIENCE × heartbeat_interval.
+        HEARTBEAT_PATIENCE = N allows for N-1 missed heartbeat before the run is
+        marked as `"completed:failed"`.
+        """
+        with self.lock:
```
danieljanes, line 610: Why do we need to lock the entire `LinkState`?

panh99: Oh, do you mean we should just lock the run-related info?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -461,8 +466,29 @@ def get_run_ids(self) -> set[int]:
         with self.lock:
             return set(self.run_ids.keys())
 
+    def _check_and_tag_inactive_run(self) -> None:
```
danieljanes, line 469: I wonder if this is too expensive to run every time we receive a heartbeat. An alternative to doing this would be to only do it for the run that is currently requested. This should be sufficient since we run this check on every interaction with the `LinkState`.

panh99: Yeah, that makes sense to me. I can remove it from the `acknowledge_app_heartbeat`.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
danieljanes: This might be something for a separate PR, but we could have `run_id`-level locks that lock only the access to a single run.

panh99: That sounds good to me. Yeah, we can do that in a separate PR.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
danieljanes: We should probably keep this check in `acknowledge_app_heartbeat`, but only check the particular `run_in` in question, not all of them.

panh99: Yeah, sure.

panh99: @danieljanes I've limited the `_check_and_tag_inactive_run` to only update queried runs.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: I've just added a `lock` field in `RunRecord` in `InMemoryLinkState`, which is a per-run lock.

danieljanes (approved):

PR #5332, opened 2025-05-07
feat(framework) Add ServerApp heartbeat

panh99:
Based on #5322


framework/py/flwr/server/serverapp/app.py
```diff
@@ -236,6 +255,38 @@ def run_serverapp(  # pylint: disable=R0914, disable=W0212, disable=R0915
             break
 
 
+def _get_send_app_heartbeat_fn(
```
jafermarq, line 258: I wonder if it would be better to refactor this function slightly and place it in `common.heartbeat` so it can be reused by `flwr-clientapp` and `flwr-simulation` eventually/soon. In the end, the only difference would be the typer of request , right?

framework/py/flwr/server/serverapp/app.py
```diff
@@ -236,6 +255,38 @@ def run_serverapp(  # pylint: disable=R0914, disable=W0212, disable=R0915
...
+            if status_code == grpc.StatusCode.DEADLINE_EXCEEDED:
+                return False
+            raise
+
+        # Check if success
+        if not res.success:
+            raise RuntimeError(
+                "Heartbeat failed unexpectedly. The SuperLink does not "
+                "recognize this SuperNode."
+            )
```
chongshenng, line 284: Should this be `# Check if not successful`? Also, the second part of the error message should not mention SuperNode since this heartbeat originates from the ServerApp. Perhaps "The SuperLink does not recognize this ServerApp.". 

What do you think can cause `res.success == false`? Is it from a bad connection? I'm wondering if it's worth indicating potential issues here for debugging.

chongshenng (commented):
Thanks @panh99. I only have one small comment.

framework/py/flwr/server/serverapp/app.py
panh99: Yeah, that makes sense to me!

framework/py/flwr/server/serverapp/app.py
panh99: As discussed in the slack, I will resolve this.

jafermarq (dismissed):

chongshenng (dismissed):

framework/py/flwr/common/heartbeat.py
```diff
@@ -106,3 +113,52 @@ def _heartbeat(self) -> None:
...
                 raise HeartbeatFailure
+
+
+def get_grpc_app_heartbeat_fn(
+    stub: ServerAppIoStub,
+    run_id: int,
+    *,
+    failure_message: str,
+) -> Callable[[], bool]:
+    """Get the function to sends a heartbeat to gRPC endpoint.
```
danieljanes, line 124:
```suggestion
    """Get the function to send a heartbeat to gRPC endpoint.
```

framework/py/flwr/common/heartbeat.py
```diff
@@ -106,3 +113,52 @@ def _heartbeat(self) -> None:
...
+    req = SendAppHeartbeatRequest(
+        run_id=run_id, heartbeat_interval=HEARTBEAT_DEFAULT_INTERVAL
+    )
+
+    def fn() -> bool:
+        # Call ServerAppIo API
+        try:
+            res = stub.SendAppHeartbeat(req)
+        except grpc.RpcError as e:
+            status_code = e.code()
```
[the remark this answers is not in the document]
panh99: No, not here. But thanks bot.

danieljanes (approved):

PR #5335, opened 2025-05-07
feat(framework) Introduce `Inflatable` class

jafermarq:


framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @property
+    def object_id(self):
+        """Get object_id."""
+        _, object_id = self.deflate()
+        return object_id
+
+    @object_id.setter
+    def object_id(self, value: str) -> None:
+        self._object_id = value
```
danieljanes, line 41: Why does `object_id` have a setter? Shouldn't it only be set by internal methods, never by the user?

framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def get_object_id(serialized: bytes) -> str:
+    """Return a SHA-256 hash of the serialized object."""
+    return hashlib.sha256(serialized).hexdigest()
```
danieljanes, line 46:
```suggestion
def get_object_id(object_content: bytes) -> str:
    """Return a SHA-256 hash of the (deflated) object content."""
    return hashlib.sha256(object_content).hexdigest()
```

framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def get_object_body(serialized: bytes, cls: type[T]) -> bytes:
```
danieljanes, line 49: Depending on the decision above, we should update the use `serialized` wording

framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,86 @@
...
+    """Return object head from object content."""
+    return serialized[: serialized.find(HEAD_BODY_DIVIDER)]
+
+
+def _get_object_body(serialized: bytes) -> bytes:
+    """Return object body from object content."""
+    return serialized[serialized.find(HEAD_BODY_DIVIDER) + 1 :]
+
+
+def object_type_from_bytes(serialized: bytes) -> str:
```
danieljanes, line 83: Maybe like that?
```suggestion
def object_type_from_object_content(serialized: bytes) -> str:
```

framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,82 @@
...
+        _, object_id = self.deflate()
+        return object_id
+
+
+def get_object_id(object_content: bytes) -> str:
+    """Return a SHA-256 hash of the (deflated) object content."""
+    return hashlib.sha256(object_content).hexdigest()
+
+
+def get_object_body(content: bytes, cls: type[T]) -> bytes:
```
danieljanes, line 45: Should we typically use `object_content` or just `content` when naming arguments/variables?

framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def get_object_id(object_content: bytes) -> str:
+    """Return a SHA-256 hash of the (deflated) object content."""
+    return hashlib.sha256(object_content).hexdigest()
+
+
+def get_object_body(object_content: bytes, cls: type[T]) -> bytes:
+    """Return object body but raise an error if object type in bytes does not match name
+    of class."""
```
danieljanes, line 47:
```suggestion
    """Return object body but raise an error if object type does not match class name class."""
```

framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,81 @@
...
+        return object_id
+
+
+def get_object_id(object_content: bytes) -> str:
+    """Return a SHA-256 hash of the (deflated) object content."""
+    return hashlib.sha256(object_content).hexdigest()
+
+
+def get_object_body(object_content: bytes, cls: type[T]) -> bytes:
+    """Return object body but raise an error if object type does not match class name class."""
```
danieljanes, line 46:
```suggestion
    """Return object body but raise an error if object type doesn't match class name."""
```

framework/py/flwr/common/inflatable.py
```diff
@@ -0,0 +1,81 @@
...
+    # Construct header
+    header = f"{cls.__class__.__qualname__} {len(object_body)}"
+    enc_header = header.encode(encoding="utf-8")
+    # Concatenate header and object body
+    return enc_header + HEAD_BODY_DIVIDER + object_body
+
+
+def _get_object_head(object_content: bytes) -> bytes:
+    """Return object head from object content."""
+    return object_content[: object_content.find(HEAD_BODY_DIVIDER)]
```
danieljanes, line 70: Should we also use `.split(HEAD_BODY_DIVIDER, 1)[0]` here?

danieljanes (dismissed):

danieljanes (approved):

PR #5340, opened 2025-05-08
fix(datasets) Fix partition inconsistencies across dataset splits

### LINKED ISSUE #5243: Inconsistent partition IDs for different dataset splits using NaturalIdPartitioner
### Describe the bug

The `NaturalIdPartitioner` creates natural IDs using `Dataset.unique()`, where the ordering of the returned IDs depends on the order of examples in the dataset. Then, the partition IDs are assigned based on the of the natural IDs:

```python
def _create_int_partition_id_to_natural_id(self) -> None:
    """Create a mapping from int indices to unique client ids from dataset.

    Natural ids come from the column specified in `partition_by`.
    """
    unique_natural_ids = self.dataset.unique(self._partition_by)
    self._partition_id_to_natural_id = dict(
        zip(range(len(unique_natural_ids)), unique_natural_ids)
    )
```

This is problematic when creating multiple partitioners (e.g., for training and test splits) as the mappings from partition ID to natural ID can be different for each. In this scenario, if the mappings are not the same the clients will be trained and tested on on different _natural_ partitions (even though the partition IDs are the same).

This could be fixed by sorting the unique natural IDs before assigning the partition IDs:

```python
def _create_int_partition_id_to_natural_id(self) -> None:
    """Create a mapping from int indices to unique client ids from dataset.

    Natural ids come from the column specified in `partition_by`.
    """
    unique_natural_ids = sorted(self.dataset.unique(self._partition_by))
    self._partition_id_to_natural_id = dict(
        zip(range(len(unique_natural_ids)), unique_natural_ids)
    )
```

### Steps/Code to Reproduce

```python
from datasets import Dataset
from flwr_datasets.partitioner import NaturalIdPartitioner

# Create datasets
train_data = {
    "features": [1, 2, 3],
    "labels": [0, 0, 1],
    "clients": ["a", "b", "a"],
}
test_data = {
    "features": [4, 5, 6],
    "labels": [1, 0, 0],
    "clients": ["b", "a", "a"],
}
train_dataset = Dataset.from_dict(train_data)
test_dataset = Dataset.from_dict(test_data)

# Create partitioners
train_partitioner = NaturalIdPartitioner(partition_by="clients")
test_partitioner = NaturalIdPartitioner(partition_by="clients")
train_partitioner.dataset = train_dataset
test_partitioner.dataset = test_dataset

# Check partitioner mappings
train_partitioner.load_partition(partition_id=0)
print("Train partition mapping", train_partitioner.partition_id_to_natural_id)

test_partitioner.load_partition(partition_id=0)
print("Test partition mapping", test_partitioner.partition_id_to_natural_id)

```

### Expected Results

```shell
Generating partition_id_to_indices: 3it [00:00, 91180.52it/s]
Train partition mapping {0: 'a', 1: 'b'}
Generating partition_id_to_indices: 3it [00:00, 96791.63it/s]
Test partition mapping {0: 'a', 1: 'b'}
```

### Actual Results

```shell
Generating partition_id_to_indices: 3it [00:00, 91180.52it/s]
Train partition mapping {0: 'a', 1: 'b'}
Generating partition_id_to_indices: 3it [00:00, 96791.63it/s]
Test partition mapping {0: 'b', 1: 'a'}
```

adamtupper:
## Issue

This pull request addresses Issue #5243.

### Description

These changes addresses a similar issue in the `NaturalIdPartitioner` and `DirichletParitioner` that results in mismatches between dataset splits (e.g., for training, validation, and testing).

- For the `NaturalIdPartitioner`, natural IDs were mapped to partition IDs based on the ordering of examples in the dataset. This resulted in different natural ID to partition ID mappings if the dataset was ordered differently, or on different subsets of the same dataset.
- For the `DirichletPartitioner`, a similar issue meant that changing the order of the dataset (or generating partitions for different subsets of the same dataset) led to different label distributions for each partition.

### Related issues/PRs

Fixes #5243

## Proposal

Sort the `unique_natural_ids` and `unique_classes` for the `NaturalIdPartitioner` and `DirichletParitioner`, respectively.

### Explanation

This fix ensures that partition generation is no longer dependent on the ordering of examples in the dataset. This means that, providing the set of natural IDs/unique classes is the same between dataset splits, the partitions are consistently labeled across the partitioners for the different splits. An example of a scenario where this is important is for personalized FL, where each client has their own unique training _and_ test data.

### Checklist

- [x] Write tests
- [x] Make CI checks pass


adamtupper:
Hey everyone!

Just a heads-up on the changes I made to the tests. I adjusted the `_dummy_setup` methods to shuffle the dataset, which changes the ordering of examples for different partitioners. This introduces some randomness and highlights the issue I mentioned earlier. The downside is that the tests aren't deterministic anymore, but I felt this was a reasonable trade-off compared to making bigger changes to the setup methods.

If anyone has suggestions or a cleaner solution, I'm all ears!

jafermarq:
> Hey everyone!
> 
> Just a heads-up on the changes I made to the tests. I adjusted the `_dummy_setup` methods to shuffle the dataset, which changes the ordering of examples for different partitioners. This introduces some randomness and highlights the issue I mentioned earlier. The downside is that the tests aren't deterministic anymore, but I felt this was a reasonable trade-off compared to making bigger changes to the setup methods.
> 
> If anyone has suggestions or a cleaner solution, I'm all ears!

Thanks for highlighting this! I agree that introducing this randomness isn't super ideal so I was wondering if we can keep the construction of the test datasets in `_dummy_setup` as is and instead create new datasets in the two new test methods introduced. For example, how about using those you presented in your https://github.com/adap/flower/issues/5243?

adamtupper:
Thanks for your feedback @jafermarq! Definitely, that seems like a better option. I'll update the tests and rollback the changes to `_dummy_setup`.

PR #5341, opened 2025-05-09
feat(framework) Introduce `inflate` functionality in `InflatableObject`

jafermarq:


framework/py/flwr/common/inflatable.py
```diff
@@ -30,6 +30,22 @@ def deflate(self) -> bytes:
         """Deflate object."""
         raise NotImplementedError()
 
+    @classmethod
+    def inflate(cls, object_content: bytes) -> "InflatableObject":
+        """Deserialize the object from bytes.
```
jafermarq, line 35:
```suggestion
        """Inflate the object from bytes.
```

danieljanes (approved):

PR #5350, opened 2025-05-11
refactor(framework) Set upper limit for `click`

jafermarq:
The latest version of `click` isn't compatible with the latest of `type`. We se an upper limit for `click`.

### DIFF
--- a/framework/pyproject.toml
+++ b/framework/pyproject.toml
@@ -76,7 +76,7 @@ pathspec = "^0.12.1"
 rich = "^13.5.0"
 pyyaml = "^6.0.2"
 requests = "^2.31.0"
-click = <8.2.0"
+click = "<8.2.0"
 # Optional dependencies (Simulation Engine)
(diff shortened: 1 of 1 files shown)


framework/pyproject.toml
```diff
@@ -76,6 +76,7 @@ pathspec = "^0.12.1"
 rich = "^13.5.0"
 pyyaml = "^6.0.2"
 requests = "^2.31.0"
+click = "<=8.1.8"
```
jafermarq, line 79:
```suggestion
click = <8.2.0"
```

framework/pyproject.toml
```diff
@@ -76,6 +76,7 @@ pathspec = "^0.12.1"
 rich = "^13.5.0"
 pyyaml = "^6.0.2"
 requests = "^2.31.0"
+click = <8.2.0"
```
jafermarq, line 79:
```suggestion
click = "<8.2.0"
```

danieljanes (approved):

PR #5360, opened 2025-05-12
docs(framework) Add Flower Components Network Interface and Comm Model

dstripelis:


framework/docs/source/ref-flower-network-communication.rst
```diff
@@ -134,3 +134,50 @@ Typical examples include:
...
+~~~~~~~~~~~~~~~~~~~
+
+During real-world deployment, the push/pull communication model adopted by each
+component can influence decisions related to resource provisioning, scaling, monitoring,
+and reliability. To support such decisions, the list below outlines the communication
+model used between the Flower components:
+
+- **SuperLink ↔ ServerApp**: ServerApp pulls tasks from the SuperLink
+- **SuperLink ↔ SuperNode**: SuperNode pulls tasks from the SuperLink
+- **SuperNode ↔ ClientApp**: ClientApp pulls tasks from the SuperNode
```
jafermarq, line 183: I would rework these based on the language used in the ##Isolation Mode section in this page. The term `task` might be too generic and it doesn't map directly to a concept in `Flower` in all three cases above. For example, for **SuperLink ↔ ServerApp**, the `ServerApp` pulls a `Run` at first and a `FAB`, the it pulls and pushes `Messages` until it finishes executing (that would be what the _canonical_ `FlowerApp` would do, but it's up to the person writing it ofc)

framework/docs/source/ref-flower-network-communication.rst
```diff
@@ -134,3 +134,50 @@ Typical examples include:
...
+Similarly, the ``SuperNode`` component includes the ``ClientAppIo API``. Each of these
+APIs serves a distinct purpose during the runtime of a Flower app, as summarized in the
+table below.
+
+.. list-table:: Flower Components Default Ports and Purpose Overview
+    :widths: 25 25 50 50
+    :header-rows: 1
+
+    - - Component
+      - Port
```
chongshenng, line 153: Should we just say "Default Port" here? Then we can remove the `(default)` in the remaining rows.

chongshenng (changes requested):
Thanks @dstripelis, a few nits and I agree with Javier's comments too.

framework/docs/source/ref-flower-network-communication.rst
```diff
@@ -134,3 +134,50 @@ Typical examples include:
...
+      - API
+      - Purpose
+    - - ``SuperLink``
+      - 9091 (default)
+      - ``ServerAppIo API``
+      - Communication between the ``SuperLink`` and the ``ServerApp``
+    - -
+      - 9092 (default)
+      - ``Fleet API``
+      - Coordinates execution across ``SuperNodes``
```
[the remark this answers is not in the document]
dstripelis: True. It is more accurate to say "used by SuperNodes". Modified.

framework/docs/source/ref-flower-network-communication.rst
```diff
@@ -134,3 +134,50 @@ Typical examples include:
   MLFlow and Weights & Biases are often used to track the progress of training runs. In
   this setting, the ``ServerApp`` typically acts as a client to the metric logging
   service.
+
+Networking Interfaces
+~~~~~~~~~~~~~~~~~~~~~
+
+Each component — ``SuperLink``, ``ServerApp``, ``SuperNode``, and ``ClientApp`` —
```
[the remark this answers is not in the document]
dstripelis: Thank you for pointing this. I changed it throughout the page.

framework/docs/source/ref-flower-network-communication.rst
dstripelis: Looked better in the header and we avoid duplication.

dstripelis:
Copying for reference how the multi-row approach can be applied for the network interface table using the grid table. However, the problem is that our formatting scripts raise the following error: ``NotImplementedError: Tables with cells that span multiple cells are not supported. Consider using the 'include' directive to include the table from another file.`` - therefore we resort to the original approach.
```
+-------------+--------------+-------------------+--------------------------------------------------------------+
| Component   | Default Port | API               | Purpose                                                      |
+=============+==============+===================+==============================================================+
| SuperLink   | 9091         | ServerAppIo API   | Communication between the SuperLink and the ``ServerApp``    |
|             +--------------+-------------------+--------------------------------------------------------------+
|             | 9092         | Fleet API         | Used by the SuperNodes to communicate with the SuperLink     |
|             +--------------+-------------------+--------------------------------------------------------------+
|             | 9093         | Exec API          | Users interface with the SuperLink via this API using the    |
|             |              |                   | `FlowerCLI <ref-api-cli.html>`_.                             |
|             +--------------+-------------------+--------------------------------------------------------------+
+-------------+--------------+-------------------+--------------------------------------------------------------+
| SuperNode   | 9094         | ClientAppIo API   | Communication between the SuperNode and the ``ClientApp``    |
|             |              |                   | process                                                      |
+-------------+--------------+-------------------+--------------------------------------------------------------+
```

framework/docs/source/ref-flower-network-communication.rst
jafermarq: Yes. but i think "necessary information/inputs" can be clarified a bit for each case. Also, it's not just pulling, there is pushing too:
- For SuperLink ↔ SuperNode (Fleet API): The SuperNode pull/push Messages. It also pulls the FAB if a new run is being executed.
- For SuperLink ↔ ServerApp (ServerAppIo): The ServerApp process pull/push Messages from/to the SuperLink. It also pull the FAB as part of the first interaction with the SuperLink. At the end of the execution it pushes the Context back to the SuperLink.
- For SuperNode ↔ ClientApp (ClientAppIo): the same as above but between SuperNode and ClientApp process

dstripelis: Thank you for the clarification @jafermarq! I added a bit more text to clarify further in the latest commit.

framework/docs/source/ref-flower-network-communication.rst
```diff
@@ -62,7 +62,42 @@ Optional Network Connections
 Depending on the SuperLink and SuperNode configuration, Flower systems can have/use a
 number of additional network connections.
 
-Isolation mode
+Networking Interfaces
+~~~~~~~~~~~~~~~~~~~~~
+
+Each component — SuperLink, ``ServerApp``, SuperNode, and ``ClientApp`` — exposes ports
+for interacting with other Flower components. The SuperLink component includes three
```
danieljanes, line 69: Technically not true. `ServerApp` and `ClientApp` don't have open ports. Also, while we often refer to `flwr-serverapp` with as just `ServerApp`, we should be more precise in reference docs. We should call it either `flwr-serverapp` or something like "the `ServerApp` process".

dstripelis: Good point. We can change `ports` to `APIs` since indeed the referenced components do not have open ports. Moreover, based on the feedback and given that this is the first time we refer to `ServerApp` and `ClientApp` in this document we can be more detailed - this is the revised version:

> All Flower components — SuperLink, ServerApp process (`flwr-serverapp`), SuperNode, and ClientApp process (`flwr-clientapp`) — expose APIs to interact with other Flower components

framework/docs/source/ref-flower-network-communication.rst
```diff
@@ -62,7 +62,42 @@ Optional Network Connections
 Depending on the SuperLink and SuperNode configuration, Flower systems can have/use a
 number of additional network connections.
 
-Isolation mode
+Networking Interfaces
+~~~~~~~~~~~~~~~~~~~~~
```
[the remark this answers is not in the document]
dstripelis: Maybe just rename the header to ``Components APIs`` to avoid introducing two new subtitles and splitting the table into two? (we only have one record for SuperNode) - proposed edit:

> Components APIs
> All Flower components — SuperLink, ServerApp process (``flwr-serverapp``), SuperNode, and ClientApp process (``flwr-clientapp``) — expose APIs to interact with other Flower components.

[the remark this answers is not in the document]
danieljanes: Good idea

danieljanes (approved):

PR #5366, opened 2025-05-14
fix(framework) Fix `devcontainer.json` config due to failing `taplo` build

chongshenng:
We include `taplo` as one of our dev dependencies. In containers of some OS, `taplo` will be built from source, but in order to build the package, it requires `maturin`, which further requires the Rust toolchain to be installed. I.e., we see this:
```
Collecting taplo
  Downloading taplo-0.9.3.tar.gz (102 kB)
  ...
```

In other words, with the current `pyproject.toml` dev configuration, users cannot install `taplo` from a source distribution, if they do not have a Rust compiler installed. Note that this issue is specific to development containers which do not have the Rust toolchain included.

This PR updates the Dockerfile for our development container to:
1. Install `maturin == 1.8.6` which allows `taplo` to be installed from `source` by bootstrapping rust (https://github.com/PyO3/maturin/pull/2421#issue-2765260905).
2. Install `taplo` from the main branch of the GitHub repo, where a valid `pyproject.toml` file is present and therefore allows `pip install`.
3. 
It also updates the `devcontainer.json` to:
1. Update the `postCreateCommand` to call a script. This script temporarily comments out the existing `taplo` dependencies in `framework/pyproject.toml` and `devtools/pyproject.toml` to prevent conflicting installation of `taplo`.


.devcontainer/devcontainer.json
```diff
@@ -18,4 +21,4 @@
   "mounts": [
     "source=flwr-bashhistory,target=/commandhistory,type=volume"
   ]
-}
+}
```
Robert-Steiner, line 24:
```suggestion
}

```

Robert-Steiner (commented):
Looks good to me. As far as I understand, this is only a temporary fix until the new Taplo version is released. Is that correct?

chongshenng:
> Looks good to me. As far as I understand, this is only a temporary fix until the new Taplo version is released. Is that correct?

@Robert-Steiner I'm not sure if `taplo` will release a new update/supported wheels anytime soon. The last release was in July 2024. IMO, this is a semi-permanent fix.

Robert-Steiner (approved):

PR #5376, opened 2025-05-15
refactor(framework) Move utils and add `SqliteStateMixin`

panh99:
This reduces code duplication between `SqliteLinkState` and future `SqliteNodeState`

### Code moved without changes
The only change is in `SqliteLinkState` where it inherits `SqliteStateMixin` now.

### DIFF
--- a/framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
+++ b/framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
@@ -18,11 +18,9 @@
 # pylint: disable=too-many-lines
 
 import json
-import re
 import sqlite3
 import time
-from collections.abc import Sequence
-from logging import DEBUG, ERROR, WARNING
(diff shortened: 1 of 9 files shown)


framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -170,90 +171,22 @@
...
-        return result
+    @property
+    def schema_setup_commands(self) -> list[str]:
+        """Return the schema setup commands."""
+        return [
+            SQL_CREATE_TABLE_RUN,
+            SQL_CREATE_TABLE_LOGS,
+            SQL_CREATE_TABLE_CONTEXT,
+            SQL_CREATE_TABLE_MESSAGE_INS,
+            SQL_CREATE_TABLE_MESSAGE_RES,
```
jafermarq, line 185: would it make sense to also move these to the new `state_utils` ?

PR #5397, opened 2025-05-21
refactor(framework) Remove grpc-bidi support from SuperNode main loop

danieljanes:


framework/py/flwr/client/start_client_internal.py
```diff
@@ -276,32 +274,36 @@ def _on_backoff(retry_state: RetryState) -> None:
...
+                delete_node,
+            )
+            get_run = cast(
+                Callable[[int], Run],
+                get_run,
+            )
+            get_fab = cast(
+                Callable[[str, int], Fab],
+                get_fab,
+            )
```
panh99, line 294: Instead of casting the returns to correct types, should we update the signature of the `_init_connection`, i.e., removing `Optional`?

danieljanes: That would be better, but it might require us to change the signature in each of the transport type implementations, which means we have to update it in the legacy stack as well (not possible for `grpc-bidi`). Or do you think we could introduce the `cast` somewhere in between?

panh99: I see wym. Let's keep it like this then. I think we can revisit this when updating the grpc stack and introducing pythonic interface later.

panh99 (dismissed):
LGTM

tanertopal (approved):

PR #5402, opened 2025-05-22
feat(framework) Extend `NodeState` ABC

panh99:


framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,126 @@
...
+        self,
+        *,
+        run_id: Optional[int | list[int]] = None,
+        is_reply: Optional[bool] = None,
+        limit: Optional[int] = None,
+    ) -> dict[str, Message]:
+        """Retrieve messages based on the specified filters.
+
+        If a filter is set to None, it is ignored.
+        If a list is provided for a filter, it is treated as an OR condition.
```
danieljanes, line 48: Why `OR`? Shouldn't it be `AND`? `run_id` must match AND `is_reply` must match as well.

panh99: How do you like this version?

framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,126 @@
...
+    def get_message(
+        self,
+        *,
+        run_id: Optional[int | list[int]] = None,
+        is_reply: Optional[bool] = None,
+        limit: Optional[int] = None,
+    ) -> dict[str, Message]:
+        """Retrieve messages based on the specified filters.
+
+        If a filter is set to None, it is ignored.
```
chongshenng, line 46: For my understanding, are filters referring to any none `None` values set for `run_id`, `is_reply`, and `limit`?

panh99: Yes, exactly — `run_id`, `is_reply`, and `limit` are the filters. If any of them is `None`, it’s ignored when retrieving messages.

framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,126 @@
...
-    @abc.abstractmethod
+    @abstractmethod
     def get_node_id(self) -> int:
         """Get the node ID."""
+
+    @abstractmethod
+    def get_message(
+        self,
+        *,
+        run_id: Optional[Union[int, list[int]]] = None,
```
danieljanes, line 40: Should we just put `list[int]` here? No need to make it `Optional` since the list can be empty.

danieljanes: & rename it to `run_ids`?

panh99: I think the signature `int | Iterable[int] | None = None` (or similar) is quite common in the APIs of popular libraries. For example:

* **NumPy**: many functions accept `axis: int | tuple[int, ...] | None`, allowing `None` for default behavior.
* **Pandas**: methods like `drop` use parameters like `labels: Hashable | Sequence[Hashable] | None = None`.

These designs offer flexibility while allowing `None` to carry a specific semantic meaning distinct from an empty collection.

In our case, using `Optional[int | list[int]]` is helpful because:

1. A raw `list[int]` doesn't support a default value of `None`, which would force us to use a mutable default, `[]`, (not ideal).
2. Semantically, we treat run IDs in the list as an OR condition. So an empty list would mean "no run ID can satisfy the condition"—which is logically different from `None`, which might mean "no filtering by run ID".

That said, I’m happy to update the signature to `Optional[int | Iterable[int]]` if you think it's better.

panh99: Also open to rename it to `run_ids`! As they are internal APIs, I think both `run_id` and `run_ids` makes sense to me. I was trying to follow a design used by other public libraries like NumPy and Pandas.

framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,126 @@
...
+    @abstractmethod
+    def store_message(self, message: Message, object_id: str) -> None:
+        """Store a message.
+
+        Parameters
+        ----------
+        message : Message
+            The message to store.
+        object_id : str
+            The object ID of the message.
```
danieljanes, line 79: What's the reason for having to pass this in addition to the `Message`? Could `NodeState` just call `.object_id`?

panh99: Oh, I assume that, when I work on the SuperNode stuff, the content-addressable message should be ready soon, so I should ensure the signatures are compatible with it. This is the same as the LinkState design we discussed earlier in the content-addressable message sync.

framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,143 @@
...
+            The message to store.
+
+        Returns
+        -------
+        Optional[str]
+            The object ID of the stored message, or None if storage failed.
+        """
+
+    @abstractmethod
+    def get_message(
```
danieljanes: Should this be:
```suggestion
    def get_messages(
```

framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,143 @@
...
+            A sequence of messages matching the specified filters.
+
+        Notes
+        -----
+        **IMPORTANT:** Retrieved messages will **NOT** be returned again by subsequent
+        calls to this method, even if the filters match them.
+        """
+
+    @abstractmethod
+    def delete_message(
```
danieljanes:
```suggestion
    def delete_messages(
```

framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,143 @@
...
+        self,
+        *,
+        message_ids: Optional[Union[str, Sequence[str]]] = None,
```
danieljanes: Should we simplify this to:
```suggestion
        message_ids: Optional[Sequence[str]] = None,
```

panh99: Sure, np. No strong pinion on this.

framework/py/flwr/supernode/nodestate/nodestate.py
panh99: That makes sense.

framework/py/flwr/supernode/nodestate/nodestate.py
```diff
@@ -15,17 +15,143 @@
...
+
+        Parameters
+        ----------
+        message_ids : Optional[Sequence[str]] (default: None)
+            Sequence of message (object) IDs to filter by. If a sequence is provided,
+            it is treated as an OR condition.
+
+        Notes
+        -----
+        **IMPORTANT:** **All messages** will be deleted if no filters are provided.
```
danieljanes, line 106:
```suggestion
        **IMPORTANT:** All messages** will be deleted if no filters are provided.
```

PR #5429, opened 2025-05-28
refactor(framework) Cache `.object_id` and introduce `.is_dirty`

panh99:


framework/py/flwr/common/record/array.py
```diff
@@ -295,3 +295,36 @@ def inflate(
...
+    def is_dirty(self) -> bool:
+        """Check if the object is dirty after the last deflation."""
+        if "_is_dirty" not in self.__dict__:
+            self.__dict__["_is_dirty"] = True
+        return cast(bool, self.__dict__["_is_dirty"])
+
+    @is_dirty.setter
+    def is_dirty(self, value: bool) -> None:
+        """Set the dirty flag."""
+        self.__dict__["_is_dirty"] = value
```
jafermarq, line 316: move to base `InflatableObject` class?

framework/py/flwr/common/inflatable.py
```diff
@@ -55,13 +55,24 @@ def inflate(
...
     def children(self) -> dict[str, InflatableObject] | None:
         """Get all child objects as a dictionary or None if there are no children."""
         return None
 
+    @property
+    def is_dirty(self) -> bool:
+        """Check if the object is dirty after the last deflation.
+
+        An object is considered dirty if its content has changed since the last
+        deflation to determine whether its object ID has changed.
```
jafermarq, line 72:
```suggestion
        An object is considered dirty if its content has changed since the last
        its object ID was computed.
```

framework/py/flwr/common/record/array.py
```diff
@@ -295,3 +295,36 @@ def inflate(
             stype=proto_array.stype,
             data=proto_array.data,
         )
+
+    @property
+    def object_id(self) -> str:
+        """Get object ID."""
+        ret = super().object_id
+        self.is_dirty = False  # Reset dirty flag
```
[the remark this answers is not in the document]
panh99: Not really. In the ABC, we don't know when the object is dirty or not. By that I mean, it cannot set `.is_dirty` back to `True` when modified, so it's a bit risky to do that in the ABC.

framework/py/flwr/common/record/array.py
panh99: Hmmm, to ensure the `.object_id` is accurate, I think it's safer to let the `ABC.is_dirty` always return `True`.

danieljanes (approved):

PR #5432, opened 2025-05-28
refactor(framework) Move `message_id` creation out of `LinkState` and use `.object_id` in Deployment

jafermarq:
This PR:
- moves the `message_id` assignment out of the `LinkState`
    - for the Simulation Runtime (i.e. `InMemoryGrid` and `VCE`), the `message_id` remains `str(uuid4())`
    - for the Deployment Runtime (i.e. `GrpcGrid` and `SuperNode` connections), the `message_id` is the `.object_id`
- The `message_id` is set at pushing time: in the `Grid` for `ServerApp`; for the `ClientApp` this is done in the `SuperNode` and the `VCE` loop for the deployment and simulation runtimes respectively.
- The `LinkState` no longer operates with `UUID`, but with `str` (main motivation is that our `SHA256`-based `Object IDs` aren't UUID compliant)


framework/py/flwr/server/grid/inmemory_grid.py
```diff
@@ -56,7 +56,8 @@ def __init__(
     def _check_message(self, message: Message) -> None:
         # Check if the message is valid
         if not (
-            message.metadata.message_id == ""
+            message.metadata.message_id != ""
+            and message.metadata.message_id == str(UUID(message.metadata.message_id))
```
panh99, line 60: May I know what's this line for? It will raise a ValueError here if the message ID is not a valid UUID.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -304,13 +294,13 @@ def delete_messages(self, message_ins_ids: set[UUID]) -> None:
...
-    def get_message_ids_from_run_id(self, run_id: int) -> set[UUID]:
+    def get_message_ids_from_run_id(self, run_id: int) -> set[str]:
         """Get all instruction Message IDs for the given run_id."""
-        message_id_list: set[UUID] = set()
+        message_id_list: set[str] = set()
         with self.lock:
             for message_id, message in self.message_ins_store.items():
                 if message.metadata.run_id == run_id:
-                    message_id_list.add(message_id)
+                    message_id_list.add(str(message_id))
```
panh99, line 303:
```suggestion
                    message_id_list.add(message_id)
```

framework/py/flwr/client/message_handler/message_handler_test.py
```diff
@@ -293,6 +293,7 @@ def test_valid_message(self) -> None:
         valid_message = make_message(
             metadata=self.valid_out_metadata, content=RecordDict()
         )
+        valid_message.metadata.__dict__["_message_id"] = valid_message.object_id
```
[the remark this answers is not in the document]
panh99: Oh, I see. thank!

framework/py/flwr/server/grid/inmemory_grid.py
panh99: Yeah, I mean, we can either remove this line, or fix the check. If we want to check the form, I think it's better to:

```
try: 
    UUID(message.metadata.message_id)
except ValueError as err:
    raise ValueError(f"Invalid message ID: {message.metadata.message_id}") from err

PR #5445, opened 2025-05-30
refactor(framework) Update gRPC utils for `InflatableObjects`

jafermarq:
- passing a `Node` -- this is required when these functions are executed by the `SuperNode` side.
- passing a list of object IDs to the push helper, so only those objects indicated by the `*Servicer` are pushed

### DIFF
--- a/framework/py/flwr/common/inflatable_grpc_utils.py
+++ b/framework/py/flwr/common/inflatable_grpc_utils.py
@@ -50,7 +50,7 @@ def push_object_to_servicer(
     obj: InflatableObject,
     stub: Union[FleetStub, ServerAppIoStub],
     node: Node,
-    object_ids_to_push: Optional[list[str]] = None,
+    object_ids_to_push: Optional[set[str]] = None,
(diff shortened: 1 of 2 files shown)


framework/py/flwr/common/inflatable_grpc_utils.py
```diff
@@ -46,40 +47,51 @@
 
 
 def push_object_to_servicer(
-    obj: InflatableObject, stub: Union[FleetStub, ServerAppIoStub]
+    obj: InflatableObject,
+    stub: Union[FleetStub, ServerAppIoStub],
+    node: Node,
+    object_ids_to_push: Optional[list[str]] = None,
```
panh99, line 53: Since we already use `set[str]` for the return type of `get_desdendant_object_ids`, I think we can also use `set` here. Wdyt?

```suggestion
    object_ids_to_push: Optional[set[str]] = None,
```

panh99 (approved):
LGTM!

PR #5449, opened 2025-05-30
refactor(framework) Switch to new messaging system

jafermarq:


framework/py/flwr/common/inflatable_grpc_utils.py
```diff
@@ -94,10 +95,14 @@ def pull_object_from_servicer(
...
-    )
-    object_content = object_proto.object_content
+    object_available = False
+    while not (object_available):
+        object_proto: PullObjectResponse = stub.PullObject(
+            PullObjectRequest(node=node, run_id=run_id, object_id=object_id)
+        )
+        object_available = object_proto.object_available
+        object_content = object_proto.object_content
+        sleep(0.1)
```
danieljanes: We might need a way to stop this. The object requested might never become available.

framework/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -285,8 +306,24 @@ def send(message: Message) -> None:
 
         # Serialize Message
         message_proto = message_to_proto(message=message)
-        request = PushMessagesRequest(node=node, messages_list=[message_proto])
-        _ = stub.PushMessages(request)
+        descendants_mapping = get_message_to_descendant_id_mapping(message)
+        request = PushMessagesRequest(
+            node=node,
+            messages_list=[message_proto],
```
panh99, line 312: The `message_proto` here is the complete serialized message, containing `.content`, right? If so, why we still call `push_object_to_servicer`?

jafermarq: We'll remove the `content` from the `Message` being sent first soon.

framework/py/flwr/server/grid/grpc_grid.py
```diff
@@ -234,6 +247,20 @@ def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
...
+            for msg, msg_ids in zip(messages, res.message_ids):
+                # If Message was added to the LinkState correctly
+                if msg_ids is not None:
+                    obj_ids_to_push = set(res.objects_to_push[msg_ids].object_ids)
+                    # Push only object that are not in the store
+                    push_object_to_servicer(
+                        msg,
+                        self._stub,
+                        node=self.node,
+                        run_id=cast(Run, self._run).run_id,
```
panh99:
```suggestion
                        run_id=run_id,
```

framework/py/flwr/server/grid/grpc_grid.py
```diff
@@ -234,6 +247,20 @@ def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
                     "passed to `push_messages`). This could be due to a malformed "
                     "message.",
                 )
+            # Push objects
+            for msg, msg_ids in zip(messages, res.message_ids):
```
panh99: It's okay-ish for now. The real problem here is the type `Iterable[Message]`. This means that users can pass a generator to create messages instead of passing a list to save memory usage. In that case, we cannot iterate over the iterator `messages` twice.

danieljanes (approved):

PR #5450, opened 2025-05-30
refactor(framework) Implement push and pull objects in servicers

jafermarq:


framework/py/flwr/common/inflatable.py
```diff
@@ -163,8 +165,12 @@ def get_object_body_len_from_object_content(object_content: bytes) -> int:
 
 def check_body_len_consistency(object_content: bytes) -> bool:
     """Check that the object body is of length as specified in the head."""
-    body_len = get_object_body_len_from_object_content(object_content)
-    return body_len == len(_get_object_body(object_content))
+    try:
+        body_len = get_object_body_len_from_object_content(object_content)
+        return body_len == len(_get_object_body(object_content))
+    except ValueError:
+        log(WARNING, "Object content does match the expected format.")
```
jafermarq:
```suggestion
        log(ERROR, "Object content does match the expected format.")
```

@panh99 , better change to error?

panh99: Yeah, that makes sense. We can't proceed with invalid object content.

panh99: ~~Btw, should we log the error here? Or would it be better to simply have the function return True or False, and let the caller handle logging the error based on the return value?~~ Nvm, I just noticed that the exception will only be raised when the format is invalid. And just `False` might not suffice.

framework/py/flwr/server/superlink/serverappio/serverappio_servicer_test.py
```diff
@@ -674,23 +682,142 @@ def test_send_app_heartbeat_not_successful(self, num_transitions: int) -> None:
...
+    def test_pull_object_fails(self) -> None:
+        """Test `PullObject` in unsuported scenarios."""
+        run_id = self.state.create_run("", "", "", {}, ConfigRecord())
+        # Run is not running
+        req = PullObjectRequest(node=Node(node_id=123), run_id=run_id)
+        with self.assertRaises(grpc.RpcError) as e:
+            self._pull_object(request=req)
+        assert e.exception.code() == grpc.StatusCode.PERMISSION_DENIED
+
+        # Run is running but node id isn't recognized
```
panh99:
```suggestion
        # Run is running but node ID isn't recognized
```

framework/py/flwr/server/superlink/serverappio/serverappio_servicer_test.py
```diff
@@ -674,23 +682,142 @@ def test_send_app_heartbeat_not_successful(self, num_transitions: int) -> None:
...
+    def test_push_object_fails(self) -> None:
+        """Test `PushObject` in unsupported scenarios."""
+        run_id = self.state.create_run("", "", "", {}, ConfigRecord())
+        # Run is not running
+        req = PushObjectRequest(node=Node(node_id=123), run_id=run_id)
         with self.assertRaises(grpc.RpcError) as e:
             self._push_object(request=req)
         assert e.exception.code() == grpc.StatusCode.PERMISSION_DENIED
+
+        # Run is running but node id isn't recognized
```
panh99:
```suggestion
        # Run is running but node ID isn't recognized
```

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer_test.py
```diff
@@ -360,3 +376,147 @@ def test_get_fab_not_successful_if_not_running(self, num_transitions: int) -> No
...
+    def test_push_object_fails(self) -> None:
+        """Test `PushObject` in unsupported scenarios."""
+        run_id = self.state.create_run("", "", "", {}, ConfigRecord())
+        # Run is not running
+        req = PushObjectRequest(node=Node(node_id=123), run_id=run_id)
+        with self.assertRaises(grpc.RpcError) as e:
+            self._push_object(request=req)
+        assert e.exception.code() == grpc.StatusCode.PERMISSION_DENIED
+
+        # Run is running but node id isn't recognized
```
panh99:
```suggestion
        # Run is running but node ID isn't recognized
```

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer_test.py
```diff
@@ -360,3 +376,147 @@ def test_get_fab_not_successful_if_not_running(self, num_transitions: int) -> No
...
+    def test_pull_object_fails(self) -> None:
+        """Test `PullObject` in unsuported scenarios."""
+        run_id = self.state.create_run("", "", "", {}, ConfigRecord())
+        # Run is not running
+        req = PullObjectRequest(node=Node(node_id=123), run_id=run_id)
+        with self.assertRaises(grpc.RpcError) as e:
+            self._pull_object(request=req)
+        assert e.exception.code() == grpc.StatusCode.PERMISSION_DENIED
+
+        # Run is running but node id isn't recognized
```
panh99:
```suggestion
        # Run is running but node ID isn't recognized
```

panh99 (dismissed):
LGTM!

framework/py/flwr/common/inflatable.py
jafermarq: yes, i agree. let's address this in the upcoming and more thorough `is_object_valid` helper

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -409,18 +409,68 @@ def PushObject(
...
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, "Unexpected node ID.")
+
+        # Init store
+        store = self.objectstore_factory.store()
+
+        # Fetch from store
+        if content := store.get(request.object_id):
+            return PullObjectResponse(object_content=content)
+
         return PullObjectResponse()
```
danieljanes: Do we want to send empty responses, or rather `context.abort` to give the client a clear error?

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -199,4 +229,26 @@ def PullObject(
...
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, "Unexpected node ID.")
+
+        # Init store
+        store = self.objectstore_factory.store()
+
+        # Fetch from store
+        if content := store.get(request.object_id):
+            return PullObjectResponse(object_content=content)
+
         return PullObjectResponse()
```
danieljanes: Same question below: it might be better to `context.abort` instead of returning an empty response.

danieljanes (dismissed):

PR #5456, opened 2025-06-01
refactor(framework) Make record protocol serialization/deserialization deterministic

panh99:
tabilize record map/dictionary key order in protobuf

### DIFF
--- a/framework/py/flwr/common/serde_test.py
+++ b/framework/py/flwr/common/serde_test.py
@@ -371,6 +371,24 @@ def test_recorddict_serialization_deserialization() -> None:
     assert original == deserialized
 
 
+def test_recorddict_key_order_stability() -> None:
+    """Test that RecordDict preserves key order."""
+    # Prepare
+    maker = RecordMaker(state=1)
+    original = maker.recorddict(10, 5, 5)
+
+    # Execute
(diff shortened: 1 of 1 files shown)


framework/py/flwr/common/serde.py
```diff
@@ -461,28 +467,33 @@ def config_record_from_proto(record_proto: ProtoConfigRecord) -> ConfigRecord:
...
+    for k, v in recorddict.items():
+        if isinstance(v, ArrayRecord):
+            items += [item_cls(key=k, array_record=array_record_to_proto(v))]
+        elif isinstance(v, MetricRecord):
+            items += [item_cls(key=k, metric_record=metric_record_to_proto(v))]
+        elif isinstance(v, ConfigRecord):
+            items += [item_cls(key=k, config_record=config_record_to_proto(v))]
+        else:
+            raise ValueError(f"Unsupported record type: {type(v)}")
+    return ProtoRecordDict(items=items)
```
danieljanes, line 481: Can we add a test for this and the following functions to check whether order is preserved?

panh99: Sure, I will add that.

danieljanes (approved):

PR #5470, opened 2025-06-03
feat(framework) Add `NodeState.create/verify/delete_token` methods and `.get_run_ids_with_pending_messages` method

panh99:


framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -148,3 +152,38 @@ def get_context(self, run_id: int) -> Optional[Context]:
...
+            }
+
+        # Remove run IDs that have tokens stored (indicating they are in progress)
+        with self.lock_token_store:
+            ret -= set(self.token_store.keys())
+            return list(ret)
+
+    def create_token(self, run_id: int) -> str:
+        """Create a token for the given run ID."""
+        token = secrets.token_hex(128)  # Generate a random token
```
chongshenng: Should we make this a constant in `flwr.common.constant`?

framework/py/flwr/supernode/nodestate/nodestate_test.py
```diff
@@ -159,6 +159,55 @@ def test_delete_message(self) -> None:
...
+        self.state.get_messages(run_ids=[4])
+        self.state.create_token(2)
+        self.state.create_token(5)
+        self.state.delete_token(5)
+
+        # Execute
+        run_ids = self.state.get_run_ids_with_pending_messages()
+
+        # Assert: only run 1 should be returned
+        self.assertEqual(set(run_ids), {1, 5})
```
chongshenng: Run 5 is also asserted in this return. Does it mean that run IDs with pending messages can also correspond to a run that had a token but was deleted? What would happen to the run then?

framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
panh99: Yeah, that makes sense!

framework/py/flwr/supernode/nodestate/nodestate_test.py
panh99: Oh, just for context, After selecting which run to handle, the scheduler first attempts to acquire a token associated with the run ID. Once the token is acquired, the SuperNode marks the run as "in progress", so `.get_run_ids_with_pending_message()` will no longer return that run ID. The token will also be used for authenticated communication between `flwr-clientapp` <> SuperNode.

As for run 5, it’s meant to mock a scenario where a token was created for that run but later deleted because the `flwr-clientapp` finished handling the message. So run 5 should still be included. For example, if message A of run N comes in and gets handled (which means a token is created and then deleted), and then message B of run N comes in again, it should also be handled.

chongshenng: Ok, makes sense. I'd suggest updating the comment on line 167 to describe the case (where a token was created for that run but later deleted because the `flwr-clientapp` finished handling the message). 
Then, the comment on line 181 should be `Assert: only runs 1 and 5 should be returned`

danieljanes (approved):

danieljanes (approved):

PR #5472, opened 2025-06-04
refactor(framework) Handle non-existent object and run stopped case

jafermarq:


framework/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -261,15 +261,20 @@ def receive() -> Optional[Message]:
...
+                        object_id=message_proto.metadata.message_id,
+                        stub=stub,
+                        node=node,
+                        run_id=message_proto.metadata.run_id,
+                    ),
+                )
+            except RunNotRunningException:
+                log(ERROR, "Run stopped running.")
+            except KeyError as ke:
+                log(ERROR, ke)
```
jafermarq, line 277: I wonder if as part of this exception handling we should send an `Error` message back to the `LinkState`. Else the run will hang indefinitely.

panh99: Why there's a key error?

jafermarq: for example:
```python
            except KeyError as ke:
                log(ERROR, ke)
                msg = message_from_proto(message_proto)
                metadata = copy(msg.metadata)
                send(Message(error=Error(code=0), reply_to=msg)) # change to a new error code
```

PR #5473, opened 2025-06-04
feat(framework) Use `NodeState` to store incoming messages

panh99:


framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -154,73 +158,79 @@ def start_client_internal(
...
-
-                run.fab_id, run.fab_version = fab_id, fab_version
+            # Ensure the run and FAB are available
+            run_id = message.metadata.run_id
+            try:
+                # Check if the message is from an unknown run
+                if (run_info := state.get_run(run_id)) is None:
+                    # Pull run info from SuperLink
+                    run_info = get_run(run_id)
+                    state.store_run(run_info)
```
danieljanes, line 199: Do we always store the run or do we need to check whether the call was successful?

panh99: It's a rare case or fatal case actually. If `get_run` fails to get the run details, it means that our SuperLink is in an unperdictable situation and it somehow stores messages of a run that doesn't exist. We don't previously have the check for whether the call is successful or not, but I think we can add it in a separate PR if you'd like an extra check here.

PR #5474, opened 2025-06-04
refactor(framework) Make pull/push object helper functions accept `Callable`

panh99:


framework/py/flwr/common/inflatable_grpc_utils.py
```diff
@@ -115,8 +151,79 @@ def pull_object_from_servicer(
...
+    run_id: int,
+) -> Callable[[str, bytes], None]:
+    """Create a push object function that uses gRPC to push objects.
+
+    Parameters
+    ----------
+    push_object_grpc : Callable[[PushObjectRequest], PushObjectResponse]
+        The gRPC function to push objects, e.g., `FleetStub.PushObject`.
+    node : Node
+        The node to which to push the object.
```
jafermarq:
```suggestion
        The node making the request.
```

framework/py/flwr/common/inflatable_grpc_utils.py
```diff
@@ -115,8 +151,79 @@ def pull_object_from_servicer(
...
+    run_id: int,
+) -> Callable[[str], bytes]:
+    """Create a pull object function that uses gRPC to pull objects.
+
+    Parameters
+    ----------
+    pull_object_grpc : Callable[[PullObjectRequest], PullObjectResponse]
+        The gRPC function to pull objects, e.g., `FleetStub.PullObject`.
+    node : Node
+        The node from which to pull the object.
```
jafermarq:
```suggestion
        The node making the request.
```

framework/py/flwr/common/inflatable_grpc_utils.py
```diff
@@ -48,61 +45,100 @@
 }
 
 
+class ObjectUnavailableError(Exception):
+    """Exception raised when an object has been pre-registered but is not yet
+    available."""
+
+    def __init__(self, object_id: str):
+        super().__init__(f"Object with ID '{object_id}' is not yet available.")
+        self.object_id = object_id
```
jafermarq: Why do we need to keep it around ? (same goes for the `ObjectIdNotPreregisteredError` below)

framework/py/flwr/common/inflatable_grpc_utils.py
```diff
@@ -48,61 +45,100 @@
...
     IDs.
+
+    Parameters
+    ----------
+    obj : InflatableObject
+        The object to push.
+    push_object_fn : Callable[[str, bytes], None]
+        A function that takes an object ID and its content as bytes, and pushes
+        it to the servicer. This function should raise `ObjectIdNotPreregisteredError`
+        if the object ID is not pre-registered.
```
jafermarq, line 81: this is not the current behaviour in the servicer. do we want it? For now, if the object is not stored a `PushObjectResponse(stored=False)` is returned.

framework/py/flwr/common/inflatable_grpc_utils.py
panh99: Oh, just in case we wanna to capture this exception and know what object ID it is in the future, e.g.:

```
try:
    xxxx
except ObjectUnavailableError as err:
    print(err.object_id)
```

Wdyt? Should we keep it?

framework/py/flwr/common/inflatable_grpc_utils.py
panh99: You are right that our servicer won't raise such an exception. But it will return `stored=False`. So, in `make_push_object_fn_grpc`, the returned function will raise an exception when `stored` is `False`.

framework/py/flwr/common/inflatable_grpc_utils.py
jafermarq: I'd say let's add it when we think its needed?

framework/py/flwr/common/inflatable_grpc_utils.py
jafermarq: right! i see. Let's keep it as is

framework/py/flwr/common/inflatable_grpc_utils.py
panh99: Make sense. I will remove them.

jafermarq (approved):
🔥

danieljanes (approved):

PR #5476, opened 2025-06-04
refactor(framework) Move into a separate function the code logic of pulling messages in SuperNode

panh99:
\


framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -302,6 +259,82 @@ def start_client_internal(
...
 
+def _pull_and_store_message(  # pylint: disable=too-many-positional-arguments
+    state: NodeState,
+    ffs: Ffs,
+    object_store: ObjectStore,  # pylint: disable=unused-argument
+    node_config: UserConfig,
+    receive: Callable[[], Optional[Message]],
+    get_run: Callable[[int], Run],
+    get_fab: Callable[[str, int], Fab],
+) -> int:
```
danieljanes: Should we turn this into an `Optional[int]` to indicate more clearly whether we have a valid `run_id` or not?

panh99: That makes sense!

danieljanes (approved):

PR #5477, opened 2025-06-04
feat(framework) Introduce validator function of object content

jafermarq:
A new helper function is added that validates the structure of an Object content (`bytes`). This can be used in different places after merging (similarly to how the `validate_message` function is used)

### DIFF
--- a/framework/py/flwr/common/inflatable.py
+++ b/framework/py/flwr/common/inflatable.py
@@ -127,12 +127,14 @@ def add_header_to_object_body(object_body: bytes, obj: InflatableObject) -> byte
 
 def _get_object_head(object_content: bytes) -> bytes:
     """Return object head from object content."""
-    return object_content.split(HEAD_BODY_DIVIDER, 1)[0]
+    index = object_content.find(HEAD_BODY_DIVIDER)
+    return object_content[:index]
 
 
 def _get_object_body(object_content: bytes) -> bytes:
     """Return object body from object content."""
-    return object_content.split(HEAD_BODY_DIVIDER, 1)[1]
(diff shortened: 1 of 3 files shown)


framework/py/flwr/common/inflatable.py
```diff
@@ -25,6 +25,16 @@
 from .logger import log
 
 
+class UnexpectedObjectContentError(Exception):
+    """Exception raised when the content of an object does not follow the structure
+    (i.e. head, body, parts in the head) expected for an InflatableObject to have."""
```
panh99:
```suggestion
    """Exception raised when the content of an object does not conform to the expected
    structure for an InflatableObject (i.e., head, body, and values within the head)."""
```

framework/py/flwr/common/inflatable.py
```diff
@@ -117,12 +127,14 @@ def add_header_to_object_body(object_body: bytes, obj: InflatableObject) -> byte
...
-    return object_content.split(HEAD_BODY_DIVIDER, 1)[0]
+    index = object_content.find(HEAD_BODY_DIVIDER)
+    return object_content[:index]
 
 
 def _get_object_body(object_content: bytes) -> bytes:
     """Return object body from object content."""
-    return object_content.split(HEAD_BODY_DIVIDER, 1)[1]
+    index = object_content.find(HEAD_BODY_DIVIDER)
+    return object_content[index + 1 :]
```
panh99:
```suggestion
    return object_content[index + len(HEAD_BODY_DIVIDER):]
```

danieljanes (approved):

danieljanes (approved):

PR #5483, opened 2025-06-06
ci(framework) Fix E2E tensorflow dataset loading

mohammadnaseri:
This would allow to load the dataset from HF.

### DIFF
--- a/intelligence/ts/package.json
+++ b/intelligence/ts/package.json
@@ -72,7 +72,7 @@
   },
   "dependencies": {
     "@huggingface/transformers": "^3.4.0",
-    "@mlc-ai/web-llm": "^0.2.78",
+    "@mlc-ai/web-llm": "^0.2.79",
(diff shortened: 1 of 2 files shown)


.github/workflows/framework-e2e.yml
```diff
@@ -162,8 +162,7 @@ jobs:
           - directory: e2e-tensorflow
             e2e: e2e_tensorflow
             dataset: |
-              import tensorflow as tf
-              tf.keras.datasets.cifar10.load_data()
+              from datasets import load_dataset;import numpy as np;dataset = load_dataset("uoft-cs/cifar10");np.stack(dataset["train"]["image"])
```
panh99:
```suggestion
              from datasets import load_dataset;import numpy as np;dataset = load_dataset(\"uoft-cs/cifar10\");np.stack(dataset[\"train\"][\"image\"])
```

PR #5492, opened 2025-06-09
refactor(framework) Make use of new object content validator

jafermarq:
- Make use of `validate_object_content` when pushing objects to either servicer. 
- The `check_body_len_consistency` function and associated tests have been removed. This function was introduced as a primitive consistency check for the object content. The check that this function performed (and the associated test) are captured by `validate_object_content` and its tests.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #5493, opened 2025-06-09
feat(framework) Limit the maximum size of FAB file

panh99:


framework/py/flwr/superexec/exec_servicer.py
```diff
@@ -75,6 +75,13 @@ def StartRun(
         """Create run ID."""
         log(INFO, "ExecServicer.StartRun")
 
+        if len(request.fab.content) > FAB_MAX_SIZE:
+            log(
+                ERROR,
+                f"FAB size exceeds maximum allowed size of {FAB_MAX_SIZE} bytes.",
+            )
```
chongshenng: Should we include some useful guide here to the user, e.g. "You can ignore files via your .gitignore or exclude unnecessary files from the build to reduce package size."

panh99: That makes sense to me, but not here, I think. This is logged on the SuperLink side, so it's better to log the guide on the CLI side.

danieljanes: @panh99 did you want to include the message in the CLI change above?

panh99: Yes, I will!

framework/py/flwr/superexec/exec_servicer.py
```diff
@@ -84,6 +85,14 @@ def StartRun(
         """Create run ID."""
         log(INFO, "ExecServicer.StartRun")
 
+        if len(request.fab.content) > FAB_MAX_SIZE:
+            log(
+                ERROR,
+                "FAB size exceeds maximum allowed size of %d bytes.",
+                FAB_MAX_SIZE,
+            )
+            return StartRunResponse()
```
danieljanes, line 94: Shouldn't we `context.abort` here?

panh99: I think we can return an empty response here, similar to line 106. In `run.py`, when it receives a response with no `run_id`, it will exit with "❌ Failed to start run"

danieljanes (approved):

PR #5503, opened 2025-06-11
refactor(framework) Rename `user_name` to `account_name`

chongshenng:


danieljanes (approved):

framework/py/flwr/superexec/exec_event_log_interceptor_test.py
```diff
@@ -95,7 +95,7 @@ def setUp(self) -> None:
         # Because shared_account_info.get() is read-only, we need to set the user info
```
panh99:
```suggestion
        # Because shared_account_info.get() is read-only, we need to set the account info
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #5509, opened 2025-06-12
feat(framework) Count references and runs in `ObjectStore`

panh99:


framework/py/flwr/server/superlink/utils.py
```diff
@@ -75,16 +75,20 @@ def store_mapping_and_register_objects(
 ) -> dict[str, ObjectIDs]:
     """Store Message object to descendants mapping and preregister objects."""
     objects_to_push: dict[str, ObjectIDs] = {}
+    msg_id_to_run_id = {
+        msg.metadata.message_id: msg.metadata.run_id for msg in request.messages_list
+    }
```
jafermarq: Do we need to explicitly construct this mapping given how we use it later in line `84`? I believe we can replace line `84` simply with this and remove the lines highlighted above:

```python
run_id = msg.metadata.run_id
```

Or even simply pass `msg.metadata.run_id` to line `91`.
wdyt?

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -28,38 +31,78 @@
...
+                        child_id = child_node.object_id
+                        self.store[child_id].ref_count += 1
+
+                    # Add the object ID to the run's mapping
+                    self.run_objects_mapping[run_id].add(obj_id)
+
+                    # Add to the list of new objects
+                    new_objects.append(obj_id)
+                else:
+                    obj_entry = self.store[obj_id]
```
jafermarq, line 91:
```suggestion
                    # Object is in store, retrieve it
                    obj_entry = self.store[obj_id]
```

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -69,41 +112,120 @@ def put(self, object_id: str, object_content: bytes) -> None:
...
+                )
 
-        self.store[object_id] = object_content
+            # Return if object is already present in the store
+            if self.store[object_id].is_available:
+                return
+
+            # Update the object entry in the store
+            self.store[object_id].is_available = True
+            self.store[object_id].content = object_content
```
jafermarq:
```suggestion
            self.store[object_id].content = object_content
            self.store[object_id].is_available = True
```

jafermarq: i thought it would be more logical to change the order. ofc no impact overal

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -69,41 +112,120 @@ def put(self, object_id: str, object_content: bytes) -> None:
...
+            # Check if the object ID is pre-registered
+            if object_id not in self.store:
+                return None
+
+            # Check if the object is available (i.e., has been put into the store)
+            object_entry = self.store[object_id]
+            if not object_entry.is_available:
+                return b""
+
+            return object_entry.content
```
jafermarq, line 165:
```suggestion
            return object_entry.content
```

We are preregistering it with `b""` so we can return `.content` directly

jafermarq:
```suggestion
            # Return content (if not yet available, it will b"")
            return self.store[object_id]
```

We are preregistering it with `b""` so we can return `.content` directly

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -69,41 +112,120 @@ def put(self, object_id: str, object_content: bytes) -> None:
...
+            for object_id in list(self.run_objects_mapping[run_id]):
+                # Check if the object is still in the store
+                if (object_entry := self.store.get(object_id)) is None:
+                    continue
+
+                # Remove the run ID from the object's runs
+                object_entry.runs.discard(run_id)
+
+                # Trigger recursive deletion if the object is no longer referenced
+                if object_entry.ref_count == 0 and not object_entry.runs:
```
jafermarq: The left part of the if statement is not needed as it will considered when calling `self.delete()`

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -69,41 +112,120 @@ def put(self, object_id: str, object_content: bytes) -> None:
...
-        if object_id in self.store:
-            del self.store[object_id]
+        """Delete an object and its unreferenced descendants from the store."""
+        with self.lock_store:
+            # If the object is not in the store, nothing to delete
+            if (object_entry := self.store.get(object_id)) is None:
+                return
+
+            # Delete the object if it has no references left
+            if object_entry.is_available and object_entry.ref_count == 0:
```
jafermarq: wouldn't we want to also delete objects that haven't been fully uploaded to the `ObjectStore`? i.e. still with content `b""`. If so, we probably need to discard the `object_entry.is_available` part in the statement. wdyt?

framework/py/flwr/supercore/object_store/object_store_test.py
```diff
@@ -211,10 +215,23 @@ def test_preregister(self) -> None:
         object_content3 = obj3.deflate()
         object_id3 = get_object_id(object_content3)
         # Execute (preregister new object)
-        not_present = object_store.preregister(get_object_tree(obj3))
+        not_present = object_store.preregister(self.run_id, get_object_tree(obj3))
         # Assert (only new message is not present)
         self.assertEqual([object_id3], not_present)
```
jafermarq: i believe we aren't testing anything new with this `obj3`, right? let's remove if you also agree

framework/py/flwr/supercore/object_store/object_store_test.py
```diff
@@ -223,32 +240,136 @@ def test_preregister_with_invalid_object_id(self, invalid_object_id) -> None:
...
+        self.assertTrue(ids[4] in object_store)  # parent2
+
+        # Execute: Delete objects in run 2
+        object_store.delete_objects_in_run(run_id=2)
+
+        # Assert: The store should be empty now
+        self.assertEqual(len(object_store), 0)
+
+
+def _create_object_hierarchy() -> tuple[list[CustomDataClass], dict[str, bytes]]:
```
jafermarq, line 340: minor comment, i wonder if it would be easier from the maintenance perspective to replace this hand crafted hierarchy with a `RecordDict` object with a bunch of `*Records`.

framework/py/flwr/server/superlink/utils.py
panh99: Oh, that's true. They should be of the same run anyway. Thanks for the suggestion!

framework/py/flwr/supercore/object_store/in_memory_object_store.py
panh99: Good catch! You are 100% right.

framework/py/flwr/supercore/object_store/in_memory_object_store.py
panh99: Sure, that makes sense.

framework/py/flwr/supercore/object_store/in_memory_object_store.py
panh99: Yes, that's true, but I think it's good to avoid unnecessary calls to `self.delete`.

framework/py/flwr/supercore/object_store/object_store_test.py
panh99: Yeah, right. Sure I will remove this.

framework/py/flwr/supercore/object_store/in_memory_object_store.py
panh99: Yeah, that makes sense to me. I will drop the condition. Thanks!

framework/py/flwr/supercore/object_store/in_memory_object_store.py
jafermarq: but wouldn't we want to trigger the deletion of an object always if `object_entry.runs` is empty? (i.e. regardless of the `ref_count`, which something that the `delete` methods considers anyway)

panh99: Oh, you mean directly delete the object when it's `.runs` is empty? I understand wym, but we don't need to delete it unless it's reference count is zero. If it has non-zero ref-count, that means it can later be deleted. Directly deleting an object with non-zero reference count may break the chain and prevent recursive deletion.

panh99: This condition here actually means: `if this object is a message of the run to be deleted`.

framework/py/flwr/supercore/object_store/object_store_test.py
panh99: Oh, I see wym. I think it's good for now. The relation of objects are clearer than `RecordDict <> *Record` IMO, so they are good for unit tests. Wdyt?

jafermarq: ok, let's keep it

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -69,41 +113,116 @@ def put(self, object_id: str, object_content: bytes) -> None:
...
     def get(self, object_id: str) -> Optional[bytes]:
         """Get an object from the store."""
-        return self.store.get(object_id)
+        with self.lock_store:
+            # Check if the object ID is pre-registered
+            if object_id not in self.store:
+                return None
+
+            # Return content (if not yet available, it will b"")
+            return self.store[object_id]
```
jafermarq:
```suggestion
            return self.store[object_id].content
```

jafermarq (approved):
🌳 🔥

danieljanes (approved):

PR #5511, opened 2025-06-12
break(framework) Add `flwr_aid`-`run_id` guard clause to `ExecServicer.ListRuns`, `.StreamLogs`, and `.StopRun`

chongshenng:
### Description

Previously, when the `auth_plugin` was enabled, users could execute `ListRuns` (via `flwr ls --runs` and `flwr ls --run-id <RUN_ID>`), `StreamLogs` (via `flwr log <RUN_ID>`), and `StopRun` (via `flwr stop <RUN_ID>`) on runs they didn’t own. This PR adds guard clauses to ensure that users can only interact with their own run_ids.

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


panh99 (dismissed):
LGTM!

framework/py/flwr/cli/utils.py
```diff
@@ -320,5 +320,7 @@ def flwr_cli_grpc_exc_handler() -> Iterator[None]:
                 fg=typer.colors.RED,
                 bold=True,
             )
+            # pylint: disable=E1101
+            typer.secho(e.details(), fg=typer.colors.RED, bold=True)
```
danieljanes, line 324: Is this intentional? We don't echo the error details above

chongshenng: Yes it's intentional. Otherwise, users will just see the generic `❌ Authorization failed ...`, but won't know that the `run_id` is not associated with their `flwr_aid`.

framework/py/flwr/superexec/exec_servicer.py
```diff
@@ -231,3 +257,22 @@ def _create_list_runs_response(run_ids: set[int], state: LinkState) -> ListRunsR
...
     )
+
+
+def _check_flwr_aid_in_run(
+    flwr_aid: Optional[str],
+    run_id: int,
+    state: LinkState,
+    context: grpc.ServicerContext,
+) -> None:
+    run = state.get_run(run_id)
```
danieljanes: We get the run before, let's just pass that into this function instead of reading it from `state` again

chongshenng: Not sure if I understand. We can pass `run_id` for `StreamLogs` and `StopRun`, but we can't pass the run for `ListRuns`. I can move the `state.get_run(run_id)` outside of this helper function, but it will introduce some repetition. 

For now, I've tweaked `StopRun` and `StreamLogs` to pass `run_id` instead of `request.run_id` to `_check_flwr_aid_in_run()`.

chongshenng: I think I get what you mean. Let me refactor this.

chongshenng: Updated to remove `state` from the function.

framework/py/flwr/superexec/exec_servicer.py
```diff
@@ -231,3 +264,15 @@ def _create_list_runs_response(run_ids: set[int], state: LinkState) -> ListRunsR
...
         now=now().isoformat(),
     )
+
+
+def _check_flwr_aid_in_run(
+    flwr_aid: Optional[str],
+    run: Run,
+    context: grpc.ServicerContext,
+) -> None:
+    if run.flwr_aid and flwr_aid is not None and run.flwr_aid != flwr_aid:
```
danieljanes: This is `False` if `run.flwr_aid` is set but `flwr_aid` is `None`, which means we won't abort if the run is associated with a `flwr_aid` but the provided `flwr_aid` is `None`.

chongshenng: Good catch! Resolved in 93429b8.

framework/py/flwr/superexec/exec_servicer.py
```diff
@@ -133,11 +141,42 @@ def ListRuns(
...
-        return _create_list_runs_response({request.run_id}, state)
+            if self.auth_plugin:
+                # If no `run_id` is specified and user auth is enabled,
+                # return run IDs for the authenticated user
+                flwr_aid = shared_account_info.get().flwr_aid
+                if flwr_aid is None:
+                    context.abort(
+                        grpc.StatusCode.PERMISSION_DENIED,
+                        "️⛔️ User authentication is enabled, but `flwr_aid` is None",
+                    )
```
panh99, line 154: Is there a case where the `flwr_aid` is None but it passes our authentication?

danieljanes: there shouldn't be. if auth is enabled, then `flwr_aid` should always be set/valid and every run should have a `flwr_aid` set. not sure if it make sense to keep this validation here though, wdyt?

PR #5516, opened 2025-06-13
fix(framework) Fix the CLI output for `flwr ls/stop/log` when run ID is not found

panh99:
# Now
image

### DIFF
--- a/framework/py/flwr/cli/utils.py
+++ b/framework/py/flwr/cli/utils.py
@@ -28,7 +28,12 @@ import typer
 
 from flwr.cli.cli_user_auth_interceptor import CliUserAuthInterceptor
 from flwr.common.auth_plugin import CliAuthPlugin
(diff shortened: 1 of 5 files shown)


framework/py/flwr/cli/utils.py
```diff
@@ -323,4 +323,11 @@ def flwr_cli_grpc_exc_handler() -> Iterator[None]:
             # pylint: disable=E1101
             typer.secho(e.details(), fg=typer.colors.RED, bold=True)
             raise typer.Exit(code=1) from None
+        if e.code() == grpc.StatusCode.NOT_FOUND and e.details() == "Run ID not found":
```
danieljanes: @panh99 if we check for `"Run ID not found"`, can we put this into a constant to avoid a change in server response from breaking the CLI?

PR #5527, opened 2025-06-16
fix(framework) Fix bugs related to objects and node availability

panh99:
- The `InMemoryLinkState` now checks the existence of node ID before accessing it, avoiding raising `KeyError`
- SuperLink-generated error messages are now correctly stored in `ObjectStore`
- `ServerAppIoServicer.PullMessages` will not include message object ID unnecessarily in `object_ids_to_pull`

### DIFF
--- a/framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
+++ b/framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
@@ -25,7 +25,7 @@ from flwr.common import Message
 from flwr.common.constant import SUPERLINK_NODE_ID, Status
 from flwr.common.inflatable import (
     UnexpectedObjectContentError,
-    get_descendant_object_ids,
+    get_all_nested_objects,
     get_object_tree,
     no_object_id_recompute,
 )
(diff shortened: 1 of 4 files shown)


framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -209,14 +209,18 @@ def PullMessages(
...
-                store.preregister(request.run_id, get_object_tree(msg_res))
+                    # Store mapping
+                    store.set_message_descendant_ids(
+                        msg_object_id=message_obj_id, descendant_ids=descendants
+                    )
+                    # Preregister
+                    store.preregister(request.run_id, get_object_tree(msg_res))
+                    # Store objects
+                    for obj_id, obj in all_objects.items():
+                        store.put(obj_id, obj.deflate())
```
jafermarq, line 223: do we need to put this block inside the `no_object_id_recompute` context?

panh99: Yes, we need it here. `.deflate()` needs the object IDs of its children, so this context avoids recomputing children object IDs.

jafermarq: ah, right. 👍

framework/py/flwr/server/superlink/serverappio/serverappio_servicer_test.py
```diff
@@ -541,9 +542,7 @@ def test_pull_message_from_expired_message_error(self) -> None:
             assert msg_res.has_error()
             # objects_to_pull is expected to be {msg_obj_id: msg_obj_id}
```
jafermarq:
```suggestion
            # objects_to_pull is expected to be {msg_obj_id: []}
```

danieljanes (approved):

PR #5543, opened 2025-06-19
feat(framework) Mask the tokens used by `flwr-clientapp`

panh99:


framework/py/flwr/common/logger.py
```diff
@@ -425,3 +425,20 @@ def print_json_error(msg: str, e: Union[typer.Exit, Exception]) -> None:
             }
         )
     )
+
+
+def mask_string(value: str, head: int = 4, tail: int = 4) -> str:
```
danieljanes: @panh99 should we move this to `supercore`?

danieljanes: & add tests for it?

panh99: Sure thing!

panh99: I am less sure about the naming. Is it okay to name it as `utils.py` or that's too vague?

danieljanes (approved):

PR #5546, opened 2025-06-20
feat(framework) Introduce `SchedulerPlugin` ABC

panh99:
### `flwr-simulation` command
image

### `flwr-serverapp` command
image

### `flwr-clientapp` command
image

## Why `SchedulerPlugin.__init__` is designed like this?
To run an app process (e.g., `flwr-simulation`, `flwr-serverapp` or `flwr-clientapp`), the scheduler needs the following things:
- `flwr_dir`: The flower directory.
- `appio-address`: The address of the corresponding servicer
- `token`: The token based on which to pull the inputs
- other arguments that we may add in the future for scheduler auth and TLS

Among all above, only `token` is different for each app process. All other arguments, e.g., `flwr_dir`, remain unchanged across app processes. Thus we can pass all other arugments when instantiating the scheduler plugin, and only pass `token` to start an app process.

## Why `SchedulerPlugin.launch_app` also accepts a `run_id`?
This is not strictly neccessary. The command to launch an app process won't need this (it only needs `token`). However, it allows the `SchedulerPlugin` to know which run will be started, which might be useful. Wdyt?

### DIFF
--- a/framework/docs/source/ref-exit-codes/101.rst
+++ b/framework/docs/source/ref-exit-codes/101.rst
@@ -8,13 +8,15 @@ The license of the SuperLink is invalid, causing it to exit prematurely on start
 
 Possible causes of invalid license include:
 
-1. ``FLWR_LICENSE_KEY`` environment variable not set. This is required when starting the
-   SuperLink Enterprise version.
-2. ``FLWR_LICENSE_KEY`` environment variable set to an invalid license key.
+1. ``FLWR_LICENSE_KEY`` environment variable set to an invalid license key.
+2. The SuperLink failed to reach the public key endpoint from the license server.
+3. The fetched public key is malformed (wrong PEM encoding or key type) or its signature
+   failed verification against signing key.
 
 How to Resolve
 --------------
 
-1. Set the ``FLWR_LICENSE_KEY`` as an environment variable when starting the SuperLink.
-2. Ensure the license key is valid.
+1. Ensure that the ``FLWR_LICENSE_KEY`` is valid.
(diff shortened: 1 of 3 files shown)


framework/py/flwr/supercore/scheduler/plugin.py
```diff
@@ -0,0 +1,68 @@
...
+
+class SchedulerPlugin(ABC):
+    """Abstract base class for Scheduler plugins."""
+
+    def __init__(
+        self,
+        appio_api_address: str,
+        flwr_dir: str,
+        get_run: Callable[[int], Run],
+        get_fab: Callable[[str], Fab],
```
jafermarq: I'm thinking we'll need a new callable to fetch the `Context`. This is applicable to both `flwr-serverapp` and `flwr-clientapp`. Worth adding it here then?

panh99: May I know why the scheduler may need the context?

jafermarq: why does it need the Fab?

panh99: Maybe it wanna check the code in the FAB? Happy to remove this now and we can bring it back later when we need it.

panh99: As discussed, the `get_fab` method has been removed. I will resolve this comment.

jafermarq (approved):

danieljanes (approved):

PR #5555, opened 2025-07-01
fix(examples) Update example dependency

yan-gao-GY:
## Proposal

Update dependency of `pytorch-from-centralized-to-federated` example as follows:

```
torch==2.6.0
torchvision==0.21.0
```


jafermarq (approved):
🔥

PR #5576, opened 2025-07-10
refactor(framework) Simplify license check ABC

chongshenng:


danieljanes (dismissed):

panh99 (commented):
The PR looks good! Btw, may I know what's the motivation for having an abstract constructor with no arguments?

PR #5585, opened 2025-07-14
refactor(framework) Rename messages in `appio.proto` phasing out serverapp-side terminology

jafermarq:
We want to use the messages in the new `appio.proto` for both `ServerAppIo` and `ClientAppIo`. Because of this, it makes sense to ditch the _server_ term used in some of them or the `ins/res` terms used in others (since they convey message directionality)

### DIFF
--- a/framework/py/flwr/common/inflatable_utils.py
+++ b/framework/py/flwr/common/inflatable_utils.py
@@ -15,6 +15,7 @@
 """InflatableObject utilities."""
 
 import concurrent.futures
+import os
 import random
 import threading
 import time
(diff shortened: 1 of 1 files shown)


framework/proto/flwr/proto/appio.proto
```diff
@@ -21,37 +21,37 @@ import "flwr/proto/message.proto";
 import "flwr/proto/fab.proto";
 import "flwr/proto/run.proto";
 
-// PushMessages messages
-message PushInsMessagesRequest {
+// PushAppMessages messages (for both SeverApp and ClientApp)
```
panh99: Could we move the comment `(for both SeverApp and ClientApp)` somewhere so we don't need to repeat it for every request?

jafermarq: ok, i'll move it at the top of the file

PR #5586, opened 2025-07-14
feat(framework) Decouple message and `ObjecStore` by storing tree structures in pre-registration

panh99:
### Summary of Changes

* Added a new `ObjectStore.get_object_tree` method.
* Child object IDs are now stored for each object.
* Fixed a bug where the Fleet API server was incorrectly adding the message ID itself to the list of descendant object IDs.

### Motivation

This PR aims to **decouple `Message` from `ObjectStore`**. The `ObjectStore` should solely serve `InflatableObject`. Currently, it is still partially coupled to `Message` through methods like `get_message_descendant_ids`, which limits its ability to generalize to non-message objects. For example, it is not currently possible to store objects like `RecordDict` or `Context` in `ObjectStore`, because there is no generic way to retrieve an inflatable object along with all of its descendants.

Additionally, having separate methods like `set/get/delete_message_descendant_ids` unnecessarily complicates the code and introduces integrity risks. External code must always remember to call both `put` + `set_message_descendant_ids`, or `delete` + `delete_message_descendant_ids`, which is error-prone and non-atomic.

### Outcome

After this PR:

* `ObjectStore` manages child object relationships directly.
* We can safely remove `set/get/delete_message_descendant_ids`.
* `ObjectStore` and `Message` are fully decoupled.
* Facilitate moving objects from SuperLink to SuperNode without inflation.


framework/py/flwr/supercore/object_store/object_store_test.py
```diff
@@ -224,6 +224,26 @@ def test_preregister(self) -> None:
...
+        )
+        obj_tree = get_object_tree(obj)
+        object_store.preregister(self.run_id, get_object_tree(obj))
+
+        # Execute
+        retrieved_tree = object_store.get_object_tree(obj_tree.object_id)
+        retrieved_tree_traversed = [
+            node.object_id for node in iterate_object_tree(retrieved_tree)
+        ]
+        obj_tree_traversed = [node.object_id for node in iterate_object_tree(obj_tree)]
```
jafermarq, line 242: we only have this because there's no other way to compare the trees right?

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -119,11 +119,11 @@ def pull_messages(
...
 
             msg_object_id = msg.metadata.message_id
-            descendants = store.get_message_descendant_ids(msg_object_id)
-            # Include the object_id of the message itself
-            objects_to_pull[msg_object_id] = ObjectIDs(
-                object_ids=descendants + [msg_object_id]
-            )
+            obj_tree = store.get_object_tree(msg_object_id)
+            descendants = [node.object_id for node in iterate_object_tree(obj_tree)]
+            descendants = descendants[:-1]  # Exclude the message itself
```
jafermarq, line 124: I wonder if it's better to change the behaviour of `iterate_object_tree` and, by means of an arg, tell it to return the head node (the `message` itself) or not. We are using this `[:-1]` quite often and seems error prone

framework/py/flwr/supercore/object_store/object_store_test.py
panh99: Yeah, exactly. This is the easiest way to compare trees, i.e., via post-order traversal

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
panh99: Yeah, I'd totally agree. But I think this is fine for now as we will change the type from object ID list to object tree very soon, and thus it seems no need to update it now.

jafermarq (approved):

danieljanes (approved):

PR #5590, opened 2025-07-15
refactor(framework) Use `ObjectTree` instead of `ObjectIDs` in the pull message response 

panh99:


framework/py/flwr/client/grpc_rere_client/connection.py
```diff
@@ -273,8 +274,9 @@ def receive() -> Optional[Message]:
         if message_proto:
             msg_id = message_proto.metadata.message_id
             run_id = message_proto.metadata.run_id
+            object_tree = response.message_object_trees[0]
             all_object_contents = pull_objects(
-                list(response.objects_to_pull[msg_id].object_ids) + [msg_id],
+                [tree.object_id for tree in iterate_object_tree(object_tree)],
```
jafermarq, line 279:
```suggestion
                [node.object_id for node in iterate_object_tree(object_tree)],
```

I suppose here it's more appropriate to use the term `node` (of a tree) ?

jafermarq: If you agree, let's change it also for the `REST` connection

panh99: Yeah, I originally use `node`, but I feel like ppl may get confused with our node, e.g., `node_id`, SuperNode etc. So I use `tree` here, as the yield type is `ObjectTree`. Wdyt?

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer_test.py
```diff
@@ -281,16 +282,17 @@ def test_pull_messages_if_running(
...
-
         if register_in_store:
+            object_tree = response.message_object_trees[0]
+            object_ids_in_response = [
+                tree.object_id for tree in iterate_object_tree(object_tree)
+            ]
             # Assert expected object_ids
             assert set(obj_ids_registered) == set(object_ids_in_response)
-            assert message_ins.object_id == object_ids_in_response[0]
+            assert message_ins.object_id == object_tree.object_id
```
jafermarq, line 276: I think adding a comment at the top here will be helpful for the future (briefly mention that the headnode is the message itself?)

panh99: yeah, that makes sense to me!

PR #5601, opened 2025-07-17
refactor(framework) Handle `grpc.StatusCode.UNAVAILABLE` in `FlowerCLI`

jafermarq:
When using the `FlowerCLI` if the connection couldn't be stablished (eg. in `flwr run`) or is lost (e.g. while streaming logs) users get a long error that might not be that useful on how to solve it. The new error message aims to guide the user on how to fix it: (1) check connection; and often more importantly (2) if `ExecAPI` address is correct.

previous error:
Screenshot 2025-07-17 at 18 10 15

previous error (while streaming logs)
Screenshot 2025-07-17 at 18 26 40

New error message:
Screenshot 2025-07-17 at 18 10 52


panh99 (approved):
LGTM!

danieljanes (approved):

PR #5613, opened 2025-07-22
fix(framework) Wait for content before pushing

jafermarq:


panh99 (dismissed):
LGTM!

framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -386,7 +386,9 @@ def _push_messages(
         # This will yield (object_id, content) pairs
         def yield_object_contents(_obj_tree: ObjectTree) -> Iterator[tuple[str, bytes]]:
             for tree in iterate_object_tree(_obj_tree):
-                while (content := object_store.get(tree.object_id)) is None:
+                while (
+                    content := object_store.get(tree.object_id)
+                ) == b"" or content is None:
```
panh99: This is not suppose to happen, right? We get the object tree from the object store.

panh99: Should we raise a `RuntimeError` when the content is `None`?

PR #5619, opened 2025-07-23
feat(framework) Add `SimpleClientAppSchedulerPlugin`

panh99:
Based on #5546


framework/py/flwr/supernode/scheduler/simple_clientapp_scheduler_plugin.py
```diff
@@ -0,0 +1,49 @@
...
+            return None
+        return candidate_run_ids[0]
+
+    def launch_app(self, token: str, run_id: int) -> None:
+        """Launch the application associated with a given run ID and token."""
+        cmds = ["flwr-clientapp", "--insecure"]
+        cmds += ["--clientappio-api-address", self.appio_api_address]
+        cmds += ["--token", token]
+        cmds += ["--parent-pid", str(os.getpid())]
+        cmds += ["--flwr-dir", self.flwr_dir]
```
jafermarq, line 44: do we want to set `--run-once`?

panh99: Good question! We don't need it here as `flwr-clientapp` will only run once if `--token` is set.

panh99: Actually plan to remove `--run-once` flag once we have `flwr-scheduler` ready, but that can be a breaking change if anyone manually run `flwr-clientapp`.

jafermarq (dismissed):

danieljanes (approved):

PR #5623, opened 2025-07-23
refactor(framework) Handle `ObjectUnavailableError` in `GrpcGrid`

jafermarq:
We need to capture the `ObjectUnavailableError` that is raised when pulling objects hits the `PULL_MAX_TRIES_PER_OBJECT` or `PULL_MAX_TIME` limit. When this happens, the `GrpcGrid` catches the exception and creates a `Message` carrying an error. It then gets returned to the `ServerApp`.


framework/py/flwr/server/grid/grpc_grid.py
```diff
@@ -307,16 +311,42 @@ def pull_messages(self, message_ids: Iterable[str]) -> Iterable[Message]:
...
+                        object_ids=[
+                            tree.object_id for tree in iterate_object_tree(msg_tree)
+                        ],
+                        pull_object_fn=make_pull_object_fn_protobuf(
+                            pull_object_protobuf=self._stub.PullObject,
+                            node=self.node,
+                            run_id=run_id,
+                        ),
+                    )
+                except ObjectUnavailableError as e:
```
panh99, line 325: Should we capture the `ObjectIdNotPreregisteredError` exception, or let it escalate?

framework/py/flwr/server/grid/grpc_grid_test.py
```diff
@@ -304,4 +315,63 @@ def test_simple_retry_mechanism_get_nodes(self) -> None:
...
+        response = Mock(object_found=True, object_available=False, object_content=None)
+        self.mock_stub.PullObject.return_value = response
+
+        # Execute
+        with patcher:
+            # Depending on the patch context, this will either hit the timeout or
+            # the limit of pulling attempts for a given object
+            msgs = list(self.grid.pull_messages([ins1.object_id]))
+
+        # Assert if msgs doesn't contain a single error message
```
panh99:
```suggestion
        # Assert: `msgs` should contain a single error message
```

framework/py/flwr/server/grid/grpc_grid.py
jafermarq: better to capture in a separate PR?

panh99 (approved):
LGTM!

danieljanes (approved):

PR #5624, opened 2025-07-23
refactor(framework) Pack all `Messages` into a single request

jafermarq:
So far the `GrpcGrid` has been pushing one `Message`  (followed by all its objects) at a time. This approach delays the pulling of _shared_ objects in the `Messages` sent to the `SuperNodes` (this is a common patter in FL and distrib AI when there is a _global_ model) until the _shared_ objects form the first `Message` are pushed.

This PR ensures all `Messages` are first registered in the `LinkState` (which enables all `SuperNodes` involved to start pulling them and their objects) and then, commence the pushing of the `Objects`. 

**This effectively masks the comms/time costs of pushing `SeverApp --> SuperLink` with the `SuperLink --> SuperNodes` pulling.**

### DIFF
--- a/framework/docs/source/build.rst
+++ b/framework/docs/source/build.rst
@@ -24,5 +24,7 @@ Problem-oriented how-to guides show step-by-step how to achieve a specific goal.
     how-to-use-differential-privacy
     how-to-implement-fedbn
     how-to-use-cli-json-output
+    how-to-authenticate-users
+    how-to-configure-audit-logging
     how-to-upgrade-to-flower-1.0
     how-to-upgrade-to-flower-1.13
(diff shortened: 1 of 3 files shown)


framework/py/flwr/server/grid/grpc_grid.py
```diff
@@ -218,37 +218,41 @@ def get_node_ids(self) -> Iterable[int]:
...
-                message_object_trees=[object_tree],
+                message_object_trees=object_trees,
             )
         )
 
         # Push objects
         # If Message was added to the LinkState correctly
-        if msg_id is not None:
+        for msg_id in res.message_ids:
             obj_ids_to_push = set(res.objects_to_push[msg_id].object_ids)
```
panh99: We should also keep track of objects that have just been pushed, right? So that we won't push repeated objects when pushing objects of the following messages.

panh99 (approved):
LGTM!

PR #5628, opened 2025-07-24
feat(framework) Make SuperNode wait for the reply message

panh99:


framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -237,16 +237,20 @@ def start_client_internal(
                 ]
                 subprocess.run(command, check=False)
 
+            # This is to ensure that only one message is processed at a time
+            # The `_push_messages` function will wait until a reply message is available
```
jafermarq:
```suggestion
            # No message has been pulled therefore we can skip the push stage.
```

I'm thinking a comment like the one suggested is more representative of what this if statement is doing.

danieljanes (approved):

PR #5648, opened 2025-07-29
feat(framework) Ensure `Context` cleanup after executing the run in `flwr-serverapp` and `flwr-simulation`

panh99:


framework/py/flwr/simulation/app.py
```diff
@@ -271,6 +272,13 @@ def run_simulation_process(  # pylint: disable=R0914, disable=W0212, disable=R09
                     )
                 )
 
+            # Clean up the Context if it exists
+            try:
+                del updated_context
+            except NameError:
+                pass
+            gc.collect()
+
```
jafermarq, line 281: Should we better run this if a run was actually executed? i.e. moving these lines inside the `if run_status` statement immediately above?

panh99: Do you see any downside to call GC at the end here? We can do a cleanup for unsuccessful execution as well.

jafermarq: no apparent downside. Feel free to resolve

jafermarq (approved):

PR #5707, opened 2025-08-12
break(framework) Deprecate long running `flwr-clientapp`

panh99:


framework/py/flwr/supercore/superexec/run_superexec.py
```diff
@@ -120,3 +126,44 @@ def get_run(run_id: int) -> Run:
...
+    parent_pid: Optional[int],
+) -> None:
+    """Log a deprecation warning and run the equivalent `flower-superexec` command.
+
+    Used for legacy long-running `flwr-*` commands (i.e., without `--token`) that will
+    be removed in favor of `flower-superexec`.
+    """
+    log(
+        WARN,
+        "Running `%s` as a long-running process without `--token` is "
```
jafermarq:
```suggestion
        "Directly executing `%s` is "
```

I wonder if we should skip mentioning the "without `--token`". with the recent change a user won't know what token to pass right? If so, should we make the change as above?

panh99: Yeah, good point. You are right that this may make users think they need to provide `--token`.

framework/py/flwr/supernode/cli/flwr_clientapp.py
```diff
@@ -35,6 +38,19 @@ def flwr_clientapp() -> None:
...
+            plugin_type=ExecPluginType.CLIENT_APP,
+            plugin_class=ClientAppExecPlugin,
+            stub_class=ClientAppIoStub,
+            appio_api_address=args.clientappio_api_address,
+            flwr_dir=args.flwr_dir,
+            parent_pid=args.parent_pid,
+        )
+        return
+
     log(INFO, "Start `flwr-clientapp` process")
```
jafermarq, line 54: should we remove this one also ?

panh99: Do you mean to remove this log? Why?

jafermarq: the `log(INFO, "Start `flwr-clientapp` process")` given that you are removing the ones for `flwr-simulation` and `flwr-serverapp`

jafermarq: nvrm, they are moved. my bad

panh99: Oh, yeah. I moved them so users won't see the logs if they ran it manually.

framework/py/flwr/server/superlink/fleet/vce/vce_api.py
```diff
@@ -180,6 +180,7 @@ def run_api(
     """Run the VCE."""
     messageins_queue: Queue[Message] = Queue()
     messageres_queue: Queue[Message] = Queue()
+    backend = None
```
panh99:
```suggestion
```

PR #5710, opened 2025-08-13
feat(framework): Introduce first `Message`-based strategy

jafermarq:


examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,220 @@
...
+        # To perform averaging
+        # TODO: implement as dict so each entry in final recorddict
+        # TODO: is averaged independently
+        counter_ar = 0
+        counter_metrics = 0
+
+        for msg in results[1:]:
+            for name, record in msg.content.items():
+                if isinstance(record, ArrayRecord):
+                    counter_ar += 1
```
panh99: I believe `FedAvg` uses the number of samples as weights, right?

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,220 @@
...
+                    for arr_name, array in record.items():
+                        # TODO: optimize (currently low memory usage, but repeated serde)
+                        rd[name][arr_name] = Array(rd[name][arr_name].numpy() + array.numpy())
+                    
+                elif isinstance(record, MetricRecord):
+                    counter_metrics += 1
```
panh99: Same here. We should use the number of examples, not equal weights.

examples/app-pytorch/app_pytorch/fedavg.py
jafermarq: yes, there is a comment about that. this is a poc not a complete FedAvg re-implementation

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,220 @@
...
+        return rd
+
+    def evaluate(
+        self, server_round: int, record: RecordDict
+    ) -> RecordDict:
+        # Evaluate the current model parameters
+        return {}
+    
+
+def run_strategy(record_dict: RecordDict, strategy: Strategy, grid: Grid, num_rounds: int, timeout: float):
```
panh99: I think it would be better to make this a method, e.g., `Strategy.run`, so that users can override it if needed. This also makes it clearer that they can customize their federated learning flow using the Message API.

examples/app-pytorch/app_pytorch/fedavg.py
panh99: I see, thanks. I will resolve this then.

examples/app-pytorch/app_pytorch/strategy.py
```diff
@@ -0,0 +1,50 @@
...
+    """Abstract base class for server strategy implementations."""
+
+    @abstractmethod
+    def initialize_parameters(
+        self, grid: Grid
+    ) -> Optional[ArrayRecord]:
+        """Initialize the (global) model parameters."""
+
+    @abstractmethod
+    def configure_fit(
```
danieljanes:
```suggestion
    def configure_train(
```

examples/app-pytorch/app_pytorch/strategy.py
```diff
@@ -0,0 +1,50 @@
...
+        self, server_round: int, record: RecordDict, grid: Grid
+    ) -> list[Message]:
+        """Configure the next round of training."""
+
+    @abstractmethod
+    def aggregate_fit(
```
danieljanes:
```suggestion
    def aggregate_train(
```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,220 @@
...
+from flwr.common import Message, ArrayRecord, RecordDict, ConfigRecord, log, MessageType, Array, MetricRecord
+from .strategy import Strategy
+
+
+class FedAvg(Strategy):
+    def initialize_parameters(self, grid: Grid) -> Optional[ArrayRecord]:
+        # Initialize global model parameters
+        return None
+
+    def configure_fit(
```
danieljanes:
```suggestion
    def configure_train(
```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,220 @@
...
+        self, server_round: int, record: RecordDict, grid: Grid
+    ) -> list[Message]:
+        # Configure the next round of training
+
+        record['fit-config'] = ConfigRecord({"server_round": server_round})
```
danieljanes:
```suggestion
        record['train-config'] = ConfigRecord({"server_round": server_round})
```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,220 @@
...
+                content=record,
+                message_type=MessageType.TRAIN,  # target method in ClientApp
+                dst_node_id=node_id,
+                group_id=str(server_round),
+            )
+            messages.append(message)
+
+        return messages
+
+    def aggregate_fit(
```
danieljanes:
```suggestion
    def aggregate_train(
```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,317 @@
...
+                        )
+
+            # For MetricRecord
+            elif isinstance(record_item, MetricRecord):
+                if name not in aggregated:
+                    aggregated[name] = MetricRecord()
+                # aggregate in-place
+                for key, value in record_item.items():
+                    if key == weighting_metric_name:
+                        continue  # ! we don't want to keep this key as part of the aggregated recorddict, do we?
```
panh99: No, we don't.

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,317 @@
...
+            if isinstance(rec, MetricRecord):
+                if weighting_metric_name not in rec:
+                    #! alert user expected key is not found ?
+                    # default to homogeneous weighting (i.e. 1.0
+                    # for all model contributions)
+                    total_examples += 1.0
+                    examples_per_record.append(1.0)
+                else:
+                    total_examples += rec[weighting_metric_name]
+                    examples_per_record.append(rec[weighting_metric_name])
```
panh99:
```suggestion
                if weighting_metric_name not in rec:
                    #! alert user expected key is not found ?
                    # default to homogeneous weighting (i.e. 1.0
                    # for all model contributions)
                weight = rec.get(weighting_metric_name, 1)
                total_examples += weight
                examples_per_record.append(weight)
```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,317 @@
...
+class FedAvg:
+
+    def __init__(
+        self,
+        fraction_train: float = 1.0,
+        fraction_evaluate: float = 1.0,
+        min_train_clients: int = 2,
+        min_evaluate_nodes: int = 2,
+        min_available_nodes: int = 2,
+        weighting_key: str = "num-examples",
```
panh99: Could we make it a `Callable[[RecordDict], float]`? The issue with a single string here is that we are kind of assuming that the `RecordDict` must only have one `MetricRecord`, or this may go wrong, right?

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,317 @@
...
+                f"Waiting for nodes to connect. Connected {len(all_node_ids)} but need at least {min_wait_for} nodes.",
+            )
+            sleep(3)
+            all_node_ids = list(grid.get_node_ids())
+
+        # Sample
+        num_to_sample = int(
+            len(all_node_ids)
+            * (self.fraction_train if is_train else self.fraction_evaluate)
+        )
```
panh99: I am a bit confused here. Does it mean that, with `min_train_clients=10, fraction_train=0.5`, it may sample 5 clients to train, not 10?

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,317 @@
...
+def aggregate(records: list[RecordDict], weighting_metric_name: str) -> RecordDict:
+    # Aggregate the given records using the specified weighting metric
+    aggregated = RecordDict()
+
+    # Count total examples
+    total_examples = 0
+    examples_per_record = []
+    for record in records:
+        for rec in record.values():
+            if isinstance(rec, MetricRecord):
```
panh99: Does this assume that the `RecordDict` must have and can only have one `MetricRecord`?

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,317 @@
...
+                record_dict.array_records[key] = record
+            # Log training metrics and append to history
+            log(INFO, "Federated Training results: %s", res.metric_records)
+            metrics_history[current_round]["train"] = res.metric_records
+
+            # Configure evaluate
+            node_ids = self.sample_nodes(grid, is_train=False)
+            messages = self.configure_evaluate(current_round, record_dict, node_ids)
+            # Communicate
+            replies = grid.send_and_receive(messages, timeout=timeout)
```
panh99: Better not to keep `messages` local variable here so that GC can release them.

examples/app-pytorch/app_pytorch/fedavg.py
jafermarq: I see what you mean. yeah this is a flaw in the current code.

examples/app-pytorch/app_pytorch/fedavg.py
jafermarq: it's a bug. i'll fix

examples/app-pytorch/app_pytorch/fedavg.py
jafermarq: I fixed this. Just note that the exact code inside each method isn't the main purpose of this PR. It's more about the overall ux. For example, more important is whether we want to pass a `RecorDict` to `<strategy>.run()` defining  the arrayrecords and configs, or if we'd want to do that for example in two separate records

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,381 @@
...
+    return aggregated_arrays, aggregated_metrics
+
+
+class FedAvg:
+
+    def __init__(
+        self,
+        fraction_train: float = 1.0,
+        fraction_evaluate: float = 1.0,
+        min_train_clients: int = 2,
```
yan-gao-GY:
```suggestion
        min_train_nodes: int = 2,
```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,390 @@
...
+                            ).tolist()
+                        else:
+                            aggregated_metrics[key] += value * weight / total_examples
+
+    return aggregated_arrays, aggregated_metrics
+
+
+def sample_nodes(
+    grid: Grid, min_sample_nodes: int, fraction_sample: float
+) -> tuple[list[int], list[int]]:
```
panh99:
```suggestion
def sample_nodes(
    grid: Grid, min_sample_nodes: int, fraction_sample: float
) -> list[int]:
```

This function looks a bit strange to me. Maybe we could make it more focused and intuitive, like returning the sampled node IDs? If users do need all node IDs, they can simply call `grid.get_node_ids()`.

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,390 @@
...
+class FedAvg:
+
+    def __init__(
+        self,
+        fraction_train: float = 1.0,
+        fraction_evaluate: float = 1.0,
+        min_train_nodes: int = 2,
+        min_evaluate_nodes: int = 2,
+        min_available_nodes: int = 2,
+        weighting_factor_key: str = "num-examples",
```
panh99:
```suggestion
        weight_factor_key: str = "num-examples",
```

Maybe this? According to ChatGPT.

examples/app-pytorch/app_pytorch/fedavg.py
jafermarq: sure, i only did it thinking about the logging happening where this function is called (the `configure_xyz` methods)

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,390 @@
...
+            )  #! Note i'm not using group_id
+            messages.append(message)
+        return messages
+
+    def configure_train(
+        self, server_round: int, arrays: ArrayRecord, config: ConfigRecord, grid: Grid
+    ) -> list[Message]:
+
+        # Sample nodes
+        node_ids, num_total = sample_nodes(
```
panh99: Looks nice!

examples/app-pytorch/app_pytorch/fedavg.py
panh99: Oh, I see wym. That's a good point, let me think.

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,390 @@
...
+
+        log(
+            INFO,
+            "aggregate_train: received %s results and %s failures",
+            len(replies) - num_errors,
+            num_errors,
+        )
+
+        replies_with_content = [msg.content for msg in replies if msg.has_content()]
+        # Check if replies have one `ArrayRecord` and at most one `MetricRecord`
```
panh99: I think we should log a warning when a message contain `ConfigRecord`, right? Plus, we also need to check whether the record keys of all messages are identical, which is the pre-requisite for aggregation.

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,390 @@
...
+    def launch(
+        self,
+        arrays: ArrayRecord,  #! Now compulsory (those that don't what a mdoel, they can pass empyt ArrayRecord)
+        grid: Grid,
+        num_rounds: int,
+        timeout: float,
+        train_config: Optional[ConfigRecord] = ConfigRecord(),
+        evaluate_config: Optional[ConfigRecord] = ConfigRecord(),
+        central_eval_fn: Callable[[int, RecordDict], MetricRecord] = None,
+    ) -> MetricRecord:
```
panh99: We will use a dataclass like `StrategyResult`, right?

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,390 @@
...
+        # log name of strategy
+        log(INFO, f"Starting {self.__class__.__name__} strategy.")
+        log(INFO, f"\t> Number of rounds: {num_rounds}")
+        log(
+            INFO,
+            f"\t> ArrayRecord: {len(arrays)} Arrays totalling {sum(len(array.data) for array in arrays.values())/(1024**2):.2f} MB",
+        )
+        log(
+            INFO,
+            f"\t> ConfigRecord for train round: {train_config if train_config else '⚠️ (empty!)'}",
```
panh99: If possible, it's better to avoid emojis as they might not be rendered in all shells.

examples/app-pytorch/app_pytorch/fedavg.py
jafermarq: I'm inclined to agree with both. Let's leave this point unresolved

examples/app-pytorch/app_pytorch/server_app.py
```diff
@@ -1,145 +1,85 @@
 """app-pytorch: A Flower / PyTorch app."""
 
-import random
-from logging import INFO, WARN
-from time import sleep
+from logging import INFO
```
yan-gao-GY:
```suggestion

```

examples/app-pytorch/app_pytorch/server_app.py
```diff
@@ -1,145 +1,85 @@
...
-    ArrayRecord,
-    Context,
-    Message,
-    MessageType,
-    RecordDict,
-    ConfigRecord,
-)
+from app_pytorch.task import Net, test
+from flwr.common import ArrayRecord, ConfigRecord, Context, RecordDict
 from flwr.common.logger import log
```
yan-gao-GY:
```suggestion

```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,387 @@
...
+    return aggregated_arrays, aggregated_metrics
+
+
+def sample_nodes(
+    grid: Grid, min_available_nodes: int, sample_size: int
+) -> tuple[list[int], list[int]]:
+
+    sampled_nodes = []
+
+    # wait for min_available_nodes to be online
```
yan-gao-GY:
```suggestion
    # Wait for min_available_nodes to be online
```

examples/app-pytorch/app_pytorch/fedavg.py
```diff
@@ -0,0 +1,387 @@
...
+        log(
+            INFO,
+            f"\t> ConfigRecord for evaluate round: {evaluate_config if evaluate_config else '(empty!)'}",
+        )
+        log(INFO, "")
+
+        metrics_history = ReturnStrategyResults()
+
+        t_start = time()
+        # do central eval with starting global parameters
```
yan-gao-GY:
```suggestion
        # Do central eval with starting global parameters
```

framework/py/flwr/serverapp/strategy_utils.py
```diff
@@ -0,0 +1,142 @@
...
+from typing import cast
+
+import numpy as np
+
+from flwr.common import Array, ArrayRecord, MetricRecord, NDArray, RecordDict, log
+from flwr.server import Grid
+
+
+@dataclass
+class ReturnStrategyResults:
```
danieljanes, line 32:
```suggestion
class StrategyResults:
```

framework/py/flwr/serverapp/strategy_utils_tests.py
```diff
@@ -0,0 +1,277 @@
...
+        ),  # Compliant
+        (
+            True,
+            RecordDict(
+                {
+                    "global-model": ArrayRecord([np.random.randn(7, 3)]),
+                    "metrics": MetricRecord({"weight": [0.123]}),
+                }
+            ),
+        ),  # No weighting key is not a scalar (BAD)
```
jafermarq:
```suggestion
        ),  # Weighting key is not a scalar (BAD)
```

examples/app-pytorch/app_pytorch/server_app.py
```diff
@@ -1,145 +1,90 @@
...
+        train_config=ConfigRecord({"lr": 0.01}),
+        num_rounds=num_rounds,
+        timeout=3600,
+        central_eval_fn=central_evaluation,
+    )
+
+    # Log resulting metrics
+    print("\nTrain metrics:")
+    pprint(strategy_results.train_metrics)
+    print("\nEvaluate metrics:")
```
yan-gao-GY:
```suggestion
    print("\nDistributed evaluate metrics:")
```

examples/app-pytorch/app_pytorch/server_app.py
```diff
@@ -1,145 +1,90 @@
...
+    def apply_transforms(batch):
+        """Apply transforms to the partition from FederatedDataset."""
+        batch["img"] = [pytorch_transforms(img) for img in batch["img"]]
+        return batch
+
+    dataset = test_dataset.with_format("torch", device=device).with_transform(
+        apply_transforms
+    )
+
+    test_dataloader = DataLoader(dataset, batch_size=32)
```
danieljanes: Can we change `task.load_dataset` to return the `DataLoader` created here?

examples/app-pytorch/app_pytorch/server_app.py
```diff
@@ -1,145 +1,90 @@
...
+    print("\nCentral evaluate metrics:")
+    pprint(strategy_results.central_evaluate_metrics)
+
+    # Save final model to disk
+    state_dict = strategy_results.arrays.to_torch_state_dict()
+    print("\nSaving final model to disk.")
+    torch.save(state_dict, "final_model.pt")
+
+
+def central_evaluation(server_round, array_record: ArrayRecord) -> MetricRecord:
```
danieljanes:
```suggestion
def central_evaluation(server_round: int, array_record: ArrayRecord) -> MetricRecord:
```

framework/py/flwr/serverapp/strategy_utils.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+@dataclass
+class StrategyResults:
+    """Data class carrying records generated during the execution of a strategy."""
+
+    arrays: ArrayRecord = field(default_factory=ArrayRecord)
+    train_metrics: dict[int, MetricRecord] = field(default_factory=dict)
+    evaluate_metrics: dict[int, MetricRecord] = field(default_factory=dict)
+    central_evaluate_metrics: dict[int, MetricRecord] = field(default_factory=dict)
```
danieljanes: Let's move this into a separate module and perhaps also rename it to `Result`

examples/app-pytorch/app_pytorch/server_app.py
jafermarq: moved to a func in `task.py`

PR #5719, opened 2025-08-14
fix(framework): Fix test for `ClientAppIoServicer`

jafermarq:
Update the `side_effect `from returning objects in a fixed order into a behaviour that returns the object specified in the request. This fix is needed to bring support for Python 3.13

close after merge: https://github.com/adap/flower/pull/5718


panh99 (approved):
LGTM!

PR #5725, opened 2025-08-15
refactor(framework): Remove `while` loop in `flwr-clientapp`

panh99:


framework/py/flwr/common/args.py
```diff
@@ -31,7 +31,7 @@ def add_args_flwr_app_common(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         "--token",
         type=str,
-        required=False,
+        required=True,
```
jafermarq: I think we shouldn't make it required for now as this will break how most people run in `--isolation=process`. In https://github.com/adap/flower/pull/5707 we are allowing executing `flwr-clientapp` (also `flwr-serverapp` and `flwr-simulation`) w/o a token but we show a log that this is deprecated and internally call `flower-superexec --plugin <...>`

framework/py/flwr/supernode/cli/flwr_clientapp.py
```diff
@@ -65,11 +64,5 @@ def _parse_args_run_flwr_clientapp() -> argparse.ArgumentParser:
         help="Address of SuperNode's ClientAppIo API (IPv4, IPv6, or a domain name)."
         f"By default, it is set to {CLIENTAPPIO_API_DEFAULT_CLIENT_ADDRESS}.",
     )
-    parser.add_argument(
-        "--run-once",
-        action="store_true",
-        help="When set, this process will start a single ClientApp for a pending "
-        "message. If there is no pending message, the process will exit.",
-    )
```
jafermarq, line 89: I wonder if instead of removing we better: (1) move it to `add_args_flwr_app_common` (since this `--run-once` is present in `flwr-serverapp` and `flwr-simulation`), (2) we show a deprecation warning when set; (3) we don't do anything even if set. 

In this way we: (1) don't break existing user code; (2) still can remove the `--run-once` logic in the `while` loops

framework/py/flwr/common/args.py
panh99: Oh, right, right, I will fixed it.

framework/py/flwr/supernode/cli/flwr_clientapp.py
panh99: I think it's fine to remove `--run-once`. It is not used in deployment. It's a thing that only used by the SuperLink/SuperNode in the subprocess mode. I will also remove `--run-once` of the `flwr-serverapp` and `flwr-simulation`.

framework/py/flwr/common/args.py
panh99:
```suggestion
        required=False,
```

framework/py/flwr/supernode/cli/flwr_clientapp.py
jafermarq: It's true we typically don't use it outside _subprocess_ mode but still, some people out there might be counting on it ? Technically speaking this is a breaking change 🤔

panh99: Makes sense. I will then bring back the `--run-once`.

framework/py/flwr/supernode/cli/flwr_clientapp.py
```diff
@@ -45,7 +45,6 @@ def flwr_clientapp() -> None:
     )
     run_clientapp(
```
jafermarq, line 63: should we throw a deprecation warning if `args.run_once` is set?

framework/py/flwr/supernode/cli/flwr_clientapp.py
```diff
@@ -68,8 +67,7 @@ def _parse_args_run_flwr_clientapp() -> argparse.ArgumentParser:
     parser.add_argument(
         "--run-once",
         action="store_true",
-        help="When set, this process will start a single ClientApp for a pending "
-        "message. If there is no pending message, the process will exit.",
+        help="This flag is deprecated and will be removed in a future release.",
```
jafermarq:
```suggestion
        help="DEPRECATED. When set, this process will start a single ClientApp for "
        "a pending message. If there is no pending message, the process will exit.",
```

better to keep the original `help` message but with a clear `DEPRECATED` ?

framework/py/flwr/common/args.py
```diff
@@ -60,6 +60,11 @@ def add_args_flwr_app_common(parser: argparse.ArgumentParser) -> None:
         help="The PID of the parent process. When set, the process will terminate "
         "when the parent process exits.",
     )
+    parser.add_argument(
+        "--run-once",
+        action="store_true",
+        help="This flag is deprecated and will be removed in a future release.",
```
jafermarq, line 66:
```suggestion
        help="DEPRECATED. When set, this process will start a single ClientApp for "
        "a pending message. If there is no pending message, the process will exit.",
```

better to keep the original `help` message but with a clear `DEPRECATED` ?

panh99: Hmmm, wdyt? We normally just mark it as deprecated it won't work anyway.

framework/py/flwr/supernode/cli/flwr_clientapp.py
panh99: Not sure if that's needed. If anyone manually start `flwr-*`, it will always log a warning and run SuperExec instead. Do you mean that we should log a warning saying SuperExec won't be able to accept `--run-once` flag?

jafermarq: yeah, just warning something like `--run-once was set, but this aregument is now deprecated and ignored`

framework/py/flwr/common/args.py
jafermarq: works for me!

framework/py/flwr/supernode/cli/flwr_clientapp.py
panh99: I will add the warning after #5707 is merged to avoid editing the same code in two PRs. Thanks

danieljanes (approved):

PR #5745, opened 2025-08-20
refactor(framework): Remove executors

panh99:


framework/py/flwr/server/app.py
```diff
@@ -198,26 +197,19 @@ def run_superlink() -> None:
     objectstore_factory = ObjectStoreFactory()
 
     # Start Control API
-    executor = load_executor(args)
+    is_simulation = args.executor == "flwr.superexec.simulation:executor"
```
jafermarq, line 200: I'm thinking we can change the `--executor` arg to be of type choice (w/ default `deployment` and the other option being `simulation`). wdyt? Note that if one launches the superlink as:

```shell
flower-superlink ... --executor somethingrandom
```
the Simulation Runtime will start

panh99: I will remove this very soon in a follow-up PR, and thus I think we don't need to work on it. Wdyt?

jafermarq (approved):

danieljanes (approved):

PR #5749, opened 2025-08-20
break(framework): Deprecate `--executor*` arguments and introduce `--simulation` flag for `flower-superlink`

panh99:
### Merge after #5745

### DIFF
--- a/framework/py/flwr/server/app.py
+++ b/framework/py/flwr/server/app.py
@@ -135,6 +135,15 @@ def run_superlink() -> None:
             WARN, "The `--flwr-dir` option is currently not in use and will be ignored."
         )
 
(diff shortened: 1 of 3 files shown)


framework/py/flwr/server/app.py
```diff
@@ -197,22 +206,21 @@ def run_superlink() -> None:
     objectstore_factory = ObjectStoreFactory()
 
     # Start Control API
-    is_simulation = args.executor == "flwr.superexec.simulation:executor"
```
danieljanes, line 200: can we just do 

```
is_simulation = args.simulation
```

here and undo the changes below?

panh99: Yes, we can.

PR #5768, opened 2025-08-26
refactor(framework): Improve `Strategy.start`

danieljanes:


examples/app-pytorch/app_pytorch/server_app.py
```diff
@@ -36,24 +36,24 @@ def main(grid: Grid, context: Context) -> None:
...
     )
 
     # Log resulting metrics
     print("\nTrain metrics:")
     pprint(result.train_metrics)
     print("\nDistributed evaluate metrics:")
     pprint(result.evaluate_metrics)
-    print("\nCentral evaluate metrics:")
-    pprint(result.central_evaluate_metrics)
+    print("\Global evaluate metrics:")
```
jafermarq:
```suggestion
    print("\nGlobal evaluate metrics:")
```

framework/py/flwr/serverapp/strategy.py
```diff
@@ -162,88 +162,121 @@ def start(
...
+            metrics and global evaluation metrics (if provided) from all rounds.
         """
         log(INFO, "Starting %s strategy:", self.__class__.__name__)
         log_strategy_start_info(
             num_rounds, initial_arrays, train_config, evaluate_config
         )
-        self.summary()
-        log(INFO, "")
+        summary = self.summary()
+        log(INFO, "%s\n", summary)
```
jafermarq: the `summary` method logs and returns none. I think it's better to keep it as it was

framework/py/flwr/serverapp/strategy.py
```diff
@@ -162,88 +162,121 @@ def start(
...
-            replies = grid.send_and_receive(
-                messages=self.configure_train(
-                    current_round, arrays, train_config, grid
-                ),
+            # -----------------------------------------------------------------
+            # --- TRAINING ----------------------------------------------------
+            # -----------------------------------------------------------------
+
+            # Call strategy to configure training round
+            train_messages = self.configure_train(
```
jafermarq: In the previous design we purposely made the `self.config_train(...)` call directly as an input arg to `send_and_receive()` so to avoid having 2xN models in memory (N sent, then N received).

framework/py/flwr/serverapp/strategy.py
```diff
@@ -162,88 +162,121 @@ def start(
...
+            # Aggregate evaluate
+            agg_evaluate_metrics = self.aggregate_evaluate(
+                current_round,
+                list(evaluate_replies),
+            )
+            del evaluate_replies
+
+            # Log training metrics and append to history
+            if agg_evaluate_metrics is None:
+                break
```
jafermarq: Do we want to end FL if there are no aggregation metrics? (this would be the case also if the run is configure to not sample nodes for evaluation). One idea is to not `break` and replace the above with:

```python
            if agg_metrics is not None:
                log(INFO, "\t└──> Aggregated MetricRecord: %s", agg_metrics)
                results.evaluate_metrics[current_round] = agg_metrics
```

jafermarq: a similar logic could be followed for `aggregate_train()`

jafermarq (approved):

PR #5798, opened 2025-09-02
feat(framework): Introduce `AppExitException` and improve `flwr-serverapp` logic

panh99:
~~### Merge after #5797~~

The PR should do:
1. Report status & prevent hanging.
2. Ensure all logs uploaded.
3. Avoid calling telemetry twice.

It will look like:

image


framework/py/flwr/common/exception.py
jafermarq, line 1: Should we place this in `flwr.supercore`?

danieljanes (approved):

PR #5805, opened 2025-09-03
refactor(framework): Add usage of providing TOML file to the help messages for `--run-config` and `--federation-config`

panh99:


framework/py/flwr/cli/constant.py
```diff
@@ -15,13 +15,26 @@
...
+# General help message for config overrides
+CONFIG_HELP_MESSAGE = (
+    "Override {0} values using one of the following formats:\n\n"
+    "--{1} '<k1>=<v1> <k2>=<v2>' | --{1} '<k1>=<v1>' --{1} '<k2>=<v2>' | "
+    "--{1} <path/to/your/toml>\n\n"
+    "When providing key-value pairs, values can be of any type supported by TOML "
+    "(e.g., bool, int, float, string). The specified keys (<k1>, <k2>, <k3> in the "
+    "example) must exist in the {0} under the `{2}` section of `pyproject.toml` to be "
+    "overridden. Alternatively, provide a TOML file containing overrides; "
+    "in that case, ensure that its keys also exist in the {0}."
```
jafermarq:
```suggestion
    "in that case, ensure that its keys also exist in the {2} section in the `pyproject.toml` of your app."
```
I'm thinking this makes it more explicit. @panh99 wdyt?

panh99: Hmmm, it feels a bit redundant tbh, repeating the sensetence before. That's also why I just mention run/federation config here.

jafermarq: i see. should we then just limit it to `Alternatively, provide a TOML file containing overrides`?

panh99: That also works for me.

jafermarq (approved):

PR #5811, opened 2025-09-04
refactor(framework): Update docker docs to use SuperExec

panh99:


framework/docker/distributed/client/compose.yml
```diff
@@ -4,7 +4,9 @@ services:
     command:
       - --superlink
       - ${SUPERLINK_IP:-127.0.0.1}:9092
-      - --clientappio-api-address
+      - --plugin-type
```
Robert-Steiner: Is that correct here to update the supernode?

framework/docker/distributed/client/compose.yml
```diff
@@ -4,7 +4,9 @@ services:
...
+      - clientapp
+      - --appio-api-address
```
panh99:
```suggestion
      - --clientappio-api-address
```

framework/docker/distributed/client/compose.yml
```diff
@@ -21,7 +21,9 @@ services:
...
```
panh99:
```suggestion
      - --clientappio-api-address
```

framework/docker/distributed/client/compose.yml
panh99: No, it's not. Good catch! I accidentally updated them. I've reverted the change. Thanks

danieljanes (approved):

PR #5820, opened 2025-09-08
fix(framework): Avoid confusing CLI prompt when `UNIMPLEMENTED` gRPC status code is detected

panh99:


framework/py/flwr/cli/utils.py
```diff
@@ -312,19 +313,28 @@ def flwr_cli_grpc_exc_handler() -> Iterator[None]:
...
+                typer.secho(
+                    "❌ User authentication is not enabled on this SuperLink.",
+                    fg=typer.colors.RED,
+                    bold=True,
+                )
+            else:
+                typer.secho(
+                    "❌ The SuperLink cannot process this request. Please verify that "
+                    "you set the address to its Control API endpoint correctly, and "
+                    "ensure you are using a compatible Flower version.",
```
chongshenng:
```suggestion
                    "you set the address to its Control API endpoint correctly in `pyproject.toml`, and "
                    "ensure that the Flower versions used by the CLI and SuperLink are compatible.",
```

I think it's better to be more specific in this error message, wdyt?

panh99: Yeah, 100%

danieljanes (approved):

PR #5827, opened 2025-09-09
refactor(examples): Update XGBoost quick-start example

yan-gao-GY:
## Proposal

Update XGBoost quick-start example to message API.


examples/xgboost-quickstart/xgboost_quickstart/client_app.py
```diff
@@ -1,139 +1,118 @@
...
     # Load model and data
     partition_id = context.node_config["partition-id"]
     num_partitions = context.node_config["num-partitions"]
     train_dmatrix, valid_dmatrix, num_train, num_val = load_data(
         partition_id, num_partitions
     )
 
+    # Load config
     cfg = replace_keys(unflatten_dict(context.run_config))
     num_local_round = cfg["local_epochs"]
```
jafermarq:
```suggestion
    # Read from run config
    num_local_round = context.run_config["local-epochs"]
    # Flatted config dict and replace "-" with "_"
    cfg = replace_keys(unflatten_dict(context.run_config))
```
I think it's clearer to read `local-epochs` directly from the run config (as it doesn't require any unflattening etc)

examples/xgboost-quickstart/xgboost_quickstart/server_app.py
```diff
@@ -1,54 +1,42 @@
...
     num_rounds = context.run_config["num-server-rounds"]
-    fraction_fit = context.run_config["fraction-fit"]
+    fraction_train = context.run_config["fraction-train"]
     fraction_evaluate = context.run_config["fraction-evaluate"]
 
-    # Init an empty Parameter
-    parameters = Parameters(tensor_type="", tensors=[])
+    # Init global model
+    global_model = b""  # Init with empty object
+    arrays = ArrayRecord([np.frombuffer(global_model, dtype=np.uint8)])
```
jafermarq: if it's not possible to initialize an randomly initialized XGB model, let's add a comment to explain what is happening in these lines and why we do it like this?

examples/xgboost-quickstart/xgboost_quickstart/server_app.py
```diff
@@ -1,54 +1,42 @@
...
 
-# Create ServerApp
-app = ServerApp(
-    server_fn=server_fn,
-)
+    # Save final model to disk
+    print("\nSaving final model to disk...")
+    model_arr = bytearray(result.arrays["0"].numpy().tobytes())
+    with open("final_model.json", "wb") as f:
+        f.write(model_arr)
```
jafermarq: is it possible to use the native `.save_model(...)` approach instead?

examples/xgboost-quickstart/xgboost_quickstart/client_app.py
```diff
@@ -1,139 +1,118 @@
...
-# Flower ClientApp
-app = ClientApp(
-    client_fn,
-)
+    # Save model
+    local_model = bst.save_raw("json")
+    model_np = np.frombuffer(local_model, dtype=np.uint8)
+
+    # Construct reply message
+    model_record = ArrayRecord([model_np])
```
jafermarq: I wonder if we should send the `local_model` bytes in a `ConfigRecord` (which natively supports `bytes`) instead of making things fit weirdly with an `ArrayRecord`. wdyt @yan-gao-GY, @panh99 @danieljanes ?

examples/xgboost-quickstart/xgboost_quickstart/server_app.py
jafermarq: One option would be doing something as you do in the `CLientApp` to materialize the xgb model and apply the gloabl model:

```
  bst = xgb.Booster(params=params)
  global_model = bytearray(msg.content["arrays"]["0"].numpy().tobytes())

  # Load global model into booster
  bst.load_model(global_model)
```

But then you'll need to remove the "division" of `clientapp` and `serverapp` settings in `pyproject.toml`

yan-gao-GY: Ok~ thanks for the suggestion!

examples/xgboost-quickstart/xgboost_quickstart/client_app.py
yan-gao-GY: IMO, it'd be better to use ArrayRecord for the model weights, which is the original design of ArrayRecord. Also, if we use ConfigRecord, all related functions from FedAvg for ArrayRecord will be invalid as we assume using ArrayRecord for model weights in default. We may have to override many methods.

Btw, in the first version, we used ConfigRecord for model weights, but I decided to use ArrayRecord if I remember correctly?

jafermarq: right. i agree keeping it with `ArrayRecord` is better so we use those extra checks in `FedAvg`

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def configure_train(
+        self, server_round: int, arrays: ArrayRecord, config: ConfigRecord, grid: Grid
+    ) -> Iterable[Message]:
+        """Configure the next round of federated training."""
+        # Keep track of array record being communicated
+        self.current_bst = arrays["0"].numpy().tobytes()
+        return super().configure_train(server_round, arrays, config, grid)
+
+    def aggregate_train(
```
jafermarq, line 53: Now that https://github.com/adap/flower/pull/5851 is merged, this method can be shortened by calling `self._check_and_log_replies`

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
```diff
@@ -0,0 +1,73 @@
...
+
+from .fedavg import FedAvg
+from .strategy_utils import aggregate_bagging
+
+
+# pylint: disable=line-too-long
+class FedXgbBagging(FedAvg):
+    """Configurable FedXgbBagging strategy implementation."""
+
+    def __init__(
```
jafermarq: @yan-gao-GY , we need to copy the entire args list and docstrings from `FedAvg`. This is because that, even though we are doing a mini edit of the `__init__` method, the docstrings aren't inherited. (because of this VSCode for instance won't show a helpful message or autocompletion). wdyt @panh99 ? is there an alternative?

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
```diff
@@ -0,0 +1,73 @@
...
+        """Aggregate ArrayRecords and MetricRecords in the received Messages."""
+        valid_replies, _ = self._check_and_log_replies(replies, is_train=True)
+
+        arrays, metrics = None, None
+        if valid_replies:
+            reply_contents = [msg.content for msg in valid_replies]
+
+            # Aggregate ArrayRecords
+            for content in reply_contents:
+                bst = content["arrays"]["0"].numpy().tobytes()  # type: ignore[union-attr]
```
jafermarq: @yan-gao-GY should we add an additional check that ensures there's only one `Array` in the `ArrayRecord`?

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
```diff
@@ -0,0 +1,73 @@
...
+# pylint: disable=line-too-long
+class FedXgbBagging(FedAvg):
+    """Configurable FedXgbBagging strategy implementation."""
+
+    def __init__(
+        self,
+        **kwargs: Any,
+    ):
+        self.current_bst: Optional[bytes] = None
+        super().__init__(**kwargs)
```
panh99:
```suggestion
    current_bst: Optional[bytes] = None
```

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
jafermarq: similarly, should we do this check in `configure_train`?

framework/py/flwr/serverapp/strategy/strategy_utils.py
```diff
@@ -249,3 +250,59 @@ def validate_message_reply_consistency(
...
+            "num_trees"
+        ]
+    )
+    # Get the number of parallel trees
+    paral_tree_num = int(
+        xgb_model["learner"]["gradient_booster"]["model"]["gbtree_model_param"][
+            "num_parallel_tree"
+        ]
+    )
+    return tree_num, paral_tree_num
```
jafermarq:
```suggestion

    # Access model parameters
    model_param = xgb_model["learner"]["gradient_booster"]["model"][
        "gbtree_model_param"
    ]
    # Return the number of trees and the number of parallel trees
    return int(model_param["num_trees"]), int(model_param["num_parallel_tree"])
```
a bit more compact

framework/py/flwr/serverapp/strategy/strategy_utils.py
```diff
@@ -249,3 +250,59 @@ def validate_message_reply_consistency(
...
+    )
+
+    # Aggregate new trees
+    trees_curr = bst_curr["learner"]["gradient_booster"]["model"]["trees"]
+    for tree_count in range(paral_tree_num_curr):
+        trees_curr[tree_count]["id"] = tree_num_prev + tree_count
+        bst_prev["learner"]["gradient_booster"]["model"]["trees"].append(
+            trees_curr[tree_count]
+        )
+        bst_prev["learner"]["gradient_booster"]["model"]["tree_info"].append(0)
```
jafermarq:
```suggestion
    previous_model = bst_prev["learner"]["gradient_booster"]["model"]
    previous_model["gbtree_model_param"]["num_trees"] = str(
        tree_num_prev + paral_tree_num_curr
    )
    iteration_indptr = previous_model["iteration_indptr"]
    previous_model["iteration_indptr"].append(
        iteration_indptr[-1] + paral_tree_num_curr
    )

    # Aggregate new trees
    trees_curr = bst_curr["learner"]["gradient_booster"]["model"]["trees"]
    for tree_count in range(paral_tree_num_curr):
        trees_curr[tree_count]["id"] = tree_num_prev + tree_count
        previous_model["trees"].append(trees_curr[tree_count])
        previous_model["tree_info"].append(0)
```

a bit more compact by defining at the top:
```python
previous_model = bst_prev["learner"]["gradient_booster"]["model"]
```

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
yan-gao-GY: GP! Added a check function

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
```diff
@@ -0,0 +1,80 @@
...
+        server_round: int,
+        replies: Iterable[Message],
+    ) -> tuple[Optional[ArrayRecord], Optional[MetricRecord]]:
+        """Aggregate ArrayRecords and MetricRecords in the received Messages."""
+        valid_replies, _ = self._check_and_log_replies(replies, is_train=True)
+
+        arrays, metrics = None, None
+        if valid_replies:
+            reply_contents = [msg.content for msg in valid_replies]
+
```
jafermarq, line 65: It would be better if we check that `self.current_bst` is not `None` so we avoid doing the casting later on. We could follow the same approach as done for example in the new `FedAdam`:https://github.com/adap/flower/blob/2fa0dd529107f5511c840f38582dff5b01d5bf56/framework/py/flwr/serverapp/strategy/fedadam.py#L140

yan-gao-GY: ok, I made the change~

examples/xgboost-quickstart/xgboost_quickstart/server_app.py
```diff
@@ -1,54 +1,55 @@
 """xgboost_quickstart: A Flower / XGBoost app."""
 
-from typing import Dict
+import numpy as np
+import xgboost as xgb
+from flwr.common import ArrayRecord, Context
+from flwr.common.config import unflatten_dict
+from flwr.server import Grid, ServerApp
+from flwr.serverapp.strategy import FedXgbBagging
+from xgboost_quickstart.task import replace_keys
```
jafermarq, line 10:
```suggestion

from xgboost_quickstart.task import replace_keys
```

examples/xgboost-quickstart/xgboost_quickstart/server_app.py
```diff
@@ -1,54 +1,55 @@
 """xgboost_quickstart: A Flower / XGBoost app."""
 
-from typing import Dict
+import numpy as np
+import xgboost as xgb
+from flwr.common import ArrayRecord, Context
+from flwr.common.config import unflatten_dict
+from flwr.server import Grid, ServerApp
```
jafermarq:
```suggestion
from flwr.serverapp import Grid, ServerApp
```

examples/xgboost-quickstart/xgboost_quickstart/client_app.py
```diff
@@ -2,137 +2,110 @@
...
+    bst = xgb.Booster(params=params)
+    global_model = bytearray(msg.content["arrays"]["0"].numpy().tobytes())
+    bst.load_model(global_model)
+
+    # Run evaluation
+    eval_results = bst.eval_set(
+        evals=[(valid_dmatrix, "valid")],
+        iteration=bst.num_boosted_rounds() - 1,
+    )
+    auc = round(float(eval_results.split("\t")[1].split(":")[1]), 4)
```
jafermarq:
```suggestion
    auc = float(eval_results.split("\t")[1].split(":")[1])
```

No need to round, If there is a way to retrieve the value without parsing the string I think that's preferable. Maybe there is a callback ?

yan-gao-GY: `No need to round` --- I'll remove it.

`If there is a way to retrieve the value without parsing the string I think that's preferable` --- No, I didn't find a xgb built-in better way to do it

framework/py/flwr/serverapp/strategy/fedxgb_bagging.py
```diff
@@ -0,0 +1,81 @@
...
+            # Aggregate ArrayRecords
+            for content in reply_contents:
+                self._ensure_single_array(cast(ArrayRecord, content["arrays"]))
```
jafermarq: Sorry i missed this. We shouldn't make assumptions about what key the `ArrayRecord` is under (`ClientApps` might use `"arrays"` or something else). Other strategies don't make this assumption. Maybe take a look at how we do it here when checking that there is only one `ArrayRecord` returned: https://github.com/adap/flower/blob/9c29a35c1efd6e97305d6cbf763a2bd74e39978b/framework/py/flwr/serverapp/strategy/strategy_utils.py#L212

yan-gao-GY: Solved~ thanks!

jafermarq (approved):

PR #5832, opened 2025-09-09
refactor(framework): Define `flwr.clientapp.mod` module for better re-exporting

panh99:


framework/py/flwr/clientapp/__init__.py
```diff
@@ -15,13 +15,9 @@
...
 
-from flwr.client import mod
 from flwr.client.client_app import ClientApp
 
-from .centraldp_mods import fixedclipping_mod
-
 __all__ = [
     "ClientApp",
-    "fixedclipping_mod",
     "mod",
```
danieljanes, line 22: Shouldn't this be removed if we remove `from flwr.client import mod` above?

danieljanes (approved):

PR #5834, opened 2025-09-09
refactor(examples): Migrate `quickstart-pytorch-lightning` to `Message API`

jafermarq:


examples/quickstart-pytorch-lightning/pytorchlightning_example/client_app.py
```diff
@@ -1,59 +1,82 @@
...
-app = ClientApp(client_fn=client_fn)
+    trainer = pl.Trainer(max_epochs=max_epochs, enable_progress_bar=False)
+    trainer.fit(model, train_loader, val_loader)
+
+    # FIT does not return values, but you can read final epoch metrics here:
+    fit_metrics = trainer.callback_metrics  # Dict[str, Tensor]
+    metrics = {
+        "train_loss": float(fit_metrics["train_loss"]),
+    }
+    print(metrics)
```
panh99:
```suggestion
    loss = float(fit_metrics["train_loss"])
```

examples/quickstart-pytorch-lightning/pytorchlightning_example/client_app.py
```diff
@@ -1,59 +1,82 @@
...
+
+    # Construct and return reply Message
+    model_record = ArrayRecord(model.state_dict())
+    metrics = {
+        **metrics,
+        "num-examples": len(train_loader.dataset),
+    }
```
panh99: A bit uncommon to see usage like this when adding a new key-value pair.

Btw, should we stick to `-` or `_`? I mean, we now have `train/test_loss`, `eval_acc`, but also `partition-id`, `num-partitions`, `num-examples`. Mixing them should be totally fine, but a bit less consistent.

```suggestion
    metrics = {
        "train_loss": loss,
        "num-examples": len(train_loader.dataset),
    }
```

jafermarq: thanks yes. It's better as you suggest

jafermarq: I think as for `-` vs `_`, It'd say it doesn't matter much here for the metrics. we use `-` for the config but is still a bit questionable why 😅

PR #5842, opened 2025-09-10
ci(framework): Update E2E test for `Control API`

jafermarq:
Now using `flwr ls` to check the status of a run.

### DIFF
--- a/framework/e2e/test_control_api.sh
+++ b/framework/e2e/test_control_api.sh
@@ -124,6 +124,7 @@ while [ "$found_success" = false ] && [ $elapsed -lt $timeout ]; do
     if [ "$status" == "finished:completed" ]; then
       found_success=true
       echo "Training worked correctly!"
+      cleanup_and_exit 0
     else
       echo "⏳ Not completed yet, retrying in 2s..."
       sleep 2


framework/e2e/test_control_api.sh
```diff
@@ -112,21 +112,22 @@ cleanup_and_exit() {
...
 while [ "$found_success" = false ] && [ $elapsed -lt $timeout ]; do
-    if grep -q "ERROR" flwr_output.log; then
-        echo "An ERROR occurred during training. Exiting."
-        cleanup_and_exit 1
-    elif grep -q "Run finished" flwr_output.log; then
-        echo "Training worked correctly!"
-        found_success=true
-        cleanup_and_exit 0
+  # Run the command and capture output
+    output=$(flwr ls . e2e --format=json)
```
danieljanes:
```suggestion
    # Run the command and capture output
    output=$(flwr ls . e2e --format=json)
```

framework/e2e/test_control_api.sh
```diff
@@ -112,21 +112,22 @@ cleanup_and_exit() {
...
+
+    # Extract status from the first run (or loop over all if needed)
+    status=$(echo "$output" | jq -r '.runs[0].status')
+
+    echo "Current status: $status"
+
+    if [ "$status" == "finished:completed" ]; then
+      found_success=true
+      echo "Training worked correctly!"
```
chongshenng, line 126:
```suggestion
      echo "Training worked correctly!"
      cleanup_and_exit 0
```
I think we should cleanup here too.

PR #5850, opened 2025-09-11
feat(framework): Add `FedAvgM`, `FedMedian`, and `FedTrimmedAvg` strategies

panh99:


framework/py/flwr/serverapp/strategy/fedavgm.py
```diff
@@ -0,0 +1,228 @@
...
+        average using the provided weight factor key.
+    evaluate_metrics_aggr_fn : Optional[callable] (default: None)
+        Function with signature (list[RecordDict], str) -> MetricRecord,
+        used to aggregate MetricRecords from training round replies.
+        If `None`, defaults to `aggregate_metricrecords`, which performs a weighted
+        average using the provided weight factor key.
+    server_learning_rate: float (default: 1.0)
+        Server-side learning rate used in server-side optimization.
+    server_momentum: float (default: 0.0)
+        Server-side momentum factor used for FedAvgM.
```
panh99, line 83: The only two new arguments. All others are the same as `FedAvg`.

framework/py/flwr/serverapp/strategy/fedtrimmedavg.py
```diff
@@ -0,0 +1,167 @@
...
+        used to aggregate MetricRecords from training round replies.
+        If `None`, defaults to `aggregate_metricrecords`, which performs a weighted
+        average using the provided weight factor key.
+    evaluate_metrics_aggr_fn : Optional[callable] (default: None)
+        Function with signature (list[RecordDict], str) -> MetricRecord,
+        used to aggregate MetricRecords from training round replies.
+        If `None`, defaults to `aggregate_metricrecords`, which performs a weighted
+        average using the provided weight factor key.
+    beta : float (default: 0.2)
+        Fraction to cut off of both tails of the distribution.
```
panh99, line 73: The only new argument. Others are from `FedAvg`. 

🙏 In the name of the @jafermarq, the `FedAvg`, and the Holy Message API.

framework/py/flwr/serverapp/strategy/fedavgm.py
```diff
@@ -0,0 +1,231 @@
...
+            INFO,
+            "\t│\t├──Minimum nodes: train (%d) | evaluate (%d)",
+            self.min_train_nodes,
+            self.min_evaluate_nodes,
+        )
+        log(INFO, "\t│\t└──Minimum available nodes: %d", self.min_available_nodes)
+        log(INFO, "\t├──> Keys in records:")
+        log(INFO, "\t│\t├── Weighted by: '%s'", self.weighted_by_key)
+        log(INFO, "\t│\t├── ArrayRecord key: '%s'", self.arrayrecord_key)
+        log(INFO, "\t│\t└── ConfigRecord key: '%s'", self.configrecord_key)
```
jafermarq: If you call `super().summary()` at the end of this method we'll use all this but from the `FedAvg.summary()` (see how we did it for `FedOpt` -> https://github.com/adap/flower/blob/2fa0dd529107f5511c840f38582dff5b01d5bf56/framework/py/flwr/serverapp/strategy/fedopt.py#L137
)

framework/py/flwr/serverapp/strategy/fedavgm.py
```diff
@@ -0,0 +1,231 @@
...
+        timeout: float = 3600,
+        train_config: Optional[ConfigRecord] = None,
+        evaluate_config: Optional[ConfigRecord] = None,
+        evaluate_fn: Optional[
+            Callable[[int, ArrayRecord], Optional[MetricRecord]]
+        ] = None,
+    ) -> Result:
+        """Execute the federated learning strategy."""
+        # Save initial parameters
+        self.current_arrays = initial_arrays
```
jafermarq: In the `FedOpt` family of startegies we also need to record the global model. We did that in `configure_train` -> https://github.com/adap/flower/blob/2fa0dd529107f5511c840f38582dff5b01d5bf56/framework/py/flwr/serverapp/strategy/fedopt.py#L164. Do you think it's better that way or how you do it here (overriding `start`)

framework/py/flwr/serverapp/strategy/fedtrimmedavg.py
jafermarq: Following our previous discussion in https://github.com/adap/flower/pull/5827#discussion_r2344187429 and your proposal, should we just set this `beta: float = 0.2` outside the `__init__`?

panh99: No, that's a bit different. `FedXgbBagging` doesn't need to override the `__init__` method of `FedAvg`; instead, it just needs an attribute. This strategy, however, has to add a new `beta` argument to the `__init__` method.

jafermarq: right right. resolving...

framework/py/flwr/serverapp/strategy/fedavgm.py
panh99: Good idea, thanks

framework/py/flwr/serverapp/strategy/fedavgm.py
panh99: That makes sense. I can set it there. To me the two solutions are effectively the same, but good to override `configure_train` than overriding `start`.

framework/py/flwr/serverapp/strategy/fedavgm.py
```diff
@@ -0,0 +1,195 @@
...
+        aggregated_arrays, aggregated_metrics = super().aggregate_train(
+            server_round, replies
+        )
+
+        # following convention described in
+        # https://pytorch.org/docs/stable/generated/torch.optim.SGD.html
+        if self.server_opt and aggregated_arrays is not None:
+            # The initial parameters should be set in `start()` method already
+            if self.current_arrays is None:
+                raise RuntimeError("No initial parameters set for FedAvgM")
```
jafermarq: In the `FedOpt` family of strategies we raise an `AggregationError`: https://github.com/adap/flower/blob/03a94793d91d9441f71c3c1cbcdf9eab40f26332/framework/py/flwr/serverapp/strategy/fedadam.py#L145.

Also, do you think it would be better to move this check before calling `super().aggregate_train`?

framework/py/flwr/serverapp/strategy/fedavgm.py
```diff
@@ -0,0 +1,195 @@
...
+            # Remember that updates are the opposite of gradients
+            pseudo_gradient = [
+                old - new for new, old in zip(aggregated_ndarrays, ndarrays)
+            ]
+            if self.server_momentum > 0.0:
+                if server_round == 1:
+                    self.momentum_vector = pseudo_gradient
+                else:
+                    if self.momentum_vector is None:
+                        raise RuntimeError("Momentum vector not initialized")
```
jafermarq: Should we raise `AggregationError` instead?

framework/py/flwr/serverapp/strategy/fedmedian.py
```diff
@@ -0,0 +1,76 @@
...
+        # Get the key for the only ArrayRecord from the first Message
+        record_key = list(valid_replies[0].content.array_records.keys())[0]
+        # Preserve keys for arrays in ArrayRecord
+        array_keys = list(valid_replies[0].content[record_key].keys())
+
+        # Retrieve all model weights as numpy arrays
+        ndarrays_list = [
+            cast(ArrayRecord, msg.content[record_key]).to_numpy_ndarrays()
+            for msg in valid_replies
+        ]
```
jafermarq: here we are doubling memory usage. could we do this conversion while freeing the ArrayRecord in each message? (same in `FedTrimmedAvg`)

framework/py/flwr/serverapp/strategy/fedtrimmedavg.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # Retrieve all model weights as numpy arrays
+        ndarrays_list = [
+            cast(ArrayRecord, msg.content[record_key]).to_numpy_ndarrays()
+            for msg in valid_replies
+        ]
+
+        # Compute trimmed mean for each layer and convert back to Array
+        trimmed_mean_arrays = [
+            Array(trim_mean(np.stack(layers), self.beta))
```
jafermarq: should we wrap this with a `try` and raise `AggregationError` if the `ValueError` is triggered?

framework/py/flwr/serverapp/strategy/fedavgm.py
panh99: Sure, that makes sense. On it.

framework/py/flwr/serverapp/strategy/fedavgm.py
panh99: Yeah, you are right. But I double checked the code and I think we can change the logic slightly. Wydt?

framework/py/flwr/serverapp/strategy/fedmedian.py
panh99: Good point. I am optimising it now.

framework/py/flwr/serverapp/strategy/fedtrimmedavg.py
panh99: Good idea, sure, I've added it.

jafermarq (dismissed):

PR #5851, opened 2025-09-11
refactor(framework): Move logging and validation logic of `FedAvg` to a helper method

panh99:
### Merge after #5845 

For most strategies, given the flexible nature of `Message`, we need to check reply consistency and log the number of received messages and errors.

Currently, this logic is duplicated in both `FedAvg.aggregate_train` and `FedAvg.aggregate_evaluate`. This PR removes the duplication and ensures that subclasses such as `FedMedian` and `Krum` can reuse the logic without having to copy it.

### DIFF
--- a/framework/py/flwr/cli/new/templates/app/code/client.huggingface.py.tpl
+++ b/framework/py/flwr/cli/new/templates/app/code/client.huggingface.py.tpl
@@ -1,41 +1,27 @@
 """$project_name: A Flower / $framework_str app."""
 
 import torch
-from flwr.client import ClientApp, NumPyClient
-from flwr.common import Context
+from flwr.app import ArrayRecord, Context, Message, MetricRecord, RecordDict
+from flwr.clientapp import ClientApp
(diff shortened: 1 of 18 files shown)


framework/py/flwr/serverapp/strategy/fedavg.py
```diff
@@ -203,16 +220,28 @@ def aggregate_train(
...
+        if validate and valid_replies:
             validate_message_reply_consistency(
-                replies=reply_contents,
+                replies=[msg.content for msg in valid_replies],
                 weighted_by_key=self.weighted_by_key,
-                check_arrayrecord=True,
+                check_arrayrecord=is_train,
             )
 
+        return valid_replies, error_replies
```
jafermarq, line 231: Better to return both or just `valid_replies` ?

panh99: I think it's good to return both. I assume some strategy implementations might want to access error replies.

PR #5856, opened 2025-09-12
ci(examples): Fix `isort` behaviour in `format.sh` and `test.sh` for Flower Examples

panh99:
Allow it to recognize first parties and third parties correctly.

### DIFF
--- a/dev/pyproject.toml
+++ b/dev/pyproject.toml
@@ -51,7 +51,7 @@ extend_exclude = [
 
 [tool.isort]
 profile = "black"
-known_third_party = ["datasets"]
+known_third_party = ["datasets"]  # `isort` treat "datasets" as first party by default
 
 [tool.black]
 line-length = 88


dev/pyproject.toml
```diff
@@ -51,6 +51,7 @@ extend_exclude = [
 
 [tool.isort]
 profile = "black"
+known_third_party = ["datasets"]
```
jafermarq: should this be `flwr-datasets` instead? `datasets` is the HF dataset package

panh99: `flwr_datasets` and `flwr` have already been recognized as third party. The `datasets` library of HF is somehow recognized as first party by default. I guess it might because more people name their module as `datasets`? No idea. So I have to explicitly mark it as a third-party.

panh99: Would you like me to add a comment here?

jafermarq: ah, i see. let's add a comment if you think it's better

panh99:
```suggestion
known_third_party = ["datasets"]  # `isort` treat "datasets" as first party by default
```

PR #5859, opened 2025-09-12
ci(examples): Fix `isort` for Flower Examples and make `framework.yml` run `./dev/test.sh` when `examples/` or `benchmarks/` changed

panh99:
### The CI for examples will NOT be triggered if we only edit examples.

### DIFF
--- a/examples/flower-authentication/README.md
+++ b/examples/flower-authentication/README.md
@@ -62,19 +62,16 @@ The `generate_cert.sh` script generates certificates for creating a secure TLS c
 
 ## Generate public and private keys for SuperNode authentication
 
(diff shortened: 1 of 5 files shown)


.github/workflows/framework.yml
```diff
@@ -31,6 +32,10 @@ jobs:
             framework:
               - 'framework/**/*'
               - '.github/workflows/framework.yml'
+            ex_bench:
+              - 'examples/**/*'
+              - 'benchmarks/**/*'
+              - 'dev/test.sh'
```
panh99, line 38:
```suggestion
              - 'dev/test.sh'
              - '.github/workflows/framework.yml'
```

PR #5863, opened 2025-09-14
feat(framework): Introduce `flwr pull`

jafermarq:
### Demo (combined with changes in EE)
1. Correct use case
2. Wrong run ID
3. Pull from a SuperLink without this feature
4. Pull an unfinished run

image
Screenshot 2025-09-17 at 13 52 53

### Design logic
This is built for Hackathon users to pull results quickly. Given the tight schedule, designs are lean—focused on current needs, with few future assumptions and a minimal level of flexibility.

### Note
This is not a big PR. Half of the change comes from generated proto files and copyright statements.


framework/proto/flwr/proto/control.proto
```diff
@@ -80,3 +83,6 @@ message GetAuthTokensResponse {
 
 message StopRunRequest { uint64 run_id = 1; }
 message StopRunResponse { bool success = 1; }
+
+message PullArtifactsRequest { uint64 run_id = 1; }
+message PullArtifactsResponse { optional string url = 1; }
```
panh99:
```suggestion
message PullArtifactsResponse { optional string url = 1; }

```

framework/py/flwr/server/app.py
```diff
@@ -199,6 +205,11 @@ def run_superlink() -> None:
         if args.enable_event_log:
             event_log_plugin = _try_obtain_control_event_log_writer_plugin()
 
+    # Load artifact provider if the args.artifact_provider_config is provided
+    artifact_provider = None
+    if cfg_path := getattr(args, "artifact_provider_config", None):
+        artifact_provider = get_ee_artifact_provider(cfg_path)
```
jafermarq, line 212: Should we log right below here that this is an experimental feature?

panh99: Yeah, makes sense.

PR #5864, opened 2025-09-15
docs(framework): Merge to `update-flower-tutorial` branch

panh99:


framework/docs/source/tutorial-series-customize-the-client-pytorch.rst
```diff
@@ -217,8 +224,8 @@ standard library. We can then send the serialized object as a ``ConfigRecord`` i
         #     training_losses={"epoch1": 0.56, "epoch2": 0.34}
         # )
 
-        # Serialize the TrainProcessMetadata object to JSON bytes
-        train_meta_bytes = json.dumps(asdict(train_metadata), indent=2).encode("utf-8")
+        # UNSAFE: Serialize the TrainProcessMetadata object to bytes
```
jafermarq: Since we decided to use `pickle`, let's remove the `..note::` a file lines below ? Or should we use it mostly to indicate that `pickle` isn't safe?

jafermarq: I wonder if it's ok to remove the `UNSAFE` line if we don't provide more context on why it is not safe. We can use the `.. note::` for that. wdyt?

panh99: Oh, yeah, good point. I will remove the note section below.

Regarding "UNSAFE" word, yeah, we can remove them as we already mention `pickle`'s potential issue in a warning block, just in case users might misunderstand it as "Flower is unsafe".

jafermarq (approved):
🔥

PR #5881, opened 2025-09-16
feat(framework): Allow YAML config for exec plugins

panh99:


framework/py/flwr/supercore/cli/flower_superexec.py
```diff
@@ -70,12 +78,28 @@ def flower_superexec() -> None:
     # Trigger telemetry event
     event(EventType.RUN_SUPEREXEC_ENTER, {"plugin_type": args.plugin_type})
 
+    # Load plugin config from YAML file if provided
+    plugin_config = None
+    if plugin_config_path := getattr(args, "plugin_config", None):
+        try:
+            with open(plugin_config_path, encoding="utf-8") as file:
+                yaml_config: dict[str, Any] = yaml.safe_load(file)
```
jafermarq:
```suggestion
                yaml_config: Optional[dict[str, Any]] = yaml.safe_load(file)
```

should this be wrapped with `Optional` as per the check done below (`is None`) ? this can happen if yaml file is empty

panh99: Right, right. Thanks for correcting this.

PR #5904, opened 2025-09-18
feat(framework): Implement `QFedAvg` strategy for Message API

panh99:
Following this algorithm:
Screenshot 2025-09-18 at 17 20 19

Approximate the Lipschitz constant `L` by `1 / client_learning_rate`.


framework/py/flwr/serverapp/strategy/qfedavg.py
```diff
@@ -0,0 +1,246 @@
...
+        log(INFO, "\t|\t├── client_learning_rate: %s", self.client_learning_rate)
+        log(INFO, "\t|\t└── train_loss_key: '%s'", self.train_loss_key)
+        super().summary()
+
+    def configure_train(
+        self, server_round: int, arrays: ArrayRecord, config: ConfigRecord, grid: Grid
+    ) -> Iterable[Message]:
+        """Configure the next round of federated training."""
+        self.current_arrays = arrays
+        return super().configure_train(server_round, arrays, config, grid)
```
jafermarq, line 145: would it make sense to send through the config record the learning rate that the clients need to use? i.e. `self.client_learning_rate`

framework/py/flwr/serverapp/strategy/qfedavg.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def get_train_loss(msg: Message, loss_key: str) -> float:
+    """Extract training loss from a Message."""
+    metrics = list(msg.content.metric_records.values())[0]
+    if (loss := metrics.get(loss_key)) is None or not isinstance(loss, (int, float)):
+        raise AggregationError(
+            "Missing or invalid local training loss in MetricRecord."
+            f" Make sure the key '{loss_key}' exists and "
+            "maps to a float value."
```
jafermarq:
```suggestion
        raise AggregationError(
            "Missing or invalid local training loss in MetricRecord."
            "The strategy was configured to receive a training loss "
            f"using key `{loss_key}`. ",
            f" Make sure the key '{loss_key}' exists and "
            "maps to a float value."
```
I think this helps connect with the `train_loss_key` arg set during construction. minor point though

framework/py/flwr/serverapp/strategy/qfedavg.py
```diff
@@ -0,0 +1,246 @@
...
+            Array(np.asarray(gw - (d / sum_h)))
+            for gw, d in zip(global_weights, sum_delta)
+        ]
+
+        # Aggregate MetricRecords
+        metrics = self.train_metrics_aggr_fn(
+            [msg.content for msg in valid_replies],
+            self.weighted_by_key,
+        )
+        return ArrayRecord(OrderedDict(zip(array_keys, array_list))), metrics
```
jafermarq, line 203: We need to update `self.current_arrays` before returning

framework/py/flwr/serverapp/strategy/qfedavg.py
```diff
@@ -0,0 +1,246 @@
...
+        average using the provided weight factor key.
+    evaluate_metrics_aggr_fn : Optional[callable] (default: None)
+        Function with signature (list[RecordDict], str) -> MetricRecord,
+        used to aggregate MetricRecords from training round replies.
+        If `None`, defaults to `aggregate_metricrecords`, which performs a weighted
+        average using the provided weight factor key.
+    q : float (default: 0.2)
+        The parameter q that controls the degree of fairness of the algorithm.
+        When set to 0, q-FedAvg is equivalent to FedAvg.
+    client_learning_rate : float (default: 0.01)
```
jafermarq: I wonder if we should make this value required and w/o a default value. Else it would be fairly easy to forget about it, then set a different LR on the client side (e.g. someone updating a `ServerApp` from `FedAvg` to `QFedAvg` could miss this important detail). wdyt?

framework/py/flwr/serverapp/strategy/qfedavg.py
panh99: I think it's better to set it here. We normally encourage users to set learning rate in the run config, which they can then pass to the strategy and the ClientApp. Besides, `q-Fedavg` doesn't consider a variable learning rate during training, so the strategy doesn't need to (and cannot) re-read the learning rate each round.

framework/py/flwr/serverapp/strategy/qfedavg.py
panh99: Makes total sense to me. Thanks!

framework/py/flwr/serverapp/strategy/qfedavg.py
panh99: I set `self.current_arrays = arrays` in every round in `.configure_train`. So we don't need to update it.

framework/py/flwr/serverapp/strategy/qfedavg.py
panh99: That's a good point. Would it make sense to move all the three additional argument to above then?

framework/py/flwr/serverapp/strategy/qfedavg.py
jafermarq: right. that works too. but i wonder what should be the common practice for all strategies that do keep track of the global model. For example in the `FedOpt` strategies we make the tracked arrays to be updated when finishing aggregation but in this `FedAvgQ` the `current_arrays` won't be updated until a new round starts. Not a big deal in principle but I wonder what's best for consistency

framework/py/flwr/serverapp/strategy/qfedavg.py
jafermarq: I think given how important is that `client_learning_rate` is correctly set, we should for sure move that one up (required if we don't have a default value). So then, moving the rest make sense to me

framework/py/flwr/serverapp/strategy/qfedavg.py
panh99: Yeah, I see your point. I explicitly set `self.current_arrays` each round because the formula relies on the global weights sent to the clients. That’s the `arrays` in `.configure_train`. I could track it only in `aggregate_train`, but strictly speaking, those aren’t the actual global weights being sent. They’re a separate set of weights that are supposed to be identical to the sent weights. Would this make sense?

framework/py/flwr/serverapp/strategy/qfedavg.py
panh99: It's very important for sure. Thanks for the idea.

framework/py/flwr/serverapp/strategy/qfedavg.py
jafermarq: that makes sense to me also. Would you then suggest not setting them as part of the `aggregate_train` method of the `FedOp`-based or other strategies (and anyther that tracks them and uses them for aggregation)?

panh99: Good question. I was thinking about this case specifically, but in general, yes. If the paper mentions that the computation is based on the difference between weights sent and weights received, then we should set `current_arrays` (or whatever the attribute) in each `.configure_train`. The exception would be if the paper explicitly states that the method relies on cross-round input, such as momentum in `FedAvgM`. Wdyt?

framework/py/flwr/serverapp/strategy/qfedavg.py
panh99: I will resolve this given that we make `client_learning_rate` a required argument.

danieljanes (approved):

PR #5905, opened 2025-09-18
feat(framework): Add Message-based adaptive clipping strategies

mohammadnaseri:
Introduces `Message`-based versions of:

`DifferentialPrivacyServerSideAdaptiveClipping`
`DifferentialPrivacyClientSideAdaptiveClipping` + `adaptiveclipping_mod`


framework/py/flwr/serverapp/strategy/dp_adaptive_clipping.py
```diff
@@ -0,0 +1,343 @@
...
+        for reply in replies_list:
+            for arr_name, record in reply.content.array_records.items():
+                reply_nd = record.to_numpy_ndarrays()
+                model_update = [
+                    np.subtract(x, y) for (x, y) in zip(reply_nd, current_nd)
+                ]
+                norm_bit = adaptive_clip_inputs_inplace(
+                    model_update, self.clipping_norm
+                )
+                clipped_indicator_count += int(bool(norm_bit))
```
jafermarq:
```suggestion
                clipped_indicator_count += int(norm_bit)
```
I think we don't need it as `norm_bit` is already a boolean?

framework/py/flwr/serverapp/strategy/dp_adaptive_clipping.py
```diff
@@ -0,0 +1,343 @@
...
+                restored = [c + u for c, u in zip(current_nd, model_update)]
```
jafermarq, line 194:
```suggestion
                # reconstruct array using clipped contribution from current round
                restored = [c + u for c, u in zip(current_nd, model_update)]
```
I thought of adding a minimal explanation here. wdyt?

framework/py/flwr/serverapp/strategy/dp_adaptive_clipping.py
```diff
@@ -0,0 +1,343 @@
...
+                reply.content[arr_name] = ArrayRecord(
+                    OrderedDict({k: Array(v) for k, v in zip(record.keys(), restored)})
+                )
+            log(
+                INFO,
+                "aggregate_fit: parameters are clipped by value: %.4f.",
```
jafermarq:
```suggestion
                "aggregate_train: arrays in `ArrayRecord` are clipped by value: %.4f.",
```

framework/py/flwr/serverapp/strategy/dp_adaptive_clipping.py
```diff
@@ -0,0 +1,343 @@
...
+        (
+            self.clipped_count_stddev,
+            self.noise_multiplier,
+        ) = compute_adaptive_noise_params(
+            noise_multiplier,
+            num_sampled_clients,
+            clipped_count_stddev,
+        )
+
+    def _validate_replies(self, replies: Iterable[Message]) -> bool:
```
jafermarq: i refactored the fixed-clipping strategy slightly so we can use the `validate_replies` function (no longer a method) from that module and avoid rewriting it here. Could you remove this method and use the function in the `aggregate_train` methods below?

framework/py/flwr/serverapp/strategy/dp_adaptive_clipping.py
```diff
@@ -0,0 +1,343 @@
...
+
+        for msg in replies:
+            found = False
+            for _, mrec in msg.content.metric_records.items():
+                if KEY_NORM_BIT in mrec:
+                    clipped_count += int(bool(mrec[KEY_NORM_BIT]))
+                    found = True
+                    break
+            if found:
+                continue
```
jafermarq: should we throw an error and hault the aggregation process if `KEY_NORM_BIT` is not found? seeing also the lines after this comment what i'm thinking we could do is add a stage at the top of `aggregate_train` that checks if `KEY_NORM_BIT` is found in the single `MetricRecord` arriving in the replies. if not, we can throw a `serverapp.exception.AggregationError` in this way we no longer need to do all these check  later on (since these keys are guaranteed to be present). wdyt?

framework/py/flwr/serverapp/strategy/dp_adaptive_clipping.py
mohammadnaseri: But is it a good practice to make this strategy dependent on the other one?

jafermarq: this is more a like a utils function imo. I think it's better to avoid duplicate code if possible (easier to maintain)

framework/py/flwr/serverapp/strategy/dp_adaptive_clipping.py
mohammadnaseri: I addressed it.

PR #5919, opened 2025-09-20
refactor(framework): Refactor `Krum` strategies

panh99:
- Made `Krum._krum` a function and renamed it to `select_multikrum` to align better with what it does.
- Vectorized the squared L2 distance computation.
- Renamed `num_nodes_to_keep` to `num_nodes_to_select` to align with the terminologies used in the original paper.
- Renamed `krum.py` to `multikrum.py` and `Krum` to `MultiKrum`.
- Added a new `krum.py` containing a `Krum` class, which is more of an alias of `MultiKrum`.
- Export both strategies.


framework/py/flwr/serverapp/strategy/multikrum.py
```diff
@@ -0,0 +1,247 @@
...
+    aggregation.
+
+    Parameters
+    ----------
+    contents : list[RecordDict]
+        List of contents from reply messages, where each content is a RecordDict
+        containing an ArrayRecord of model parameters from a node (client).
+    num_malicious_nodes : int (default: 0)
+        Number of malicious nodes in the system. Defaults to 0.
+    num_nodes_to_select : int (default: 1)
```
jafermarq:
```suggestion
def select_multikrum(
    contents: list[RecordDict],
    num_malicious_nodes: int,
    num_nodes_to_select: int,
) -> list[RecordDict]:
    """Select the set of RecordDicts to aggregate using the Krum or MultiKrum algorithm.

    For each node, computes the sum of squared L2 distances to its n-f-2 closest
    parameter vectors, where n is the number of nodes and f is the number of
    malicious nodes. The node(s) with the lowest score(s) are selected for
    aggregation.

    Parameters
    ----------
    contents : list[RecordDict]
        List of contents from reply messages, where each content is a RecordDict
        containing an ArrayRecord of model parameters from a node (client).
    num_malicious_nodes : int
        Number of malicious nodes in the system.
    num_nodes_to_select : int
```

danieljanes (approved):

PR #5920, opened 2025-09-20
feat(framework): Implement `Bulyan` for Message API

panh99:
- Fixed a few efficient issues with the old Bulyan
- Use MultiKrum for Bulyan instead of Krum for efficacy. (The output remain identical)


framework/py/flwr/serverapp/strategy/bulyan.py
```diff
@@ -0,0 +1,239 @@
...
+from typing import Callable, Optional, cast
+
+import numpy as np
+
+from flwr.common import (
+    Array,
+    ArrayRecord,
+    Message,
+    MetricRecord,
+    NDArray,
```
jafermarq:
```suggestion
```

PR #5939, opened 2025-09-23
feat(framework): Add `ArrayRecord` constructor for state dict

charlesbvll:
### Description

`Torch` doesn't type their `state_dict` object as `OrderedDict[str, Tensor]` but as `dict[str, Any]`, therefore my linter gives me the following error when I try to pass a `state_dict` to the `ArrayRecord` constructor:

image

Here is the `state_dict` definition for reference: https://github.com/pytorch/pytorch/blob/6e5dddba64b0bbe6ee149f9cf21e4ab4bf7db5b8/torch/nn/modules/module.py#L2184

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/common/record/arrayrecord.py
```diff
@@ -147,6 +147,15 @@ def __init__(  # noqa: E704
         keep_input: bool = True,
     ) -> None: ...
 
+    # This is also required for Torch state dict because they are not strongly typed
```
panh99:
```suggestion
    # This is also required for PyTorch state dict because they are not strongly typed
```

PR #5953, opened 2025-09-30
feat(framework): Introduce new `RPCs` for `flwr supernode`

jafermarq:
This PR introduce the RPCs (but not their functionality) for the new `flwr supernode add/delete/ls` command.


framework/proto/flwr/proto/control.proto
```diff
@@ -44,6 +45,15 @@ service Control {
 
   // Pull artifacts generated during a run (flwr pull)
   rpc PullArtifacts(PullArtifactsRequest) returns (PullArtifactsResponse) {}
+
+  // Add Supernode
```
danieljanes:
```suggestion
  // Add SuperNode
```

framework/proto/flwr/proto/control.proto
```diff
@@ -44,6 +45,15 @@ service Control {
...
+  rpc AddNode(AddNodeRequest) returns (AddNodeResponse) {}
+
+  // Remove Supernode
```
danieljanes:
```suggestion
  // Remove SuperNode
```

framework/proto/flwr/proto/control.proto
```diff
@@ -44,6 +45,15 @@ service Control {
...
+  rpc RemoveNode(RemoveNodeRequest) returns (RemoveNodeResponse) {}
+
+  // List Supernodes
```
danieljanes:
```suggestion
  // List SuperNodes
```

framework/proto/flwr/proto/node.proto
```diff
@@ -18,3 +18,11 @@ syntax = "proto3";
...
 
 message Node { uint64 node_id = 1; }
+
+message NodeInfo {
+  uint64 node_id = 1;
+  string owner = 2;
+  string registered_at = 3;
+  optional string last_connected_at = 4;
+  optional string deleted_at = 5;
+}
```
danieljanes:
```suggestion
}

```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -392,6 +398,27 @@ def PullArtifacts(
...
+        self, request: RemoveNodeRequest, context: grpc.ServicerContext
+    ) -> RemoveNodeResponse:
+        """Remove a node."""
+        log(INFO, "ControlServicer.RemoveNode")
+        return RemoveNodeResponse()
+
+    def ListNodes(
+        self, request: ListNodesRequest, context: grpc.ServicerContext
+    ) -> ListNodesResponse:
+        """List all nodes."""
```
danieljanes:
```suggestion
        """Add a SuperNode."""
        log(INFO, "ControlServicer.AddNode")
        return AddNodeResponse()

    def RemoveNode(
        self, request: RemoveNodeRequest, context: grpc.ServicerContext
    ) -> RemoveNodeResponse:
        """Remove a SuperNode."""
        log(INFO, "ControlServicer.RemoveNode")
        return RemoveNodeResponse()

    def ListNodes(
        self, request: ListNodesRequest, context: grpc.ServicerContext
    ) -> ListNodesResponse:
        """List all SuperNodes."""
```

framework/proto/flwr/proto/control.proto
```diff
@@ -86,3 +96,15 @@ message StopRunResponse { bool success = 1; }
 
 message PullArtifactsRequest { uint64 run_id = 1; }
 message PullArtifactsResponse { optional string url = 1; }
+
+message AddNodeRequest { string pub_key = 1; }
+message AddNodeResponse { optional uint64 node_id = 1; }
```
danieljanes: How about this?
```suggestion
message AddNodeResponse { optional NodeInfo node_info = 1; }
```

framework/proto/flwr/proto/control.proto
```diff
@@ -86,3 +96,15 @@ message StopRunResponse { bool success = 1; }
...
+
+message RemoveNodeRequest { uint64 node_id = 1; }
+message RemoveNodeResponse {}
+
+message ListNodesRequest {}
+message ListNodesResponse {
+  map<uint64, NodeInfo> node_dict = 1;
+  string now = 2;
```
danieljanes, line 109: Is `now` the SuperLink timestamp? What would we use it for?

jafermarq: yes, it's the timestamp when the response is created. in `flwr ls` we use it to compute the elapsed time for an ongoing run before displaying the table (https://github.com/adap/flower/blob/757c9fae2b56522624733a59efded61f3e63594d/framework/py/flwr/cli/ls.py#L194). I was thinking keeping this would be useful to potentially show how long a `SuperNode` has been connected for.

framework/proto/flwr/proto/control.proto
panh99: ~~We already have a `Node` message. Would you prefer to keeping it, or renaming it to `NodeInfo`? @danieljanes~~ 

Nvm, I should read the RFC first.

jafermarq: ok for me to use `Node` instead but it will need some work to also introduce its pythonic representation (e.g. as a dataclass).

panh99: I am okay with the change. Just a note, in `flwr run`, the `StartRun` RPC will only return a run ID, with full details returned by `flwr ls`. I'd prefer to having the behaviours of run management and node management aligned.

jafermarq: my original proposal was to also return the `node_id` (which is essentially what the _protobuf_ `Node` carries). I think the argument for alignment with `flwr run` is a valid one!

jafermarq: The response now returns `optional Node`

framework/proto/flwr/proto/control.proto
```diff
@@ -86,3 +96,15 @@ message StopRunResponse { bool success = 1; }
...
+
+message AddNodeRequest { string pub_key = 1; }
+message AddNodeResponse { optional Node node = 1; }
+
+message RemoveNodeRequest { Node node = 1; }
+message RemoveNodeResponse {}
+
+message ListNodesRequest {}
+message ListNodesResponse {
+  repeated NodeInfo node_dict = 1;
```
panh99:
```suggestion
  repeated NodeInfo node_info_list = 1;
```

panh99: Or maybe just `nodes`?

framework/proto/flwr/proto/node.proto
```diff
@@ -18,3 +18,14 @@ syntax = "proto3";
...
+
+message NodeInfo {
+  uint64 node_id = 1;
+  string owner_aid = 2;
+  string registered_at = 3;
+  optional string activated_at = 4;
+  optional string deactivated_at = 5;
+  optional string deleted_at = 6;
+  optional float online_until = 7;
+  optional float heartbeat_interval = 8;
```
panh99: May I know the motivation for adding the `optional` keywords? As discussed, we gonna use `""` for NA timestamps.

jafermarq: i'll change it.

framework/proto/flwr/proto/control.proto
jafermarq: i'll leave it as `nodes_info`

framework/proto/flwr/proto/control.proto
panh99: Oh, we don't need `optional` for message fields. It only works for scalar fields like `string`.

panh99: @danieljanes Should we return just a `node_id` to be consistent with `flwr run`? Or we can return `NodeInfo`. FYI, `Node` message is defined as `message Node { uint64 node_id = 1; }`.

panh99 (dismissed):
LGTM!

framework/proto/flwr/proto/control.proto
```diff
@@ -86,3 +96,15 @@ message StopRunResponse { bool success = 1; }
 
 message PullArtifactsRequest { uint64 run_id = 1; }
 message PullArtifactsResponse { optional string url = 1; }
+
+message AddNodeRequest { string pub_key = 1; }
```
danieljanes:
```suggestion
message AddNodeRequest { string public_key = 1; }
```

framework/proto/flwr/proto/node.proto
```diff
@@ -18,3 +18,14 @@ syntax = "proto3";
 package flwr.proto;
 
 message Node { uint64 node_id = 1; }
+
+message NodeInfo {
+  uint64 node_id = 1;
+  string owner_aid = 2;
+  string registered_at = 3;
```
danieljanes:
```suggestion
  string created_at = 3;
```

danieljanes (approved):

PR #5955, opened 2025-09-30
feat(framework): Add `NoOp*Plugin`s

panh99:
The plugins will be loaded when user auth is disabled to keep code logic thin. (Not implemented in this PR yet)

I am a bit undecided between `NoOp*` and `Default*`, given that we should consider auth mode as default.

@danieljanes I tried to avoid using `user auth` in this PR, and I will open a follow-up PR for renaming all `user auth` to `account auth`.


framework/py/flwr/common/constant.py
```diff
@@ -281,3 +282,8 @@ def all() -> list[str]:
         """Return all SuperExec plugin types."""
         # Filter all constants (uppercase) of the class
         return [v for k, v in vars(ExecPluginType).items() if k.isupper()]
+
+
+# Constants for No-op auth plugins
+NOOP_FLWR_AID = "admin"
+NOOP_ACCOUNT_NAME = "admin"
```
jafermarq: How about using `default-user` or something along those terms instead of `admin`? Maybe there is a term that better conveys the AID comes from a constant ?

panh99:
```suggestion
NOOP_FLWR_AID = "sys_noauth"
NOOP_ACCOUNT_NAME = "sys_noauth"
```

Wdyt?

jafermarq: looks good to me! (or `sys_noauth`)

framework/py/flwr/cli/auth_plugin/noop_auth_plugin.py
```diff
@@ -0,0 +1,69 @@
...
+
+import typer
+
+from flwr.common.typing import UserAuthCredentials, UserAuthLoginDetails
+from flwr.proto.control_pb2_grpc import ControlStub
+
+from .auth_plugin import CliAuthPlugin
+
+
+class NoOpCliAuthPlugin(CliAuthPlugin):
```
jafermarq, line 28: Q: would it be better to make this plugin to inherit form `OidcCliPlugin` and just override `login()` as you do below? motivation is to keep the rest of the behaviour the same

panh99: Not really. If the `NoOp` implementation inherits from `OidcCliPlugin`, we would also need to override methods like `store_tokens` to prevent tokens from being saved with the auth type set to `OIDC`, making it not really a simplification.

Also, conceptually the two are unrelated. For consistency, note that we didn’t take this approach with `ControlAuthPlugin` either, so I don’t think it’s necessary here.

IMO, `NoOp*` should stand as an independent implementation. It only needs to conform to the interface defined by the abstract base class, not the specifics of the OIDC implementation. Wdyt?

jafermarq (approved):

framework/py/flwr/cli/auth_plugin/noop_auth_plugin.py
danieljanes: Keeping them independent also has the advantage of avoiding any accidental inheritance.

framework/py/flwr/cli/auth_plugin/noop_auth_plugin.py
```diff
@@ -0,0 +1,69 @@
...
+        # Warn the account that no authentication is being performed
+        typer.secho(
+            "Warning: No authentication is being performed. "
+            "Proceeding without account authentication.",
+            fg=typer.colors.YELLOW,
+        )
+        return UserAuthCredentials(
+            access_token="",
+            refresh_token="",
+        )
```
danieljanes: Instead of returning an empty object here, should we rather raise an error?

panh99: That will break `flwr login` and result in a long traceback. Wdyt? We can also raise a `typer.Exit` here  to avoid the long traceback if you prefer.

framework/py/flwr/superlink/auth_plugin/noop_auth_plugin.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def get_login_details(self) -> Optional[UserAuthLoginDetails]:
+        """Get the login details."""
+        return UserAuthLoginDetails(
+            auth_type=AuthType.NOOP,  # No operation auth type
+            device_code="",
+            verification_uri_complete="",
+            expires_in=0,
+            interval=0,
+        )
```
danieljanes, line 54: Should we actually return something here? Or raise a specific exception that causes the interceptor to return a specific error code?

framework/py/flwr/superlink/auth_plugin/noop_auth_plugin.py
```diff
@@ -0,0 +1,84 @@
...
+    ) -> tuple[bool, Optional[AccountInfo]]:
+        """Return valid for no-op plugin."""
+        return True, NOOP_ACCOUNT_INFO
+
+    def get_auth_tokens(self, device_code: str) -> Optional[UserAuthCredentials]:
+        """Get authentication tokens."""
+        return UserAuthCredentials(
+            access_token="",
+            refresh_token="",
+        )
```
danieljanes: We could also raise an error here because this code should never be triggered

framework/py/flwr/cli/auth_plugin/noop_auth_plugin.py
danieljanes: Hm... I'm not a fan of using `typer` at this level. Could we raise an exception, catch it on the caller side and use `typer.Exit` there?

panh99: Yeah, we can. I can also adjust our `OidcCliAuthPlugin` to raise an exception if fails.

PR #5957, opened 2025-09-30
refactor(examples): Migrate `federated-kaplan-meier-fitter` example to `Message API`

jafermarq:
This example is more about Fed Analytics more than actual Fed Learning, because of this i've decided to now use a strategy and instead use the lower level `Message` with `MessageType.Query` messages.


panh99 (approved):
LGTM!

PR #5959, opened 2025-10-01
fix(framework): Correct superexec docker entrypoint

rohboz:
The entrypoint notation in the the different compose files was not aligned with the docker build file for superexec: flower instead of flwr. Causes docker compose to fail during deployment for multi-machine setup.


jafermarq (approved):
great catch @rohboz ! many thanks

PR #5967, opened 2025-10-02
feat(framework): Update flwr run and superlink

yan-gao-GY:
## Proposal

Update flwr run and superlink.

### DIFF
--- a/framework/py/flwr/superlink/servicer/control/control_servicer.py
+++ b/framework/py/flwr/superlink/servicer/control/control_servicer.py
@@ -625,7 +625,7 @@ def _request_download_link(
     identifier: str, context: grpc.ServicerContext
 ) -> tuple[str, str, list[dict[str, str]]]:
     """Request download link from Flower platform API."""
-    url = f"{PLATFORM_API_URL}/hub/fetch"
+    url = f"{PLATFORM_API_URL}/hub/fetch-fab"
     headers = {
         "Content-Type": "application/json",
         "Accept": "application/json",


framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -86,6 +86,7 @@ def start_client_internal(
     isolation: str = ISOLATION_MODE_SUBPROCESS,
     clientappio_api_address: str = CLIENTAPPIO_API_DEFAULT_SERVER_ADDRESS,
     health_server_address: Optional[str] = None,
+    trust_entity: Optional[list] = None,
```
jafermarq: Would this be a `list`? or better `dict[str, bytes]`, where `str` is the entity_id and `bytes` is the public key?

Also, should we use `trust_entities`?

framework/py/flwr/supernode/servicer/clientappio/clientappio_servicer.py
```diff
@@ -151,7 +151,7 @@ def PullClientAppInputs(
         # Retrieve context, run and fab for this run
         context = cast(Context, state.get_context(run_id))
         run = cast(Run, state.get_run(run_id))
-        fab = Fab(run.fab_hash, ffs.get(run.fab_hash)[0])  # type: ignore
+        fab = Fab(run.fab_hash, ffs.get(run.fab_hash)[0], {})  # type: ignore
```
jafermarq, line 154: Here we don't want to send the FAB's metadata ?

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -16,13 +16,17 @@
 
 
 import hashlib
+import json
+import re
 import time
+import typer
```
jafermarq:
```suggestion
```
unused

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -103,7 +111,30 @@ def StartRun(  # pylint: disable=too-many-locals
...
         ffs = self.ffs_factory.ffs()
 
-        if len(request.fab.content) > FAB_MAX_SIZE:
+        verification = None
+        if request.fab.content == b"":
+            identifier = request.fab.meta["identifier"]
+            m = re.match(r"^(?P<user>[^/]+)/(?P<app>[^/]+)$", identifier)
+            if not m:
+                raise ValueError(
+                    "Invalid remote app ID. Expected format: 'user_name/app_name'."
```
jafermarq: In `cli/run` you define the app identifier as `@user_name/app_name` but then the `@` is dropped before making the `StartRun` request. I wonder if it's better to not drop it so the regex logic is identical in both places. It will also help in keeping logs identical

yan-gao-GY: Yes good catch! I've changed to send the original string instead

framework/py/flwr/supernode/servicer/clientappio/clientappio_servicer.py
yan-gao-GY: We don't need to send the FAB's verification to `clientappio_servicer`, since the verification has been done on SuperNode (precisely, in `start_client_internal.py`)

PR #5973, opened 2025-10-03
refactor(framework): Improve `flwr ls` and introduce `flwr list` alias

jafermarq:
- Now the `Run ID` will show without introducing a line break. the width of the column is adjusted automatically.
- `flwr list` is introduced as an alias to `flwr ls`
- Now `flwr --help` shows `list` and `ls` is a hidden alias.


panh99 (dismissed):
🤩

danieljanes (approved):

PR #5974, opened 2025-10-03
feat(framework): Use `NoOp*Plugin` by default

panh99:


framework/py/flwr/common/constant.py
```diff
@@ -256,6 +256,17 @@ def __new__(cls) -> AuthnType:
         raise TypeError(f"{cls.__name__} cannot be instantiated.")
 
 
+class AuthzType:
+    """Account authorization types."""
+
+    NOOP = "noop"
+    FPGA = "fpga"
```
panh99:
```suggestion
    NOOP = "noop"
```

framework/py/flwr/server/app.py
```diff
@@ -465,18 +479,14 @@ def _load_plugin(
...
                 sys.exit(
                     f"{yaml_key}: {auth_plugin_name} is not supported. "
                     f"Please provide a valid {section} type in the configuration."
                 )
             sys.exit(f"No {section} type is provided in the configuration.")
-        except NotImplementedError:
-            sys.exit(f"No {section} plugins are currently supported.")
 
     # Warn deprecated authn_type key
     if "authn_type" in config["authentication"]:
```
jafermarq: side note, i think this should say `auth_type`

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -224,20 +221,15 @@ def ListRuns(
...
-                # If no `run_id` is specified and no account auth is enabled,
-                # return all run IDs
-                run_ids = state.get_run_ids(None)
+            # If no `run_id` is specified and account auth is enabled,
+            # return run IDs for the authenticated account
+            flwr_aid = shared_account_info.get().flwr_aid
+            if flwr_aid is None:
+                context.abort(
+                    grpc.StatusCode.PERMISSION_DENIED,
+                    "️⛔️ Account authentication is enabled, but `flwr_aid` is None",
```
jafermarq: when would this `context.abort` be reached ? Can it happen?

framework/py/flwr/server/app.py
panh99: Gosh, you are right. Good catch.

panh99: Must be me get messed up with merging the main with renaming.

framework/py/flwr/superlink/servicer/control/control_servicer.py
panh99: Right, right. It should not happen now. Good catch.

framework/py/flwr/superlink/auth_plugin/__init__.py
```diff
@@ -15,12 +15,41 @@
...
+def get_control_authn_plugins() -> dict[str, type[ControlAuthnPlugin]]:
+    """Return all Control API authentication plugins."""
+    ee_dict: dict[str, type[ControlAuthnPlugin]] = get_control_authn_ee_plugins()
+    return ee_dict | {AuthnType.NOOP: NoOpControlAuthnPlugin}
+
+
+def get_control_authz_plugins() -> dict[str, type[ControlAuthzPlugin]]:
+    """Return all Control API authorization plugins."""
+    ee_dict: dict[str, type[ControlAuthzPlugin]] = get_control_authz_ee_plugins()
+    return ee_dict | {AuthzType.NOOP: NoOpControlAuthzPlugin}
```
jafermarq, line 45: could we move the `try` inside these functions and then avoid having to define new functions using `_ee_` ? (as in `get_control_authz_ee_plugins`)

panh99: Do you mean moving the `import` inside the function?

jafermarq (approved):

PR #5968, opened 2025-10-02
feat(framework): Implement dummy behaviour `flwr supernode list/ls`

jafermarq:
This PR introduces the logic on the Flower CLI side for `flwr supernode list/ls` and can now display the info about nodes connected to the `SuperLink` as a table or in JSON format.

Concretely, info about 4 hardcoded nodes is shown:
- node 1: a `SuperNode` created via `flwr supernode add` but never connected
- node 2: same as node 1 but the `SuperNode` actually connected to the `SuperLink` and remains connected
- node 3: same as node 1 but the admin did `flwr supernode delete` before it was ever connected
- node 4: a node that was added, the, connected, then disconnected, and finally deleted.

```shell
flwr supernode list . remote-federation
Loading project configuration...
Success
📄 Listing all nodes...
┏━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┓
┃       Node ID        ┃    Owner    ┃ Status  ┃ Elapsed  ┃   Status Changed @   ┃
┡━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩
│ 15390646978706312628 │ owner_aid_1 │ created │          │ N/A                  │
├──────────────────────┼─────────────┼─────────┼──────────┼──────────────────────┤
│ 2941141058168602545  │ owner_aid_2 │ online  │ 23:30:00 │ 2025-10-06 11:54:54Z │
├──────────────────────┼─────────────┼─────────┼──────────┼──────────────────────┤
│ 1781174086018058152  │ owner_aid_4 │ offline │          │ 2025-10-06 12:24:54Z │
└──────────────────────┴─────────────┴─────────┴──────────┴──────────────────────┘
```

or as a JSON:
```
flwr supernode list . remote-federation --format=json
{
  "success": true,
  "nodes": [
    {
      "node-id": 15390646978706312628,
      "owner-aid": "owner_aid_1",
      "status": "created",
      "created-at": "2025-10-06 11:32:57Z",
      "online-at": "N/A",
      "online-elapsed": "00:00:00",
      "offline-at": "N/A",
      "deleted-at": "N/A"
    },
    {
      "node-id": 2941141058168602545,
      "owner-aid": "owner_aid_2",
      "status": "online",
[...]

PR #5977, opened 2025-10-05
feat(framework): Fuse `Linkstate.create_node` and `Linkstate.set_node_public_key`

panh99:
The `set_node_public_key` and `get_node_id(public_key)` methods can be removed in a follow-up PR.


framework/py/flwr/server/superlink/fleet/grpc_rere/node_auth_server_interceptor.py
```diff
@@ -113,50 +110,34 @@ def intercept_service(  # pylint: disable=too-many-return-statements
...
-            method_handler, expected_node_id, node_pk_bytes
-        )
+        return self._wrap_method_handler(method_handler, node_pk_bytes)
 
     def _wrap_method_handler(
         self,
         method_handler: grpc.RpcMethodHandler,
-        expected_node_id: Optional[int],
-        node_public_key: bytes,
+        expected_public_key: bytes,
```
panh99, line 118: FYI, as node ID is the official identifier of the node, I switch to this approach that retrieves the public key by node ID instead, which is more logically sound.

### What we did:
1. Retrieve node ID by the provided public key
2. Check if provided node ID matches the retrieved one

### What we do now:
1. Retrieve the public key by the provided node ID
2. Check if the provided public key matches the retrieved one

framework/py/flwr/client/rest_client/connection.py
```diff
@@ -292,7 +293,12 @@ def send_node_heartbeat() -> bool:
 
     def create_node() -> Optional[int]:
         """Set create_node."""
-        req = CreateNodeRequest(heartbeat_interval=HEARTBEAT_DEFAULT_INTERVAL)
+        req = CreateNodeRequest(
+            # REST does not support node authentication;
+            # random bytes are used instead
+            public_key=secrets.token_bytes(32),
```
jafermarq, line 299: Are we making a strong assumption elsewhere in the system about using `32` bytes ?

panh99: Not really. We don't yet put any restrictions on the length of the public key.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -330,36 +330,47 @@ def num_message_res(self) -> int:
...
-                time.time() + heartbeat_interval,
-                heartbeat_interval,
+            # Mark the node online until now().timestamp() + heartbeat_interval
+            current = now()
+            self.nodes[node_id] = NodeInfo(
+                node_id=node_id,
+                owner_aid="",  # Unused for now
+                created_at=current.isoformat(),  # Unused for now
+                activated_at=current.isoformat(),  # Unused for now
+                deactivated_at="",  # Unused for now
```
jafermarq, line 354:
```suggestion
                last_activated_at=current.isoformat(),  # Unused for now
                last_deactivated_at="",  # Unused for now
```

Let's make this change? I believe https://github.com/adap/flower/pull/5968 will be merged before this PR

jafermarq:
```suggestion
                owner_aid="",  # Unused for now
                status="created",
                created_at=current.isoformat(),  # Unused for now
                last_activated_at=current.isoformat(),  # Unused for now
                last_deactivated_at="",  # Unused for now
```

Let's make this change? I believe https://github.com/adap/flower/pull/5968 will be merged before this PR

framework/py/flwr/server/superlink/fleet/grpc_rere/node_auth_server_interceptor.py
jafermarq: thanks for the explanation !

framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -613,18 +612,21 @@ def create_node(self, heartbeat_interval: float) -> int:
...
+                    public_key,
                 ),
             )
-        except sqlite3.IntegrityError:
+        except sqlite3.IntegrityError as e:
+            if "UNIQUE constraint failed: node.public_key" in str(e):
+                raise ValueError("Public key already in use.") from None
+            # Must be node ID conflict, almost impossible unless system is compromised
             log(ERROR, "Unexpected node registration failure.")
             return 0
```
jafermarq, line 645: Should another error be raised if it's not due to public key already used? Else we'd be returning `node_id=0` to a `SuperNode`, right? Should we make:

```protobuf
message CreateNodeResponse { optional Node node = 1; } # note the `optional`
```

or better leave until node life cycle review?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: Sure. Make sense to me.

framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
panh99: Hmm, I didn't change the original logic here, but yes, we can improve it. This error only happens when the same random uint64 is generated, which is practically impossible. However, we can re-generate a random uint64 for the node ID, or we raise an warning saying that the system may be compromised.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: I tweaked your suggestion a bit since #5968 modifies the proto definition, which is not yet available in this PR.

PR #5981, opened 2025-10-06
refactor(framework): Implement `flwr supernode create` functionality

jafermarq:
## ~~Merge after https://github.com/adap/flower/pull/5977~~

An `ed25519.Ed25519PublicKey` is expected to be passed to `flwr supernode create`. If public key already in used, a context abort will trigger the following error message to be shown in Flower CLI:

```
(flwr-py3.11) ➜  ttt flwr supernode create .testkey.pub . remote-federation
Loading project configuration...
Success
❌ The provided public key is already assigned to another supernode.
```

### DIFF
--- a/framework/py/flwr/superlink/servicer/control/control_servicer.py
+++ b/framework/py/flwr/superlink/servicer/control/control_servicer.py
@@ -410,8 +410,12 @@ class ControlServicer(control_pb2_grpc.ControlServicer):
         # Init link state
         state = self.linkstate_factory.state()
         node_id = 0
+
+        flwr_aid = shared_account_info.get().flwr_aid
+        flwr_aid = _check_flwr_aid_exists(flwr_aid, context)
         try:
             node_id = state.create_node(
+                owner_aid=flwr_aid,
                 public_key=request.public_key,
                 heartbeat_interval=HEARTBEAT_DEFAULT_INTERVAL,
             )
@@ -514,7 +518,7 @@ def _create_list_runs_response(
 
 def _check_flwr_aid_exists(
     flwr_aid: Optional[str], context: grpc.ServicerContext
-) -> None:
+) -> str:
     """Guard clause to check if `flwr_aid` exists."""
     if flwr_aid is None:
         context.abort(
@@ -522,6 +526,7 @@ def _check_flwr_aid_exists(
             "️⛔️ Failed to fetch the account information.",
         )
         raise RuntimeError  # This line is unreachable
+    return flwr_aid
 
 
 def _check_flwr_aid_in_run(
--- a/framework/py/flwr/superlink/servicer/control/control_servicer_test.py
+++ b/framework/py/flwr/superlink/servicer/control/control_servicer_test.py
@@ -194,7 +194,9 @@ class TestControlServicer(unittest.TestCase):
         """Test CreateNodeCli method of ControlServicer."""
         # Prepare
         if pre_register_key:
-            self.state.create_node(public_key=pub_key, heartbeat_interval=10)
+            self.state.create_node(
+                owner_aid="fake_aid", public_key=pub_key, heartbeat_interval=10
+            )
 
         # Execute
         req = CreateNodeCliRequest(public_key=pub_key)


framework/proto/flwr/proto/control.proto
```diff
@@ -97,7 +97,10 @@ message StopRunResponse { bool success = 1; }
 message PullArtifactsRequest { uint64 run_id = 1; }
 message PullArtifactsResponse { optional string url = 1; }
 
-message CreateNodeCliRequest { string public_key = 1; }
+message CreateNodeCliRequest {
+  bytes public_key = 1;
+  double heartbeat_interval = 2;
```
panh99: We don’t need this field actually. It’s kinda a tech debt. We can be pass a random value as heartbeat_interval to LinkState.create_node. If you check the code, it makes no difference: no matter what is the initial interval, the NodeStatus will be “created” upon creation, which will later be switched to activated when receiving the first heartbeat, refreshing both heartbeat_interval and online_until fields of the node. Wdyt?

framework/py/flwr/cli/supernode/create.py
```diff
@@ -14,25 +14,46 @@
 # ==============================================================================
 """Flower command line interface `supernode create` command."""
 
+import io
```
panh99, line 18:
```suggestion

import io
```

framework/proto/flwr/proto/control.proto
jafermarq: True that it makes no difference now. I thought of moving it as part of the _create_ request so we give users the control of what this heartbeat should be (note that the person doing `flwr supernode create` might not be the same one actually launching the supernode). But i'm happy to leave this out for now if you think it's best.

panh99: I see, that’s a valid point! Since we haven’t yet decided how to support a custom heartbeat interval, and given that the `heartbeat_interval` field in the request isn’t currently used, I think it’s better not to introduce it for now. We can always add it later if we decide it’s necessary. Would this make sense to you?

framework/py/flwr/supercore/primitives/asymmetric.py
```diff
@@ -107,3 +107,11 @@ def verify_signature(
         return True
     except InvalidSignature:
         return False
+
+
+def check_public_key_is_nist_ec(public_key: ec.EllipticCurvePublicKey) -> bool:
+    """Check if the provided public key is a NIST EC curve."""
```
panh99: Minor tweak. Btw, just a note, this function can also be used to check the curve type of a private key (nothing needs changing now).

```suggestion
def uses_nist_ec_curve(public_key: ec.EllipticCurvePublicKey) -> bool:
    """Return True if the provided key uses a NIST EC curve."""
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -406,7 +412,37 @@ def CreateNodeCli(
...
+        try:
+            # Attempt to deserialize public key
+            pub_key = bytes_to_public_key(request.public_key)
+            # Check if it's a NIST EC curve public key
+            if not check_public_key_is_nist_ec(pub_key):
+                log(ERROR, "The provided public key is not a NIST EC curve public key.")
+                raise ValueError()
+        except (ValueError, AttributeError):
+            log(ERROR, "The provided public key could not be deserialized.")
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, PUBLIC_KEY_NOT_VALID)
```
panh99, line 408: Do you think we need any disambiguation here, since we have several public keys for different uses? Or maybe this is the only public key used on the CLI side. Wdyt?

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -406,7 +412,37 @@ def CreateNodeCli(
...
+            node_id = state.create_node(
+                public_key=request.public_key,
+                heartbeat_interval=request.heartbeat_interval,
+            )
+
+        except ValueError:
+            # Public key already in use
+            log(ERROR, PUBLIC_KEY_ALREADY_IN_USE_MESSAGE)
+            context.abort(
+                grpc.StatusCode.FAILED_PRECONDITION, PUBLIC_KEY_ALREADY_IN_USE_MESSAGE
```
panh99, line 427: Same question as above.

framework/py/flwr/cli/utils.py
```diff
@@ -386,4 +388,22 @@ def flwr_cli_grpc_exc_handler() -> Iterator[None]:
                     bold=True,
                 )
                 raise typer.Exit(code=1) from None
+            if (
+                e.details() == PUBLIC_KEY_ALREADY_IN_USE_MESSAGE
+            ):  # pylint: disable=E1101
+                typer.secho(
+                    "❌ The provided public key is already assigned to another "
```
panh99: Just a thought. Not sure which is clearer.

```suggestion
                    "❌ The provided public key is already in use by another "
```

framework/py/flwr/cli/utils.py
```diff
@@ -386,4 +388,22 @@ def flwr_cli_grpc_exc_handler() -> Iterator[None]:
...
+                    "supernode.",
+                    fg=typer.colors.RED,
+                    bold=True,
+                )
+                raise typer.Exit(code=1) from None
+            if e.details() == PUBLIC_KEY_NOT_VALID:  # pylint: disable=E1101
+                typer.secho(
+                    "❌ The provided public key is not valid. Please provide a valid "
```
```suggestion
                    "❌ The provided public key is invalid. Please provide a valid "
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
jafermarq: by _disambiguation_ you mean in the name of the constant? yes, this is the only public key used/passed from the CLI side for now.

framework/proto/flwr/proto/control.proto
jafermarq: make sense to me, i'll revert this

framework/py/flwr/superlink/servicer/control/control_servicer.py
panh99: Yeah, I mean the name and the value of the constant. Since it's the only public key, we can keep it as it is. I will resolve this thread and the following thread.

framework/py/flwr/cli/utils.py
```diff
@@ -370,4 +372,22 @@ def flwr_cli_grpc_exc_handler() -> Iterator[None]:
                     bold=True,
                 )
                 raise typer.Exit(code=1) from None
+            if (
+                e.details() == PUBLIC_KEY_ALREADY_IN_USE_MESSAGE
+            ):  # pylint: disable=E1101
+                typer.secho(
+                    "❌ The provided public key is already in use by another "
+                    "supernode.",
```
danieljanes: Always `SuperNode` in user-facing messages (docs/logs/...):
```suggestion
                    "SuperNode.",
```

PR #5982, opened 2025-10-06
feat(framework): Verify FAB on SuperNode

psfoley:
WIP


framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -305,6 +307,37 @@ def _pull_and_store_message(  # pylint: disable=too-many-positional-arguments
...
+                log(
+                    WARN,
+                    f"The SHA256 hash computed on the FAB ({verified_fab_hash}) did",
+                    f" not match the expected hash ({fab.hash_string})",
+                    " Experiment will not execute",
+                )
+                return None
+
+            # FAB must be signed if trust entity provided
+            if trust_entity:
```
yan-gao-GY: Shall we add a flag argument `enable-entity-verification`? Only when it's true, we do the verification, and the `trust_entity` must not be empty.

psfoley: Done

framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -15,18 +15,20 @@
...
 from collections.abc import Iterator
 from contextlib import contextmanager
 from functools import partial
-from logging import INFO
+from logging import INFO, WARN
 from pathlib import Path
 from typing import Callable, Optional, Union, cast
 
 import grpc
 from cryptography.hazmat.primitives.asymmetric import ec
```
yan-gao-GY, line 32:
```suggestion
from cryptography.hazmat.primitives import serialization
from cryptography.hazmat.primitives.asymmetric import ec
```

framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -305,6 +313,41 @@ def _pull_and_store_message(  # pylint: disable=too-many-positional-arguments
             # Verify the received FAB
             #########################
 
+            # FAB must be signed if trust entity provided
+            if enable_entity_verification:
+                if len(trust_entity) == 0:
```
yan-gao-GY: How about having a check in the CLI side (`flower_supernode.py`) instead? Something like:
```
parser = _parse_args_run_supernode()
args = parser.parse_args()

    if args.enable_entity_verification and not args.trust_entity:
        parser.error(
            "Entity verification is enabled, but no trusted entities were provided. " \
            "The Flower SuperNode should be restarted with a trusted entities list."
        )
```

psfoley: Done!

PR #5985, opened 2025-10-07
feat(framework): Add `public_key` field to `NodeInfo` and require `owner_aid` for node creation

panh99:
### Merge after #5977

- Also removed `set_node_public_key` and `get_node_id` methods.
- An index is created on `owner_aid` to ensure faster lookup for future `flwr supernode list`.

### DIFF
--- a/framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
+++ b/framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
@@ -395,7 +395,7 @@ class InMemoryLinkState(LinkState):  # pylint: disable=R0902,R0904
         """Get `public_key` for the specified `node_id`."""
         with self.lock:
             if (node := self.nodes.get(node_id)) is None:
-                raise ValueError(f"Node {node_id} not found")
+                raise ValueError(f"Node ID {node_id} not found")
 
             return node.public_key
 


framework/py/flwr/server/grid/inmemory_grid_test.py
```diff
@@ -46,7 +47,9 @@ def push_messages(grid: InMemoryGrid, num_nodes: int) -> tuple[Iterable[str], in
     """Help push messages to state."""
     for _ in range(num_nodes):
         node_id = grid.state.create_node(
-            secrets.token_bytes(32), heartbeat_interval=HEARTBEAT_MAX_INTERVAL
+            NOOP_FLWR_AID,
```
jafermarq: in the other tests files we don't use `NOOP_FLWR_AID`, instead we set some ad-hoc ids `mock_flwr_aid` and `mock_owner`. Should we use `NOOP_FLWR_AID` throughout ?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -341,24 +343,25 @@ def create_node(self, public_key: bytes, heartbeat_interval: float) -> int:
...
-            if public_key in self.public_key_to_node_id:
+            if public_key in self.registered_node_public_keys:
                 raise ValueError("Public key already in use")
 
             # Mark the node online until now().timestamp() + heartbeat_interval
             current = now()
             self.nodes[node_id] = NodeInfo(
                 node_id=node_id,
-                owner_aid="",  # Unused for now
+                owner_aid=owner_aid,  # Unused for now
```
jafermarq, line 353:
```suggestion
                owner_aid=owner_aid,
```
Should we remove this comment, it's technically being used (even if `NOOP`). Same applies for `SQLiteLinkState`.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -391,29 +391,13 @@ def get_nodes(self, run_id: int) -> set[int]:
...
-                raise ValueError("Public key already in use")
-
-            self.public_key_to_node_id[public_key] = node_id
-            self.node_id_to_public_key[node_id] = public_key
-
     def get_node_public_key(self, node_id: int) -> Optional[bytes]:
         """Get `public_key` for the specified `node_id`."""
         with self.lock:
-            if node_id not in self.nodes:
-                raise ValueError(f"Node {node_id} not found")
```
jafermarq, line 410: Should we leave this here? checking if the node exists is the first stage also in other methods. (same Q applies to `SQLiteLinkState`

framework/py/flwr/server/grid/inmemory_grid_test.py
panh99: Hmmm, since it's not strictly needed for now, I think we can set some ad-hoc IDs?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: I think it's okay to keep it now. I will remove it in a follow-up PR: https://github.com/adap/flower/pull/5986.

I added this comment b/c, in this PR, we cannot yet access the owner ID, nor filter nodes by the owner ID. Thus it's effectively not in use.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: Sure, np. Then I think we should update its return type to `bytes` instead of `Optional[bytes]`. We used to allow NoneType b/c we still allow registering SuperNode without a public key by that time.

framework/py/flwr/server/grid/inmemory_grid_test.py
panh99: I set it to `"mock_owner"` for now. Lmkwyt!

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -391,29 +391,13 @@ def get_nodes(self, run_id: int) -> set[int]:
...
-            self.public_key_to_node_id[public_key] = node_id
-            self.node_id_to_public_key[node_id] = public_key
-
-    def get_node_public_key(self, node_id: int) -> Optional[bytes]:
+    def get_node_public_key(self, node_id: int) -> bytes:
         """Get `public_key` for the specified `node_id`."""
         with self.lock:
-            if node_id not in self.nodes:
+            if (node := self.nodes.get(node_id)) is None:
                 raise ValueError(f"Node {node_id} not found")
```
jafermarq:
```suggestion
                raise ValueError(f"Node ID {node_id} not found")
```

Just so to have the exact same message here.

PR #5986, opened 2025-10-07
feat(framework): Add `LinkState.get_node_info` and ensure node statuses updated

panh99:


framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -656,22 +658,24 @@ def create_node(
...
 
-        try:
-            with self.conn:
-                rows = self.conn.execute(query, params)
-                if rows.rowcount < 1:
-                    raise ValueError(f"Node {node_id} not found")
-        except KeyError as exc:
-            log(ERROR, {"query": query, "data": params, "exception": exc})
+        row = self.query(query, params)
+        if not row:
```
panh99:
```suggestion
        rows = self.query(query, params)
        if not rows:
```

framework/py/flwr/server/superlink/linkstate/linkstate.py
```diff
@@ -147,6 +149,38 @@ def get_nodes(self, run_id: int) -> set[int]:
...
+        owner_aids : Optional[Sequence[str]] (default: None)
+            Sequence of owner account IDs to filter by. If a sequence is provided,
+            it is treated as an OR condition.
+        statuses : Optional[Sequence[str]] (default: None)
+            Sequence of node status values (e.g., "created", "activated")
+            to filter by. If a sequence is provided, it is treated as an OR condition.
+
+        Returns
+        -------
+        Sequence[NodeInfo]
```
jafermarq, line 179: What's the main motivation for using `Sequence` as opposed to `list` (like in other methods in this class)?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -346,17 +349,16 @@ def create_node(
...
-                last_deactivated_at="",  # Unused for now
-                deleted_at="",  # Unused for now
-                online_until=current.timestamp() + heartbeat_interval,
+                owner_aid=owner_aid,
+                status=NodeStatus.CREATED,
+                created_at=now().isoformat(),
+                last_activated_at="",
+                last_deactivated_at="",
+                deleted_at="",
+                online_until=0.0,
```
danieljanes: For unset values, instead of using a "magic" value with a different meaning (e.g., `str ""` instead of a full `str`), we should rather use `None` b/c it allows the type system to raise errors if some other component forgets to validate if a value is actually set or not.

framework/py/flwr/supercore/constant.py
```diff
@@ -15,5 +15,20 @@
...
 # Top-level key in YAML config for exec plugin settings
 EXEC_PLUGIN_SECTION = "exec_plugin"
+
+
+class NodeStatus:
+    """Event log writer types."""
+
+    CREATED = "created"
+    ACTIVATED = "activated"
+    DEACTIVATED = "deactivated"
```
danieljanes: Shouldn't this be `ONLINE` and `OFFLINE`?

PR #5990, opened 2025-10-10
refactor(framework): Implement `flwr supernode list` functionality

jafermarq:
## Merge after https://github.com/adap/flower/pull/5986

- Implement basic `flwr supernode list` functionality. It shows `SuperNodes` previously created by the same account
- With `-v` _deleted_ supernodes are also shown
- A new flag `--dry-run` is added. When enabled, the `Control API` will return a dummy list of `SuperNodes` with different statuses. This can be used to implement future functionality w/o having to actually create & connect `SuperNodes`.

### DIFF
--- a/framework/py/flwr/cli/supernode/ls.py
+++ b/framework/py/flwr/cli/supernode/ls.py
@@ -77,7 +77,6 @@ def ls(  # pylint: disable=R0914, R0913, R0917
         bool,
         typer.Option(
             "--dry-run",
-            "-dr",
             help="Simulate the command without contacting any SuperNodes",
         ),
     ] = False,


framework/py/flwr/cli/supernode/ls.py
```diff
@@ -73,6 +73,14 @@ def ls(  # pylint: disable=R0914
...
         ),
     ] = False,
+    smoke: Annotated[
+        bool,
+        typer.Option(
+            "--smoke-test",
+            "-st",
+            help="Enable smoke testing",
+        ),
+    ] = False,
```
tanertopal: I wouldn't include this in the public API. I've never seen anything like it before. A better alternative might be a `--dry-run` option, which simulates the commands behavior without making real network calls or state changes. That pattern is much more common in CLI's.

```python
dry_run: Annotated[
    bool,
    typer.Option(
        "--dry-run",
        help="Simulate the command without contacting any SuperNodes",
    ),
] = False,
```

framework/py/flwr/cli/supernode/ls.py
```diff
@@ -73,12 +73,12 @@ def ls(  # pylint: disable=R0914, R0913, R0917
...
     ] = False,
-    smoke: Annotated[
+    dry_run: Annotated[
         bool,
         typer.Option(
-            "--smoke-test",
-            "-st",
-            help="Enable smoke testing",
+            "--dry-run",
+            "-dr",
```
tanertopal: I'd skip the short version here. Short flags are usually reserved for options that users invoke frequently, and `--dry-run` tends to be used only occasionally.

PR #5991, opened 2025-10-10
refactor(framework): Check for `ssh` public key upon `flwr supernode create`

jafermarq:
- Node auth expects NIST EC keys in ssh format, but the check wasn't ready for it. 
- We call `flwr_exit` pointing users to an error page

### DIFF
--- a/framework/docs/source/build.rst
+++ b/framework/docs/source/build.rst
@@ -25,6 +25,7 @@ Problem-oriented how-to guides show step-by-step how to achieve a specific goal.
     how-to-use-differential-privacy
     how-to-implement-fedbn
     how-to-use-cli-json-output
+    how-to-migrate-from-openfl
     how-to-upgrade-to-flower-1.0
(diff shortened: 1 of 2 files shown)


framework/py/flwr/common/exit/exit_code.py
```diff
@@ -116,6 +119,13 @@ def __new__(cls) -> ExitCode:
     ExitCode.SUPEREXEC_INVALID_PLUGIN_CONFIG: (
         "The YAML configuration for the SuperExec plugin is invalid."
     ),
+    # FlowerCLI-specific exit codes (500-599)
+    ExitCode.NODE_AUTH_PUBLIC_KEY_INVALID: (
```
danieljanes: Let's namespace CLI exit codes like the other ones:
```suggestion
    ExitCode.FLWRCLI_NODE_AUTH_PUBLIC_KEY_INVALID: (
```

PR #5997, opened 2025-10-13
refactor(framework): Remove `CSV`-based node auth in `SuperLink`

jafermarq:
## Merge after https://github.com/adap/flower/pull/5980

Node authentication can be enabled by passing flag `--enable-supernode-auth` to the `SuperLink`. Then `SuperNodes` need to be registered via `flwr supernode create <pub-key>` before they can connect.

### DIFF
--- a/framework/py/flwr/cli/utils.py
+++ b/framework/py/flwr/cli/utils.py
@@ -24,7 +24,6 @@ from pathlib import Path
 from typing import Any, Callable, Optional, Union, cast
 
 import grpc
-
 import typer
 
 from flwr.common.constant import (


framework/e2e/test_control_api.sh
```diff
@@ -79,6 +79,12 @@ timeout 2m flower-superlink $combined_args &
 sl_pid=$(pgrep -f "flower-superlink")
 sleep 2
 
+if [ "$2" = "client-auth" ]; then
+  # Create SuperNodes using the Flower CLI
```
jafermarq:
```suggestion
  # Create two SuperNodes using the Flower CLI
```

framework/py/flwr/server/app.py
```diff
@@ -738,24 +687,13 @@ def _add_args_common(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         "--auth-list-public-keys",
         type=str,
-        help="A CSV file (as a path str) containing a list of known public "
-        "keys to enable authentication.",
+        help="This argument is deprecated and will be removed in a future release.",
```
panh99, line 694: Would it be possible to point users to a page explaining how to do node auth with the new version of Flower?

panh99: But it's also true that we will print a helpful prompt when this flag is enabled. I think there's no need to change the help message here. Wdyt? Feel free to resolve this.

framework/py/flwr/server/app.py
```diff
@@ -223,6 +219,15 @@ def run_superlink() -> None:
             "connections from any SuperNode.",
         )
 
+    if args.auth_list_public_keys:
+        flwr_exit(
+            ExitCode.SUPERLINK_INVALID_ARGS,
```
panh99, line 227: Should we create a new error code for this, say, `SUPERLINK_DEPRECATED_NODE_AUTH`? In the error page, we can just say the old approach is no longer supported, and link to the updated node auth page.

framework/py/flwr/server/app.py
jafermarq: yes, we could point to the docs page (which is going to be updated soon): https://flower.ai/docs/framework/how-to-authenticate-supernodes.html

framework/py/flwr/server/app.py
jafermarq: I think it's better to add link to docs page in the `flwr_exit` comment instead of creating a new error (since this error code will be remove once we remove the old CSV flag altogether). wdyt?

framework/py/flwr/server/app.py
```diff
@@ -223,6 +220,19 @@ def run_superlink() -> None:
             "connections from any SuperNode.",
         )
 
+    if args.auth_list_public_keys:
+        _, v = _check_package("flwr")
```
panh99: Quick question: why not use package_version from flwr.common.version? 

Btw, do you think it makes sense to create a helper function to build the link to docs? Like return f"https://flower.ai/docs/framework/v{v}/en/" for flwr, and “https://flower.ai/docs/framework/main/en/" for flwr-nightly

framework/py/flwr/server/app.py
panh99: As discussed, we won't introduce a error code. Instead, a help URL will be printed.

framework/py/flwr/server/app.py
panh99: As discussed, we will print a help URL if the flag is passed and keep this line as it is.

PR #6003, opened 2025-10-14
fix(framework): Ensure no nodes with same identity can be connected simultaneously

jafermarq:


framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -88,6 +89,17 @@ def CreateNode(
 
             # Check if public key is already in use
             if node_id := state.get_node_id_by_public_key(request.public_key):
+
+                node_info = state.get_node_info(node_ids=[node_id])[0]
+                if node_info.status == NodeStatus.ONLINE:
```
panh99: Btw, `InMemoryLinkState.get_node_id_by_public_key` can return public keys of deleted nodes. Could you also fix that?

panh99: This does not really support concurrency, IMO. Say, two SuperNodes start at the same time, and one would die when sending the first heartbeat. Should we invoke `LinkState.acknowledge_node_heartbeat` here?

jafermarq: good point, yes, let's do that

panh99: As discussed, this will be done in a separate PR.

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -88,8 +89,24 @@ def CreateNode(
...
+                    )
+                    raise ValueError("Public key already in use by an active SuperNode")
+
                 # Prepare response with existing node_id
                 response = CreateNodeResponse(node=Node(node_id=node_id))
+                # Awknowledge heartbeat to mark node as online
+                state.acknowledge_node_heartbeat(
+                    node_id=node_id,
+                    heartbeat_interval=request.heartbeat_interval,
+                )
```
panh99: Maybe it's good to do this right after the check above? I am still a bit worried about the race condition:

Node 1 and node 2 send requests in the same time, and the `node_info.status` in both RPCs show the non-online status, and they both call `acknowledge_node_heartbeat`.

But we might over-think this. This can only happen when the same private key is used by both the nodes, which is more of a user problem. Wdyt?

jafermarq: handled with a threading lock

danieljanes (approved):

PR #6005, opened 2025-10-14
feat(framework): Verify FAB hash on SuperNode

psfoley:


framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -296,6 +297,14 @@ def _pull_and_store_message(  # pylint: disable=too-many-positional-arguments
             fab = get_fab(run_info.fab_hash, run_id)
             ffs.put(fab.content, {})
 
+            # Verify the FAB hash
+            verified_fab_hash = hashlib.sha256(fab.content).hexdigest()
+            if verified_fab_hash != fab.hash_str:
+                raise RuntimeError(
```
psfoley, line 303: Note: a `RuntimeError` is thrown for consistency with SuperLink when it receives a FAB contents that can't be verified: https://github.com/adap/flower/blob/main/framework/py/flwr/superlink/servicer/control/control_servicer.py#L136-L138

panh99: Thanks for pointing me to the `ControlServicer`. I think there’s a key difference between how it’s run by the SuperLink and how it works here: raising a `RuntimeError` won’t crash the SuperLink, but in this case, it could crash the SuperNode since the exception is raised in the main thread.

More broadly, one of our goals is to ensure the robustness of the SuperNode, i.e., it shouldn’t crash due to **recoverable** issues. So, we have two possible approaches here:

1. **Handle gracefully:** Generate an error message as the reply to the original instruction message and suppress the exception, allowing the SuperNode to continue running.
2. **Fail fast:** Crash the SuperNode (either by raising an exception or by setting an exit code and calling the `flwr_exit` function). Receiving an invalid FAB from the SuperLink could indicate that the SuperLink is compromised, in which case it might not be safe to maintain the connection.

Personally, I lean toward option 2, which is essentially what you’re doing here, but I’m curious to hear your thoughts.
Cc @danieljanes and @jafermarq

PR #6020, opened 2025-10-20
refactor(framework): Adjust constants for `Object` communication

jafermarq:
New defaults, all can be adjusted via env vars:

- `MAX_CONCURRENT_PUSHES` defaults to 2 (down from 8)
- `MAX_CONCURRENT_PUSHES` defaults to 2 (down from 8)
- `MAX_ARRAY_CHUNK_SIZE` defaults to 5 MB (down from 20MB)

### DIFF
--- a/framework/dev/update_changelog.py
+++ b/framework/dev/update_changelog.py
@@ -217,6 +217,14 @@ def _extract_changelog_entry(
     """Extract the changelog entry from a pull request's body."""
     # Use regex search to find matches
     match = re.search(PATTERN, pr_info.title)
+
+    # Extract the topic label
+    topic = ""
+    for label in pr_info.labels:
+        if label.name not in ["Maintainer", "Contributor", "Bot", "General"]:
+            topic = label.name
+            break
+
     if match:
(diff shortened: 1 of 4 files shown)


framework/py/flwr/common/constant.py
```diff
@@ -144,8 +146,12 @@
 FLWR_APP_TOKEN_LENGTH = 128  # Length of the token used
 
 # Constants for object pushing and pulling
-MAX_CONCURRENT_PUSHES = 8  # Default maximum number of concurrent pushes
-MAX_CONCURRENT_PULLS = 8  # Default maximum number of concurrent pulls
+MAX_CONCURRENT_PUSHES = int(
+    os.getenv("MAX_CONCURRENT_PUSHES", "1")
```
jafermarq:
```suggestion
    os.getenv("MAX_CONCURRENT_PUSHES", "2")
```

As suggested by @panh99

panh99:
The changes look good to me. However, are we sure to introduce three more env vars?

PR #6040, opened 2025-10-22
refactor(framework): Remove `SqliteMixin.transaction` and add unit tests

panh99:
I just realized that sqlite's native `self._conn` already do the same thing... No need to have the method.

### DIFF
--- a/framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
+++ b/framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
@@ -52,6 +52,7 @@ from flwr.proto.recorddict_pb2 import RecordDict as ProtoRecordDict
 from flwr.server.utils.validator import validate_message
 from flwr.supercore.constant import NodeStatus
 from flwr.supercore.sqlite_mixin import SqliteMixin
+from flwr.supercore.utils import int64_to_uint64, uint64_to_int64
 
 from .linkstate import LinkState
 from .utils import (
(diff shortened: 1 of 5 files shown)


framework/py/flwr/supercore/sqlite_mixin_test.py
```diff
@@ -0,0 +1,70 @@
...
+    def initialize(self, log_queries: bool = False) -> list[tuple[str]]:
+        """Initialize the database with a simple test table."""
+        return self._ensure_initialized(
+            "CREATE TABLE IF NOT EXISTS test"
+            "(id INTEGER PRIMARY KEY AUTOINCREMENT, value INTEGER)"
+        )
+
+
+def test_transaction_serialization_with_tempfile() -> None:
+    """Verify that .transaction() runs inside real SQLite transactions."""
```
jafermarq: Should we rephrase this given that we removed `.transaction()`?

panh99:
```suggestion
    """Verify that .conn runs inside real SQLite transactions."""
```

framework/py/flwr/supercore/sqlite_mixin_test.py
```diff
@@ -0,0 +1,70 @@
...
+    def initialize(self, log_queries: bool = False) -> list[tuple[str]]:
+        """Initialize the database with a simple test table."""
+        return self._ensure_initialized(
+            "CREATE TABLE IF NOT EXISTS test"
+            "(id INTEGER PRIMARY KEY AUTOINCREMENT, value INTEGER)"
+        )
+
+
+def test_transaction_serialization_with_tempfile() -> None:
+    """Verify that .conn runs inside real SQLite transactions."""
```
panh99:
```suggestion
    """Verify that `.conn` runs inside real SQLite transactions."""
```

jafermarq (approved):

danieljanes (approved):

PR #6047, opened 2025-10-23
feat(framework): Add meta to FAB

yan-gao-GY:
## Proposal

This PR adds a meta attribute to both FAB dataclass and protobuf, and updates all related references accordingly.

### DIFF
--- a/framework/dev/update_changelog.py
+++ b/framework/dev/update_changelog.py
@@ -175,6 +175,7 @@ def _get_pull_requests_since_tag(
 
     print("Retrieving contributors...")
     contributors = _get_contributors_from_commits(api, commits)
+    print(f"Found following contributors:\n{', '.join(sorted(contributors))}\n")
 
     print("Retrieving pull requests...")
     commit_shas = {commit.hexsha for commit in commits}
@@ -279,16 +280,6 @@ def _update_changelog(prs: set[PullRequest], tag: str, new_tag: str) -> bool:
         for pr_info in prs:
             parsed_title = _extract_changelog_entry(pr_info)
 
-            # Create the topic if not found
-            if (topic := parsed_title["topic"]) and topic not in topic_to_section:
-                section = f"### {topic}"
-                if content.find(section, unreleased_index, end_index) == -1:
-                    content = (
-                        content[:end_index] + f"\n{section}\n\n" + content[end_index:]
-                    )
-                    end_index = content.find(f"## {tag}", end_index)
-                topic_to_section[topic] = section
-
             # Skip if the PR is already in changelog
             if f"#{pr_info.number}]" in content:
                 continue
@@ -296,6 +287,16 @@ def _update_changelog(prs: set[PullRequest], tag: str, new_tag: str) -> bool:
             # Skip Flower Intelligence PRs
             if parsed_title["project"] == "intelligence":
                 continue
+
+            # Create the topic if not found
+            if (topic := parsed_title.get("topic")) and topic not in topic_to_section:
+                section = f"### {topic}"
+                if content.find(section, unreleased_index, end_index) == -1:
+                    content = (
+                        content[:end_index] + f"\n{section}\n\n" + content[end_index:]
+                    )
+                    end_index = content.find(f"## {tag}", end_index)
+                topic_to_section[topic] = section
 
             # Find section to insert
             pr_type = parsed_title.get("type", "unknown")


framework/py/flwr/client/rest_client/connection.py
```diff
@@ -398,11 +398,12 @@ def get_fab(fab_hash: str, run_id: int) -> Fab:
         # Send the request
         res = _request(req, GetFabResponse, PATH_GET_FAB)
         if res is None:
-            return Fab("", b"")
+            return Fab("", b"", {})
 
         return Fab(
             res.fab.hash_str,
             res.fab.content,
+            dict(res.fab.meta),
```
panh99: Is this to make `mypy` happy? The gRPC mapping is already a dict-like object.

panh99: Btw, I think we can use `serde.fab_from_proto`, right?

framework/py/flwr/common/serde_test.py
```diff
@@ -138,9 +146,17 @@ def test_fab_to_proto() -> None:
 
 def test_fab_from_proto() -> None:
     """Test Fab deserialization."""
-    proto_fab = ProtoFab(hash_str="fab_test_hash", content=b"fab_test_content")
+    proto_fab = ProtoFab(
+        hash_str="fab_test_hash",
+        content=b"fab_test_content",
+        meta={"fab_test_meta": "fab_test_meta"},
```
panh99:
```suggestion
        meta={"meta_key": "meta_value"},
```

framework/py/flwr/common/serde_test.py
```diff
@@ -138,9 +146,17 @@ def test_fab_to_proto() -> None:
...
+    )
 
-    py_fab = typing.Fab(hash_str="fab_test_hash", content=b"fab_test_content")
+    py_fab = typing.Fab(
+        hash_str="fab_test_hash",
+        content=b"fab_test_content",
+        meta={"fab_test_meta": "fab_test_meta"},
```
panh99:
```suggestion
        meta={"meta_key": "meta_value"},
```

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -316,7 +316,7 @@ def GetFab(
 
         ffs: Ffs = self.ffs_factory.ffs()
         if result := ffs.get(request.hash_str):
-            fab = Fab(request.hash_str, result[0])
+            fab = Fab(request.hash_str, result[0], {})
```
panh99: Why are we inserting an empty dict here instead of `result[1]`?

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -343,7 +343,7 @@ def PullAppInputs(
             fab = None
             if run and run.fab_hash:
                 if result := ffs.get(run.fab_hash):
-                    fab = Fab(run.fab_hash, result[0])
+                    fab = Fab(run.fab_hash, result[0], {})
```
panh99: Same question here.

framework/py/flwr/server/superlink/simulation/simulationio_servicer.py
```diff
@@ -150,7 +150,7 @@ def PullAppInputs(
             fab = None
             if run and run.fab_hash:
                 if result := ffs.get(run.fab_hash):
-                    fab = Fab(run.fab_hash, result[0])
+                    fab = Fab(run.fab_hash, result[0], {})
```
panh99: Also here.

framework/py/flwr/supercore/ffs/disk_ffs.py
```diff
@@ -52,10 +52,11 @@ def put(self, content: bytes, meta: dict[str, str]) -> str:
             The key (sha256hex hash) of the content.
         """
         content_hash = hashlib.sha256(content).hexdigest()
+        meta_dict = dict(meta)
```
panh99: Why adding the conversion here given that the method already specifies `meta: dict[str, str]`? Do you see a place where a non-dict is passed to `Ffs.put` without triggering `mypy` errors?

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -129,7 +129,11 @@ def StartRun(  # pylint: disable=too-many-locals
                 )
 
             # Create run
-            fab = Fab(hashlib.sha256(fab_file).hexdigest(), fab_file)
+            fab = Fab(
+                hashlib.sha256(fab_file).hexdigest(),
+                fab_file,
+                {},
```
panh99: Let's pass the metadata from the request to the `Fab` instance here. Though it's 100% empty for now, it's good to still pass whatever in the request to the `Fab` constructor. Wdyt?

framework/py/flwr/supernode/servicer/clientappio/clientappio_servicer.py
```diff
@@ -151,7 +151,7 @@ def PullClientAppInputs(
         # Retrieve context, run and fab for this run
         context = cast(Context, state.get_context(run_id))
         run = cast(Run, state.get_run(run_id))
-        fab = Fab(run.fab_hash, ffs.get(run.fab_hash)[0])  # type: ignore
+        fab = Fab(run.fab_hash, ffs.get(run.fab_hash)[0], {})  # type: ignore
```
panh99: Same here.

framework/py/flwr/client/rest_client/connection.py
yan-gao-GY: Yes it will get this `Argument 3 to "Fab" has incompatible type "ScalarMap[str, str]"; expected "dict[str, str]"  [arg-type]` without `dict()`.

and yes we can use `serde.fab_from_proto` instead.

Do you think the implementation of `serde.fab_from_proto` below is correct?

```
def fab_from_proto(fab: ProtoFab) -> typing.Fab:
    """Create a Python Fab object from a proto Fab."""
    return typing.Fab(fab.hash_str, fab.content, dict(fab.meta))
```

panh99: Thanks for the clarification. The implementation looks good, and I believe you include a unit test for it. Do you have any other concerns?

framework/py/flwr/supercore/ffs/disk_ffs.py
yan-gao-GY: You're right, I've removed the conversion.

PR #6052, opened 2025-10-24
refactor(framework): Make `InMemoryLinkState.get_node_info` faster

panh99:
Also separate the status checking logic to a new method in `SqliteLinkState`.

### DIFF
--- a/framework/py/flwr/cli/build.py
+++ b/framework/py/flwr/cli/build.py
@@ -17,6 +17,7 @@
 
 import hashlib
 import zipfile
+from collections.abc import Iterable
 from io import BytesIO
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -99,6 +99,10 @@
 CREATE INDEX IF NOT EXISTS idx_node_owner_aid ON node(owner_aid);
 """
 
+SQL_CREATE_INDEX_NODE_STATUS = """
+CREATE INDEX IF NOT EXISTS idx_node_status ON node(status);
```
jafermarq, line 103: is this being used?

PR #6058, opened 2025-10-28
refactor(framework): Refactor `build_fab` into `build_fab_from_files` and `build_fab_from_disk`

panh99:


framework/py/flwr/cli/build.py
```diff
@@ -140,18 +143,53 @@ def build_fab(app: Path) -> tuple[bytes, str, dict[str, Any]]:
...
+        Dictionary mapping relative file paths to their contents.
+        - Keys: Relative paths (strings with forward slashes)
+        - Values: Either bytes (file contents) or Path (will be read)
+        Must include "pyproject.toml" and optionally ".gitignore".
+
+    Returns
+    -------
+    bytes
+        The FAB as bytes.
+    """
```
tanertopal, line 197: Can you add an example of the expected data structure - especially the str keys in the dict

panh99: Sure, that makes sense. I've just added an example. Lmkwyt!

PR #6060, opened 2025-10-28
feat(framework): Add `LinkState.activate_node` and `LinkState.deactivate_node` methods

panh99:


framework/py/flwr/server/superlink/linkstate/linkstate.py
```diff
@@ -139,6 +139,45 @@ def create_node(
     def delete_node(self, owner_aid: str, node_id: int) -> None:
         """Remove `node_id` from the link state."""
 
+    @abc.abstractmethod
+    def activate_node(self, node_id: int, heartbeat_interval: float) -> bool:
```
danieljanes, line 143: Shouldn't it be `public_key` instead of `node_id`?

panh99: Good question. I was thinking to use `.get_node_id_by_public_key` and then `.activate_node`, but it is indeed more straight-forward to take the public key directly here. After this change, should the method return the node ID, or still return whether the activation is successful?

panh99: General question: Should I raise exceptions when activation/deactivation fails? I make the methods return True/False for now.

danieljanes: Got it, using `node_id` seems ok then. I prefer returning `True/False`, especially if the error is something we expect to happen in certain situations.

PR #6061, opened 2025-10-28
feat(framework): Implement `Fleet.Register/Activate/Deactivate/UnregisterNode` RPCs

panh99:
### ~~Merge after #6051 and #6060~~

PR #6063, opened 2025-10-30
feat(framework): Use new node RPCs

panh99:
### Merge after #6061


framework/py/flwr/client/rest_client/connection.py
```diff
@@ -447,12 +476,13 @@ def confirm_message_received(run_id: int, object_id: str) -> None:
         fn(object_id)
 
     try:
+        register_node()
```
jafermarq: Should this be instead:

```python
        if self_registered:
            register_node()
```
as done in `grpc_rere`? (similarly later for `unregister_node()`)

framework/py/flwr/server/superlink/fleet/grpc_rere/node_auth_server_interceptor_test.py
```diff
@@ -293,48 +319,48 @@ def _test_get_fab(self, metadata: list[Any]) -> Any:
...
+    @parameterized.expand(rpcs)  # type: ignore
     def test_successful_rpc_with_metadata(
         self, rpc: Callable[[Any, list[Any]], Any]
     ) -> None:
         """Test server interceptor for RPC."""
+        # Skip registration and unregistration tests when node auth is enabled
+        if self.enable_node_auth and rpc.__name__ in [
+            self._test_register_node.__name__,
+            self._test_unregister_node.__name__,
+        ]:
```
jafermarq, line 354: Should we have a test that checks for a `context.abort` in the scenario when these RPCs are called and node auth is enabled?

panh99: That has been covered by `fleet_servicer_test.py`. And it's also good to make this file focuses on unit tests solely for the interceptor, as the checks for `enable_supernode_auth` lives in `FleetServicer`.

jafermarq (approved):

framework/py/flwr/supernode/cli/flower_supernode.py
```diff
@@ -61,6 +61,13 @@ def flower_supernode() -> None:
     root_certificates = try_obtain_root_certificates(args, args.superlink)
     authentication_keys = _try_setup_client_authentication(args)
 
+    # Warn if authentication keys are provided but transport is not grpc-rere
+    if authentication_keys is not None and args.transport != TRANSPORT_TYPE_GRPC_RERE:
```
danieljanes, line 65: Shouldn't we immediately exit here?

panh99: Should we allow `flwr supernode register` when SuperLink uses REST API?

danieljanes: Not sure, either option could be fine

danieljanes (approved):

PR #6067, opened 2025-10-31
feat(framework): Introduce base `FederationManager`

jafermarq:


framework/py/flwr/superlink/federation/federation_manager.py
```diff
@@ -0,0 +1,43 @@
...
+from abc import ABC, abstractmethod
+
+from flwr.app import Message
+
+
+class FederationManager(ABC):
+    """Abstract base class for FederationManager."""
+
+    @abstractmethod
+    def exists(self, federation_name: str) -> bool:
```
panh99: Quick question: could we just refer to it as federation? The suffix seems a bit long and unnecessary.

```suggestion
    def exists(self, federation: str) -> bool:
```

framework/py/flwr/superlink/federation/federation_manager.py
```diff
@@ -0,0 +1,43 @@
...
+        """Check if a federation exists."""
+
+    @abstractmethod
+    def is_member(self, federation_name: str, flwr_aid: str) -> bool:
+        """Check if a member of the federation."""
+
+    @abstractmethod
+    def filter_nodes(self, node_ids: list[int], federation_name: str) -> list[int]:
```
panh99: If we wanna use this with `LinkState.get_nodes(...) -> set[int]`, it's better to use the same type.

```suggestion
    def filter_nodes(self, node_ids: set[int], federation_name: str) -> set[int]:
```

framework/py/flwr/superlink/federation/federation_manager.py
```diff
@@ -0,0 +1,43 @@
...
+        """Given a list of node IDs, return sublist with nodes in federation."""
+
+    @abstractmethod
+    def filter_messages(
+        self, messages: list[Message], federation_name: str
+    ) -> list[Message]:
```
panh99: Could we replace this with `has_node`? This doesn't work well with our current logic here: https://github.com/adap/flower/blob/main/framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py#L204

framework/py/flwr/supercore/constant.py
```diff
@@ -27,6 +27,9 @@
 APP_ID_PATTERN = r"^@(?P<user>[^/]+)/(?P<app>[^/]+)$"
 PLATFORM_API_URL = "https://api.flower.ai/v1"
 
+# Constants for federations
+NOOP_FEDERATION_NAME = "default"
```
panh99: Should we following the naming convention?

```suggestion
NOOP_FEDERATION = "default"
```

panh99 (dismissed):
LGTM!

framework/py/flwr/superlink/federation/noop_federation_manager.py
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""NoOp implementation of FederationManager."""
+
+from flwr.app import Message
+from flwr.supercore.constant import NOOP_FEDERATION_NAME
```
panh99:
```suggestion
from flwr.supercore.constant import NOOP_FEDERATION
```

framework/py/flwr/superlink/federation/noop_federation_manager.py
```diff
@@ -0,0 +1,40 @@
...
+
+from .federation_manager import FederationManager
+
+
+class NoOpFederationManager(FederationManager):
+    """No-Op FederationManager implementation."""
+
+    def exists(self, federation: str) -> bool:
+        """Check if a federation exists."""
+        return federation == NOOP_FEDERATION_NAME
```
panh99:
```suggestion
        return federation == NOOP_FEDERATION
```

danieljanes (approved):

PR #6077, opened 2025-11-03
feat(framework): Add fab error code

yan-gao-GY:
## Proposal

Add fab error code for SuperNode trusted entities.


framework/py/flwr/common/exit/exit_code.py
```diff
@@ -115,6 +116,10 @@ def __new__(cls) -> ExitCode:
         "The private key for SuperNode authentication was provided, but TLS is not "
         "enabled. Node authentication can only be used when TLS is enabled."
     ),
+    ExitCode.SUPERNODE_TRUST_ENTITY_REQUIRED: (
+        "Entity verification is enabled, but no trusted entities were provided. "
+        "The Flower SuperNode should be restarted with a trusted entities list."
```
panh99:
```suggestion
        "Please ensure that a trusted entities list is provided."
```

framework/docs/source/ref-exit-codes/304.rst
```diff
@@ -0,0 +1,33 @@
...
+  .. code-block:: yaml
+
+      alice-2025: |
+        -----BEGIN PUBLIC KEY-----
+        ...
+        -----END PUBLIC KEY-----
+      bob-2025: |
+        -----BEGIN PUBLIC KEY-----
+        ...
+        -----END PUBLIC KEY-----
```
panh99: This seems a bit different than what we actually use. Would you mind changing the key IDs like `alice-2025` to real key ID format we use in the Platform API? I think it's UUID with the account ID as prefix?

Plus, IIUC, we now only accept ssh format ed25519 public keys for entity verification, right? So the format should be:

```
ssh-ed25519 <base64-encoded-key> [comment]
```

The format you used here is actually PEM, not ssh.

framework/docs/source/ref-exit-codes/304.rst
```diff
@@ -0,0 +1,33 @@
...
+SuperNode should be restarted with a valid list of trusted entities.
+
+How to Resolve
+--------------
+
+To fix this issue, you can either:
+
+- Provide a trusted entities list with ``--trust-entities`` when launching the
+  SuperNode. Supply a YAML file containing the trusted entities' public key IDs and
+  their corresponding public keys when launching the SuperNode. The file should follow
```
panh99: Should we add an example usage?

```suggestion
  their corresponding public keys when launching the SuperNode, e.g., `--trust-entities ./pks.yaml`. The file should follow
```

framework/py/flwr/common/exit/exit_code.py
```diff
@@ -115,6 +116,10 @@ def __new__(cls) -> ExitCode:
         "The private key for SuperNode authentication was provided, but TLS is not "
         "enabled. Node authentication can only be used when TLS is enabled."
     ),
+    ExitCode.SUPERNODE_TRUST_ENTITY_REQUIRED: (
+        "Entity verification is enabled, but no trusted entities were provided. "
+        "Please ensure that a trusted entities list is provided."
```
panh99:
```suggestion
        "Please ensure that a trusted entities list is provided via `--trust-entities`."
```

framework/docs/source/ref-exit-codes/304.rst
yan-gao-GY: Yes! that's good

framework/docs/source/ref-exit-codes/304.rst
yan-gao-GY: Yes you're right~ I just double checked with platform team. The format of public key is ed25519 OpenSSH, and the format of public key ID is `fpk_<UUID>`, where `fpk` is an acronym for Flower Public Key. I've updated the related text.

PR #6083, opened 2025-11-04
feat(framework): Update `flwr run`

yan-gao-GY:
## Proposal

Update `flwr run`.

### DIFF
--- a/framework/py/flwr/cli/run/run.py
+++ b/framework/py/flwr/cli/run/run.py
@@ -103,7 +103,7 @@ def run(
         if suppress_output:
             redirect_output(captured_output)
 
-        app_id_str = str(app) if app is not None else ""
+        # Determine if app is remote
         app_id = None
         if (app_str := str(app)).startswith("@"):
             if not re.match(r"^@(?P<user>[^/]+)/(?P<app>[^/]+)$", app_str):
@@ -111,18 +111,17 @@ def run(
                     "Invalid remote app ID. Expected format: '@user_name/app_name'."
                 )
             app_id = app_str
+        is_remote_app = app_id is not None
 
         typer.secho("Loading project configuration... ", fg=typer.colors.BLUE)
 
-        # Disable the validation due to the local empty project
-        if is_remote_app:
-            config = load_toml(Path("pyproject.toml"))
-            if not config:
-                raise typer.BadParameter("A 'pyproject.toml' file is required.")
-        else:
-            pyproject_path = app / "pyproject.toml" if app else None
-            config, errors, warnings = load_and_validate(path=pyproject_path)
-            config = process_loaded_project_config(config, errors, warnings)
+        # Disable the validation for remote apps
+        pyproject_path = app / "pyproject.toml" if not is_remote_app else None
+        # `./pyproject.toml` will be loaded when `pyproject_path` is None
+        config, errors, warnings = load_and_validate(
+            pyproject_path, check_module=not is_remote_app
+        )
+        config = process_loaded_project_config(config, errors, warnings)
 
         federation, federation_config = validate_federation_in_project_config(
             federation, config, federation_config_overrides
@@ -136,8 +135,7 @@ def run(
                 run_config_overrides,
                 stream,
                 output_format,
-                app_id_str,
-                is_remote_app,
+                app_id,
             )
         else:
             _run_without_control_api(
@@ -168,26 +166,27 @@ def _run_with_control_api(
     config_overrides: Optional[list[str]],
     stream: bool,
     output_format: str,
-    app_id_str: str,
-    is_remote_app: bool,
+    app_id: Optional[str],
 ) -> None:
     channel = None
+    is_remote_app = app_id is not None
     try:
         auth_plugin = load_cli_auth_plugin(app, federation, federation_config)
         channel = init_channel(app, federation_config, auth_plugin)
         stub = ControlStub(channel)
 
-        # Build fab only if not a remote reference
-        fab_id = fab_version = fab_hash = ""
-        if is_remote_app:
-            # Skip build; send a placeholder Fab containing the remote reference
-            fab = Fab(fab_hash, b"", {})
-        else:
+        # Build FAB if local app
+        if not is_remote_app:
             fab_bytes = build_fab_from_disk(app)
             fab_hash = hashlib.sha256(fab_bytes).hexdigest()
             config = cast(dict[str, Any], load_toml(app / FAB_CONFIG_FILE))
             fab_id, fab_version = get_metadata_from_config(config)
             fab = Fab(fab_hash, fab_bytes, {})
+        # Skip FAB build if remote app
+        else:
+            # Use empty values for FAB
+            fab_id = fab_version = fab_hash = ""
+            fab = Fab(fab_hash, b"", {})
 
         real_federation: str = federation_config.get("federation", NOOP_FEDERATION)
 
@@ -200,7 +199,7 @@ def _run_with_control_api(
             override_config=user_config_to_proto(parse_config_args(config_overrides)),
             federation=real_federation,
             federation_options=config_record_to_proto(c_record),
-            app_id=app_id_str,
+            app_id=app_id or "",
         )
         with flwr_cli_grpc_exc_handler():
             res = stub.StartRun(req)


framework/py/flwr/cli/run/run.py
```diff
@@ -100,27 +101,53 @@ def run(
     try:
         if suppress_output:
             redirect_output(captured_output)
+
+        original_app_str = str(app) if app is not None else ""
+        remote_app_ref: Optional[str] = None  # "user_name/app_name" if given with "@"
```
jafermarq: we only use this variable as a boolean. Should we define it as such and set it to `True` if the app starts with `@`?

framework/py/flwr/cli/run/run.py
```diff
@@ -100,27 +101,53 @@ def run(
...
+                raise typer.BadParameter(
+                    "Invalid remote app ID. Expected format: '@user_name/app_name'."
+                )
+            app_name = m.group("app")
+            user_name = m.group("user")
+
+            # Use local folder named {app_name} for pyproject lookup
+            # and downstream calls
+            app_path = Path(app_name)
+            remote_app_ref = f"{user_name}/{app_name}"
```
jafermarq: We don't use this string except as a boolean. Let's set it to `True`?

framework/py/flwr/cli/run/run.py
```diff
@@ -100,27 +101,53 @@ def run(
     try:
         if suppress_output:
             redirect_output(captured_output)
+
+        original_app_str = str(app) if app is not None else ""
```
jafermarq: Do we need to introduce this variable? we already pass `app_path: Path = app` around and to `_run_with_control_api`. I think we can remove `original_app_str`

framework/py/flwr/cli/run/run.py
```diff
@@ -178,23 +213,34 @@ def _run_with_control_api(
...
-                    "fab-hash": fab_hash[:8],
-                    "fab-filename": get_fab_filename(config, fab_hash),
-                }
-            )
+            # Only include FAB metadata if we actually built a local FAB
+            payload: dict[str, Any] = {
+                "success": res.HasField("run_id"),
+                "run-id": res.run_id if res.HasField("run_id") else None,
+            }
+            if not remote_app_ref and fab_id and fab_version and fab_hash:
```
jafermarq:
```suggestion
            if not remote_app_ref:
```

Would this be sufficient?

framework/py/flwr/cli/run/run.py
yan-gao-GY: I agree~ thanks!

framework/py/flwr/cli/run/run.py
yan-gao-GY: We need to pass the `original_app_str` to SuperLink and further to Platform API for verification. I renamed it to `app_id_str`. Wdyt?

jafermarq: sounds good!

framework/py/flwr/cli/run/run.py
yan-gao-GY: I changed it back. We need `fab_id and fab_version and fab_hash` to avoid some mypy test errors.

framework/py/flwr/cli/run/run.py
```diff
@@ -148,20 +170,28 @@ def _run_with_control_api(
...
-        fab_bytes = build_fab_from_disk(app)
-        fab_hash = hashlib.sha256(fab_bytes).hexdigest()
-        config = cast(dict[str, Any], load_toml(app / FAB_CONFIG_FILE))
-        fab_id, fab_version = get_metadata_from_config(config)
-        real_federation: str = federation_config.get("federation", NOOP_FEDERATION)
+        # Build fab only if not a remote reference
+        fab_id = fab_version = fab_hash = None
+        if is_remote_app:
+            # Skip build; send a placeholder Fab containing the remote reference
+            fab = Fab("", b"", {})
```
jafermarq:
```suggestion
        fab_id = fab_version = fab_hash = ""
        if is_remote_app:
            # Skip build; send a placeholder Fab containing the remote reference
            fab = Fab(fab_hash, b"", {})
```

framework/py/flwr/cli/run/run.py
```diff
@@ -181,23 +212,34 @@ def _run_with_control_api(
...
-                    "fab-hash": fab_hash[:8],
-                    "fab-filename": get_fab_filename(config, fab_hash),
-                }
-            )
+            # Only include FAB metadata if we actually built a local FAB
+            payload: dict[str, Any] = {
+                "success": res.HasField("run_id"),
+                "run-id": res.run_id if res.HasField("run_id") else None,
+            }
+            if not is_remote_app and fab_id and fab_version and fab_hash:
```
jafermarq:
```suggestion
            if not is_remote_app:
```

framework/py/flwr/cli/run/run.py
jafermarq: I left a suggestion about how to go around this

framework/py/flwr/cli/run/run.py
```diff
@@ -101,11 +102,30 @@ def run(
...
-        config = process_loaded_project_config(config, errors, warnings)
+        # Disable the validation due to the local empty project
+        if is_remote_app:
+            config = load_toml(Path("pyproject.toml"))
+            if not config:
+                raise typer.BadParameter("A 'pyproject.toml' file is required.")
+        else:
+            pyproject_path = app / "pyproject.toml" if app else None
+            config, errors, warnings = load_and_validate(path=pyproject_path)
+            config = process_loaded_project_config(config, errors, warnings)
```
panh99: If the goal is to disable module checking when executing remote app, I think we can just do this.
```suggestion
        pyproject_path = app / "pyproject.toml" if app else None
        config, errors, warnings = load_and_validate(pyproject_path, check_module=not is_remote_app)
        config = process_loaded_project_config(config, errors, warnings)
```

framework/py/flwr/cli/run/run.py
```diff
@@ -101,11 +102,30 @@ def run(
     try:
         if suppress_output:
             redirect_output(captured_output)
+
+        app_id_str = str(app) if app is not None else ""
```
panh99: May I know what's an App ID? Is it `'@user_name/app_name'` or `'user_name/app_name'`?

yan-gao-GY: It's '@user_name/app_name'.

framework/py/flwr/cli/run/run.py
```diff
@@ -101,11 +102,30 @@ def run(
...
+        is_remote_app: bool = False
+
+        if app_id_str.startswith("@"):
+            m = re.match(r"^@(?P<user>[^/]+)/(?P<app>[^/]+)$", app_id_str)
+            if not m:
+                raise typer.BadParameter(
+                    "Invalid remote app ID. Expected format: '@user_name/app_name'."
+                )
+            is_remote_app = True
```
panh99: Is `app_id` used when it's not a remote app? If not, I think it would be nice to make `app_id` an `Optional[str]` instead of introducing a new boolean variable `is_remote_app`. Wdyt? For example:

```suggestion
        app_id = None
        if (app_str := str(app)).startswith("@"):
            if not re.match(r"^@(?P<user>[^/]+)/(?P<app>[^/]+)$", app_str):
                raise typer.BadParameter(
                    "Invalid remote app ID. Expected format: '@user_name/app_name'."
                )
            app_id = app_str
```

framework/py/flwr/cli/run/run.py
yan-gao-GY: Ah ok, we have a `check_module` argument. That's good~

But for the `pyproject_path`, we want it to be "./pyproject.toml" if it's a remote app, as @jafermarq suggested.

framework/py/flwr/cli/run/run.py
panh99: Could we store `str(app)` in a separate variable instead? The idea is to avoid storing a local app path in `app_id`, since that’s technically not an App ID but a file path. It would be clearer to keep `app_id` as `None` when a local path (rather than an App ID) is provided as the argument. Related to https://github.com/adap/flower/pull/6083/files#r2503047464

framework/py/flwr/cli/run/run.py
panh99: Oh, I see wym.

framework/py/flwr/cli/run/run.py
yan-gao-GY: I see, good point! i'll make the change.

danieljanes (approved):

PR #6085, opened 2025-11-04
feat(framework): Update control servicer for `flwr run` with app ID

yan-gao-GY:
## Proposal

Update control servicer.

### DIFF
--- a/examples/quickstart-numpy/pyproject.toml
+++ b/examples/quickstart-numpy/pyproject.toml
@@ -10,7 +10,7 @@ build-backend = "hatchling.build"
 [project]
 name = "quickstart-numpy"
 version = "1.0.0"
-description = ""
+description = "Federated Learning with Numpy and Flower (Quickstart Example)"
 license = "Apache-2.0"
 # Dependencies for your Flower App
 dependencies = [
--- a/examples/xgboost-quickstart/README.md
+++ b/examples/xgboost-quickstart/README.md
@@ -14,21 +14,24 @@ This project provides a minimal code example to enable you to get started quickl
 
 ## Set up the project
 
-### Clone the project
+### Fetch the app
 
-Start by cloning the example project:
+Install Flower:
 
 ```shell
-git clone --depth=1 https://github.com/adap/flower.git _tmp \
-        && mv _tmp/examples/xgboost-quickstart . \
-        && rm -rf _tmp \
-        && cd xgboost-quickstart
+pip install flwr
 ```
 
-This will create a new directory called `xgboost-quickstart` with the following structure:
+Fetch the app:
 
 ```shell
-xgboost-quickstart
+flwr new @flwrlabs/quickstart-xgboost
+```
+
+This will create a new directory called `quickstart-xgboost` with the following structure:
+
+```shell
+quickstart-xgboost
 ├── xgboost_quickstart
 │   ├── __init__.py
 │   ├── client_app.py   # Defines your ClientApp
--- a/examples/xgboost-quickstart/pyproject.toml
+++ b/examples/xgboost-quickstart/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [project]
-name = "xgboost_quickstart"
+name = "quickstart-xgboost"
 version = "1.0.0"
 description = "Federated Learning with XGBoost and Flower (Quickstart Example)"
 license = "Apache-2.0"


framework/py/flwr/cli/run/run.py
```diff
@@ -212,7 +212,8 @@ def _run_with_control_api(
             if is_remote_app:
                 typer.secho(
                     "❌ Failed to start run. Please check that the provided "
-                    "app identifier (@user_name/app_name) is correct.",
+                    "app identifier (@user_name/app_name) is correct, "
```
panh99: Given the recent [renaming](https://flower.ai/docs/framework/ref-changelog.html#incompatible-changes) to "account" throughout the codebase, I think it's better to use `account_name` here. Wdyt?

```suggestion
                    "app identifier (@account_name/app_name) is correct, "
```

framework/py/flwr/cli/run/run.py
```diff
@@ -215,7 +216,8 @@ def _run_with_control_api(
             if is_remote_app:
                 typer.secho(
                     "❌ Failed to start run. Please check that the provided "
-                    "app identifier (@account_name/app_name) is correct.",
+                    "app identifier (@account_name/app_name) is correct, "
+                    "or that it is not using an unsupported gRPC adapter.",
```
jafermarq: `unsupported gRCP adapter` might be a bit to technical? When `flwr run ` fails becaues the connection cannot be stablished with a `SuperLink`, [this error message](https://github.com/adap/flower/blob/e7ad112974ba98d6d01990a188ccea14df806b06/framework/py/flwr/cli/utils.py#L484) shown is:

> Connection to the SuperLink is unavailable. Please check your network connection and 'address' in the federation configuration.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -96,23 +107,70 @@ def __init__(  # pylint: disable=R0913, R0917
...
+            verification_dict = _format_verification(verifications, verification_dict)
+
+            # Download FAB from Flower platform API
+            try:
+                r = requests.get(url, timeout=60)
+                r.raise_for_status()
+            except requests.RequestException as e:
+                log(ERROR, "FAB download failed: %s", str(e))
+                return StartRunResponse()
+            fab_file = r.content
```
jafermarq: should we move this to a separate function? it will be easier to write test for it in that way.

jafermarq: or maybe into a new module (e.g. `remote_fab.py`) where also the utility functions at the end of the file get moved ot?

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -96,23 +107,70 @@ def __init__(  # pylint: disable=R0913, R0917
...
+            app_id, app_version = _parse_app_spec(request.app_spec)
+
+            # Validate app version format
+            if app_version:
+                if not re.match(APP_VERSION_PATTERN, app_version):
+                    log(
+                        ERROR,
+                        "Invalid app version. Expected format: x.y.z (digits only).",
+                    )
+                    return StartRunResponse()
```
jafermarq: I think we need a more informative handling of this. Else the user doing `flwr run @....` will not know what the issue is. imo it would be better to replace all the 

```python
return StartRunResponse()
```

with a `context.abort(<grpc.StatusCode>, "informative error message"`. For example as we do in https://github.com/adap/flower/blob/e7ad112974ba98d6d01990a188ccea14df806b06/framework/py/flwr/superlink/servicer/control/control_servicer.py#L596

framework/py/flwr/cli/run/run.py
```diff
@@ -113,12 +113,13 @@ def run(
...
             app_spec = app_str
+            app = Path(".")
         is_remote_app = app_spec is not None
 
         typer.secho("Loading project configuration... ", fg=typer.colors.BLUE)
 
         # Disable the validation for remote apps
-        pyproject_path = app / "pyproject.toml" if not is_remote_app else None
+        pyproject_path = app / "pyproject.toml"
         # `./pyproject.toml` will be loaded when `pyproject_path` is None
```
panh99, line 125: May I know why this change? It seems that, with the comment ``# `./pyproject.toml` will be loaded when `pyproject_path` is None``, it may be more straightforward to not set the `app`, and provide `None` when it's a remote app. Given the title of this PR is "Update control servicer", is this file changed accidentally?

framework/py/flwr/cli/run/run.py
panh99: Agree with above. As this is a user-facing prompt, better to avoid mentioning tech details.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -608,3 +666,92 @@ def _check_flwr_aid_in_run(
...
+
+
+def _parse_app_spec(app_spec: str) -> tuple[str, str | None]:
+    """Parse app specification string into app ID and version."""
+    if "==" in app_spec:
+        app_id, app_version = app_spec.split("==")
+    else:
+        app_id = app_spec
+        app_version = None
+    return app_id, app_version
```
panh99: I think we have such a util function in `cli/utils.py`, right? Should we move that function to `supercore/utils.py` and make both the CLI and the ControlServicer use the same util function?

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -608,3 +666,92 @@ def _check_flwr_aid_in_run(
...
+        context.abort(
+            grpc.StatusCode.NOT_FOUND,
+            error_message,
+        )
+    if not resp.ok:
+        context.abort(
+            grpc.StatusCode.UNAVAILABLE,
+            f"Flower platform API request failed with status {resp.status_code}. "
+            f"Details: {resp.text}",
+        )
```
panh99: Could we use a different status code here? This is very often used by gRPC itself to indicate a gRPC connection error, and our gRPC clients will retry if it gets this error code. But this error should suggest an wrong connection between the SuperLink and the Platform API, not the gRPC connection error.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -608,3 +666,92 @@ def _check_flwr_aid_in_run(
             grpc.StatusCode.PERMISSION_DENIED,
             "⛔️ Run ID does not belong to the account",
         )
+
+
+def _request_download_link(
```
panh99: Is this similar to the `request_download_link` in `utils.py`? If so, could we merge the two and move it to `supercore/utils.py`?

framework/py/flwr/cli/run/run.py
yan-gao-GY: @panh99 We have to set `app = Path(".")`, otherwise `app="@account_name/app_name"` will be passed to `_run_with_control_api`, and then directory `@account_name/app_name` will created when doing auth check to save the credentials, which is not intended.

If we set `app = Path(".")`, then we don't need `pyproject_path = app / "pyproject.toml" if not is_remote_app else None`, instead we can only leave `pyproject_path = app / "pyproject.toml"`. Does it make sense?

framework/py/flwr/superlink/servicer/control/control_servicer.py
yan-gao-GY: Yes GP! I've moved it to a separate function. It's maybe not necessary to have a new module imo.

framework/py/flwr/superlink/servicer/control/control_servicer.py
yan-gao-GY: I see. Changed to `grpc.StatusCode.FAILED_PRECONDITION`

framework/py/flwr/superlink/servicer/control/control_servicer.py
yan-gao-GY: @panh99 The error handling part is different. For example, in `cli/utils.py`, we raise `typer.Exit(code=1)` ([link](https://github.com/adap/flower/blob/main/framework/py/flwr/cli/utils.py#L745)), but here we use `context.abort`. Maybe we can add a flag to raise different errors based on the usage case, like:

```
def parse_app_spec(app_spec: str, typer_error: bool = True, context: grpc.ServicerContext | None = None) -> tuple[str, str | None]:
    if "==" in app_spec:
        app_id, app_version = app_spec.split("==")

        # Validate app version format
        if not re.match(APP_VERSION_PATTERN, app_version):
            error_message = "Invalid app version. Expected format: x.y.z (digits only)."
            if typer_error:
                typer.secho(
                    error_message,
                    fg=typer.colors.RED,
                    err=True,
                )
                raise typer.Exit(code=1)
            if context:
                context.abort(
                    grpc.StatusCode.NOT_FOUND,
                    error_message,
                )
    else:
        app_id = app_spec
        app_version = None

    # Validate app_id format
    if not re.match(APP_ID_PATTERN, app_id):
        error_message = "Invalid remote app ID. Expected format: '@account_name/app_name'."
        if typer_error:
            typer.secho(
                error_message,
                fg=typer.colors.RED,
                err=True,
            )
            raise typer.Exit(code=1)
        if context:
            context.abort(
                grpc.StatusCode.NOT_FOUND,
                error_message,
            )

    return app_id, app_version
```

Wdyt?

framework/py/flwr/superlink/servicer/control/control_servicer.py
yan-gao-GY: It's in the same case as `parse_app_spec` function above.

framework/py/flwr/cli/run/run.py
panh99: Ah, I see wym. So, for now, due to the lack of global SuperLink connection configuration, we have to assume that users execute the command from the app directory when using app ID, right?

framework/py/flwr/cli/run/run.py
```diff
@@ -113,6 +113,7 @@ def run(
             # Validate app version and ID format
             _ = parse_app_spec(app_str)
             app_spec = app_str
+            app = Path(".")
```
panh99, line 118: Could you add a comment here?

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -612,3 +627,71 @@ def _check_flwr_aid_in_run(
...
+    app_spec: str,
+    context: grpc.ServicerContext,
+    verification_dict: dict[str, str],
+) -> bytes:
+    """Get remote FAB from Flower platform API."""
+    if fleet_api_type == TRANSPORT_TYPE_GRPC_ADAPTER:
+        context.abort(
+            grpc.StatusCode.FAILED_PRECONDITION,
+            "Connection to the SuperLink is unavailable.",
+        )
```
panh99, line 657: The error message seems a bit unclear. It's b/c we don't want a SuperLink with grpc-adapter to connect to the Flower platform. Should we say something like "The selected SuperLink transport type is not supported for connecting to Flower Platform."?

yan-gao-GY: ok sounds good!

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -96,23 +101,33 @@ def __init__(  # pylint: disable=R0913, R0917
...
         log(INFO, "ControlServicer.StartRun")
         state = self.linkstate_factory.state()
         ffs = self.ffs_factory.ffs()
 
-        if len(request.fab.content) > FAB_MAX_SIZE:
+        verification_dict: dict[str, str] = {}
+        if request.app_spec.startswith("@"):
+            fab_file = _get_remote_fab(
+                self.fleet_api_type, request.app_spec, context, verification_dict
+            )
```
panh99, line 126: May I know why passing an empty dictionary in the function? If the `verification_dict` is also built inside the `_get_remote_fab` function, should we remove this input argument and return it alongside with fab_file_bytes?

yan-gao-GY: Ah my bad. It's a bug when I refactored the functions. Good catch! I've pushed the fix.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -96,23 +101,33 @@ def __init__(  # pylint: disable=R0913, R0917
...
         self, request: StartRunRequest, context: grpc.ServicerContext
     ) -> StartRunResponse:
         """Create run ID."""
         log(INFO, "ControlServicer.StartRun")
         state = self.linkstate_factory.state()
         ffs = self.ffs_factory.ffs()
 
-        if len(request.fab.content) > FAB_MAX_SIZE:
+        verification_dict: dict[str, str] = {}
+        if request.app_spec.startswith("@"):
```
panh99: IIUC, the `app_spec` should either be a valid one when `flwr run <app-id>`, or it will be an empty string, right? Should we use this condition to be clear?


```suggestion
        if request.app_spec:
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #6090, opened 2025-11-05
feat(framework): Add `federation` argument to `LinkState.create_run`

panh99:


framework/py/flwr/superlink/servicer/control/control_servicer_test.py
```diff
@@ -316,6 +315,11 @@ def tearDown(self) -> None:
         """Clean up after tests."""
         self.tmp_dir.cleanup()
 
+    def _create_dummy_run(self, flwr_aid: Optional[str]) -> int:
+        return self.state.create_run(
+            "flwr/demo", "v0.0.1", "hash123", {}, "mock-fed", ConfigRecord(), flwr_aid
+        )
+
```
jafermarq, line 322:
```suggestion
```

Seems to be duplicated

framework/py/flwr/server/superlink/linkstate/linkstate.py
```diff
@@ -273,10 +273,40 @@ def create_run(  # pylint: disable=too-many-arguments,too-many-positional-argume
...
+
+        Returns
+        -------
+        int
+            The run ID of the newly created run.
+
+        Notes
+        -----
+        This method will not verify if the account has permission to create
+        a run in the federation.
```
jafermarq, line 308: should we leave this here? we are going to handle this very soon

panh99: Oh, I thought we gonna check membership outside. Or we gonna check it inside? If so, happy to remove it.

framework/py/flwr/superlink/servicer/control/control_servicer_test.py
panh99: We have two test classes here... So one for each.

framework/py/flwr/server/superlink/linkstate/linkstate.py
jafermarq: outside. I see your point now. let's keep it

jafermarq (approved):

PR #6092, opened 2025-11-06
refactor(framework): Use `FederationManager` in `ControlServicer.StartRun`

jafermarq:
A run can only start if (1) federation exist and (2) user submitting the request is a member of such federation

### DIFF
--- a/framework/py/flwr/superlink/servicer/control/control_servicer_test.py
+++ b/framework/py/flwr/superlink/servicer/control/control_servicer_test.py
@@ -48,7 +48,7 @@ from flwr.proto.control_pb2 import (  # pylint: disable=E0611
     UnregisterNodeRequest,
 )
 from flwr.server.superlink.linkstate import LinkStateFactory
-from flwr.supercore.constant import FLWR_IN_MEMORY_DB_NAME
+from flwr.supercore.constant import FLWR_IN_MEMORY_DB_NAME, NOOP_FEDERATION
 from flwr.supercore.ffs import FfsFactory
 from flwr.supercore.primitives.asymmetric import generate_key_pairs, public_key_to_bytes
 from flwr.superlink.auth_plugin import NoOpControlAuthnPlugin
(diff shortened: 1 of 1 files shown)


framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -128,6 +128,22 @@ def StartRun(  # pylint: disable=too-many-locals
                     "Federation options doesn't contain key `num-supernodes`."
                 )
 
+            # If it's not simulation, check (1) federation exists
+            # and (2) the flwr_aid is a member
+            if not self.is_simulation:
```
panh99: Why use the condition here? I think it's fine to apply to the simulation runtime, no? I mean, users should not pass `--federation-config` when running SuperLink with simulation runtime if they don't want such check.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -128,6 +128,22 @@ def StartRun(  # pylint: disable=too-many-locals
...
+
+                federation = request.federation
+
+                if not state.federation_manager.exists(federation):
+                    raise ValueError(f"Federation '{federation}' does not exist.")
+
+                if not state.federation_manager.has_member(flwr_aid, federation):
+                    raise ValueError(
+                        f"Account with flwr_aid '{flwr_aid}' is not a member of "
```
panh99: `flwr_aid` is more like a var name for Flower account ID, and I think this might be more understandable to users.

```suggestion
                        f"Account with ID '{flwr_aid}' is not a member of the "
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -128,6 +128,22 @@ def StartRun(  # pylint: disable=too-many-locals
...
+                        f"federation '{federation}'. Login with another account "
```
```suggestion
                        f"federation '{federation}'. Please log in with another account "
```

PR #6097, opened 2025-11-07
feat(framework): Add app publish command

yan-gao-GY:
## Proposal

Add app publish command.

### DIFF
--- a/framework/docs/locales/fr/LC_MESSAGES/framework-docs.po
+++ b/framework/docs/locales/fr/LC_MESSAGES/framework-docs.po
@@ -3,7 +3,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Flower Docs\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2025-11-18 00:08+0000\n"
+"POT-Creation-Date: 2025-11-21 00:08+0000\n"
 "PO-Revision-Date: 2025-02-05 10:02+0000\n"
 "Last-Translator: lco <Leandro.COLLIER@student.umons.ac.be>\n"
 "Language: fr\n"
@@ -3182,6 +3182,7 @@ msgstr "Augmente la version mineure de ``pyproject.toml`` d'une unité."
 #: ../../source/docker/run-quickstart-examples-docker-compose.rst:63
 #: ../../source/docker/tutorial-quickstart-docker.rst:310
 #: ../../source/how-to-enable-tls-connections.rst:127
+#: ../../source/how-to-manage-flower-federations.rst:69
 #: ../../source/how-to-run-flower-with-deployment-engine.rst:179
 #: ../../source/how-to-upgrade-to-message-api.rst:132
 msgid "pyproject.toml"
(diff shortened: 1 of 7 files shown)


framework/py/flwr/cli/hub_app/publish.py
```diff
@@ -0,0 +1,332 @@
...
+def publish(
+    app: Annotated[
+        Path,
+        typer.Argument(
+            help="Project directory to upload (defaults to current directory)."
+        ),
+    ] = Path("."),
+    federation: Annotated[
+        Optional[str],
+        typer.Argument(help="Name of the federation used for app publishing."),
```
jafermarq: Would it be better to change the help message to say that this "federation" is the one the user has first `flwr login` into ? As you indicate in error in line 290?

framework/py/flwr/cli/hub_app/publish.py
```diff
@@ -0,0 +1,332 @@
...
+    # Build and POST multipart
+    with ExitStack() as stack:
+        files_param = _build_multipart_files_param(files, stack)
+        try:
+            resp = _post_files(files_param, token)
+        except requests.RequestException as err:
+            typer.secho(f"❌ Network error: {err}", fg=typer.colors.RED, err=True)
+            raise typer.Exit(code=1) from err
+
+    if resp.status_code // 100 == 2:
```
jafermarq: This looks a bit strange.  Would it be better to check for a specific status code instead of doing `// 100`?

framework/py/flwr/cli/hub_app/publish.py
```diff
@@ -0,0 +1,332 @@
...
+        )
+        raise typer.Exit(code=1) from err
+
+    required_keys = [AUTHN_TYPE_JSON_KEY, ACCESS_TOKEN_KEY, REFRESH_TOKEN_KEY]
+    missing = [key for key in required_keys if key not in content]
+
+    if missing:
+        typer.secho(
+            f"Credentials file '{creds_path}' is missing "
+            f"required key(s): {', '.join(missing)}.",
```
jafermarq: When would this happen? Should we extend the error message telling the user to login again?

framework/py/flwr/cli/hub_app/publish.py
```diff
@@ -0,0 +1,332 @@
...
+def _validate_credentials_content(creds_path: Path) -> dict[str, str]:
+    """Load and validate the credentials file content.
+
+    Ensures required keys exist:
+      - AUTHN_TYPE_JSON_KEY
+      - ACCESS_TOKEN_KEY
+      - REFRESH_TOKEN_KEY
+    """
+    try:
+        content = json.loads(creds_path.read_text(encoding="utf-8"))
```
jafermarq: Can we set the type of this `content` so we don't have to use `# type: ignore` in the return?

jafermarq: Alternatively, since only the `ACCESS_TOKEN_KEY ` is needed to continue with the publishing process, we could change the return type to just a `str` and remove the line `token = creds[ACCESS_TOKEN_KEY]` later in the code. wdyt?

framework/py/flwr/supercore/constant.py
```diff
@@ -27,6 +27,19 @@
 APP_ID_PATTERN = r"^@(?P<user>[^/]+)/(?P<app>[^/]+)$"
 PLATFORM_API_URL = "https://api.flower.ai/v1"
 
+# App spec
+ALLOWED_EXTS = [".py", ".toml", ".md"]
+MAX_TOTAL_BYTES = 10 * 1024 * 1024  # 10 MB
+MAX_FILE_BYTES = 1 * 1024 * 1024  # 1 MB
```
jafermarq, line 44: I wonder if this is too big. a file with code thats' 1MB is huge.

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -0,0 +1,332 @@
...
+    for p in root.rglob("*"):
+        try:
+            if ignored(p):
+                if p.is_file():
+                    typer.echo(
+                        typer.style(f"Skip (gitignore): {p}", fg=typer.colors.YELLOW)
+                    )
+                continue
+        except Exception:  # pylint: disable=broad-exception-caught
+            # If ignore callable crashes for some weird FS entry, just continue safely
```
jafermarq: is the exception raised because of how the `ignored` function works?

framework/py/flwr/cli/hub_app/publish.py
yan-gao-GY: Yeah good to tell the user to login again~

framework/py/flwr/cli/hub_app/publish.py
yan-gao-GY: Yeah make sense to me to remove the line token = creds[ACCESS_TOKEN_KEY] later in the code.

framework/py/flwr/cli/app_cmd/publish.py
yan-gao-GY: It's just used to handle some cases from FS issues. For example:

- Broken symlinks — if `rglob` yields a symlink that points to nowhere, p.relative_to(root) or p.is_dir() could raise a FileNotFoundError.
- Unreadable filesystem entries — permission errors, weird virtual FS mounts, or transient files.

For these cases, we just skip them and continue the publishing flow. Do you think it makes sense?

framework/py/flwr/supercore/constant.py
yan-gao-GY: Yeah it could be. I'm not sure. What value do you suggest?

framework/py/flwr/cli/app_cmd/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Flower command line interface `app` command."""
+
+from .publish import publish as publish
```
panh99, line 18:
```suggestion


from .publish import publish as publish
```

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+def _collect_files(root: Path) -> list[tuple[Path, Path]]:
+    """Return list of (abs_path, rel_path) that pass allowed list + ignore logic."""
+    ignored = _compile_gitignore(root)
+    files: list[tuple[Path, Path]] = []
+
+    for p in root.rglob("*"):
+        try:
+            if ignored(p):
```
panh99: We also have very similar logic in `flwr build`. Should we move shared logic to `utils.py` so both `flwr build` and `flwr app publish` can share the logic for loading and using gitignore file?

https://github.com/adap/flower/blob/main/framework/py/flwr/cli/build.py#L280
https://github.com/adap/flower/blob/main/framework/py/flwr/cli/build.py#L155

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -0,0 +1,333 @@
...
+        flwr_dir_rel += "/"
+    patterns.append(flwr_dir_rel)
+
+    spec = pathspec.PathSpec.from_lines("gitwildmatch", patterns)
+
+    def ignored(p: Path) -> bool:
+        rel = p.relative_to(root).as_posix()
+        # Pathspec expects directory entries with trailing slash for dir matches
+        if p.is_dir() and not rel.endswith("/"):
+            rel += "/"
```
panh99: May I know what's the purpose of this part? Should we also update [this](https://github.com/adap/flower/blob/main/framework/py/flwr/cli/build.py#L226) accordingly if necessary?

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -0,0 +1,333 @@
...
+    """Detect files' MIME."""
+    return MIME_MAP.get(path.suffix.lower(), "text/plain; charset=utf-8")
+
+
+def _collect_files(root: Path) -> list[tuple[Path, Path]]:
+    """Return list of (abs_path, rel_path) that pass allowed list + ignore logic."""
+    ignored = _compile_gitignore(root)
+    files: list[tuple[Path, Path]] = []
+
+    for p in root.rglob("*"):
```
panh99: Could we use `path` or similar var names, instead of `p`? That would be more readable and maintainable.

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -0,0 +1,333 @@
...
+            # If ignore callable crashes for some weird FS entry, just continue safely
+            continue
+
+        if p.is_dir():
+            continue
+
+        # Only include allowed extensions
+        if p.suffix.lower() not in ALLOWED_EXTS:
+            typer.echo(typer.style(f"Skip (ext): {p}", fg=typer.colors.YELLOW))
+            continue
```
panh99: Should we also use `pathspec` here, like gitignore? For example: https://github.com/adap/flower/blob/main/framework/py/flwr/cli/build.py#L275

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -0,0 +1,333 @@
...
+    parts = relpath.as_posix().split("/")
+    return max(0, len(parts) - 1)
+
+
+def _detect_mime(path: Path) -> str:
+    """Detect files' MIME."""
+    return MIME_MAP.get(path.suffix.lower(), "text/plain; charset=utf-8")
+
+
+def _collect_files(root: Path) -> list[tuple[Path, Path]]:
```
panh99: May I know the purpose of having both absolute path and relative path returned? Could we just return the relative path? I think it's more common to use relative path, and the absolute path can easily be derived via `path.absolute()` as it's of type `Path`, right?

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -0,0 +1,333 @@
...
+
+def _depth_of(relpath: Path) -> int:
+    """Return depth that is number of parts (directories) in the relative path
+    (excluding filename).
+
+    Example: "a/b/c.py" -> depth 3
+    Interpret "directory depth" as number of directories: len(parts) - 1
+    """
+    parts = relpath.as_posix().split("/")
+    return max(0, len(parts) - 1)
```
panh99:
```suggestion
    return max(0, len(relpath.parts) - 1)
```

panh99 (commented):
A general comment: It would be better to use the full variable names for improved readability, for example `absolute_path` and `relative_path`. While `abs_p` and `rel_p/rel/p` are shorter and still understandable given the context, in a collaborative open-source project readability and clarity should take precedence over brevity. Would you mind updating the variable names to be more descriptive? I’d really appreciate that.

framework/py/flwr/cli/app_cmd/publish.py
yan-gao-GY: Thanks for the feedback. The main reason to keep both absolute path and relative path is that the absolute path will be used in the later functions `_validate_files` and `_build_multipart_files_param`:

```
size = absolute_path.stat().st_size
absolute_path.read_text(...)
open(absolute_path, "rb")
```

This avoids recomputing absolute path every time `absolute_path = root / relative_path`, and need to have `root` available in every downstream call.

framework/py/flwr/cli/app_cmd/publish.py
yan-gao-GY: Good question! This part is only useful when the path is a directory, but in both cases of `flwr build` and `flwr app publish`, we only include the path that is a file (e.g. `all_files = [f for f in app.rglob("*") if f.is_file()]`). As a result, we don't need this part.

framework/py/flwr/cli/app_cmd/publish.py
panh99: Oh, I mean, it seems that you are using the `Path` objects like `str`. The `path.absolute()` call should be very fast. And it's also less intuitive to return two paths in the same time. Based on the code, should we just return the relative path, and only convert it to absolute path when necessary?

panh99: Btw, may I know why these?
```
size = absolute_path.stat().st_size
absolute_path.read_text(...)
open(absolute_path, "rb")
```

Why not use:
```
size = relative_path.stat().st_size
relative_path.read_text(...)
relative_path.read_bytes()
```

The two are strictly identical.

PR #6100, opened 2025-11-07
refactor(framework): Add app version & refactor downloading function for flwr new

yan-gao-GY:
## Proposal

Update `flwr new`:

- Add app version option.
- Refactor downloading function for re-using.

### DIFF
--- a/framework/py/flwr/cli/app.py
+++ b/framework/py/flwr/cli/app.py
@@ -63,6 +63,14 @@ supernode_app.command("list")(supernode_list)
 supernode_app.command(hidden=True)(supernode_list)
 app.add_typer(supernode_app, name="supernode")
 
+# Create federation command group
+federation_app = typer.Typer(help="Manage Federations")
+# Make it appear as "list"
+federation_app.command("list")(supernode_list)
+# Hide "ls" command (left as alias)
+federation_app.command(hidden=True)(supernode_list)
+app.add_typer(federation_app, name="federation")
+
 typer_click_object = get_command(app)
 
 
--- a/framework/py/flwr/cli/federation/__init__.py
+++ b/framework/py/flwr/cli/federation/__init__.py
@@ -1,0 +1,22 @@
+# Copyright 2025 Flower Labs GmbH. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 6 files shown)


framework/py/flwr/cli/new/new.py
```diff
@@ -287,14 +263,21 @@ def new(
         Optional[str],
         typer.Option(case_sensitive=False, help="The Flower username of the author"),
     ] = None,
+    version: Annotated[
+        Optional[str],
+        typer.Option(
+            "--version",
+            help="Version of the app to download (e.g., '1.0.0').",
```
panh99: If this is not specified, the CLI will download the latest version, right? If so, I think it's better to mention it in the help message.

framework/py/flwr/cli/new/new.py
```diff
@@ -249,7 +225,7 @@ def download_remote_app_via_api(app_id: str) -> None:
             bold=True,
         )
     )
-    presigned_url = _request_download_link(app_id)
+    presigned_url = _request_download_link(app_id, version)
```
panh99: Could we use `request_download_link` from `flwr.cli.utils` directly and get rid of `_request_download_link`? The implementation of `_request_download_link` seems pretty thin, and I think it may be good to avoid unnecessary nesting. Wdyt?

framework/py/flwr/cli/new/new.py
yan-gao-GY: Good point!

framework/py/flwr/cli/new/new.py
yan-gao-GY: okk make sense

panh99 (dismissed):
LGTM!

framework/py/flwr/cli/new/new.py
```diff
@@ -277,7 +268,10 @@ def download_remote_app_via_api(app_id: str) -> None:
 def new(
     app_name: Annotated[
         Optional[str],
-        typer.Argument(help="The name of the Flower App"),
+        typer.Argument(
+            help="Flower app name. For remote apps, use '@user/app==1.0.0'. "
```
panh99:
```suggestion
            help="Flower app name. For remote apps, use the format '@user/app==1.0.0'. "
```

PR #6101, opened 2025-11-10
feat(framework): Add app review command

yan-gao-GY:
## Proposal

Add app review command.

### DIFF
--- a/framework/py/flwr/cli/app_cmd/publish.py
+++ b/framework/py/flwr/cli/app_cmd/publish.py
@@ -23,7 +23,7 @@ import requests
 import typer
 from requests import Response
 
-from flwr.common.constant import CREDENTIALS_DIR, FLWR_DIR
+from flwr.common.constant import FAB_CONFIG_FILE
 from flwr.common.version import package_version as flwr_version
 from flwr.supercore.constant import (
     APP_PUBLISH_EXCLUDE_PATTERNS,
@@ -37,13 +37,17 @@ from flwr.supercore.constant import (
     UTF8,
 )
 
-from ..utils import (
-    build_pathspec,
-    load_gitignore_patterns,
-    validate_credentials_content,
+from ..auth_plugin.oidc_cli_plugin import OidcCliPlugin
+from ..config_utils import (
+    load_and_validate,
+    process_loaded_project_config,
+    validate_federation_in_project_config,
 )
+from ..constant import FEDERATION_CONFIG_HELP_MESSAGE
+from ..utils import build_pathspec, load_cli_auth_plugin, load_gitignore_patterns
 
 
+# pylint: disable=too-many-locals
 def publish(
     app: Annotated[
         Path,
@@ -57,11 +61,11 @@ def publish(
             help="Name of the federation used for login before publishing app."
         ),
     ] = None,
-    token: Annotated[
-        str | None,
+    federation_config_overrides: Annotated[
+        list[str] | None,
         typer.Option(
-            "--token",
-            help="Bearer token for Platform API.",
+            "--federation-config",
+            help=FEDERATION_CONFIG_HELP_MESSAGE,
         ),
     ] = None,
 ) -> None:
@@ -70,28 +74,27 @@ def publish(
     This command uploads your app project to the Flower Platform. Files are filtered
     based on .gitignore patterns and allowed file extensions.
     """
-    # Check the credentials path
-    if not token:
-        if not federation:
-            typer.secho(
-                "❌ Please specify the federation used for "
-                "login before publishing app.",
-                fg=typer.colors.RED,
-                err=True,
-            )
-            raise typer.Exit(code=1)
+    # Load configs
+    pyproject_path = app / FAB_CONFIG_FILE if app else None
+    config, errors, warnings = load_and_validate(pyproject_path, check_module=False)
+    config = process_loaded_project_config(config, errors, warnings)
+    federation, federation_config = validate_federation_in_project_config(
+        federation, config, federation_config_overrides
+    )
 
-        creds_path = app.absolute() / FLWR_DIR / CREDENTIALS_DIR / f"{federation}.json"
-        if not creds_path.is_file():
-            typer.secho(
-                "❌ Please log in before publishing app.",
-                fg=typer.colors.RED,
-                err=True,
-            )
-            raise typer.Exit(code=1)
+    # Load the authentication plugin
+    auth_plugin = load_cli_auth_plugin(app, federation, federation_config)
+    auth_plugin.load_tokens()
+    if not isinstance(auth_plugin, OidcCliPlugin) or not auth_plugin.access_token:
+        typer.secho(
+            "❌ Please log in before publishing app.",
+            fg=typer.colors.RED,
+            err=True,
+        )
+        raise typer.Exit(code=1)
 
-        # Load and validate credentials
-        token = validate_credentials_content(creds_path)
+    # Load token from the plugin
+    token = auth_plugin.access_token
 
     # Collect & validate app files
     file_paths = _collect_file_paths(app)


framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+            err=True,
+        )
+        raise typer.Exit(code=1) from e
+    return r.content
+
+
+def _load_private_key(path: Path) -> ed25519.Ed25519PrivateKey:
+    """Load a private key (Ed25519) using cryptography."""
+    try:
+        pem = path.read_bytes()
```
panh99: This is not PEM bytes, no? It should be ssh format
```suggestion
        private_key_bytes = path.read_bytes()
```

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+        typer.secho(
+            f"❌ FAB download failed: {e}",
+            fg=typer.colors.RED,
+            err=True,
+        )
+        raise typer.Exit(code=1) from e
+    return r.content
+
+
+def _load_private_key(path: Path) -> ed25519.Ed25519PrivateKey:
```
panh99: Could you move this function to `framework/py/flwr/supercore/primitives/asymmetric_ed25519.py`? And prolly just do:

```python

def load_private_key(path: Path) -> ed25519.Ed25519PrivateKey:
    """Load an SSH-format private key (Ed25519) using cryptography."""
    key_bytes= path.read_bytes()
    private_key = serialization.load_ssh_private_key(key_bytes, password=None)
    if not isinstance(private_key, ed25519.Ed25519PrivateKey):
        raise ValueError("Private key is not Ed25519")
```

And then in this file:
```
try:
    load_private_key(path)
except (OSError, ValueError, UnsupportedAlgorithm) as e:
    typer.secho(
        f"❌ Failed to load the private key: {e}", fg=typer.colors.RED, err=True
    )
    raise typer.Exit(code=1) from e

File: framework/py/flwr/cli/app_cmd/review.py [hunk 3]
```diff
@@ -0,0 +1,252 @@
...
+    # Error path:
+    msg = f"❌ Review submission failed (HTTP {resp.status_code})"
+    if resp.text:
+        msg += f": {resp.text}"
+    typer.secho(msg, fg=typer.colors.RED, err=True)
+    raise typer.Exit(code=1)
+
+
+# pylint: disable-next=too-many-locals
+def review(
```
panh99, line 54: Would you mind moving this function to the top? That would be consistent with other CLI command files.

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+    app_name: Annotated[
+        str,
+        typer.Argument(
+            help="App identifier (e.g., '@user/app==1.0.0'). "
+            "Version is optional; defaults to the latest."
```
panh99: This should be `app_id`, right?

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+        ),
+    ],
+    token: Annotated[
+        Optional[str],
+        typer.Option(
+            "--token",
+            help="Bearer token for Platform API.",
+        ),
+    ] = None,
```
panh99: I am less sure about this. Though I think this will be temporary until we have global config, but it would be slightly strange to require an extra `--token` option, as it's inconsistent with the current `flwr` CLI commands. Should we follow other CLI commands, and take two positional arguments, i.e., `APP`, and `FEDERATION`? And we can locate the token based on the two. This is also how all other `flwr` CLI load tokens. Wdyt?

FYI, we can use `load_cli_auth_plugin` [here](https://github.com/adap/flower/blob/main/framework/py/flwr/cli/utils.py#L250) and then find the access token as follows:

```python
# Load the plugin
auth_plugin = load_cli_auth_plugin(...)

# The command only work when logged in
if not isinstance(auth_plugin, OidcCliPlugin):
    raise Exception

# Retrieve the access token
auth_plugin.load_tokens()
token = auth_plugin.access_token
```

framework/py/flwr/cli/app_cmd/review.py
panh99: Or if you have any concerns, should we name it as `app_ref` or `app_spec`?

panh99:
```suggestion
            help="App identifier (e.g., '@user/app' or '@user/app==1.0.0'). "
            "Version is optional; defaults to the latest."
```

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+            "Please run `flwr login` to generate one.",
+            fg=typer.colors.RED,
+            err=True,
+        )
+        raise typer.Exit(code=1)
+
+    # Validate version format
+    if "==" in app_name:
+        app_id = app_name.split("==")[0]
+        version = app_name.split("==")[1]
```
panh99:
```suggestion
        app_id, app_version = app_name.split("==")
```

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+
+        # Validate app version format
+        m = re.match(APP_VERSION_PATTERN, version)
+        if not m:
+            typer.secho(
+                "❌ Invalid app version. Expected format: x.x.x (digits only).",
```
panh99:
```suggestion
                "❌ Invalid app version. Expected format: x.y.z (digits only).",
```

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+                err=True,
+            )
+            raise typer.Exit(code=1)
+    else:
+        app_id = app_name
+        version = None
+        typer.secho(
+            "No app version specified. Downloading the latest version.",
+            fg=typer.colors.YELLOW,
+        )
```
panh99: Should we remove this prompt? It’s quite common for users to install the latest version, isn’t it? The yellow styling makes it look like a warning, even though there’s nothing users actually need to be warned about. `pip install flwr` also installs the latest Flower version without any warning, right?

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+
+    # Validate app_id format
+    m = re.match(APP_ID_PATTERN, app_id)
+    if not m:
+        typer.secho(
+            "❌ Invalid remote app ID. Expected format: '@user_name/app_name'.",
```
panh99: Just for consistency. We've renamed all "User ID" and "User name" to "Account ID" and "Account name".
```suggestion
            "❌ Invalid remote app ID. Expected format: '@account_name/app_name'.",
```

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+    # Ask for private key path
+    key_path_str = typer.prompt(
+        "Please specify the path of Ed25519 OpenSSH private key for signing"
+    )
+    key_path = Path(key_path_str).expanduser().resolve()
+    if not key_path.is_file():
+        typer.secho(
+            f"❌ Private key not found: {key_path}", fg=typer.colors.RED, err=True
+        )
+        raise typer.Exit(code=1)
```
panh99: Should we prompt users to retry when an operation fails? This would help tolerate simple mistakes such as typos or an incorrect key path. Without a retry option, users might have to restart the entire process if they enter something wrong. They can still abort at any time with Ctrl+C.

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: Yes good catch!

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: I found that the `run` function for `flwr run` is on the top, but the `new` function for `flwr new` is on the bottom.  I not sure which one we should follow.

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: We defined app id as `@user/app`, not '@user/app==1.0.0'. I think `app_ref` would be better in this case.

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: Make sense!

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: @panh99 Thanks for the suggestion! yes it's a temporary thing, but IMO, it's a bit wired to introduce two more arguments `APP` and `FEDERATION`, because these two arguments are for an existing local project, which is unrelated to the reviewing app. 

At the end when we have global config, I think users can do `flwr login` first, then the token will be saved as an env variable. Then, the `app review` command will load the token automatically if it's not provided. Wdyt?

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: Let's use '@user/app' for consistency?

framework/py/flwr/cli/app_cmd/review.py
panh99: It's generally in the top. I think we'd also move the `new()` function to the top.

framework/py/flwr/cli/app_cmd/review.py
panh99: Yeah, I see your point. We also had similar discussion when introducing `flwr stop/ls/log/login`, as none of these commands require an existing app project. And I was suggesting to use the similar signature b/c: 1. `flwr login` must be executed in a project for now, meaning that the user should be already in a project when executing `flwr app review`; 2. We haven't tell our users how to find the access token yet, and the access token is a bit too long to copy and paste, tbh.

And you are right that all these will be resolved quite smoothly once we have global configuration.

framework/py/flwr/cli/app_cmd/review.py
panh99: Do we use '@user/app' anywhere in the code base? If so, I think it's better to use "account" instead. We've renamed user auth, user name, etc., all to account auth, account name in the last release, and therefore better to keep the wording consistent,

yan-gao-GY: Ok i see. I believe '@user/app' is used in a few places. I can change all of them to '@account/app', but I'm just a bit concerned if '@account/app' is a good naming?

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: Based on the discussion, use `app_spec` for now.

framework/py/flwr/cli/app_cmd/review.py
yan-gao-GY: Based on the discussion, use '@account/app'.

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,263 @@
...
+    federation: Annotated[
+        str | None,
+        typer.Argument(
+            help="Name of the federation used for login before reviewing app."
+        ),
+    ] = None,
+) -> None:
+    """Download a FAB for <APP-ID>, unpack it for manual review, and upon confirmation
+    sign & submit the review to the Platform."""
+    if not app_dir_login or not federation:
```
panh99: Should we use the default federation when this field is not specified, like other Flower CLI commands?

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,263 @@
...
+    )
+    if not creds_path.is_file():
+        typer.secho(
+            "❌ Please log in before reviewing app.",
+            fg=typer.colors.RED,
+            err=True,
+        )
+        raise typer.Exit(code=1)
+
+    token = validate_credentials_content(creds_path)
```
panh99: Just noticed that you've implemented a well-structured `validate_credentials_content` function. Should we follow the similar logic as in other Flower CLI to avoid inconsistency? This could also help us to refactor our logic easier when introducing global configuration. This also load the default federation when none is provided. For example:

```python
# Load configs (this logic is shared across all Flower CLI commands)
pyproject_path = app / FAB_CONFIG_FILE if app else None
config, errors, warnings = load_and_validate(pyproject_path, check_module=False)
config = process_loaded_project_config(config, errors, warnings)
federation, federation_config = validate_federation_in_project_config(
    federation, config, federation_config_overrides
)

# Load the authentication plugin (This is also shared util for loading auth plugin)
auth_plugin = load_cli_auth_plugin(app, federation, federation_config)
if not isinstance(auth_plugin, OidcCliPlugin):
    typer.secho(
        "❌ Please log in before reviewing app.",
        fg=typer.colors.RED,
        err=True,
    )
    raise typer.Exit(code=1)

# Load token from the plugin
auth_plugin.load_tokens()
token = auth_plugin.access_token
```

This code should be able to replace line 72-92. Wdyt?

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,263 @@
...
+        app_version = None
+
+    # Validate app_id format
+    if not re.match(APP_ID_PATTERN, app_id):
+        typer.secho(
+            "❌ Invalid remote app ID. Expected format: '@account/app'.",
+            fg=typer.colors.RED,
+            err=True,
+        )
+        raise typer.Exit(code=1)
```
panh99: Should we move this to a internal function? And then we can do the following here:

```python
# def parse_app_spec(app_spec: str) -> tuple[str, str | None]
app_id, app_version = parse_app_spec(app_spec)
```

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,263 @@
...
+        break  # valid
+
+    # Load private key and sign FAB
+    try:
+        private_key = load_private_key(key_path)
+    except (OSError, ValueError, UnsupportedAlgorithm) as e:
+        typer.secho(
+            f"❌ Failed to load the private key: {e}", fg=typer.colors.RED, err=True
+        )
+        raise typer.Exit(code=1) from e
```
panh99: Can we also move this part to the while loop at line 155? Just to prompt users to re-enter the key if the provided key path is invalid or the key file itself is invalid.

framework/py/flwr/cli/app_cmd/review_test.py
```diff
@@ -0,0 +1,146 @@
...
+import requests
+import typer
+from cryptography.hazmat.primitives.asymmetric import ed25519
+
+from .review import _download_fab, _sign_fab, _submit_review
+
+
+def test_download_fab_success_and_failure(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
```
panh99: Should we split this into two tests, one for the success case and one for the failure case, or you'd suggest to keep the two in the same function?

framework/py/flwr/cli/app_cmd/review_test.py
```diff
@@ -0,0 +1,146 @@
...
+    assert before <= ts <= after
+
+
+def test_submit_review_success_and_errors(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Test _submit_review() on success, HTTP error, and network exception."""
+    captured: dict[str, Any] = {}
+
+    class FakeResp:
```
panh99: This seems similar to the class implementation in the above test function. Should we create a `FakeResp` class in the module-level and use it in all test functions to avoid duplication?

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,263 @@
...
+    return r.content
+
+
+def _sign_fab(
+    fab_bytes: bytes, private_key: ed25519.Ed25519PrivateKey
+) -> tuple[bytes, int]:
+    """Sign the given FAB hash bytes."""
+    # Get current timestamp
+    timestamp = int(time.time())
```
panh99: Though effectively the same in Python, could you use the `now` function from `flwr.common` module? We should use this function to ensure consistency across our codebase. 

```suggestion
    timestamp = int(now().timestamp())
```

framework/py/flwr/cli/app_cmd/review_test.py
yan-gao-GY: Yeah it'd be good to have two tests for success and failure case separately

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+
+TRY_AGAIN_MESSAGE = "Please try again or press CTRL+C to abort.\n"
+
+
+# pylint: disable-next=too-many-locals, too-many-statements
+def review(
+    app_spec: Annotated[
+        str,
+        typer.Argument(
+            help="App identifier (e.g., '@account/app' or '@account/app==1.0.0'). "
```
panh99:
```suggestion
            help="App specifier (e.g., '@account/app' or '@account/app==1.0.0'). "
```

framework/py/flwr/cli/app_cmd/review.py
```diff
@@ -0,0 +1,252 @@
...
+    sign & submit the review to the Platform."""
+    # Load configs
+    pyproject_path = app_dir_login / FAB_CONFIG_FILE if app_dir_login else None
+    config, errors, warnings = load_and_validate(pyproject_path, check_module=False)
+    config = process_loaded_project_config(config, errors, warnings)
+    federation, federation_config = validate_federation_in_project_config(
+        federation, config, federation_config_overrides
+    )
+
+    # Load the authentication plugin (This is also shared util for loading auth plugin)
```
panh99:
```suggestion
    # Load the authentication plugin
```

PR #6102, opened 2025-11-10
refactor(framework): Make `import flwr` faster

panh99:
### Reduce import time to `0.13` second from `1` second

This makes all `flwr` commands and `flower-*` commands faster! Feel free to try it out, e.g., `python -X importtime -c "import flwr"`.

Core idea: do not eagerly initialize all modules if not necessary, especially the modules that import `ray`.

### DIFF
--- a/framework/py/flwr/__init__.py
+++ b/framework/py/flwr/__init__.py
@@ -37,7 +37,7 @@ _lazy_imports = {"simulation", "server", "client", "common"}
 def __getattr__(name: str) -> object:
     """Lazy import for legacy support."""
     if name in _lazy_imports:
-        module = importlib.import_module(f"flwr.{name}")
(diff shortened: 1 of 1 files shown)


framework/py/flwr/__init__.py
```diff
@@ -15,18 +15,29 @@
...
+
+
+# Lazy imports for legacy support
+_lazy_imports = {"simulation", "server", "client", "common"}
+
+
+def __getattr__(name: str) -> object:
+    """Lazy import for legacy support."""
+    if name in _lazy_imports:
+        module = importlib.import_module(f"flwr.{name}")
```
jafermarq:
```suggestion
        module = importlib.import_module(f"{__name__}.{name}")
```

danieljanes (approved):

PR #6103, opened 2025-11-10
feat(framework): Prepare `flwr federation list`

jafermarq:

PR #6107, opened 2025-11-11
refactor(framework): Add capability to `FederationManager` to list federations an account is a member of

jafermarq:
Merge after #6103

### DIFF
--- a/framework/py/flwr/superlink/federation/federation_manager.py
+++ b/framework/py/flwr/superlink/federation/federation_manager.py
@@ -38,5 +38,5 @@ class FederationManager(ABC):
         """Given a node ID, check if it is in the federation."""
 
     @abstractmethod
-    def ls(self, flwr_aid: str) -> list[str]:
-        """List federations the account is member of."""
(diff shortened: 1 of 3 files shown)


framework/py/flwr/cli/app.py
```diff
@@ -63,6 +63,7 @@
 supernode_app.command(hidden=True)(supernode_list)
 app.add_typer(supernode_app, name="supernode")
 
+
```
jafermarq:
```suggestion
```

framework/py/flwr/superlink/federation/federation_manager.py
```diff
@@ -36,3 +36,7 @@ def filter_nodes(self, node_ids: set[int], federation: str) -> set[int]:
     @abstractmethod
     def has_node(self, node_id: int, federation: str) -> bool:
         """Given a node ID, check if it is in the federation."""
+
+    @abstractmethod
+    def ls(self, flwr_aid: str) -> list[str]:
```
panh99: I think this is not a common Python method name. Could we rename it to something like `get_federations`?

PR #6114, opened 2025-11-12
refactor(framework): Send `flwr` version in `flwr new`

yan-gao-GY:
## Proposal

Send `flwr` version in `flwr new`.


framework/py/flwr/cli/utils.py
```diff
@@ -419,6 +420,7 @@ def request_download_link(
     body = {
         "app_id": app_id,  # send raw string of app_id
         "version": version,
```
```suggestion
        "app_version": version,
```

yan-gao-GY: Hi @panh99 we need to make the change ("version" -> "app_version") on platform-api first, otherwise it will get an error.

PR #6126, opened 2025-11-15
refactor(datasets): Adjust `datasets` versions

jafermarq:
HF's `datasets` 4.0 has been out for a couple of months, but it's not supported by `flower-datasets` officially yet.

- In HF datasets 4.x, the `audio` extra requires `torch` and `torchcodec`. Therefore, these have been added as optional extras. 
- Minor tweaks to test due to changes in HF datasets 4.x
- CI tests now need to install `torch` which requires first freeing some space and installing `ffmpeg`  (hence the changes in the workflows)

### DIFF
--- a/framework/py/flwr/cli/flower_config.py
+++ b/framework/py/flwr/cli/flower_config.py
@@ -15,9 +15,12 @@
 """Flower command line interface configuration utils."""
 
 
+import re
+from pathlib import Path
 from typing import Any, cast
 
 import tomli
+import tomli_w
 import typer
 
 from flwr.cli.constant import (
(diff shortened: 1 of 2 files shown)


datasets/pyproject.toml
```diff
@@ -52,7 +52,7 @@ exclude = ["./**/*_test.py"]
 [tool.poetry.dependencies]
 python = "^3.10"
 numpy = ">=1.26.0,<3.0.0"
-datasets = ">=2.14.6 <=3.1.0"
+datasets = ">=3.0.0, <5.0.0"
```
jafermarq:
```suggestion
datasets = ">=2.14.6, <5.0.0"
```

PR #6134, opened 2025-11-16
feat(framework): Support Python 3.14

danieljanes:


garciadias:
Hi @danieljanes and Flower friends, is there anything still blocking this one that I could help with?

Some context on why we're asking. We build [FLIP](https://github.com/londonaicentre/FLIP), and our FL images build FROM flwr/base:<version>-py<X.Y>-ubuntu24.04. We've just finished evaluating a move to 3.13/3.14 across the platform, and the published base image is the single thing keeping the Flower backend on 3.12.

On timing: 3.15 reached rc1 on 4 August and is due final on 1 October, so `SUPPORTED_PYTHON_VERSIONS` topping out at 3.13 will shortly be N-2.

The docformatter/untokenize issue @panh99 raised also looks resolved from both direction, PyCQA/docformatter#327 was closed in April and 1.7.8 dropped the untokenize dependency, and main has since moved to docstrfmt regardless.

What we've seen on 3.14 so far is encouraging. Installed from PyPI, flwr 1.33.0 imports cleanly on 3.14 including ClientApp/ServerApp, and the unit suite for our own Flower-facing API service passes unchanged there. torch 2.11 (cu128) publishes cp314 and cp314t wheels, so the ML side of our apps is covered too. What we haven't been able to do is exercise a real federation on 3.14. We could build our own image, but we would prefer to keep it consistent and follow your lead on this.

The branch looks like it has conflicts with main now. If it's mainly a question of someone finding the time, I'm happy to rebase it and take it through CI, or open a fresh PR with the same changes if you'd rather leave yours as it stands. Whichever is more useful to you.

Many thanks. 🙏🏾

PR #6198, opened 2025-11-25
refactor(framework): Update CLI token loading methods

yan-gao-GY:
## Proposal

Update CLI token loading methods for consistency.

### DIFF
--- a/framework/py/flwr/common/exit/exit_handler.py
+++ b/framework/py/flwr/common/exit/exit_handler.py
@@ -58,5 +58,9 @@ def trigger_exit_handlers() -> None:
     """Trigger all registered exit handlers in LIFO order."""
     with _lock_handlers:
         for handler in reversed(registered_exit_handlers):
-            handler()
+            try:
+                handler()
+            except Exception:  # pylint: disable=broad-exception-caught
+                # Ignore exceptions in exit handlers
+                pass
         registered_exit_handlers.clear()
--- a/framework/py/flwr/common/exit/exit_handler_test.py
+++ b/framework/py/flwr/common/exit/exit_handler_test.py
@@ -1,0 +1,88 @@
+# Copyright 2025 Flower Labs GmbH. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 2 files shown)


framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -269,7 +272,7 @@ def _build_multipart_files_param(
 
 def _post_files(
     files_param: list[tuple[str, tuple[str, IO[bytes], str]]],
-    token: str,
+    token: str | None,
```
danieljanes: Should we avoid adding the `Authorization` header if `token` is `None`?

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -70,28 +74,27 @@ def publish(
...
-                "❌ Please log in before publishing app.",
-                fg=typer.colors.RED,
-                err=True,
-            )
-            raise typer.Exit(code=1)
+    # Load the authentication plugin (This is also shared util for loading auth plugin)
+    auth_plugin = load_cli_auth_plugin(app, federation, federation_config)
+    if not isinstance(auth_plugin, OidcCliPlugin):
+        typer.secho(
+            "❌ Please log in before reviewing app.",
```
panh99:
```suggestion
            "❌ Please log in before publishing app.",
```

framework/py/flwr/cli/app_cmd/publish.py
```diff
@@ -70,28 +74,27 @@ def publish(
...
 
-        creds_path = app.absolute() / FLWR_DIR / CREDENTIALS_DIR / f"{federation}.json"
-        if not creds_path.is_file():
-            typer.secho(
-                "❌ Please log in before publishing app.",
-                fg=typer.colors.RED,
-                err=True,
-            )
-            raise typer.Exit(code=1)
+    # Load the authentication plugin (This is also shared util for loading auth plugin)
```
panh99:
```suggestion
    # Load the authentication plugin
```

framework/py/flwr/cli/app_cmd/publish.py
panh99: Good catch! I think this was caused by another minor issue. I've fixed it and we don't need to amend the type annotation here. If the token is None, the command will prompt users to log in.

PR #6199, opened 2025-11-25
refactor(framework): Update `app_spec` parsing

yan-gao-GY:
## Proposal

Add `app_spec` parser function that can be used across CLIs.

### DIFF
--- a/framework/py/flwr/cli/utils.py
+++ b/framework/py/flwr/cli/utils.py
@@ -682,7 +682,7 @@ def parse_app_spec(app_spec: str) -> tuple[str, str | None]:
 
     Raises
     ------
-    typer.Exit
+    typer.Exit(code=1)
         If the app specification format is invalid.
     """
     if "==" in app_spec:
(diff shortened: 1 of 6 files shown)


framework/py/flwr/cli/utils.py
```diff
@@ -663,3 +664,49 @@ def validate_credentials_content(creds_path: Path) -> str:
...
+        '@account/app==x.y.z' (digits only).
+
+    Returns
+    -------
+    tuple[str, str | None]
+        A tuple containing the app ID and optional version.
+
+    Raises
+    ------
+    typer.Exit(code=1)
```
panh99, line 685:
```suggestion
    typer.Exit
```

framework/py/flwr/supercore/constant.py
```diff
@@ -26,7 +26,7 @@
 FLWR_IN_MEMORY_DB_NAME = ":flwr-in-memory:"
 
 # Constants for Hub
-APP_ID_PATTERN = r"^@(?P<account>[^/]+)/(?P<app>[^/]+)$"
+APP_ID_PATTERN = r"^@[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$"
```
panh99, line 29: Is this a bug fix, or we are changing the app ID pattern in this PR?

yan-gao-GY: It's not a bug, but we want to exclude all punctuations other than "_.-" in app ID naming.

PR #6201, opened 2025-11-26
feat(framework): Allow fetching retrieved messages in `NodeState`

panh99:
This will be used internally to fetch all retrieved messages for a given run and insert error messages when the ClientApp heartbeat is missing.


framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -90,9 +91,10 @@ def get_messages(
                 entry = self.msg_store[object_id]
                 message = entry.message
 
-                # Skip messages that have already been retrieved
-                if entry.is_retrieved:
-                    continue
+                # Filter by retrieved status if specified
+                if is_retrieved is not None:
```
jafermarq, line 95: how about handling this as a boolean? so having an intermediate var that's `False` if the input `is_retrieved` is None or False?

PR #6202, opened 2025-11-26
refactor(framework): Add app version not found warning

yan-gao-GY:
## Proposal

- List available app versions if the specified version doesn't exist.
- Refactor error message for consistency.

### DIFF
--- a/framework/py/flwr/cli/utils.py
+++ b/framework/py/flwr/cli/utils.py
@@ -269,6 +269,7 @@ def get_account_auth_config_path(root_dir: Path, federation: str) -> Path:
             f"Please check the permissions of `{gitignore_path}` and try again.",
             fg=typer.colors.RED,
             bold=True,
+            err=True,
         )
         raise typer.Exit(code=1) from err
 
@@ -439,6 +440,7 @@ def flwr_cli_grpc_exc_handler() -> Iterator[None]:  # pylint: disable=too-many-b
                 " to authenticate and try again.",
                 fg=typer.colors.RED,
                 bold=True,
+                err=True,
             )
             raise typer.Exit(code=1) from None
         if e.code() == grpc.StatusCode.UNIMPLEMENTED:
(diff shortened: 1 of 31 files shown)


framework/py/flwr/cli/utils.py
```diff
@@ -572,19 +572,41 @@ def request_download_link(
...
+            err=True,
+        )
+        raise typer.Exit(code=1) from e
 
     if resp.status_code == 404:
-        raise typer.BadParameter(f"'{app_id}' not found in Platform API")
+        available_app_versions = resp.json()["detail"]["available_app_versions"]
+        available_versions_str = ", ".join(map(str, available_app_versions))
+        typer.secho(
+            f"{app_id} not found in Platform API. "
```
panh99: Is this correct? It's a wrong-version error, not an  invalid app ID, right?

framework/py/flwr/cli/utils.py
```diff
@@ -572,19 +572,41 @@ def request_download_link(
     try:
         resp = requests.post(in_url, headers=headers, data=json.dumps(body), timeout=20)
     except requests.RequestException as e:
-        raise typer.BadParameter(f"Unable to connect to Platform API: {e}") from e
+        typer.secho(
+            f"Unable to connect to Platform API: {e}",
+            fg=typer.colors.RED,
+            err=True,
```
panh99, line 591: Should we remove this `err=True`? Just went through our CLI code, and we rarely print the error message to stderr.

framework/py/flwr/cli/utils.py
yan-gao-GY: Good catch! it's for `app_id` and `app_version` actually. I'll adjust the message.

framework/py/flwr/cli/utils.py
yan-gao-GY: I used `err=True` in `flwr new`, `flwr app publish`, and `flwr app review`. Do you think we should remove all of them?

danieljanes (approved):

PR #6203, opened 2025-11-26
fix(framework): Handle `AppExitException` in `flwr-serverapp`

jafermarq:


framework/py/flwr/app/exception.py
panh99:
```suggestion
   When raised (and not suppressed), the process will exit and report a telemetry event with the associated
   exit code.
```

framework/py/flwr/app/exception.py
```diff
@@ -15,11 +15,12 @@
...
 
 
-class AppExitException(BaseException):
+class AppExitException(Exception):
     """Base exception for all application-level errors in ServerApp and ClientApp.
 
     When raised, the process will exit and report a telemetry event with the associated
-    exit code. This is not intended to be caught by user code.
+   When raised (and not suppressed), the process will exit and report a telemetry event with the associated
+   exit code.
```
panh99:
```suggestion
   When raised (and not suppressed), the process will exit and report a telemetry event with the associated
   exit code.
```

PR #6208, opened 2025-11-27
refactor(framework): Add `err=True` to all `typer.secho` error message

yan-gao-GY:
## Proposal

Add `err=True` to all `typer.secho` error message, to ensure consistency, and output errors to stderr.


panh99 (approved):
LGTM!

PR #6240, opened 2025-12-04
feat(framework): Add utility functions for reading global connection config

panh99:


framework/py/flwr/supercore/utils.py
```diff
@@ -50,3 +56,14 @@ def int64_to_uint64(signed: int) -> int:
...
+
+def get_flwr_home() -> Path:
+    """Get the Flower home directory path.
+
+    Returns FLWR_HOME environment variable if set, otherwise returns a default
+    subdirectory in the user's home directory.
+    """
+    if flwr_home := os.getenv(FLWR_HOME):
+        return Path(flwr_home)
+    return Path.home() / FLWR_DIR
```
panh99, line 76: We also have a similar `get_flwr_dir` function. I wonder if we should merge this with it.

The original function is a bit unnecessarily complicated:

```python
def get_flwr_dir(provided_path: str | None = None) -> Path:
    """Return the Flower home directory based on env variables."""
    if provided_path is None or not Path(provided_path).is_dir():
        return Path(
            os.getenv(
                FLWR_HOME,
                Path(f"{os.getenv('XDG_DATA_HOME', os.getenv('HOME'))}") / FLWR_DIR,
            )
        )
    return Path(provided_path).absolute()
```

framework/py/flwr/supercore/constant.py
```diff
@@ -69,3 +69,18 @@ class NodeStatus:
...
+
+
+# CLI connection configuration file name
+CONNECTION_CONFIG_FILE = "config.yml"
+
+
+class ConnectionConfigYamlKey:
+    """YAML keys for connection configuration."""
+
+    CURRENT_CONNECTION = "current-connection"
```
jafermarq: should we still include this in the v1?

framework/py/flwr/supercore/utils.py
jafermarq: As discussed, we won't merge them but we'll add a depcwarning when using `XDG_DATA_HOME` and recommend using `FLWR_HOME` instead.

danieljanes (approved):

PR #6243, opened 2025-12-05
feat(framework): Store traffic in `LinkState`

chongshenng:
Merge after:
* #6244

### DIFF
--- a/framework/docs/source/contributor-how-to-install-development-versions.rst
+++ b/framework/docs/source/contributor-how-to-install-development-versions.rst
@@ -63,15 +63,14 @@ Install ``flwr`` from a specific GitHub branch (``branch-name``):
  Open Jupyter Notebooks on Google Colab
 ****************************************
 
-Open the notebook
-``framework/docs/source/tutorial-series-get-started-with-flower-pytorch.ipynb``:
+Open the notebook ``examples/flower-in-30-minutes/tutorial.ipynb``:
 
-- https://colab.research.google.com/github/adap/flower/blob/main/framework/docs/source/tutorial-series-get-started-with-flower-pytorch.ipynb
+- https://colab.research.google.com/github/adap/flower/blob/main/examples/flower-in-30-minutes/tutorial.ipynb
 
 Open a development version of the same notebook from branch `branch-name` by changing
 ``main`` to ``branch-name`` (right after ``blob``):
 
-- https://colab.research.google.com/github/adap/flower/blob/branch-name/framework/docs/source/tutorial-series-get-started-with-flower-pytorch.ipynb
+- https://colab.research.google.com/github/adap/flower/blob/branch-name/examples/flower-in-30-minutes/tutorial.ipynb
 
 Install a `whl` on Google Colab:
 
--- a/framework/docs/source/ref-faq.rst
+++ b/framework/docs/source/ref-faq.rst
@@ -9,8 +9,7 @@ Flower.
 
     Yes, it can! Flower even comes with a few under-the-hood optimizations to make it work even better on Colab. Here's a quickstart example:
 
-    * `Flower simulation PyTorch <https://colab.research.google.com/github/adap/flower/blob/main/examples/simulation-pytorch/sim.ipynb>`_
-    * `Flower simulation TensorFlow/Keras <https://colab.research.google.com/github/adap/flower/blob/main/examples/simulation-tensorflow/sim.ipynb>`_
+    * `Flower 30-minute tutorial <https://colab.research.google.com/github/adap/flower/blob/main/examples/flower-in-30-minutes/tutorial.ipynb>`_
 
 .. dropdown:: :fa:`eye,mr-1` How can I run Federated Learning on a Raspberry Pi?
 
--- a/framework/docs/source/tutorial-series-get-started-with-flower-pytorch.rst
+++ b/framework/docs/source/tutorial-series-get-started-with-flower-pytorch.rst
@@ -96,7 +96,6 @@ It should have the following structure:
 .. code-block:: shell
 
     flower-tutorial
-    ├── README.md
     ├── flower_tutorial
     │   ├── __init__.py
     │   ├── client_app.py   # Defines your ClientApp


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -778,3 +778,39 @@ def store_traffic(
...
+                run_id,
+                bytes_sent,
+                bytes_recv,
+            )
+            return
+
+        # Skip if no traffic to record
+        if bytes_sent == 0 and bytes_recv == 0:
+            log(WARNING, "No traffic data to store for run_id %d", run_id)
+            return
```
jafermarq: when would this be the case?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -778,3 +778,39 @@ def store_traffic(
...
+        # Validate non-negative values
+        if bytes_sent < 0 or bytes_recv < 0:
+            log(
+                WARNING,
+                "Negative traffic values for run %d: bytes_sent=%d, bytes_recv=%d",
+                run_id,
+                bytes_sent,
+                bytes_recv,
+            )
+            return
```
jafermarq: should this be treated as an error? the function that would call this will not know an issue (negative bytes) happened.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
chongshenng: This will be a very edge case, if the `ServerApp` and `ClientApp` is not sending anything, e.g. if some computation failed in the `ClientApp` and the user opts to return nothing in the `Message`. IMO, it's necessary to log this.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
jafermarq: I think logging as error is good but what i meant is to actually treat is as an error and raise an exception that can be handled in the function that calls this one. Without raising an excp. the caller will have no idea something was wrong (since the return type is always `None`)

jafermarq: likewise for the other scenarios where `None` is returned after logging an `ERROR`

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
jafermarq: should we raise an exception in addition to log an error? else the caller doesn't know somethign is very wrong (negative bytes)

chongshenng: @jafermarq I think your comment was meant for [the other thread](https://github.com/adap/flower/pull/6243#discussion_r2592851272) where the code originally logged an `ERROR`. IMO it's better to raise an exception since it already contains the error message. We can handle the error in `FleetServicer` when `state.store_traffic()` is called. Wdyt?

chongshenng: Updated to raise `ValueError` for negative values and when `run_id` is not found.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
chongshenng: Updated to raise `ValueError` for negative values and when `run_id` is not found.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
jafermarq, line 1: for completeness. should we include a test that attempts to call `.store_traffic()` to a run that doesn't exist?

danieljanes: We also need a test that prevents negative values from being added

chongshenng: Added

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -776,3 +776,24 @@ def get_serverapp_log(
 
     def store_traffic(self, run_id: int, *, bytes_sent: int, bytes_recv: int) -> None:
```
panh99, line 777: Quick q: when calling `LinkState.store_traffic`, will we ever provide non-zero values for both `bytes_sent` and `bytes_recev`? If not, should we set default values for the two arguments?

framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -1263,6 +1263,40 @@ def get_valid_message_ins(self, message_id: str) -> dict[str, Any] | None:
...
+            if not rows:
+                raise ValueError(f"Run {run_id} not found")
+
+            # Perform the update
+            update_query = """
+                    UPDATE run
+                    SET bytes_sent = bytes_sent + ?,
+                        bytes_recv = bytes_recv + ?
+                    WHERE run_id = ?;
+                """
```
panh99:
```suggestion
                    UPDATE run
                    SET bytes_sent = bytes_sent + ?,
                        bytes_recv = bytes_recv + ?
                    WHERE run_id = ?
                    RETURNING run_id;
                """
```

framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -1263,6 +1263,40 @@ def get_valid_message_ins(self, message_id: str) -> dict[str, Any] | None:
...
+            # Check if run exists
+            query = """
+                    SELECT run_id
+                    FROM run
+                    WHERE run_id = ?;
+                """
+            rows = self.conn.execute(query, (sint64_run_id,)).fetchall()
+
+            if not rows:
+                raise ValueError(f"Run {run_id} not found")
```
panh99: It would be good to reduce the number of queries when possible. We can use `RETURNING` keyword to check if the row exist. See the change below.

framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
panh99: And then we can check if returned rows is not empty to see if the run row exists.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
chongshenng: No, we don't ever set both to zero at the same time. @danieljanes brought up a point about not setting default values here as it could be introduce subtle bugs, which I tend to agree. So, the acceptable usage here would be
* `LinkState.store_traffic(run_id: ..., bytes_sent=<val>, bytes_recv=0)`
* `LinkState.store_traffic(run_id: ..., bytes_sent=0, bytes_recv=<val>)`

framework/py/flwr/server/superlink/linkstate/sqlite_linkstate.py
```diff
@@ -1263,6 +1263,35 @@ def get_valid_message_ins(self, message_id: str) -> dict[str, Any] | None:
...
+        sint64_run_id = uint64_to_int64(run_id)
+
+        with self.conn:
+            # Check if run exists, performing the update only if it does
+            update_query = """
+            UPDATE run
+            SET bytes_sent = bytes_sent + ?,
+                bytes_recv = bytes_recv + ?
+            WHERE run_id = ?
+            RETURNING run_id;
```
panh99:
```suggestion
                UPDATE run
                SET bytes_sent = bytes_sent + ?,
                    bytes_recv = bytes_recv + ?
                WHERE run_id = ?
                RETURNING run_id;
```

PR #6244, opened 2025-12-05
feat(framework): Introduce `store_traffic` to `LinkState`

chongshenng:
Merge after
* #6242


framework/py/flwr/server/superlink/linkstate/linkstate.py
```diff
@@ -480,3 +480,22 @@ def get_serverapp_log(
...
+    ) -> None:
+        """Store traffic data for the specified `run_id`.
+
+        Parameters
+        ----------
+        run_id : int
+            The identifier of the run for which to store traffic data.
+        bytes_sent : int (default: 0)
+            The number of bytes sent from SuperLink to SuperNodes to add to the run's
+            total (default: 0). A value of 0 means no bytes are added to the counter.
```
jafermarq:
```suggestion
            The number of bytes pulled by SuperNodes from the SuperLink to add to the run's
            total (default: 0). A value of 0 means no bytes are added to the counter.
```

I wonder if it's better to rephrase it to keep it consistent with how blobs are communicated.

chongshenng: Makes sense. Updated.

PR #6245, opened 2025-12-05
feat(framework): Record traffic in `FleetServicer`

chongshenng:
Merge after:
* #6243


framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
jafermarq, line 1: this looks good! but we also need the same logic in `Pull/PushMessage` (although it's true they will contribute minimally to the total `bytes` communicated).  But, given that some messages don't have objects (e.g. those that don't carry `.content` --i.e. carrying an `Error`) it would be good to account for them.

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -288,8 +288,12 @@ def PushObject(
                 state=self.state_factory.state(),
                 store=self.objectstore_factory.store(),
             )
-        except InvalidRunStatusException as e:
-            abort_grpc_context(e.message, context)
+            # Record traffic
+            bytes_recv = len(request.object_content)
+            state = self.state_factory.state()
```
jafermarq: Since we are also calling `.state()` a few lines above, is it better if we do it just once? wdyt @panh99 ?

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer_test.py
```diff
@@ -652,16 +652,18 @@ def test_pull_object_successful(self) -> None:
         # Preregister object
         self.store.preregister(run_id, get_object_tree(obj))
 
-        # Pull
-        req = PullObjectRequest(
-            node=Node(node_id=node_id), run_id=run_id, object_id=obj.object_id
-        )
-        res: PullObjectResponse = self._pull_object(req)
+        # Mock store_traffic to avoid validation error when object_content is empty
```
jafermarq, line 655:
```suggestion
        # Mock store_traffic to avoid validation error when object_content is empty
        # This is because the object has been preregistered but not yet pushed
```

jafermarq: Extending the comment slightly. Maybe worth doing in the other places too?

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -152,12 +154,26 @@ def pull_messages(
...
             log(ERROR, e.message)
             # Delete message ins from state
             state.delete_messages(message_ins_ids={msg_object_id})
 
-    return PullMessagesResponse(messages_list=msg_proto, message_object_trees=trees)
+    response = PullMessagesResponse(messages_list=msg_proto, message_object_trees=trees)
+
+    # Record traffic only if message was successfully processed
+    if run_id_to_record is not None:
+        bytes_sent = len(response.SerializeToString())
```
jafermarq, line 171: just a note that the way the measurement is is done for `Messages` and `Objects` is different. Here for `Messages` we are counting the `bytes` of the entire request but for `Objects` we don't.

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -152,12 +154,26 @@ def pull_messages(
             # Add Message and its object tree to the response
             msg_proto.append(message_to_proto(msg))
             trees.append(obj_tree)
+
+            # Track run_id for traffic recording
+            run_id = msg.metadata.run_id
```
jafermarq: Should this `run_id` be instead `run_id_to_record`? (it will be left unset otherwise)

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
chongshenng: Makes sense. Updated.

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
chongshenng: Yes, that's correct. Doing `response.SerializeToString()` could incur some overhead for `Objects`, which is why we can directly count `bytes_sent/recv` from the `object_content`.

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
chongshenng: Good catch!

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -281,15 +281,19 @@ def PushObject(
...
+                state=state,
                 store=self.objectstore_factory.store(),
             )
-        except InvalidRunStatusException as e:
-            abort_grpc_context(e.message, context)
+            # Record traffic
+            bytes_recv = len(request.object_content)
+            state.store_traffic(request.run_id, bytes_sent=0, bytes_recv=bytes_recv)
+        except (InvalidRunStatusException, ValueError) as e:
+            abort_grpc_context(str(e), context)
```
panh99: `abort_grpc_context` will raise a permission denied error. Would it be a bit strange to raise this error when the bytes sent/received are wrong? Btw, are we capturing the `ValueError` from `push_object`, or from `state.store_traffic`?

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -313,8 +317,12 @@ def PullObject(
...
                 store=self.objectstore_factory.store(),
             )
-        except InvalidRunStatusException as e:
-            abort_grpc_context(e.message, context)
+            # Record traffic
+            bytes_sent = len(res.object_content)
+            state = self.state_factory.state()
+            state.store_traffic(request.run_id, bytes_sent=bytes_sent, bytes_recv=0)
+        except (InvalidRunStatusException, ValueError) as e:
+            abort_grpc_context(str(e), context)
```
panh99: Same as above.

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -281,15 +281,19 @@ def PushObject(
...
                 request=request,
-                state=self.state_factory.state(),
+                state=state,
                 store=self.objectstore_factory.store(),
             )
-        except InvalidRunStatusException as e:
-            abort_grpc_context(e.message, context)
+            # Record traffic
+            bytes_recv = len(request.object_content)
+            state.store_traffic(request.run_id, bytes_sent=0, bytes_recv=bytes_recv)
```
panh99: Should we move this to `message_handler.push_object`?

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
```diff
@@ -313,8 +317,12 @@ def PullObject(
                 state=self.state_factory.state(),
                 store=self.objectstore_factory.store(),
             )
-        except InvalidRunStatusException as e:
-            abort_grpc_context(e.message, context)
+            # Record traffic
+            bytes_sent = len(res.object_content)
+            state = self.state_factory.state()
+            state.store_traffic(request.run_id, bytes_sent=bytes_sent, bytes_recv=0)
```
panh99: Should we move this to `message_handler.pull_object`?

framework/py/flwr/server/superlink/fleet/grpc_rere/fleet_servicer.py
chongshenng: Good point. My original implementation was to capture `ValueError` from `state.store_traffic`. I've updated this to store the traffic in `message_handler.push/pull_object` and capture the `ValueError` there.

Question: Do you think it's sufficient to log the error and not abort the gRPC context? IMO not being able to store the traffic is less severe than `InvalidRunStatusException`. Wdyt?

chongshenng: Removed gRPC abort check in latest commit.

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -152,12 +154,23 @@ def pull_messages(
...
         except NoObjectInStoreError as e:
             log(ERROR, e.message)
             # Delete message ins from state
             state.delete_messages(message_ins_ids={msg_object_id})
 
-    return PullMessagesResponse(messages_list=msg_proto, message_object_trees=trees)
+    response = PullMessagesResponse(messages_list=msg_proto, message_object_trees=trees)
+
+    # Record traffic only if message was successfully processed
+    if run_id_to_record is not None:
```
panh99, line 170: Should we call store_traffic for each message? Cuz they may belong to different runs.

panh99: Nvm, I noticed that we’ve already implicitly assumed that all messages in the request belong to the same run. Would you mind adding a comment here about the assumption?

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -193,6 +209,10 @@ def push_messages(
         results={str(message_id): 0},
         objects_to_push=objects_to_push,
     )
+
+    # Record traffic
+    state.store_traffic(run_id, bytes_sent=0, bytes_recv=bytes_recv)
```
panh99, line 216: Same as above.

PR #6260, opened 2025-12-08
ci(framework): Update `framework` CI tests to use examples

jafermarq:
Stop using `flwr new` templates and instead run some `quickstart-` examples. Setting upper limit for `datasets` (that of the current version of `flwr-datasets`)  in `quickstart-jax` so CI test can resolve dependencies within the time limit for the job.

### DIFF
--- a/examples/quickstart-pytorch/pyproject.toml
+++ b/examples/quickstart-pytorch/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [project]
-name = "pytorchexample"
+name = "quickstart-pytorch"
 version = "1.0.0"
 description = "Federated Learning with PyTorch and Flower (Quickstart Example)"
 license = "Apache-2.0"
--- a/framework/py/flwr/client/grpc_adapter_client/connection.py
+++ b/framework/py/flwr/client/grpc_adapter_client/connection.py
@@ -45,7 +45,7 @@ def grpc_adapter(  # pylint: disable=R0913,too-many-positional-arguments
     tuple[
         int,
         Callable[[], tuple[Message, ObjectTree] | None],
-        Callable[[Message, ObjectTree], set[str]],
+        Callable[[Message, ObjectTree, float], set[str]],
         Callable[[int], Run],
         Callable[[str, int], Fab],
         Callable[[int, str], bytes],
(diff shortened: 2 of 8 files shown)


.github/workflows/framework-e2e.yml
```diff
@@ -303,15 +303,21 @@ jobs:
...
     strategy:
       matrix:
-        framework: ["numpy", "pytorch", "PyTorch (Legacy API, deprecated)","tensorflow", "jax", "sklearn"]
+        include:
+          - example: quickstart-pytorch
+          - example: quickstart-tensorflow
+          - example: quickstart-jax
+          - example: quickstart-sklearn-tabular
+          - example: quickstart-numpy
+          - example: quickstart-pytorch-deprecated
```
panh99: I noticed that `quickstart-pytorch-deprecated` is not on the Flower Hub, right?

.github/workflows/framework-e2e.yml
```diff
@@ -327,22 +333,19 @@ jobs:
...
         run: |
           python -m pip install https://${{ env.ARTIFACT_BUCKET }}/py/${{ needs.wheel.outputs.dir }}/${{ needs.wheel.outputs.short_sha }}/${{ needs.wheel.outputs.whl_path }}
-      - name: Create project and install it
+      - name: Install project dependencies
         if: ${{ needs.changes.outputs.framework == 'true' }}
+        working-directory: examples/${{ matrix.example }}
         run: |
-          CLEAN_FRAMEWORK=$(echo "${{ matrix.framework }}" | sed 's/[^a-zA-Z0-9]//g' | tr '[:upper:]' '[:lower:]')
-          flwr new tmp-${CLEAN_FRAMEWORK} --framework="${{ matrix.framework }}" --username gh_ci
-          cd tmp-${CLEAN_FRAMEWORK}
```
panh99, line 335: I think we need to add a new `flwr new` command here, no?

jafermarq: no, these test will use the  content in `examples/`

.github/workflows/framework-e2e.yml
```diff
@@ -303,15 +303,21 @@ jobs:
...
     timeout-minutes: 10
     needs: [changes, wheel]
     strategy:
       matrix:
-        framework: ["numpy", "pytorch", "PyTorch (Legacy API, deprecated)","tensorflow", "jax", "sklearn"]
+        include:
+          - example: quickstart-pytorch
+          - example: quickstart-tensorflow
+          - example: quickstart-jax
+          - example: quickstart-sklearn-tabular
```
jafermarq:
```suggestion
          - example: quickstart-sklearn
```

PR #6269, opened 2025-12-09
feat(framework): Extend `NodeState` base class

chongshenng:

PR #6270, opened 2025-12-09
feat(framework): Implement `InMemoryNodeState` runtime methods 

chongshenng:
Merge after:
* #6269

### DIFF
--- a/framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
+++ b/framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
@@ -253,12 +253,10 @@ class InMemoryNodeState(
             duration = entry.finished_at - entry.starting_at
             return duration
 
-    def _cleanup_old_message_times(
-        self, max_age_seconds: float = MESSAGE_TIME_ENTRY_MAX_AGE_SECONDS
-    ) -> None:
-        """Remove time entries older than max_age_seconds."""
+    def _cleanup_old_message_times(self) -> None:
+        """Remove time entries older than MESSAGE_TIME_ENTRY_MAX_AGE_SECONDS."""
         with self.lock_time_store:
-            cutoff = now().timestamp() - max_age_seconds
+            cutoff = now().timestamp() - MESSAGE_TIME_ENTRY_MAX_AGE_SECONDS
             to_delete = [
                 msg_id
                 for msg_id, entry in self.time_store.items()


framework/py/flwr/supernode/nodestate/nodestate_test.py
```diff
@@ -213,6 +214,102 @@ def test_get_error_reply_when_token_expires(self) -> None:
...
+        msg_id = "test_msg_123"
+
+        # Execute: record start time
+        self.state.record_message_processing_start(msg_id)
+
+        # Simulate some processing time
+        sleep(0.01)
+
+        # Execute: record end time
+        self.state.record_message_processing_end(msg_id)
```
panh99: Should we patch `datatime.datetime` to simulate time passing? For example:

```python
        # Execute
        patched_dt = now() + timedelta(seconds=10)

        with patch("datetime.datetime") as mock_dt:
            mock_dt.now.return_value = patched_dt
```

framework/py/flwr/supernode/nodestate/nodestate_test.py
```diff
@@ -213,6 +214,102 @@ def test_get_error_reply_when_token_expires(self) -> None:
...
+
+        # Execute: record timing for first message
+        self.state.record_message_processing_start(msg1_id)
+        sleep(0.01)
+        self.state.record_message_processing_end(msg1_id)
+
+        # Execute: record timing for second message
+        self.state.record_message_processing_start(msg2_id)
+        sleep(0.02)
+        self.state.record_message_processing_end(msg2_id)
```
panh99: Same as above.

framework/py/flwr/supernode/nodestate/nodestate_test.py
```diff
@@ -213,6 +214,102 @@ def test_get_error_reply_when_token_expires(self) -> None:
...
+        self.assertGreater(duration2, duration1)
+
+    def test_cleanup_old_message_times(self) -> None:
+        """Test that old message timing entries are cleaned up."""
+        # Prepare
+        old_msg_id = "old_msg"
+        recent_msg_id = "recent_msg"
+
+        # Record timing for an "old" message
+        with patch("flwr.supernode.nodestate.in_memory_nodestate.now") as mock_now:
```
panh99: Just a note, it would be easier to patch `"datetime.datetime"` as mentioned above, but this also works fine.

framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -211,9 +223,47 @@ def _on_tokens_expired(self, expired_records: list[tuple[int, float]]) -> None:
...
+                    f"Start time or end time for message ID {msg_id} is missing."
+                )
+
+            starting_at = datetime.fromisoformat(entry.starting_at)
+            finished_at = datetime.fromisoformat(entry.finished_at)
+            duration = (finished_at - starting_at).total_seconds()
+
+            return duration
+
+    def _cleanup_old_message_times(self, max_age_seconds: float = 3600) -> None:
```
panh99: Should we make `max_age_seconds` configurable, or just use a constant?

chongshenng: Good idea re: constant. Updated.

framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -211,10 +224,49 @@ def _on_tokens_expired(self, expired_records: list[tuple[int, float]]) -> None:
 
     def record_message_processing_start(self, message_id: str) -> None:
         """Record the start time of message processing based on the message ID."""
+        with self.lock_time_store:
+            self.time_store[message_id] = TimeEntry(starting_at=now().isoformat())
```
panh99: Since we just need to calculate the duration, should we simply save .timestamp() instead of .isoformat()?

panh99: Then we don’t need to convert them back to datetime and calculate.

chongshenng: Nice suggestion, thanks. I've implemented it.

framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -211,10 +223,46 @@ def _on_tokens_expired(self, expired_records: list[tuple[int, float]]) -> None:
...
+            if entry.starting_at is None or entry.finished_at is None:
+                raise ValueError(
+                    f"Start time or end time for message ID {message_id} is missing."
+                )
+
+            duration = entry.finished_at - entry.starting_at
+            return duration
+
+    def _cleanup_old_message_times(
+        self, max_age_seconds: float = MESSAGE_TIME_ENTRY_MAX_AGE_SECONDS
```
panh99: Should we make the max age an argument? Or just use the constant?

chongshenng: Should be a constant, you're right.

PR #6271, opened 2025-12-09
feat(framework): Introduce runtime in `connection`

chongshenng:
Merge after:
* #6266 
* #6272

### DIFF
--- a/framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
+++ b/framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
@@ -801,4 +801,7 @@ class InMemoryLinkState(LinkState, InMemoryCoreState):  # pylint: disable=R0902,
 
     def add_clientapp_runtime(self, run_id: int, runtime: float) -> None:
         """Add ClientApp runtime to the cumulative total for the specified `run_id`."""
-        raise NotImplementedError("Method not yet implemented.")
+        with self.lock:
+            if run_id not in self.run_ids:
+                raise ValueError(f"Run {run_id} not found")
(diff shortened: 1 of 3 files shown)


framework/py/flwr/client/grpc_adapter_client/connection.py
```diff
@@ -45,7 +46,7 @@ def grpc_adapter(  # pylint: disable=R0913,too-many-positional-arguments
     tuple[
         int,
         Callable[[], tuple[Message, ObjectTree] | None],
-        Callable[[Message, ObjectTree], set[str]],
+        Callable[[MessageSendContext, ObjectTree], set[str]],
```
panh99: Just curious, why not just add a new argument?

```suggestion
        Callable[[Message, ObjectTree, float], set[str]],
```

chongshenng: It's easier to modify it without changing the signature, e.g. if we want to pass any other parameter to `send()`- suggestion by Daniel.

panh99: Hmmm, looks a bit strange. For example, when updating `LinkState.create_node` to include a public key, we add a `public_key` argument instead of packing all arguments (`owner_aid`, `owner_name`, etc.) into a `CreateNodeInput` class, right? Wdyt? @danieljanes 

In the long run, I think the correct solution is to refactor this into `FleetConnection`.

Btw, if we want to introduce a `MessageSendContext`, I think it makes sense to also include the `ObjectTree` in it, right?

chongshenng: @panh99 Makes sense to keep explicit parameters rather than packing them. I'll push an update.

PR #6279, opened 2025-12-10
fix(framework): Fix time entry cleanup logic

chongshenng:


framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
panh99, line 54: Should we make `starting_at` compulsory? It will always be set, right?

```suggestion
class TimeEntry:
"""Data class to represent a time entry."""

starting_at: float
finished_at: float | None = None
```

framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -257,10 +257,14 @@ def _cleanup_old_message_times(self) -> None:
...
             cutoff = now().timestamp() - MESSAGE_TIME_ENTRY_MAX_AGE_SECONDS
+            # Find message IDs for entries that have a starting and finished time
+            # and that started before the cutoff
             to_delete = [
                 msg_id
                 for msg_id, entry in self.time_store.items()
-                if entry.starting_at and entry.starting_at < cutoff
+                if entry.starting_at
+                and entry.finished_at
+                and entry.finished_at < cutoff
```
panh99: In addition to this, should we also remove all entries that use message IDs that doesn't exist in the `message_store`? Cuz that may suggest the message has been replied due to clientapp crash. Wydt?

chongshenng: Added, thanks for the suggestion!

framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -257,10 +257,26 @@ def _cleanup_old_message_times(self) -> None:
...
```
panh99:
```suggestion
                if entry.finished_at and entry.finished_at < cutoff
```

panh99 (approved):
LGTM!

PR #6299, opened 2025-12-12
refactor(framework): Stop retrying when TLS handshake failure detected

panh99:
image


framework/py/flwr/common/args.py
```diff
@@ -89,7 +89,7 @@ def try_obtain_root_certificates(
     else:
         # Load the certificates if provided, or load the system certificates
         if root_cert_path is None:
-            log(INFO, "Using system certificates")
+            log(INFO, "Using system certificates for TLS connection.")
```
panh99:
```suggestion
            log(INFO, "Using system certificates for TLS connection")
```

danieljanes (dismissed):

danieljanes (approved):

PR #6314, opened 2025-12-22
feat(examples): Add Federated Analytics example

chongshenng:


examples/federated-analytics/federated_analytics/server_app.py
```diff
@@ -0,0 +1,85 @@
...
+        }
+    )
+    recorddict = RecordDict({"config": config})
+    messages = []
+    for node_id in node_ids:  # one message for each node
+        message = Message(
+            content=recorddict,
+            message_type=MessageType.QUERY,  # target `query` method in ClientApp
+            dst_node_id=node_id,
+            group_id="1",
```
jafermarq:
```suggestion
```

imo we can remove this

examples/federated-analytics/federated_analytics/client_app.py
```diff
@@ -0,0 +1,61 @@
...
+def query(msg: Message, context: Context) -> Message:
+    """Query PostgreSQL database and report aggregated results to `ServerApp`."""
+
+    # Get database connection details from node config
+    db_host: str = context.node_config.get("db-host", "localhost")
+    db_port: int = context.node_config.get("db-port", 5432)
+    db_name: str = context.node_config.get("db-name", "flwrlabs")
+    db_user: str = context.node_config.get("db-user", "flwrlabs")
+    db_password: str = context.node_config.get("db-password", "flwrlabs")
+    table_name: str = context.node_config.get("table-name", "person_measurements")
```
jafermarq, line 25: I wonder if it's good to add some minimal comment in the `README.md` about the fact that there are many hard-coded parameters here. Maybe it would make sense to pass them from `--nod-config` also? That will also remove the bit of having the user/password right there

jafermarq: potentailly passing `context.node_config` straight into `query_database` could save some lines of code

examples/federated-analytics/federated_analytics/server_app.py
```diff
@@ -0,0 +1,85 @@
...
+    aggregated_stats = aggregate_features(
+        replies, selected_features, feature_aggregation
+    )
+
+    # Display final aggregated stats
+    print("\n" + "=" * 40)
+    print("FINAL AGGREGATED STATISTICS".center(40))
+    print("=" * 40)
+    print(json.dumps(aggregated_stats, indent=2))
+    print("=" * 40 + "\n")
```
jafermarq, line 89: Could we also add a couple of lines to save `aggregated_stats` into a JSON file? (as we do in other examples with the "global model")

examples/federated-analytics/federated_analytics/client_app.py
chongshenng: Makes sense. Updated.

examples/federated-analytics/README.md
```diff
@@ -0,0 +1,179 @@
+---
+tags: [basic, tabular, federated analytics]
```
jafermarq:
```suggestion
tags: [tabular, federated analytics, omop]
```

PR #6785, opened 2026-03-17
fix(framework): Make message delivery safe for multiple LinkState replicas

psfoley:
### Checklist

- [x] Write tests
- [x] Make CI checks pass


framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -252,47 +252,58 @@ def get_message_ins(self, node_id: int, limit: int | None) -> list[Message]:
...
+                        WHERE dst_node_id = :node_id
+                        AND delivered_at = ''
+                        AND (created_at + ttl) > CAST(strftime('%s', 'now') AS REAL)
+                        ORDER BY rowid
+                        LIMIT :limit
+                    )
+                    AND delivered_at = ''
+                    RETURNING *
+                """
                 params["limit"] = limit
```
panh99: Let's unify the common part of the two queries

```suggestion
            condition = """
                dst_node_id = :node_id
                AND delivered_at = ''
                AND (created_at + ttl) > CAST(strftime('%s', 'now') AS REAL)
            """
            if limit is not None:
                condition = f"""
                    message_id IN (
                        SELECT message_id
                        FROM message_ins
                        WHERE {condition}
                        ORDER BY rowid
                        LIMIT :limit
                    )
                """
                params["limit"] = limit

            query = f"""
                UPDATE message_ins
                SET delivered_at = :delivered_at
                WHERE {condition}
                RETURNING *
            """

```

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -252,47 +252,58 @@ def get_message_ins(self, node_id: int, limit: int | None) -> list[Message]:
...
+            else:
+                query = """
+                    UPDATE message_ins
+                    SET delivered_at = :delivered_at
+                    WHERE message_id IN (
+                        SELECT message_id
+                        FROM message_ins
+                        WHERE dst_node_id = :node_id
+                        AND delivered_at = ''
+                        AND (created_at + ttl) > CAST(strftime('%s', 'now') AS REAL)
```
panh99: Should we do `AND (created_at + ttl) > :current` and use `now().timestamp()`? I am less sure if `CAST(strftime('%s', 'now') AS REAL)` will use UTC+0 timezone as we do elsewhere.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -434,15 +445,18 @@ def get_message_res(self, message_ids: set[str]) -> list[Message]:
             )
             ret.update(tmp_ret_dict)
 
-            # Find all reply Messages
+            # Atomically claim all eligible reply Messages.
```
panh99:
```suggestion
            # Atomically claim all eligible reply Messages
```

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
msheller: The SQL approach should be UTC. I think the more important difference may be "app-time" vs "DB-time". Do we do app-side logic on these values in our code? If so, it makes sense to me to normalize to "app-time", but if we don't, "DB-time" seems appropriate to me.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
msheller: @panh99 , curious about your take. I would want this to read something like:
common_condition = <what condition is right now>
if limit is None:
   condition = common_condition
else:
   condition = ... WHERE {common_condition} ...

It's a few more lines, adds an extra check, but makes it easier to pick up that the initial condition is used in both cases. Wdyt? Is this something you'd consider important or too nitpicky (or outright incorrect)?

msheller (dismissed):

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
panh99: Yeah, sure, sounds good to me! Let's use common_condition.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
panh99: Yeah, I think it's better to use `app-time` in the sense that we use `now()` everywhere (or at least we're supposed to use it everywhere).

Regarding "The SQL approach should be UTC", do you mean the DB will use UTC+0 timestamp even though the local timezone is, say, New York time?

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -1933,6 +1937,199 @@ def state_factory(self) -> SqlLinkState:
...
+                exceptions.append(ex)
+            for thread in threads:
+                thread.join(timeout=self._CONCURRENT_TEST_TIMEOUT)
+            alive_threads = [thread for thread in threads if thread.is_alive()]
+            if alive_threads:
+                self.fail("Timed out waiting for concurrent pull_res threads to finish")
+
+            if exceptions:
+                raise exceptions[0]
+            assert all(message_list is not None for message_list in results)
```
panh99: I think the big bulk of code above is very similar to `_run_in_parallel` function. Should we create a helper function (not a static method) to just do:

```python
def run_in_parallel(fns: Iterable[Callable[[], Any]], timeout: float | None = None) -> list[Any]:
    """Invoke all callables in the same time and return results as a list."""
    fns = list(fns)
    barrier = threading.Barrier(len(fns) + 1)
    results = [None] * len(fns)
    exceptions = []

    def _run(i, fn):
        try:
            barrier.wait()
            results[i] = fn()
        except Exception as e:
            exceptions.append(e)

    threads = [threading.Thread(target=_run, args=(i, fn)) for i, fn in enumerate(fns)]
    for t in threads:
        t.start()
    barrier.wait()  # release all threads at once

    for t in threads:
        t.join(timeout)

    if exceptions:
        raise exceptions[0]

    assert all(r is not None for r in results)
    return results
```

Then we can do:
```python
results = run_in_parallel(
    lambda: state_0.get_message_ins(node_id=node_id, limit=1),
    lambda: state_1.get_message_ins(node_id=node_id, limit=1),
)
```

Also:
```python
msg_id = pulled_ins.metadata.message_id

results = run_in_parallel(
    lambda: state_0.get_message_res({msg_id}),
    lambda: state_1.get_message_res({msg_id}),
)
```

We can also wait until https://github.com/flwrlabs/flower/pull/6877 is merged to see if we can reuse/refactor some of the unit testing code.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -1933,6 +1937,199 @@ def state_factory(self) -> SqlLinkState:
...
+
+            node_id = create_dummy_node(state_0)
+            assert state_0.store_message_ins(
+                message_from_proto(
+                    create_ins_message(
+                        src_node_id=SUPERLINK_NODE_ID,
+                        dst_node_id=node_id,
+                        run_id=create_dummy_run(state_0),
+                    )
+                )
```
panh99: Let's create a better function for this... Though it's consistent with the rest of the code, but I feel It's a bit wild now...

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -1933,6 +1937,199 @@ def state_factory(self) -> SqlLinkState:
...
+        state_1.initialize()
+        return state_0, state_1
+
+    @staticmethod
+    def _run_in_parallel(
+        state_0: SqlLinkState,
+        state_1: SqlLinkState,
+        fn: Callable[[SqlLinkState], list[Message]],
+        timeout: float = _CONCURRENT_TEST_TIMEOUT,
+    ) -> list[list[Message]]:
```
panh99: I feel like we have similar `_create_shared_sql_states` and `_run_in_parallel` in another PR?

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
panh99: `get_message_res` shouldn't return `None`. Let's skip this assertion.
```suggestion
```

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -1933,6 +1937,199 @@ def state_factory(self) -> SqlLinkState:
...
+            for thread in threads:
+                thread.join(timeout=self._CONCURRENT_TEST_TIMEOUT)
+            alive_threads = [thread for thread in threads if thread.is_alive()]
+            if alive_threads:
+                self.fail("Timed out waiting for concurrent pull_res threads to finish")
+
+            if exceptions:
+                raise exceptions[0]
+            assert all(message_list is not None for message_list in results)
+            assert self._count_successful_replies(results) == 1
```
panh99: Should we simplify this? For this test, `LinkState.get_message_res` shouldn’t return error messages. We can assume everything else is working correctly and focus only on the behavior the unit test is targeting.

```suggestion
            assert sum(len(res) for res in results) == 1
```

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
msheller: @panh99 , That's my understanding. I believe SQL always uses UTC+0 in that pattern.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
msheller: Resolved in [20e76a2](https://github.com/flwrlabs/flower/pull/6785/commits/20e76a29e4f3f4bf8dfa7e776647d4892924eb79)

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
msheller: Resolved in [20e76a2](https://github.com/flwrlabs/flower/pull/6785/commits/20e76a29e4f3f4bf8dfa7e776647d4892924eb79)

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
msheller: Resolved in [20e76a2](https://github.com/flwrlabs/flower/pull/6785/commits/20e76a29e4f3f4bf8dfa7e776647d4892924eb79)

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
msheller: Resolved in [20e76a2](https://github.com/flwrlabs/flower/pull/6785/commits/20e76a29e4f3f4bf8dfa7e776647d4892924eb79)

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
msheller: Resolved in [20e76a2](https://github.com/flwrlabs/flower/pull/6785/commits/20e76a29e4f3f4bf8dfa7e776647d4892924eb79)

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -250,48 +250,18 @@ def get_message_ins(self, node_id: int, limit: int | None) -> list[Message]:
...
-                AND delivered_at = ''
-                AND (created_at + ttl) > CAST(strftime('%s', 'now') AS REAL)
-            """
-
-            if limit is not None:
-                query += " LIMIT :limit"
-                params["limit"] = limit
-
-            rows = self.query(query, params)
+            rows = self._claim_message_ins_rows(sint64_node_id, limit)
```
panh99: Let's move `sint64_node_id = uint64_to_int64(node_id)` inside `self._claim_message_ins_rows`.

msheller: Done! I just moved the conversion inside the params assignment since it's not used elsewhere.

panh99 (approved):

danieljanes (approved):

PR #6996, opened 2026-04-10
fix(framework): Fix Windows remote app parsing in `flwr run`

mohammadnaseri:
This prevents `flwr run @account/app` from being treated as a Windows path before validation.


jafermarq (approved):

panh99 (approved):
LGTM!

PR #7020, opened 2026-04-21
fix(framework): Validate run_id

danieljanes:


panh99 (dismissed):
LGTM!

framework/py/flwr/supercore/interceptors/appio_token_interceptor.py
```diff
@@ -57,6 +63,14 @@ def _terminate(_request: GrpcMessage, context: grpc.ServicerContext) -> GrpcMess
     return grpc.unary_unary_rpc_method_handler(_terminate)
 
 
+def _get_request_run_id(request: GrpcMessage) -> int | None:
+    descriptor = getattr(request, "DESCRIPTOR", None)
+    if descriptor is None or "run_id" not in descriptor.fields_by_name:
+        return None
+    request_any = cast(Any, request)
+    return cast(int, request_any.run_id)
```
panh99: Wdyt? If `run_id` is a field of the request,  `getattr(request, "run_id")` will return its value, and return its default value `0` if the field is not set. This change make no functional changes.

```suggestion
    return cast(int, getattr(request, "run_id", None))
```

PR #7021, opened 2026-04-21
refactor(framework): Make FAB required in PullAppInputs

danieljanes:
### Motivation

- Ensure the client application run always receives and installs a FAB rather than treating it as optional, simplifying downstream code paths.
- Make the RPC contract and local API consistent by reflecting that `PullAppInputs` always returns a `Fab` in the client runtime.

### Description

- Unconditionally install the FAB in `run_clientapp` by removing the `if fab:` guard and always calling `install_from_fab` and related setup code. 
- Update `pull_appinputs` signature to return `Fab` (non-optional) and always call `fab_from_proto(res.fab)` instead of conditionally handling a missing FAB. 
- Adjust client app loading to pass `fab.hash_str` without an optional check. 
- Update unit test `TestClientAppIoServicer.test_pull_appinputs` to assert the returned `fab` fields directly instead of guarding the assertions behind an `if`.

### Testing

- Ran the unit tests in `framework/py/flwr/supernode/servicer/clientappio/clientappio_servicer_test.py` which include `test_pull_appinputs`, and they passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_69e6f824da88832b9e5a73e2d8aad7f1)

### DIFF
--- a/framework/py/flwr/supernode/runtime/run_clientapp.py
+++ b/framework/py/flwr/supernode/runtime/run_clientapp.py
@@ -118,38 +118,37 @@ def run_clientapp(  # pylint: disable=R0913, R0914, R0917
         heartbeat_sender = HeartbeatSender(make_app_heartbeat_fn_grpc(stub, token))
         heartbeat_sender.start()
 
-        # Pull Message, Context, Run and (optional) FAB from SuperNode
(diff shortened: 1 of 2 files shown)


panh99 (approved):
LGTM!

PR #7031, opened 2026-04-23
refactor(framework): Add `task` table to `CoreState`

jafermarq:


danielnugraha (dismissed):

framework/py/flwr/supercore/typing.py
panh99, line 1: Should we directly use `Task` proto instead? No need to introduce a corresponding `Task` dataclass, imo.

jafermarq: I see your point. i added it to follow the `Run` dataclass style (which includes fields like `pending_at` etc -- which aren't part of the `Task` proto)

panh99 (dismissed):
LGTM!

panh99 (approved):

danieljanes (approved):

PR #7038, opened 2026-04-24
feat(framework): Add runtime version check for simulation ServerAppIo

mohammadnaseri:
This PR adds runtime version metadata helpers and enables the first runtime version check for simulation -> ServerAppIo.


framework/py/flwr/common/constant.py
panh99, line 1: Let's move the changes from `flwr.common` to `flwr.supercore`. The `flwr.common` is a legacy module and it's better not to grow it further.

framework/py/flwr/common/runtime_version.py
```diff
@@ -0,0 +1,286 @@
...
+)
+from flwr.supercore.version import package_name, package_version
+
+RuntimeCompatibilityStatus = Literal[
+    "missing",
+    "disabled",
+    "invalid",
+    "compatible",
+    "incompatible",
+]
```
panh99: Should we create a `RuntimeCompatibilityStatus(str, Enum)` in `flwr.supercore.constant`?

framework/py/flwr/common/runtime_version.py
```diff
@@ -0,0 +1,286 @@
...
+    component_name: str
+
+
+@dataclass(frozen=True, order=True)
+class ParsedFlowerVersion:
+    """A parsed `major.minor.patch` Flower version."""
+
+    major: int
+    minor: int
+    patch: int
```
panh99: Since `packaging` is a required dep of `flwr`, should we directly use `Version` from `packaging.version`, or implement`ParsedFlowerVersion` as a subclass of `Version`? That way, we don't need to build and maintain the util ourselves. With `Version(package_version)`, it already provides `.major`, `.minor`, and `.micro` (patch version), and it also supports comparison.

framework/py/flwr/common/runtime_version.py
panh99, line 1: Nit: the module name `runtime_version.py` is a little broad. Most of this module is about runtime version metadata and compatibility checks, so something like `runtime_version_compatibility.py` or `runtime_version_metadata.py` might describe the responsibility more clearly.

framework/py/flwr/common/runtime_version.py
```diff
@@ -0,0 +1,286 @@
...
+)
+
+
+@dataclass(frozen=True)
+class RuntimeVersionMetadata:
+    """Flower runtime version metadata attached to a caller."""
+
+    package_name: str
+    package_version: str
+    component_name: str
```
panh99: The current shape spreads one concept across `build_runtime_version_metadata`, `runtime_version_metadata_to_dict`, `read_runtime_version_metadata`, `evaluate_runtime_version_compatibility`, and the interceptor glue.

Since this is used with gRPC metadata, can we make the API consistently operate on gRPC metadata sequences instead of converting through dicts? For example, `.from_grpc_metadata(...)` could parse incoming metadata, and `.append_to_grpc_metadata(existing_grpc_metadata)` could return a new metadata tuple with the Flower runtime metadata added/replaced. That would avoid mixing dict and sequence representations, preserve the fact that gRPC metadata is a `Sequence`, and make the call sites less fragmented. The proposal is just to move the code around a bit. Lmkwyt!

Proposal:
```python
@dataclass(frozen=True)
class RuntimeVersionMetadata:
    component_name: str
    package_name: str = flwr_package_name
    package_version: str = flwr_package_version

    @classmethod
    def from_grpc_metadata(
        cls,
        grpc_metadata: Sequence[tuple[str, str | bytes]] | None,
    ) -> tuple[RuntimeVersionMetadata | None, str | None]:
        ...

    def append_to_grpc_metadata(
        self,
        grpc_metadata: Sequence[tuple[str, str | bytes]] | None,
    ) -> tuple[tuple[str, str | bytes], ...]:
        ...

    def is_compatible_with(
        self,
        other: RuntimeVersionMetadata,
    ) -> CompatibilityResult:
        ...

```

Usage:
```python
class RuntimeVersionClientInterceptor(grpc.UnaryUnaryClientInterceptor):

    def __init__(self, component_name: str) -> None:
        self._metadata = RuntimeVersionMetadata(component_name)

    def intercept_unary_unary(
        self,
        continuation: Callable[[Any, Any], Any],
        client_call_details: grpc.ClientCallDetails,
        request: GrpcMessage,
    ) -> grpc.Call:
        """Add or replace the runtime version metadata headers."""
        metadata = client_call_details.metadata
        details = client_call_details._replace(
            metadata=self._metadata.append_to_grpc_metadata(metadata)
        )
        return continuation(details, request)


class RuntimeVersionServerInterceptor(grpc.ServerInterceptor):

    def intercept_service(
        self,
        continuation: Callable[[Any], Any],
        handler_call_details: grpc.HandlerCallDetails,
    ) -> grpc.RpcMethodHandler:
        """Reject explicit invalid or incompatible peer metadata before handling."""
        ...

        peer_metadata, err_msg = RuntimeVersionMetadata.from_grpc_metadata(
            handler_call_details.invocation_metadata
        )
        if err_msg:
            ...
        compatibility = self._local_metadata.is_compatible_with(peer_metadata)
```

framework/py/flwr/common/runtime_version.py
```diff
@@ -0,0 +1,286 @@
...
+    missing_keys = [key for key in relevant_keys if key not in metadata_map]
+    if missing_keys:
+        missing_keys_str = ", ".join(sorted(missing_keys))
+        return None, f"Missing required Flower runtime metadata: {missing_keys_str}."
+
+    values = {key: metadata_map[key].strip() for key in relevant_keys}
+    empty_keys = [key for key, value in values.items() if value == ""]
+    if empty_keys:
+        empty_keys_str = ", ".join(sorted(empty_keys))
+        return None, f"Flower runtime metadata contains empty values: {empty_keys_str}."
```
panh99: Should we use `get_metadata_str` from `flwr.supercore.utils`? It's designed to extract a string value from the grpc metadata and reject duplicate keys or empty values.

panh99 (commented):
Nice feature! I left a few comments mostly around simplifying the structure and keeping the metadata representation consistent with gRPC metadata. Functionally, the approach looks good to me.

framework/py/flwr/common/runtime_version.py
mohammadnaseri: Hmm, I’d keep this local for now. These statuses are internal to the runtime version compatibility helper/interceptor flow and don’t look like general `supercore.constant` material yet. wdyt?

framework/py/flwr/common/runtime_version.py
mohammadnaseri: I don’t think `get_metadata_str` is enough for this code path. It collapses missing, duplicate, empty, and wrong type cases into `None`, but the runtime version rollout logic needs to distinguish missing from invalid metadata. This parser also accepts `bytes` metadata values. So I’d keep the dedicated parsing here. wdyt?

framework/py/flwr/common/runtime_version.py
mohammadnaseri: I agree with the direction. The current API is more fragmented than it needs to be, and this is really one runtime metadata concept. I moved parsing/serialization onto `RuntimeVersionMetadata` and updated the interceptors to work directly with grpc metadata sequences, while keeping the existing compatibility result behavior unchanged.

framework/py/flwr/common/runtime_version.py
mohammadnaseri: Good point, made the change.

framework/py/flwr/common/runtime_version.py
mohammadnaseri: I pushed the change.

framework/py/flwr/common/constant.py
mohammadnaseri: I pushed the change.

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,376 @@
...
+        package_name_value=package_name_value,
+        package_version_value=package_version_value,
+    )
+
+
+def runtime_version_metadata_to_dict(
+    metadata: RuntimeVersionMetadata,
+) -> dict[str, str]:
+    """Serialize runtime version metadata to a string dictionary."""
+    return metadata.to_dict()
```
panh99: This function is no longer in use.

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,376 @@
...
+    local_version: ParsedFlowerVersion | None
+    peer_version: ParsedFlowerVersion | None
+
+
+def build_runtime_version_metadata(
+    component_name: str,
+    *,
+    package_name_value: str = flwr_package_name,
+    package_version_value: str = flwr_package_version,
+) -> RuntimeVersionMetadata:
```
panh99: It seems that this function is no longer in use.

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,355 @@
...
+        )
+
+
+@dataclass(frozen=True, order=True)
+class ParsedFlowerVersion:
+    """A parsed `major.minor.patch` Flower version."""
+
+    major: int
+    minor: int
+    patch: int
```
panh99: Oh, I mean, we don't need this dataclass entirely. We can remove it including the `parse_flower_version` and `_normalize_flower_version`, just simply use `Version` directly. And `Version` supports version string like `1.30.0.dev20260425`.

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,363 @@
...
+            package_name=package_name_value,
+            package_version=package_version_value,
+            component_name=component_name,
+        )
+
+    @classmethod
+    def from_grpc_metadata(
+        cls,
+        grpc_metadata: Sequence[tuple[str, str | bytes]] | None,
+    ) -> tuple[RuntimeVersionMetadata | None, str | None]:
```
panh99, line 87: This function feels heavier than necessary for this feature. These runtime metadata keys are regular gRPC metadata keys, not `-bin` keys, so we should expect string values and should not decode bytes here. Accepting bytes seems to broaden the contract in a way we don’t need and could hide invalid metadata. Could we still rely on the existing `get_metadata_str` helper instead? Or generating different error messages for missing/invalid/empty keys is vital in this design?

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,363 @@
...
+        self,
+        grpc_metadata: Sequence[tuple[str, str | bytes]] | None,
+    ) -> tuple[tuple[str, str | bytes], ...]:
+        """Return gRPC metadata with runtime version values added or replaced."""
+        metadata = tuple(grpc_metadata or ())
+        runtime_keys = self.to_dict()
+        filtered_metadata = tuple(
+            (key, value) for key, value in metadata if key not in runtime_keys
+        )
+        return filtered_metadata + tuple(runtime_keys.items())
```
panh99:
```suggestion
        """Return gRPC metadata with runtime version values added or replaced."""
        return tuple(grpc_metadata or ()) + (
            (FLWR_PACKAGE_NAME_METADATA_KEY, self.package_name),
            (FLWR_PACKAGE_VERSION_METADATA_KEY, self.package_version),
            (FLWR_COMPONENT_NAME_METADATA_KEY, self.component_name),
        )
```

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,363 @@
...
+            None,
+        )
+
+    def to_dict(self) -> dict[str, str]:
+        """Serialize runtime version metadata to a string dictionary."""
+        return {
+            FLWR_PACKAGE_NAME_METADATA_KEY: self.package_name,
+            FLWR_PACKAGE_VERSION_METADATA_KEY: self.package_version,
+            FLWR_COMPONENT_NAME_METADATA_KEY: self.component_name,
+        }
```
panh99: Let's remove this method. If possible, we'd avoid mixing `dict` and `Sequence`.

framework/py/flwr/supercore/runtime_version_compatibility.py
panh99: Or we can add `default` argument in `get_metadata_str`, and raise an error if the key is missing and `default` is not set.

framework/py/flwr/common/runtime_version.py
panh99: I see wym. That's a valid point. Though I still think it's a bit heavy for a single feature. Should we then update `get_metadata_str` to make it more flexible? Related to this: https://github.com/flwrlabs/flower/pull/7038#discussion_r3148140820

framework/py/flwr/supercore/runtime_version_compatibility.py
mohammadnaseri: You are right, I tightened this to require `str` values for the runtime-version keys and reject non string values as invalid metadata. I didn’t switch to `get_metadata_str` because this path still needs to distinguish missing metadata from invalid metadata.

framework/py/flwr/supercore/runtime_version_compatibility.py
mohammadnaseri:
```
filtered_metadata = tuple(
    (key, value) for key, value in metadata if key not in runtime_keys
)
```


this is necessary. no?

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,317 @@
...
+    ),
+) -> CompatibilityResult:
+    """Evaluate whether a peer is runtime-compatible with the local component."""
+    peer: RuntimeVersionMetadata | None
+    metadata_error: str | None
+    if isinstance(peer_metadata, RuntimeVersionMetadata):
+        peer = peer_metadata
+        metadata_error = None
+    else:
+        peer, metadata_error = read_runtime_version_metadata(peer_metadata)
```
panh99:
```suggestion
        peer, metadata_error = RuntimeVersionMetadata.from_grpc_metadata(peer_metadata)
```

framework/py/flwr/supercore/runtime_version_compatibility.py
```diff
@@ -0,0 +1,228 @@
...
+        runtime_metadata = (
+            (FLWR_PACKAGE_NAME_METADATA_KEY, self.package_name),
+            (FLWR_PACKAGE_VERSION_METADATA_KEY, self.package_version),
+            (FLWR_COMPONENT_NAME_METADATA_KEY, self.component_name),
+        )
+        runtime_keys = {key for key, _ in runtime_metadata}
+        filtered_metadata = tuple(
+            (key, value) for key, value in metadata if key not in runtime_keys
+        )
+        return filtered_metadata + runtime_metadata
```
panh99: Replacing existing keys may suppress system errors, as the version-related metadata key-value pairs should only be appended here. Let's simply append the key-value pairs without replacing any, and then the server interceptor can detect the error and warn us if anything goes wrong.

```suggestion
        """Return gRPC metadata with runtime version values added."""
        return tuple(grpc_metadata or ()) + (
            (FLWR_PACKAGE_NAME_METADATA_KEY, self.package_name),
            (FLWR_PACKAGE_VERSION_METADATA_KEY, self.package_version),
            (FLWR_COMPONENT_NAME_METADATA_KEY, self.component_name),
        )
```

mohammadnaseri: what if they are already present in the outgoing metadata?

panh99: I see your point, and that’s exactly the situation I’d like to avoid suppressing. The concern with the current approach is that it can mask underlying bugs.

Right now, the flow would be:
1. Keys are already present in the metadata (which is actually a bug)
2. The client interceptor replaces the keys
3. The server interceptor sees valid keys and proceeds without issue

In this case, even though step 1 indicates something is wrong, it goes unnoticed because it gets overwritten downstream.

In contrast, if we only *add* metadata keys without replacing existing ones, the issue becomes visible earlier:
1. Keys are already present in the metadata (still a bug)
2. The client interceptor adds the keys without overwriting
3. The server interceptor detects duplicate keys and raises an error

This way, we surface the bug instead of silently hiding it.

mohammadnaseri: I agree that silently masking preexisting runtime version keys is not ideal. but I don’t think intentionally appending duplicates is the right fix, especially since the server interceptor is now observe only and would not stop the call anyway.

panh99: I agree we shouldn’t intentionally introduce duplicates, especially if the server interceptor won’t block the call.

My concern is more about replacing existing keys, which can silently hide upstream bugs. If a key is already present, something is likely wrong, and overwriting it makes it harder to detect. We had similar discussions when implementing other interceptors.

We could instead check and raise an error on the client side. It might be a bit redundant with server-side validation, but it surfaces the issue earlier and closer to the source.

mohammadnaseri: I would actually  keep the interceptor authoritative for these keys and, if we want to surface upstream misuse, add an explicit client side check or log when runtime-version keys are already present before replacement. That catches the bug at the source without sending malformed metadata. wdyt?

panh99: I’m still a slightly hesitant to add validation directly into this function.

That said, if we do want this kind of check, maybe we can keep it separate and make it explicit, e.g. a small utility like `has_any_metadata_key(grpc_metadata, keys: Iterable[str]) -> bool`. Then callers can decide whether to validate, log, or ignore, depending on their needs. Wdyt?

Like

```python
if has_any_metadata_key(metadata, _RUNTIME_METADATA_KEYS):
    raise RuntimeError
```

mohammadnaseri: Yes, I think that separation makes sense. I pushed a change.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -0,0 +1,147 @@
...
+)
+
+
+class RuntimeVersionClientInterceptor(grpc.UnaryUnaryClientInterceptor):  # type: ignore
+    """Attach Flower runtime version metadata to outbound unary RPCs."""
+
+    def __init__(self, component_name: str) -> None:
+        self._metadata = RuntimeVersionMetadata.from_local_component(
+            component_name, package_version_value="1.32.0"
+        )
```
panh99:
```suggestion
        self._metadata = RuntimeVersionMetadata.from_local_component(component_name)
```

panh99 (dismissed):
LGTM!

framework/py/flwr/simulation/simulationio_connection.py
```diff
@@ -82,7 +85,10 @@ def _connect(self) -> None:
             server_address=self._addr,
             insecure=self._insecure,
             root_certificates=self._cert,
-            interceptors=[AppIoTokenClientInterceptor(token=self._token)],
+            interceptors=[
+                RuntimeVersionClientInterceptor(component_name="simulation"),
```
panh99:
```suggestion
                RuntimeVersionClientInterceptor(component_name="flwr-simulation"),
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7041, opened 2026-04-24
ci(framework): Check Alembic migrations in CI

chongshenng:
## Issue
Schema changes can currently land without a matching Alembic migration, leaving the committed migrations out of sync with the SQLAlchemy table definitions.

## Proposal
Add `framework/dev/check-migrations.sh`, which upgrades an isolated temporary SQLite database to all Alembic heads and runs alembic check against the current models.

Run this check in `framework-test.yml` for framework changes on the Python 3.10 matrix entry, so CI fails when Alembic detects pending migration operations.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #6986, opened 2026-04-09
feat(framework): Add `--root-certificates` argument for `flwr-*` commands and `flower-superexec`

panh99:

PR #7046, opened 2026-04-26
feat(framework): Add TLS for appio servicers

msheller:
Completes the TLS implementation ini #6986 by adding servicer-side changes. 

### Checklist

- [x] Write tests
- [x] Make CI checks pass


framework/py/flwr/server/app.py
```diff
@@ -446,7 +447,8 @@ def run_superlink() -> None:
         # bound_address contains the actual address when the port is set to :0
         # which means let the OS choose a free port.
         appio_address = resolve_bind_address(serverappio_server.bound_address)
-        command = ["flower-superexec", "--insecure"]
+        command = ["flower-superexec"]
+        command += get_superexec_appio_tls_args(certificates, args.ssl_ca_certfile)
```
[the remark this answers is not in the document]
msheller: No longer relevant as certs are split.

framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -223,7 +232,10 @@ def start_client_internal(
 
     # Launch the SuperExec if the isolation mode is `subprocess`
     if isolation == ISOLATION_MODE_SUBPROCESS:
-        command = ["flower-superexec", "--insecure"]
+        command = ["flower-superexec"]
+        command += get_superexec_appio_tls_args(
+            clientappio_certificates, clientappio_root_certificates_path
+        )
```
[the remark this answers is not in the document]
msheller: Adding this to help text.

framework/py/flwr/common/args.py
```diff
@@ -152,3 +152,40 @@ def try_obtain_server_certificates(
...
+        and args.appio_ssl_ca_certfile
+    ):
+        if not isfile(args.appio_ssl_ca_certfile):
+            sys.exit(
+                "Path argument `--appio-ssl-ca-certfile` does not point to a file."
+            )
+        if not isfile(args.appio_ssl_certfile):
+            sys.exit("Path argument `--appio-ssl-certfile` does not point to a file.")
+        if not isfile(args.appio_ssl_keyfile):
+            sys.exit("Path argument `--appio-ssl-keyfile` does not point to a file.")
```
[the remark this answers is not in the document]
msheller: Fixed

framework/py/flwr/common/args.py
```diff
@@ -152,3 +152,40 @@ def try_obtain_server_certificates(
...
+    and return `None`. This helper also intentionally does not inspect `args.insecure`
+    because callers such as `flower-supernode` use `--insecure` for an outbound
+    client connection, not for the local AppIO server this certificate tuple secures.
+    """
+    if (
+        args.appio_ssl_certfile
+        and args.appio_ssl_keyfile
+        and args.appio_ssl_ca_certfile
+    ):
+        if not isfile(args.appio_ssl_ca_certfile):
```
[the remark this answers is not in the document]
msheller: fixed

framework/py/flwr/common/args.py
```diff
@@ -152,3 +152,43 @@ def try_obtain_server_certificates(
         "in insecure mode using '--insecure' if you understand the risks.",
     )
     sys.exit(1)
+
+
+def try_obtain_optional_appio_server_certificates(
```
panh99: Would it be possible to move the utility function to `supercore` module? The `flwr.common` is the legacy module and it will be better if we can avoid adding new code in it.

framework/py/flwr/common/args.py
```diff
@@ -152,3 +152,43 @@ def try_obtain_server_certificates(
...
+            sys.exit("Path argument `--appio-ssl-keyfile` does not point to a file.")
+        return (
+            appio_ssl_ca_certfile.read_bytes(),
+            appio_ssl_certfile.read_bytes(),
+            appio_ssl_keyfile.read_bytes(),
+        )
+    if args.appio_ssl_certfile or args.appio_ssl_keyfile or args.appio_ssl_ca_certfile:
+        sys.exit(
+            "You need to provide valid file paths to `--appio-ssl-certfile`, "
+            "`--appio-ssl-keyfile`, and `--appio-ssl-ca-certfile` to create a "
```
panh99: Should we create one exit code, and use `flwr_exit` instead of `sys.exit`?

framework/py/flwr/server/app.py
```diff
@@ -487,6 +490,37 @@ def _format_address(address: str) -> tuple[str, str, int]:
     return (f"[{host}]:{port}" if is_v6 else f"{host}:{port}", host, port)
 
 
+def _obtain_superlink_certificates(
+    args: argparse.Namespace,
+) -> tuple[tuple[bytes, bytes, bytes] | None, tuple[bytes, bytes, bytes] | None]:
+    """Return Fleet/Control and ServerAppIo certificate tuples."""
+    certificates = try_obtain_server_certificates(args)
```
panh99, line 507: Should we move the following code in this function, and then the two `try_obtain_*` functions can just load the certificates?

```python
    if args.insecure:
        log(
            WARN,
            "Option `--insecure` was set. Starting insecure HTTP server with "
            "unencrypted communication (TLS disabled). Proceed only if you understand "
            "the risks.",
        )
        return None
```

framework/py/flwr/server/app.py
```diff
@@ -711,20 +745,20 @@ def _add_args_common(parser: argparse.ArgumentParser) -> None:
...
         "--ssl-keyfile",
-        help="Fleet API server SSL private key file (as a path str) "
+        help="SuperLink API server TLS private key file (as a path str) "
         "to create a secure connection.",
         type=str,
     )
     parser.add_argument(
         "--ssl-ca-certfile",
-        help="Fleet API server SSL CA certificate file (as a path str) "
+        help="SuperLink API server CA certificate file (as a path str) "
```
panh99: We don't use the term "SuperLink API". Prolly we can say something like "Server TLS xxx file for Fleet API and Control API"?

framework/py/flwr/server/app.py
```diff
@@ -791,6 +825,28 @@ def _add_args_serverappio_api(parser: argparse.ArgumentParser) -> None:
         "`--simulationio-api-address` is accepted as a deprecated alias. "
         f"By default, it is set to {SERVERAPPIO_API_DEFAULT_SERVER_ADDRESS}.",
     )
+    parser.add_argument(
+        "--appio-ssl-certfile",
+        help="ServerAppIo API server TLS certificate file (as a path str) "
+        "to create a secure connection. The certificate must be valid for the "
+        "AppIO API address used by SuperExec, typically including loopback SANs "
+        "such as 127.0.0.1 and ::1 for auto-launched local SuperExec.",
```
panh99: Will this sound like a recommendation? Should we remove it since this is the help message for an argument?

framework/py/flwr/server/app.py
```diff
@@ -791,6 +825,28 @@ def _add_args_serverappio_api(parser: argparse.ArgumentParser) -> None:
...
+        "to create a secure connection.",
+        type=str,
+    )
+    parser.add_argument(
+        "--appio-ssl-ca-certfile",
+        help="Path to the PEM-encoded CA certificate file used by SuperExec to verify "
+        "the ServerAppIo API server certificate. This is not a client certificate "
+        "for mTLS.",
+        type=str,
+    )
```
panh99, line 857: Is this required to allow SuperExec and task executors to connect to, when it's in subprocess mode? I recall you mentioned that this is not needed if we just need to start the server with TLS.

framework/py/flwr/supercore/tls.py
```diff
@@ -19,6 +19,31 @@
...
+    The `certificates` tuple is the AppIO server's TLS key material. Its presence
+    tells the auto-launched SuperExec that the AppIO server is listening with TLS
+    enabled; its absence means the AppIO server is plaintext and SuperExec must be
+    launched with `--insecure`.
+
+    When AppIO TLS is enabled and `root_certificates_path` is provided, pass that
+    path to SuperExec as `--root-certificates` so the SuperExec client channel can
+    verify the AppIO server certificate. When the path is omitted, SuperExec uses
+    TLS without explicit trust roots, which lets gRPC use its default trust store.
+    """
```
panh99: Should we simplify the docstring? It feels a bit long relative to the function body, which is already quite self-explanatory.

framework/py/flwr/supercore/tls.py
```diff
@@ -19,6 +19,31 @@
 
 from flwr.common.exit import ExitCode, flwr_exit
 
+ServerCertificates = tuple[bytes, bytes, bytes]
+
+
+def get_superexec_appio_tls_args(
+    certificates: ServerCertificates | None,
+    root_certificates_path: str | None,
+) -> list[str]:
```
panh99: Could we make the name more generic? I think this is a good utility that we could also reuse for starting `flwr-*` commands.

Also, should we switch to using an `insecure: bool` flag instead of passing the full certificate just to check `if certs is None`? That logic could live on the caller side, for example:
`get_client_tls_args(certificates is None, root_certificates_path)`.

```suggestion
def get_client_tls_args(
    insecure: bool,
    root_certificates_path: str | None,
) -> list[str]:
```

framework/py/flwr/server/app.py
msheller: This is a bit confusing, but it is needed specifically because it is going to pass it to the SuperExec so that it can verify the appio servicer certificate. So the key/cert pair is for the SuperLink/Node to use itself when starting the appio servicer in tls mode, and the ca cert is what signed its server cert, so it needs to pass it to the SuperExec so that it can verify the appio server's certificate.

framework/py/flwr/supercore/tls.py
msheller: Accepted and updated.

framework/py/flwr/supercore/tls.py
msheller: Accepted and updated.

framework/py/flwr/server/app.py
msheller: Accepted and updated.

framework/py/flwr/common/args.py
msheller: Accepted and added.

framework/py/flwr/server/app.py
msheller: accepted and added.

panh99 (dismissed):
LGTM!

framework/py/flwr/common/args.py
msheller: accepted and applied!

framework/docs/source/ref-exit-codes/250.rst
panh99, line 1: Is this added accidentally? We don't seem to have a `250` exit code in `exit_code.py`.

msheller: Ah, yes! This is for the next PR. Good catch. Will remove.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7051, opened 2026-04-27
refactor(framework): Make `FlowerError` return `JSON`-like message and introduce `EntitlementError`

jafermarq:


framework/py/flwr/cli/utils.py
```diff
@@ -84,6 +84,21 @@ def print_json_to_stdout(data: str | Any) -> None:
...
 
+def _extract_error_message(err: Exception) -> str:
+    """Return the message field from JSON-encoded errors when available."""
+    err_message = str(err)
+    try:
+        parsed = json.loads(err_message)
+        if isinstance(parsed, dict) and "public_message" in parsed:
+            msg = str(parsed["public_message"])
+            if details := parsed.get("public_details"):
+                msg += f" Details: {details}"
```
panh99: I tried to compare ` - `, ` Details: `, `: `, ` `, and some others, printing the details in a new line might be better given that ClickException supports multi-lines and that the details is normally a full sentence as well. Wdyt?

```suggestion
                msg += f"\n{details}"
```

framework/py/flwr/cli/utils.py
```diff
@@ -108,7 +123,7 @@ def cli_output_handler(
     except Exception as err:  # pylint: disable=broad-except
         if is_json:
             restore_output()
-            print_json_error(captured_output.getvalue(), err)
+            print_json_error(captured_output.getvalue(), Exception(str(err)))
```
panh99: Why is this change?

framework/py/flwr/cli/utils.py
```diff
@@ -407,62 +422,61 @@ def flwr_cli_grpc_exc_handler(  # pylint: disable=too-many-branches
     except grpc.RpcError as e:
         if custom_handler is not None:
             custom_handler(e)
+        # pylint: disable-next=E1101
+        details = _extract_error_message(Exception(e.details()))
```
```suggestion
        details = _format_grpc_error(e)
```

framework/py/flwr/cli/utils.py
jafermarq: ah, that's wrong. i was reverting my previous changes to `cli_output_handler` and added that extra _wrapping_ around `err` by mistake

framework/py/flwr/cli/utils.py
```diff
@@ -84,6 +84,21 @@ def print_json_to_stdout(data: str | Any) -> None:
         Console(file=sys.__stdout__).print_json(data=data)
 
 
+def _extract_error_message(err: Exception) -> str:
+    """Return the message field from JSON-encoded errors when available."""
+    err_message = str(err)
```
```suggestion
def _extract_error_message(err: grpc.RpcError) -> str:
    """Return a user-facing message from a gRPC error.

    This function parses FlowerError JSON in `err.details()` when present,
    otherwise falls back to the raw gRPC details string.
    """
    err_message = err.details()  # pylint: disable=E1101
```

panh99 (commented):
After reading the code, I think it might be cleaner to update the formatting function like this so the function is focused. Wdyt?

framework/py/flwr/supercore/error/base.py
```diff
@@ -14,17 +14,51 @@
...
 class FlowerError(Exception):
     """Base exception that carries an internal error code and debug message."""
 
-    def __init__(self, code: int, message: str) -> None:
+    def __init__(
+        self,
+        code: int,
+        message: str,
+        public_details: str | None = None,
+    ) -> None:
```
panh99: Since the error class is growing, should we add docstring for the arguments? Like saying code is the API error code to which public message and grpc status code is mapped; message is only for internal logging, can be sensitive; and public_details is additional details shown to the caller side (cannot contain sensitive info).

framework/py/flwr/supercore/error/base.py
```diff
@@ -14,17 +14,98 @@
...
-    def __init__(self, code: int, message: str) -> None:
+        Parameters
+        ----------
+        code : int
+            Internal numeric error code used to look up the API error contract.
+        message : str
+            Sensitive diagnostic message intended for server-side logs.
+        public_details : str | None
+            Optional client-safe details to include in the serialized error payload.
+        """
```
panh99:
```suggestion
    """Base exception for API errors exposed through client-safe responses.

    Parameters
    ----------
    code : int
        Internal numeric error code used to look up the API error contract.
    message : str
        Sensitive diagnostic message intended for server-side logs.
    public_details : str | None
        Optional client-safe details to include in the serialized error payload.
    """

    def __init__(
        self,
        code: int,
        message: str,
        public_details: str | None = None,
    ) -> None:
```

framework/py/flwr/supercore/error/base.py
```diff
@@ -14,17 +14,98 @@
...
+            {
+                "code": self.code,
+                "public_message": public_message,
+                "public_details": self.public_details,
+            }
+        )
+
+
+class EntitlementError(FlowerError):
+    """API error for actions blocked by entitlement checks."""
```
panh99: Should we move the error somewhere else? The `base.py` may not be the best fit.

framework/py/flwr/supercore/error/base.py
```diff
@@ -14,17 +14,98 @@
 # ==============================================================================
 """Base error types for API-facing error translation."""
 
-
+import json
```
panh99, line 18:
```suggestion

import json
```

framework/py/flwr/supercore/error/exceptions.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2025 Flower Labs GmbH. All Rights Reserved.
```
panh99:
```suggestion
# Copyright 2026 Flower Labs GmbH. All Rights Reserved.
```

panh99 (approved):
LGTM!

PR #7056, opened 2026-04-28
refactor(framework): Use `EntitlementError`

jafermarq:
To be merged after https://github.com/flwrlabs/flower/pull/7051


framework/py/flwr/cli/utils_test.py
panh99, line 1: Good catch!

panh99 (approved):
LGTM!

PR #7058, opened 2026-04-28
refactor(framework): Implement `claim_task`, `activate_task`, `finish_task`, `acknowledge_task_heartbeat` and `get_task_id_by_token` in `CoreState`

jafermarq:


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -114,6 +114,91 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+
+        Returns
+        -------
+        bool
+            True if the task existed and transitioned from starting to running,
+            otherwise False.
+        """
+
+    @abstractmethod
+    def finish_task(self, task_id: int, substatus: str, detail: str) -> bool:
```
panh99: Should we use `sub_status` for consistency?
```suggestion
    def finish_task(self, task_id: int, sub_status: str, detail: str) -> bool:
```

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -119,6 +124,131 @@ def test_get_task_returns_copy(self) -> None:
...
+
+        self.assertIsNotNone(token)
+        assert token is not None
+        self.assertEqual(state.get_task_id_by_token(token), task_id)
+        tasks = state.get_tasks(task_ids=[task_id])
+        self.assertEqual(len(tasks), 1)
+        self.assertEqual(tasks[0].status.status, Status.STARTING)
+        self.assertTrue(tasks[0].starting_at)
+        self.assertEqual(tasks[0].running_at, "")
+        self.assertEqual(tasks[0].finished_at, "")
```
panh99, line 146: Should we check directly if the timestamp fields are not set? Cuz we don't expect `.running_at` to be set to `""`; it should not be set.
```suggestion
        self.assertFalse(tasks[0].HasField("running_at"))
        self.assertFalse(tasks[0].HasField("finished_at"))
```

panh99: Nvm, the timestamp fields are not `optional` in `Task` proto. I'll resolve this.

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -114,6 +114,91 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+        """Move an unfinished task to finished.
+
+        Parameters
+        ----------
+        task_id : int
+            The ID of the task to finish.
+        substatus : str
+            Terminal task sub-status, such as completed, failed, or stopped.
```
panh99, line 160: Should we add that only running status can be transition to finished:completed?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -114,6 +114,91 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+        detail : str
+            Additional terminal status details.
```
panh99, line 163:
```suggestion
            Additional terminal status detail.
```

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -119,6 +124,131 @@ def test_get_task_returns_copy(self) -> None:
...
+        state = self.state_factory()
+        task_id = state.create_task(task_type="flwr-model", run_id=42)
+        assert task_id is not None
+
+        # Task does not exist.
+        self.assertFalse(state.finish_task(61016, SubStatus.FAILED, "missing"))
+        # Task is pending, so it can be finished.
+        self.assertTrue(state.finish_task(task_id, SubStatus.FAILED, "boom"))
+        # Task is already finished, so it cannot be finished again.
+        self.assertFalse(state.finish_task(task_id, SubStatus.COMPLETED, "again"))
```
panh99: COMPLETED substatus is a bit different. Let's still use FAILED here. Btw, should we use `@parameterized.expand([(SubStatus.FAILED,), (SubStatus.STOPPED,)])` to check both substatus?
```suggestion
        self.assertFalse(state.finish_task(task_id, SubStatus.FAILED, "again"))
```

jafermarq: sure, and not include `SubStatus.COMPLETED`?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+
+    def get_task_id_by_token(self, token: str) -> int | None:
+        """Return the task ID associated with the task token, if valid."""
+        with self.lock_task_store:
+            # Resolve tokens after cleanup so callers never receive expired claims.
+            self._cleanup_expired_task_tokens_locked()
+            return self.task_token_to_task_id.get(token)
+
+    def _cleanup_expired_task_tokens_locked(self) -> None:
+        """Remove expired task tokens while holding `lock_task_store`."""
```
panh99: What does `_locked` mean here? The method body doesn't seem to use any lock.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
                 result.append(task_copy)
             return result
 
+    def claim_task(self, task_id: int) -> str | None:
+        """Atomically claim a pending task."""
+        token = secrets.token_hex(FLWR_APP_TOKEN_LENGTH)
+        with self.lock_task_store:
+            # Expire abandoned claims before deciding whether this task can be claimed.
+            self._cleanup_expired_task_tokens_locked()
```
panh99: I don't think we need to clean up the token here. If a token has expired, the task should simply fail, which doesn’t inherently block the task-claiming logic. (This is different from non-bijective run <> token mapping). Reducing the frequency of cleanup could actually be beneficial.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+            task = self.task_store.get(task_id)
+            if task is None or task_id in self.task_token_store:
+                return None
+            if determine_task_status(task).status != Status.PENDING:
```
panh99: If I am not mistaken, the `status` field of the `task` should be updated, right? If so, should we also just use `task.status` instead of calling `determine_task_status`?
```suggestion
            if task.status.status != Status.PENDING:
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
jafermarq: yes, this was referring to the fact that this helper method is called from other methods in the class after adquiring the lock. For example:

```python
    def acknowledge_task_heartbeat(self, task_id: int) -> bool:
        """Extend heartbeat state for the claimed task."""
        with self.lock_task_store:
            # Heartbeats are accepted only for active, unexpired task claims.
            self._cleanup_expired_task_tokens_locked()
            ....
```

I can probably adjust that docstring to avoid confusion

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+            if task is None or determine_task_status(task).status == Status.FINISHED:
+                return False
+
+            task.finished_at = now().isoformat()
+            task.status.CopyFrom(
+                TaskStatus(
+                    status=Status.FINISHED,
+                    sub_status=substatus,
+                    details=detail,
+                )
```
panh99, line 247: Would this look more consistent with other methods?
```suggestion
                TaskStatus(
                    status=Status.FINISHED, sub_status=substatus, details=detail
                )
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+            )
+            # Finished tasks no longer need claim/heartbeat bookkeeping.
+            if (record := self.task_token_store.pop(task_id, None)) is not None:
+                self.task_token_to_task_id.pop(record.token, None)
```
panh99: I think we don't have to clear tokens now (before, it's to allow a new token to be created for a run). Or you think it may help reduce potential token conflicts and avoid overheads?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+            self._cleanup_expired_task_tokens_locked()
+            return self.task_token_to_task_id.get(token)
+
+    def _cleanup_expired_task_tokens_locked(self) -> None:
+        """Remove expired task tokens while holding `lock_task_store`."""
+        current = now().timestamp()
+        for task_id, record in list(self.task_token_store.items()):
+            if record.active_until < current:
+                del self.task_token_store[task_id]
+                self.task_token_to_task_id.pop(record.token, None)
```
panh99, line 302: Should we transition the task status to `finished:failed` with detail saying "No heartbeat received from the task"?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+            self.task_token_store[task_id] = TokenRecord(
+                token=token,
+                active_until=claimed_at.timestamp() + HEARTBEAT_DEFAULT_INTERVAL,
+            )
+            self.task_token_to_task_id[token] = task_id
+            return token
+
+    def activate_task(self, task_id: int) -> bool:
+        """Move a task from starting to running."""
+        with self.lock_task_store:
```
panh99, line 212:
```suggestion
        with self.lock_task_store:
            # Expire non-responsive tasks before transitioning task status.
            self._cleanup_expired_task_tokens_locked()

```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+
+            task.running_at = now().isoformat()
+            task.status.CopyFrom(
+                TaskStatus(status=Status.RUNNING, sub_status="", details="")
+            )
+            return True
+
+    def finish_task(self, task_id: int, substatus: str, detail: str) -> bool:
+        """Move an unfinished task to finished."""
+        with self.lock_task_store:
```
panh99, line 229:
```suggestion
        with self.lock_task_store:
            # Expire non-responsive tasks before transitioning task status.
            self._cleanup_expired_task_tokens_locked()

```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+            )
+            # Finished tasks no longer need claim/heartbeat bookkeeping.
+            if (record := self.task_token_store.pop(task_id, None)) is not None:
+                self.task_token_to_task_id.pop(record.token, None)
+            return True
+
+    def acknowledge_task_heartbeat(self, task_id: int) -> bool:
+        """Extend heartbeat state for the claimed task."""
+        with self.lock_task_store:
+            # Heartbeats are accepted only for active, unexpired task claims.
```
```suggestion
            # Heartbeats are accepted only for starting and running tasks
```

framework/py/flwr/supercore/corestate/corestate_test.py
panh99: Yes! Cuz we only allow transition from running to finished:completed sub status, which is different from other sub statuses.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -181,6 +185,100 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+            )
+            return True
+
+    def finish_task(self, task_id: int, substatus: str, detail: str) -> bool:
+        """Move an unfinished task to finished."""
+        with self.lock_task_store:
+            # Transition task to FINISHED
+            task = self.task_store.get(task_id)
+            if task is None or determine_task_status(task).status == Status.FINISHED:
+                return False
```
panh99: Let's do a special check for `substatus == SubStatus.COMPLETED`: the current status must be running in order to be transitioned to completed sub status.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
jafermarq: I've adjusted the description of the method.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
jafermarq: @panh99 this is just formatting no? our formatting script does allow this change

framework/py/flwr/supercore/corestate/in_memory_corestate.py
jafermarq: i agree it's not that needed. after all we are using mappings which should be fast. I'll remove these lines to keep the methods thinner

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -177,10 +181,125 @@ def get_tasks(  # pylint: disable=too-many-arguments
             for task in tasks:
                 task_copy = Task()
                 task_copy.CopyFrom(task)
-                task_copy.status.CopyFrom(determine_task_status(task))
+                task_copy.status.CopyFrom(task.status)
```
panh99: This is redundant given the above line.
```suggestion
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
panh99: It allows, no? I tested it locally. Maybe you can delete the comma after `details=detail` and then run `black` again? It will be collapsed into one line.

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -114,6 +114,92 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+
+        Returns
+        -------
+        bool
+            True if the task existed and transitioned from starting to running,
+            otherwise False.
+        """
+
+    @abstractmethod
+    def finish_task(self, task_id: int, sub_status: str, detail: str) -> bool:
```
panh99: After checking the proto, it seems that we've used the word `details`. Should we also update here to keep consistency? I guess we use the word b/c gRPC also uses `details` instead of `detail`.

```suggestion
    def finish_task(self, task_id: int, sub_status: str, details: str) -> bool:
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -177,10 +181,124 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+        """
+        expired_at = now()
+        current = expired_at.timestamp()
+        for task_id, record in list(self.task_token_store.items()):
+            if record.active_until < current:
+                # The task is considered expired. Mark it as finished with a failed
+                # status if it's not already finished, and remove the token.
+                if (
+                    task := self.task_store.get(task_id)
+                ) is not None and task.status.status != Status.FINISHED:
```
```suggestion
                task = self.task_store.get(task_id)
                if task and task.status.status != Status.FINISHED:
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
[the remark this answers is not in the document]
panh99: The task will not stay at `STARTING` status; it will be transitioned to `FINISHED:FAILED`.

framework/py/flwr/supercore/corestate/sql_corestate.py
panh99, line 54: We use task status as conditions quite often, and I think we may introduce a table mapping status to table conditions to improve readability and maintainability.
```suggestion
# Define SQL conditions for task statuses to ensure consistency across queries
STATUS_CONDITIONS = {
    Status.PENDING: "(starting_at IS NULL AND finished_at IS NULL)",
    Status.STARTING: "(starting_at IS NOT NULL AND running_at IS NULL "
    "AND finished_at IS NULL)",
    Status.RUNNING: "(running_at IS NOT NULL AND finished_at IS NULL)",
    Status.FINISHED: "(finished_at IS NOT NULL)",
}

```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+                    UPDATE task
+                    SET token = :token,
+                        active_until = :active_until,
+                        starting_at = :starting_at
+                    WHERE task_id = :task_id
+                    AND pending_at IS NOT NULL
+                    AND starting_at IS NULL
+                    AND running_at IS NULL
+                    AND finished_at IS NULL
+                    AND token IS NULL
```
```suggestion
                    f"""
                    UPDATE task
                    SET token = :token,
                        active_until = :active_until,
                        starting_at = :starting_at
                    WHERE task_id = :task_id AND token IS NULL
                    AND {STATUS_CONDITIONS[Status.PENDING]}
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
         """Return SQLAlchemy MetaData needed for CoreState tables."""
         return create_corestate_metadata()
 
+    def claim_task(self, task_id: int) -> str | None:
+        """Atomically claim a pending task."""
+        token = secrets.token_hex(FLWR_APP_TOKEN_LENGTH)
+        claimed_at = now()
+        sint64_task_id = uint64_to_int64(task_id)
+        try:
+            with self.session():
```
panh99: No need to use `with self.session()` to explicitly for a single `self.query` call after removing `self._cleanup_expired_task_tokens()`. The call itself is atomic.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+                # Expire abandoned claims before trying to claim this task.
+                self._cleanup_expired_task_tokens()
```
panh99: As in in-memory corestate, expired tokens won't impact task claiming. Let's remove it to reduce the cleanup frequency.

```suggestion
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+                        ),
+                        "starting_at": claimed_at.isoformat(),
+                    },
+                )
+                if not rows:
+                    return None
+
+                return token
+        except IntegrityError:
+            return None
```
panh99, line 281: Am I right?
```suggestion
        except IntegrityError:
            # Rare failure: generated token already exists (duplicate)
            return None
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+
+            # Activation is a strict STARTING -> RUNNING transition.
+            rows = self.query(
+                """
+                UPDATE task
+                SET running_at = :running_at
+                WHERE task_id = :task_id
+                AND starting_at IS NOT NULL
+                AND running_at IS NULL
+                AND finished_at IS NULL
```
```suggestion
                f"""
                UPDATE task
                SET running_at = :running_at
                WHERE task_id = :task_id AND {STATUS_CONDITIONS[Status.STARTING]}
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+            updated = self.query(
+                f"""
+                UPDATE task
+                SET finished_at = :finished_at,
+                    sub_status = :sub_status,
+                    details = :details,
+                    active_until = NULL,
+                    token = NULL
+                WHERE task_id = :task_id
+                AND finished_at IS NULL{completion_constraint}
```
```suggestion
                AND finished_at IS NULL {completion_constraint}
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+
+    def finish_task(self, task_id: int, sub_status: str, detail: str) -> bool:
+        """Move an unfinished task to finished."""
+        sint64_task_id = uint64_to_int64(task_id)
+        with self.session():
+            self._cleanup_expired_task_tokens()
+            # FINISHED:COMPLETED is only valid from RUNNING.
+            completion_constraint = ""
+            if sub_status == SubStatus.COMPLETED:
+                completion_constraint = " AND running_at IS NOT NULL"
```
```suggestion
                completion_constraint = "AND running_at IS NOT NULL"
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+
+            updated = self.query(
```
panh99: Should we use `rows` for consistency? Plus, the return type is not really a `bool`.
```suggestion
            rows = self.query(
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+        """Extend heartbeat state for the claimed task."""
+        # Heartbeats are accepted only for active, unexpired task claims.
+        self._cleanup_expired_task_tokens()
+        current = now().timestamp()
+        rows = self.query(
+            """
+            UPDATE task
+            SET active_until = :active_until
+            WHERE task_id = :task_id
+            AND token IS NOT NULL
```
panh99: Token is checked in the auth layer. It would be nice if we can avoid replying on `token` in task methods. The only exception is `claim_task` where we inevitably use token to ensure the atomic operation of status transition + token creation.
```suggestion
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+                    current + HEARTBEAT_PATIENCE * HEARTBEAT_DEFAULT_INTERVAL
+                ),
+            },
+        )
+        return len(rows) > 0
+
+    def get_task_id_by_token(self, token: str) -> int | None:
+        """Return the task ID associated with the task token, if valid."""
+        # Resolve tokens after cleanup so callers never receive expired claims.
+        self._cleanup_expired_task_tokens()
```
panh99: Since this function will be called in every RPC, including `Push/PullObject`, should we remove this full table cleanup line? The condition of `active_until >= :current AND finished_at IS NULL` should suffice: no expired token will be returned.

```suggestion
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+
+        rows = self.query(
+            """
+            SELECT task_id FROM task
+            WHERE token = :token
+            AND active_until >= :current
+            AND finished_at IS NULL
+            LIMIT 1
```
panh99: `LIMIT 1` is unnecessary given the uniqueness constraint of `token`.
```suggestion
            WHERE token = :token AND active_until >= :current AND finished_at IS NULL
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +237,176 @@ def get_metadata(self) -> MetaData:
...
+            """
+            UPDATE task
+            SET token = NULL,
+                active_until = NULL,
+                finished_at = :finished_at,
+                sub_status = :sub_status,
+                details = :details
+            WHERE token IS NOT NULL
+            AND active_until < :current
+            AND finished_at IS NULL
```
panh99: Once set to finished, the token will be cleared, and thus we don't need to check the `finished_at` timestamp specifically, right?
```suggestion
            WHERE token IS NOT NULL AND active_until < :current
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -366,8 +539,8 @@ def determine_task_status(row: dict[str, Any]) -> TaskStatus:
         if row["finished_at"]:
             return TaskStatus(
                 status=Status.FINISHED,
-                sub_status=row.get("sub_status", ""),
-                details=row.get("details", ""),
+                sub_status=row.get("sub_status") or "",
+                details=row.get("details") or "",
```
panh99: Why this change? Will it be better if we directly set the `default` argument when calling `.get` method?

panh99 (commented):
Pretty good already! A few comments around sql queries.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -234,6 +245,170 @@ def get_metadata(self) -> MetaData:
...
+                RETURNING task_id
+                """,
+                {
+                    "task_id": sint64_task_id,
+                    "finished_at": now().isoformat(),
+                    "sub_status": sub_status,
+                    "details": details,
+                },
+            )
+            if not updated:
```
jafermarq:
```suggestion
            if not rows:
```

framework/py/flwr/supercore/corestate/sql_corestate.py
jafermarq: not sure, it was added by @danielnugraha

jafermarq: but given we're defining already the cols with a default:
```
        Column("sub_status", String, nullable=False, server_default=text("''")),
        Column("details", String, nullable=False, server_default=text("''")),
```
we don't need that `or ""`

jafermarq: updated to simply do `.get(..)`

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -114,6 +114,92 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+
+        Parameters
+        ----------
+        task_id : int
+            The ID of the task to finish.
+        sub_status : str
+            Terminal task sub-status, such as completed, failed, or stopped.
+            Only RUNNING status can be transitioned to FINISHED:COMPLETED
+        details : str
+            Additional terminal status detail.
```
```suggestion
            Additional terminal status details.
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -177,10 +181,124 @@ def get_tasks(  # pylint: disable=too-many-arguments
...
+                    return False
+
+            task.finished_at = now().isoformat()
+            task.status.CopyFrom(
+                TaskStatus(
+                    status=Status.FINISHED,
+                    sub_status=sub_status,
+                    details=detail,
+                )
+            )
```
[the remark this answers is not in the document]
panh99: @jafermarq Should we clear the token?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -366,8 +530,8 @@ def determine_task_status(row: dict[str, Any]) -> TaskStatus:
         if row["finished_at"]:
             return TaskStatus(
                 status=Status.FINISHED,
-                sub_status=row.get("sub_status", ""),
-                details=row.get("details", ""),
+                sub_status=cast(str, row.get("sub_status")),
+                details=cast(str, row.get("details")),
```
```suggestion
                sub_status=row["sub_status"],
                details=row["details"],
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7061, opened 2026-04-29
fix(framework): Report clientapp appio communication failures as errors

msheller:
Minor fix for error checking corner case that could cause silent failures (success reported despite error).

This addresses the codex comment remaining in already merged PR 6986

### Checklist

- [x] Write tests
- [x] Make CI checks pass


framework/docs/source/ref-exit-codes/250.rst
panh99, line 1: The implementation looks good to me, but I think the new exit-code docs are a little too user-action-oriented for an internal process. In normal runtime usage, users do not start `flwr-clientapp` directly; it is launched by SuperExec with the AppIO address/token/TLS settings already derived from the SuperNode/SuperExec setup.

Could we adjust this page to say that this means the internal ClientApp process (`flwr-clientapp`) could not communicate with the SuperNode ClientAppIo API, and that users should check the surrounding SuperNode/ClientApp logs for the underlying gRPC error? A likely cause is that the ClientApp process took too long to start, for example on a very slow or overloaded system, causing the short-lived token/heartbeat window to expire. If the log message is not enough to diagnose it, users should contact the Flower team with the relevant logs. Wdyt?

msheller: Updated! LMK if you want further changes!

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7067, opened 2026-05-01
feat(framework): Support interceptor unary stream

mohammadnaseri:
This PR extends the runtime version interceptor to support unary-stream RPCs, in addition to the existing unary-unary support.


framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Runtime version metadata interceptors."""
 
-
 from __future__ import annotations
```
```suggestion

from __future__ import annotations
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -62,8 +66,35 @@ def intercept_unary_unary(
...
+                incompat_message = get_metadata_str(
+                    done.trailing_metadata(),
+                    VERSION_INCOMPATIBILITY_MESSAGE_METADATA_KEY,
+                )
+                if incompat_message:
+                    self._compatibility_warning_logged = True
+                    log(WARN, incompat_message)
 
-class RuntimeVersionServerInterceptor(grpc.ServerInterceptor):  # type: ignore
+        call.add_done_callback(_log_incompat_warning)
```
panh99: Does this mean the incompatibility note will only be printed upon the RPC stream is ended? Is it possible to log it earlier, e.g., in the first response of the stream?

mohammadnaseri: Yes and yes, I pushed a change.

panh99 (dismissed):

panh99 (commented):
This feels larger than the requested change. If the goal is only to extend runtime-version metadata handling to unary-stream RPCs, could we first look for a minimal gRPC-native implementation?

My biggest concern is `_UnaryStreamCallWrapper`: it partially proxies a gRPC call object, while unary-stream calls are expected to behave both as iterators and as `grpc.Call`/future-like objects. Recreating that surface with `__iter__`, `__next__`, and `__getattr__` seems fragile and hard to review, especially for callers or future interceptors that may rely on gRPC’s real call semantics. I’m not convinced this maintenance cost is justified for logging a version warning.

Also, `send_initial_metadata()` can be risky because initial metadata can only be sent once and may conflict with other interceptors/handlers. If an early warning cannot be done simply and safely, I think we should keep using trailing metadata and log when the stream completes.

mohammadnaseri:
> This feels larger than the requested change. If the goal is only to extend runtime-version metadata handling to unary-stream RPCs, could we first look for a minimal gRPC-native implementation?
> 
> My biggest concern is `_UnaryStreamCallWrapper`: it partially proxies a gRPC call object, while unary-stream calls are expected to behave both as iterators and as `grpc.Call`/future-like objects. Recreating that surface with `__iter__`, `__next__`, and `__getattr__` seems fragile and hard to review, especially for callers or future interceptors that may rely on gRPC’s real call semantics. I’m not convinced this maintenance cost is justified for logging a version warning.
> 
> Also, `send_initial_metadata()` can be risky because initial metadata can only be sent once and may conflict with other interceptors/handlers. If an early warning cannot be done simply and safely, I think we should keep using trailing metadata and log when the stream completes.

Agreed. I removed the unary-stream call wrapper and the early initial-metadata path. The unary-stream interceptor now returns the original gRPC call object unchanged and only registers a done callback, when available, to inspect trailing metadata after the stream completes. On the server side, incompatibility warnings continue to use trailing metadata only, so we avoid calling `send_initial_metadata()` and preserve gRPC’s normal call semantics.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -52,18 +70,35 @@ def intercept_unary_unary(
...
-                log(WARN, incompat_message)
+        )
+        call: grpc.Call = continuation(details, request)
+
+        def _log_incompat_warning(done: Any) -> None:
+            self._maybe_log_incompat_warning(done.trailing_metadata())
+
+        add_done_callback = getattr(call, "add_done_callback", None)
+        if callable(add_done_callback):
+            add_done_callback(_log_incompat_warning)
```
panh99: Shall we use stable API from `grpc.Call`?
```suggestion
        def _log_incompat_warning() -> None:
            self._maybe_log_incompat_warning(call.trailing_metadata())

        call.add_callback(_log_incompat_warning)
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -94,24 +127,54 @@ def intercept_service(
...
+            ((VERSION_INCOMPATIBILITY_MESSAGE_METADATA_KEY, incompat_message),)
+            if incompat_message
+            else None
         )
 
+        def maybe_set_incompat_trailing_metadata(
+            context: grpc.ServicerContext,
+        ) -> None:
+            if incompat_metadata:
+                context.set_trailing_metadata(incompat_metadata)
```
panh99: Should we call it `trailing_metadata` since we might add error message in the metadata soon? Also rename the function name to `maybe_set_trailing_metadata`.
```suggestion
        # Prepare trailing metadata
        trailing_metadata = ()
        if incompat_details:
            incompat_message = (
                "Runtime version compatibility check failed for "
                f"{self._connection_name}. {incompat_details}"
            )
            trailing_metadata += (
                (VERSION_INCOMPATIBILITY_MESSAGE_METADATA_KEY, incompat_message),
            )

        def maybe_set_trailing_metadata(
            context: grpc.ServicerContext,
        ) -> None:
            if trailing_metadata:
                context.set_trailing_metadata(trailing_metadata)
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```suggestion

from __future__ import annotations
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -94,23 +128,52 @@ def intercept_service(
...
+        ) -> None:
+            if trailing_metadata:
+                context.set_trailing_metadata(trailing_metadata)
+
+        if method_handler.unary_unary is not None:
+
+            def wrapped(
+                request: GrpcMessage, context: grpc.ServicerContext
+            ) -> GrpcMessage:
+                maybe_set_incompat_trailing_metadata(context)
```
panh99:
```suggestion
                maybe_set_trailing_metadata(context)
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -94,23 +128,52 @@ def intercept_service(
...
+                request_deserializer=method_handler.request_deserializer,
+                response_serializer=method_handler.response_serializer,
+            )
+
+        if method_handler.unary_stream is not None:
+
+            def wrapped_stream(
+                request: GrpcMessage, context: grpc.ServicerContext
+            ) -> Any:
+                maybe_set_incompat_trailing_metadata(context)
```
panh99:
```suggestion
                maybe_set_trailing_metadata(context)
```

panh99 (approved):
LGTM!

PR #7068, opened 2026-05-01
feat(framework): Allow task token in appio token auth interceptors

panh99:


framework/py/flwr/supercore/interceptors/appio_token_interceptor.py
```diff
@@ -141,6 +158,11 @@ def _authenticated_handler(
         )
 
 
+def get_authenticated_task_id() -> int:
+    """Get the authenticated task ID from the current context."""
+    return _current_task_id.get()
```
jafermarq: Will this be used for other than for the tests?

panh99: Yes, the RPC handlers will use this function to extract the identity of the caller instead of relying on the identity set in the request body.

danieljanes (approved):

PR #7079, opened 2026-05-05
refactor(framework): Create task in `SuperNode` when a `Message` is pulled

jafermarq:


framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -408,6 +410,33 @@ def _pull_and_store_message(  # pylint: disable=too-many-positional-arguments
...
+                    "processed.",
+                    run_info.fab_hash,
+                    run_id,
+                )
+                return None
+            fab_hash_str = fab_.hash_str
+
+        # Create task
+        task_id = state.create_task(
+            task_type=TaskType.CLIENT_APP, run_id=run_id, fab_hash=fab_hash_str
```
panh99: Should we just use `run_info.fab_hash`? We don't need the actual FAB object, right?
```suggestion
            task_type=TaskType.CLIENT_APP, run_id=run_id, fab_hash=run_info.fab_hash
```

framework/py/flwr/supernode/start_client_internal.py
```diff
@@ -408,6 +410,33 @@ def _pull_and_store_message(  # pylint: disable=too-many-positional-arguments
...
+        )
+        if task_id is None:
+            log(
```
panh99, line 420:
```suggestion
            # Task creation can fail if the generated uint64 task ID collides
            log(
```

framework/py/flwr/supernode/start_client_internal.py
jafermarq: totally. i didn't remember we had that in `run_info`

panh99 (approved):

PR #7084, opened 2026-05-05
feat(framework): Add runtime version metadata to ServerAppIo Grid client

mohammadnaseri:
This PR adds `RuntimeVersionClientInterceptor` to `GrpcGrid`, which is used by `flwr-serverapp` to call the SuperLink ServerAppIo API.


framework/py/flwr/server/grid/grpc_grid.py
```diff
@@ -159,7 +162,10 @@ def _connect(self) -> None:
             server_address=self._addr,
             insecure=self._insecure,
             root_certificates=self._cert,
-            interceptors=[AppIoTokenClientInterceptor(token=self._token)],
+            interceptors=[
+                RuntimeVersionClientInterceptor(component_name=TaskType.SERVER_APP),
```
panh99: Should we hard-code it here? The `TaskType` is not intended to use like this, and it's better to avoid relying on it.

```suggestion
                RuntimeVersionClientInterceptor(component_name="flwr-serverapp"),
```

framework/py/flwr/server/grid/grpc_grid_test.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Tests for grid SDK."""
 
-
 import threading
```
panh99, line 18:
```suggestion

import threading
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7089, opened 2026-05-06
refactor(framework): Make `ClientAppIoServicer` claim and finish tasks

jafermarq:


framework/py/flwr/supernode/runtime/run_clientapp.py
```diff
@@ -252,6 +253,12 @@ def push_appoutputs(
         with no_object_id_recompute():
             # Get object tree and all objects to push
             object_tree = get_object_tree(message)
+            has_error = message.has_error()
```
panh99: The `push_appoutputs` is just a helper function. Would it be better to pass `substatus` and `details` to this function directly, and determine them in the main body of `run_clientapp`? Btw, `details` should be `""` when it's completed successfully. The description of `details` for now is the details of the failure.

framework/py/flwr/supernode/runtime/run_clientapp.py
```diff
@@ -122,6 +122,8 @@ def on_exit() -> None:
         # Pull Message, Context, Run and FAB from SuperNode
         message, context, run, fab = pull_appinputs(stub=stub, token=token)
 
+        sub_status = SubStatus.COMPLETED
+        details = ""
```
panh99: Should we initialise it with failed substatus? Feel free to tweak the details, btw.

```suggestion
        sub_status = SubStatus.FAILED
        details = "ClientApp task failed due to unknown reason"
```

jafermarq: isn't it better to initialize it as successful and edit it when it fails? (we can give in that way a more informative `details`)

panh99: I see wym. I wonder if we should move `push_appoutputs` invocation to `on_exit`, and then default the substatus to failed. Wdyt?

panh99: The default values will just be the last resort to use when somehow the process fails to set the correct `details`. For example, the `flwr-clientapp` is killed by `pkill`.

panh99 (approved):
LGTM!

PR #7090, opened 2026-05-06
feat(framework): Make `ServerAppIo` switch to task token

panh99:


framework/py/flwr/server/serverapp/app.py
```diff
@@ -260,46 +247,52 @@ def on_exit() -> None:
...
-        log(DEBUG, "[flwr-serverapp] Will push ServerAppOutputs")
-        out_req = PushAppOutputsRequest(
-            token=token, run_id=run.run_id, context=context_proto
-        )
-        _ = grid._stub.PushAppOutputs(out_req)
-
-        run_status = RunStatus(Status.FINISHED, SubStatus.COMPLETED, "")
-
-    # Raised when the run is already stopped by the user
-    except RunNotRunningException:
```
danielnugraha, line 283: `RunNotRunningException` is raised by `retry_invoker`. Do we need to handle the error here?

panh99: Good question. The `RunNotRunningException` was used when an executor still tried to talk to ServerAppIo API when the run is stopped by the user. Since we've enabled token authentication, and a token will be cleared when the run is stopped, so the caller (executor) will directly get a unauthenticated error instead of triggering `RunNotRunningException`.

danielnugraha (approved):

jafermarq (approved):

danieljanes (approved):

PR #7091, opened 2026-05-06
feat(framework): Add runtime version check for `ClientAppIo`

mohammadnaseri:
This PR adds runtime-version metadata wiring for the `flwr-clientapp `<-> SuperNode `ClientAppIo` API link.


framework/py/flwr/supernode/runtime/run_clientapp.py
```diff
@@ -93,7 +97,10 @@ def run_clientapp(  # pylint: disable=R0913, R0914, R0915, R0917
         server_address=clientappio_api_address,
         insecure=insecure,
         root_certificates=certificates,
-        interceptors=[AppIoTokenClientInterceptor(token)],
+        interceptors=[
+            RuntimeVersionClientInterceptor(component_name=TaskType.CLIENT_APP),
```
panh99: 😆 Good catch! But let’s hard-code the component name here. `TaskType` isn’t intended for this use case, so it’s better to avoid relying on it.
```suggestion
            RuntimeVersionClientInterceptor(component_name="flwr-clientapp"),
```

panh99 (approved):
LGTM!

PR #7101, opened 2026-05-07
refactor(framework): Remove `create_token` from state

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -180,6 +183,18 @@ def get_context(self, run_id: int) -> Context | None:
 
     def get_run_ids_with_pending_messages(self) -> Sequence[int]:
```
panh99: Could we remove this as well? No need to update it now. It was used in the removed `PullAppsToLaunch` RPC

framework/py/flwr/supernode/nodestate/in_memory_nodestate.py
```diff
@@ -189,27 +204,68 @@ def get_run_ids_with_pending_messages(self) -> Sequence[int]:
...
-        # Remove run IDs that have tokens stored (indicating they are in progress)
-        with self.lock_token_store:
-            ret -= set(self.token_store.keys())
-            return list(ret)
+        # Exclude runs with an active claim already in progress.
+        ret -= active_task_run_ids
+        ret -= active_run_token_ids
+        return list(ret)
+
+    def _cleanup_expired_task_tokens_locked(self) -> None:
```
panh99: Let's avoid overriding `_cleanup_expired_task_tokens_locked` this way, since it would duplicate the cleanup logic.

Could we instead add a hook in `CoreState`, something like:

```python
def _on_task_token_expires(self, tasks: list[Task]) -> None:
    pass
```

Then `_cleanup_expired_task_tokens` can call it after the expired task tokens have been cleaned up. That way, `LinkState`/`NodeState` can override `_on_task_token_expires` only, without duplicating the logic for clearing task tokens.

panh99: Btw, is this change required to remove `.create_token`?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -314,22 +315,23 @@ def _cleanup_expired_task_tokens_locked(self) -> None:
...
+        if expired_records:
+            self._on_task_tokens_expired(expired_records)
 
-            active_until = int(now().timestamp()) + HEARTBEAT_DEFAULT_INTERVAL
-            self.token_store[run_id] = TokenRecord(
-                token=token, active_until=active_until
-            )
-            self.token_to_run_id[token] = run_id
-        return token
+    def _on_task_tokens_expired(self, expired_records: list[tuple[int, int]]) -> None:
```
panh99: In this callback method, since we now have `Task`, and given what `LinkState` and `NodeState` need, I think the following signature is sufficient:

```python
def _on_task_tokens_expired(self, tasks: list[Task]) -> None:
```

I previously used the rather awkward `expired_records: list[tuple[int, int]]` because we needed that information at the time. But I don’t think it’s necessary anymore; `list[Task]` is cleaner and should be enough.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7104, opened 2026-05-07
feat(framework): Add FlowerError JSON deserialization

mohammadnaseri:


framework/py/flwr/supercore/error/base.py
```diff
@@ -66,6 +66,39 @@ def to_json(self, public_message: str) -> str:
             }
         )
 
+    @classmethod
+    def from_json(cls, value: str | None) -> "FlowerError | None":
```
panh99: Let's add `from __future__ import annotations` and then we don't need the double quotes.

panh99 (dismissed):
LGTM!

panh99 (approved):

PR #7107, opened 2026-05-07
feat(framework): Log runtime version FlowerError rejections in client interceptor

mohammadnaseri:


framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -68,10 +83,22 @@ def intercept_unary_unary(
...
-        call: grpc.Call = continuation(details, request)
+        try:
+            call: grpc.Call = continuation(details, request)
+        except grpc.RpcError as err:
+            self._maybe_log_incompat_error(err)
+            raise
+
+        def _log_incompat_warning() -> None:
+            self._maybe_log_incompat_warning(call.trailing_metadata())
+            self._maybe_log_incompat_error(call)
```
panh99: Why we call it in this function?

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -68,10 +83,22 @@ def intercept_unary_unary(
...
+
+        if isinstance(call, grpc.RpcError):
+            _log_incompat_warning()
+            return call
```
panh99: why we log a warning when the grpc error is raised?

Given that `context.abort` will be called on the server side immediately (if it wants to), I think we could just do:

```python
try:
    call: grpc.Call = continuation(details, request)
except grpc.RpcError as err:
    if flwr_err := FlowerError.from_json(err.details()):
        # Format and log the error
        # Call `flwr_exit` or `os.kill(os.getpid(), SIGINT)` to terminate
    else:
        raise

# If the call works, log warning message if exists
```

panh99 (commented):
I might be missing part of the interceptor logic here. My understanding is that the client-side flow should be:

1. If the call fails, check whether it failed due to an incompatibility error. If so, log the error message and exit.
2. If the call succeeds, check whether there is an incompatibility warning attached in the trailing metadata, and log it if present.

It looks to me like this PR may be mixing warning handling and error handling slightly. Could we keep those two paths clean and separate?

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -56,6 +57,20 @@ def _maybe_log_incompat_warning(
             self._compatibility_warning_logged = True
             log(WARN, incompat_message)
 
+    def _maybe_log_incompat_error(self, grpc_error: Any) -> None:
+        """Log runtime-version rejections encoded as FlowerError JSON."""
+        if self._compatibility_warning_logged:
+            return
+
```
panh99: Should we remove this flag in this PR?
```suggestion
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Runtime version metadata interceptors."""
 
-
 from __future__ import annotations
```
```suggestion

from __future__ import annotations
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -39,63 +40,76 @@ class RuntimeVersionClientInterceptor(
...
+            self._maybe_log_incompat_warning(call.trailing_metadata())
+
+        if isinstance(call, grpc.RpcError) and (
+            not isinstance(call, grpc.Future) or call.done()
+        ):
+            self._maybe_log_incompat_error(call)
+            self._maybe_log_incompat_warning(
+                call.trailing_metadata() if hasattr(call, "trailing_metadata") else None
+            )
+            return call
```
panh99: Once the server reject connection, we should use `flwr_exit` to exit the process, no? If we return the `call` here, the client side will still see the unformatted raw grpc error with JSON.
```suggestion
        if isinstance(call, grpc.RpcError):
            self._maybe_handle_incompat_error(call)
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -39,63 +40,76 @@ class RuntimeVersionClientInterceptor(
...
-    def intercept_unary_unary(
+    def _maybe_log_incompat_error(self, grpc_error: Any) -> None:
+        """Log runtime-version rejections encoded as FlowerError JSON."""
+        details = grpc_error.details() if hasattr(grpc_error, "details") else None
+        flower_error = FlowerError.from_json(details)
+        if (
+            flower_error is not None
+            and flower_error.code == ApiErrorCode.RUNTIME_VERSION_INCOMPATIBLE
+        ):
+            log(WARN, flower_error.public_details or flower_error.message)
```
panh99: Let's add an ExitCode and use `flwr_exit` here. After receiving the rejection from the server, the client should log the message and exit.

```suggestion
            flwr_exit(
                ExitCode.RUNTIME_VERSION_INCOMPATIBLE, 
                f"{flower_error.message}\n{flower_error.public_details}"
            )
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
mohammadnaseri: @panh99 addressed, I kept the not `isinstance(call, grpc.Future)` or `call.done()` guard to avoid blocking pending .`future()` calls.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -39,63 +41,85 @@ class RuntimeVersionClientInterceptor(
...
         incompat_message = get_metadata_str(
             grpc_metadata,
             VERSION_INCOMPATIBILITY_MESSAGE_METADATA_KEY,
         )
         if incompat_message:
-            self._compatibility_warning_logged = True
             log(WARN, incompat_message)
 
-    def intercept_unary_unary(
+    def _maybe_log_incompat_error(self, grpc_error: Any) -> None:
```
panh99: Is this name a bit less clear given what the method is doing?

panh99: Should we use the exact type here?
```suggestion
    def _maybe_log_incompat_error(self, grpc_error: grpc.RpcError) -> None:
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -39,63 +41,85 @@ class RuntimeVersionClientInterceptor(
...
+            and flower_error.code == ApiErrorCode.RUNTIME_VERSION_INCOMPATIBLE
+        ):
+            flwr_exit(
+                ExitCode.RUNTIME_VERSION_INCOMPATIBLE,
+                "\n".join(
+                    detail
+                    for detail in (flower_error.message, flower_error.public_details)
+                    if detail
+                ),
+            )
```
panh99: Let's make the exit message construction a bit clearer. The original code looks a bit like black magic, even though we're not doing anything fancy here.

```suggestion
            exit_message = flower_error.message
            if flower_error.public_details:
                exit_message += f"\n{flower_error.public_details}"
            flwr_exit(ExitCode.RUNTIME_VERSION_INCOMPATIBLE, exit_message)
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
mohammadnaseri: renamed it to `_maybe_exit_on_incompat_error`

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
panh99: I might still be missing something, but I don't think the `grpc.Future` / `done()` guard is needed for the blocking concern here.

For unary-unary/stream interceptors, `continuation(...)` returns a `Call-Future`. If the RPC later fails, the future's exception value is an `RpcError`. So if `call` itself is already a `grpc.RpcError`, we're already on the failed-RPC path, not just holding a normal pending future.

I agree that `details()` can block on a generic `grpc.Call` until the status is available, but in this branch we are explicitly handling an RPC error. At that point I think waiting for the error details is the correct behavior; returning the raw error would just surface the unformatted gRPC/JSON error to the user.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -39,63 +41,77 @@ class RuntimeVersionClientInterceptor(
...
-        # Log the incompatibility message from the response metadata
-        self._maybe_log_incompat_warning(call.trailing_metadata())
+        def _handle_completion() -> None:
+            self._maybe_log_incompat_warning(call.trailing_metadata())
+
+        if isinstance(call, grpc.RpcError):
+            self._maybe_exit_on_incompat_error(call)
+            self._maybe_log_incompat_warning(
+                call.trailing_metadata() if hasattr(call, "trailing_metadata") else None
+            )
```
panh99: Curious. Is this to best-effort log warning even if the RPC fails? I guess it's not really necessary?

mohammadnaseri: Yep, now with `flwr_exit`, it is not required anymore.

framework/docs/source/ref-exit-codes/606.rst
```diff
@@ -0,0 +1,16 @@
...
+*************
+
+The Flower runtime versions are incompatible.
+
+****************
+ How to Resolve
+****************
+
+Use compatible Flower runtime versions for the components that communicate with each
+other. Check the error message for the specific compatibility issue.
```
panh99:
```suggestion
Check the error message for details. It may tell you which Flower version is
required.

If a specific version is shown, install that version:

.. code-block:: shell

    $ pip install flwr==<version>

Otherwise, upgrade Flower:

.. code-block:: shell

    $ pip install -U flwr
```

framework/py/flwr/common/exit/exit_code.py
```diff
@@ -183,6 +184,9 @@ def __new__(cls) -> ExitCode:
     ExitCode.COMMON_TLS_SERVER_CERTIFICATES_INVALID: (
         "TLS server certificate configuration is incomplete or invalid."
     ),
+    ExitCode.RUNTIME_VERSION_INCOMPATIBLE: (
+        "The Flower runtime versions are incompatible."
+    ),
```
panh99, line 190: Since the main error message already contains the specific version incompatibility details, I wonder if we should adjust the short help message for this exit code.

Otherwise the output can become a bit repetitive: the detailed message explains that the local Flower version is incompatible with the target server version, and then the short help message repeats the same thing in a more generic way.

For this case, I think the most useful structure would be:

```console
    Exit Code: 606
    <specific incompatibility message from the server>
    Upgrade your Flower version to the required version, or contact the server administrator.

    For more information, visit: <help-page-url>
```


```suggestion
    ExitCode.RUNTIME_VERSION_INCOMPATIBLE: (
        "Upgrade your Flower version to the required version, or contact the server administrator."
    ),
```

panh99 (dismissed):
LGTM!

danieljanes (dismissed):

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -39,63 +41,72 @@ class RuntimeVersionClientInterceptor(
...
             )
         )
         call: grpc.Call = continuation(details, request)
 
-        # Log the incompatibility message from the response metadata
-        self._maybe_log_incompat_warning(call.trailing_metadata())
+        def _handle_completion() -> None:
+            self._maybe_log_incompat_warning(call.trailing_metadata())
+            if isinstance(call, grpc.RpcError):
+                self._maybe_exit_on_incompat_error(call)
```
panh99, line 89: Should we handle error before handling warning?

mohammadnaseri: I’d keep warning first. If we exit first, `flwr_exit` may terminate before the trailing metadata warning is logged.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -39,63 +41,72 @@ class RuntimeVersionClientInterceptor(
...
                 client_call_details.metadata
             )
         )
         call: grpc.Call = continuation(details, request)
 
-        # Log the incompatibility message from the response metadata
-        self._maybe_log_incompat_warning(call.trailing_metadata())
+        def _handle_completion() -> None:
+            self._maybe_log_incompat_warning(call.trailing_metadata())
+            if isinstance(call, grpc.RpcError):
```
panh99, line 88:
```suggestion
            # Some successful call objects (e.g., unary-stream) can also be
            # subclasses of grpc.RpcError. Do not treat RpcError alone as failure;
            # _maybe_exit_on_incompat_error checks the actual RPC details.
            if isinstance(call, grpc.RpcError):
```

panh99 (approved):

danieljanes (approved):

PR #7108, opened 2026-05-07
feat(framework): Add reject mode for runtime version incompatibility

mohammadnaseri:


framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -192,49 +223,124 @@ def wrapped_stream(
...
         local_metadata=RuntimeVersionMetadata.from_local_component("SuperLink"),
         send_warning_metadata=send_warning_metadata,
+        reject_incompatible=reject_incompatible,
     )
 
 
 def create_clientappio_runtime_version_server_interceptor(
     connection_name: str = "Caller <-> SuperNode ClientAppIo API",
     send_warning_metadata: bool = False,
+    reject_incompatible: bool = False,
```
panh99, line 227: I think we plan to enforce the version check for AppIO API, right?

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -192,49 +223,124 @@ def wrapped_stream(
...
+    return (
+        local_version.major != peer_version.major
+        or local_version.minor != peer_version.minor
+    )
+
+
 def create_serverappio_runtime_version_server_interceptor(
     connection_name: str = "Caller <-> SuperLink ServerAppIo API",
     send_warning_metadata: bool = False,
+    reject_incompatible: bool = False,
```
panh99, line 213: Should this be `True` here cuz we plan to enforce the check for ServerAppIo API?

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -144,17 +149,41 @@ def intercept_service(
...
+            warning_message = _format_runtime_version_warning(
+                local_metadata=self._local_metadata,
+                peer_metadata=peer_metadata,
+                fallback_details=incompat_details,
+            )
+            error_details = _format_runtime_version_error(
+                local_metadata=self._local_metadata,
+                peer_metadata=peer_metadata,
+                fallback_details=incompat_details,
+            )
```
panh99: I think using the same `incompat_details` for both warning and error is acceptable here because the difference between warning and error is already represented by how the message is handled: warnings go through `log(WARN, ...)`, while errors go through `flwr_exit`, which formats the exit output. Adding separate formatting helpers feels like extra complexity for a temporary patch, especially since the proper long-term fix should probably be handled in `.check_compatibility` through a clearer compatibility policy. Wdyt?

panh99: To clarify my earlier comment: I agree that reusing the exact same message for both warning and error is not ideal in principle. My point is that I think it is acceptable for this temporary patch. The message already contains the relevant incompatibility details, and the two paths still present it differently.

Given that, I would prefer not to add extra helper functions or re-check the incompatibility reasons just to produce slightly different messages here. The longer-term fix should probably be handled when we update `.check_compatibility` with a proper policy.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
mohammadnaseri: yes, that will be in a different PR.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
mohammadnaseri: I will have a separate PR for that.

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -147,17 +149,33 @@ def intercept_service(
         if incompat_details is None:
             incompat_details = self._local_metadata.check_compatibility(peer_metadata)
 
+        warning_message = None
+        error_details = None
+        if incompat_details:
+            warning_message = incompat_details
+            error_details = incompat_details
```
panh99: Curious, is this part to remind us that we should have different messages in the future? Cuz it seems not necessary in this function.

mohammadnaseri: yep, but I think better to simplify it. pushed an update.

panh99 (approved):
LGTM!

PR #7117, opened 2026-05-08
feat(framework): Add `task_logs` table to `CoreState` and methods to read/write

jafermarq:


danielnugraha (dismissed):

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -61,6 +62,15 @@ def other_task_run_id(self, _state: CoreState) -> int:
...
         return 123
 
+    def _patch_task_log_datetime_now(self, *timestamps: Any) -> ExitStack:
+        """Patch the shared datetime source used for task-log timestamps."""
+        stack = ExitStack()
+        mock_datetime = stack.enter_context(
+            patch("flwr.supercore.date.datetime.datetime")
+        )
+        mock_datetime.now.side_effect = list(timestamps)
+        return stack
```
panh99: This is a very useful helper function for patching dateimes. The input argument type should be `datetime`, not `Any`, right? And `side_effect` accepts any iterable, including `timestamps` of type `tuple[datetime, ...]`, I think?


```suggestion
    def _patch_task_log_datetime_now(self, *timestamps: datetime) -> ExitStack:
        """Patch the shared datetime source used for task-log timestamps."""
        stack = ExitStack()
        mock_datetime = stack.enter_context(
            patch("flwr.supercore.date.datetime.datetime")
        )
        mock_datetime.now.side_effect = timestamps
        return stack
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -108,6 +108,72 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+                                SELECT MAX(timestamp) + :timestamp_increment
+                                FROM task_logs
+                                WHERE task_id = :task_id
+                            ),
+                            :current_ts
+                        )
+                    ),
+                    :task_id,
+                    :log
+                )
```
panh99: This looks a bit complicated. I understand it’s trying to avoid inserting multiple logs for the same task with identical timestamps, but I don’t think we need that guarantee here. In production, the chance of two calls getting the exact same timestamp should be negligible, and we’re likely moving to an explicit log index soon anyway, which would be a more robust way to preserve insertion order.

So I think we can keep this simpler for now and just insert rows directly. Wdyt?

```suggestion
                INSERT INTO task_logs (timestamp, task_id, log)
                VALUES (:current_ts, :task_id, :log)
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -108,6 +108,72 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+        """Get task logs for the specified `task_id`."""
+        sint64_task_id = uint64_to_int64(task_id)
+
+        with self.session():
+            rows = self.query(
+                "SELECT task_id FROM task WHERE task_id = :task_id",
+                {"task_id": sint64_task_id},
+            )
+            if not rows:
+                raise ValueError(f"Task {task_id} not found")
```
panh99: This check seems unnecessary. The task ID should already be validated in the auth layer, so we probably don’t need an extra DB query here. Plus, it does no harm to the database, I think.

```suggestion
```

framework/py/flwr/supercore/corestate/sql_corestate.py
jafermarq: fine with me!

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -108,6 +108,52 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+            rows = self.query(
+                """
+                SELECT log, timestamp FROM task_logs
+                WHERE task_id = :task_id AND timestamp > :after_timestamp
+                ORDER BY timestamp
+                """,
+                {"task_id": sint64_task_id, "after_timestamp": after_timestamp},
+            )
+            latest_timestamp = rows[-1]["timestamp"] if rows else 0.0
+            return "".join(row["log"] for row in rows), latest_timestamp
```
```suggestion
        # We don't check if the task exists before querying logs
        # because the task_id is validated by the authz layer

        if after_timestamp is None:
            after_timestamp = 0.0

        # Polling is strict-after: entries at the checkpoint timestamp have
        # already been delivered.
        rows = self.query(
            """
            SELECT log, timestamp FROM task_logs
            WHERE task_id = :task_id AND timestamp > :after_timestamp
            ORDER BY timestamp
            """,
            {"task_id": sint64_task_id, "after_timestamp": after_timestamp},
        )
        latest_timestamp = rows[-1]["timestamp"] if rows else 0.0
        return "".join(row["log"] for row in rows), latest_timestamp
```

panh99 (approved):
LGTM!

PR #7120, opened 2026-05-09
docs: improve documentation and add examples

AkhiChalasani:
## Summary

- Added Quick Start section after Installation


danieljanes:
Thanks for the PR @AkhiChalasani, but we don't think the proposed change improves the README.

PR #7122, opened 2026-05-11
feat(framework): Update `create_run` to include `primary_task_id`

danielnugraha:
…le sql query

### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99:
```suggestion
            return super().create_task(
```

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -126,13 +127,6 @@ def create_task(  # pylint: disable=too-many-arguments,too-many-positional-argum
...
                 connector_ref=connector_ref,
             )
-            if task_id is None:
-                return None
-
-            run_record = self.run_ids.get(run_id)
-            if run_record is not None and run_record.run.primary_task_id is None:
-                run_record.run.primary_task_id = task_id
-
             return task_id
```
panh99:
```suggestion
```

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -583,19 +576,25 @@ def create_run(
...
+                pending_at = now().isoformat()
                 run_record = RunRecord(
                     run=Run(
                         run_id=run_id,
                         fab_id=fab_id if fab_id else "",
                         fab_version=fab_version if fab_version else "",
                         fab_hash=fab_hash if fab_hash else "",
                         override_config=override_config,
-                        pending_at=now().isoformat(),
+                        pending_at=pending_at,
```
panh99: After linking run status to task status, we no longer need to set the pending_at field here.
```suggestion
                        pending_at="",
```

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
panh99: Should we remove this unit test? It was introduced first to check if `.create_task` sets the primary task of a run. Later, it is changed to test if `.create_task` will not override existing primary task of a run. But now, since `.create_task` no longer does anything to the run table, I think we don't need this unit test, right?

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -841,63 +827,88 @@ def create_run(  # pylint: disable=R0913, R0914, R0917
...
+             active_until, pending_at, starting_at, running_at, finished_at,
+             sub_status, details)
+            VALUES
+            (:task_id, :type, :run_id, :fab_hash, :model_ref, :connector_ref, :token,
+             :active_until, :pending_at, :starting_at, :running_at, :finished_at,
+             :sub_status, :details)
+        """
+        override_config_json = json.dumps(override_config)
+        uint64_run_id = generate_rand_int_from_bytes(RUN_ID_NUM_BYTES)
+        uint64_task_id = generate_rand_int_from_bytes(TASK_ID_NUM_BYTES)
```
panh99: Should we remove `uint64_` prefix, since uint64 ID is the actual ID?
```suggestion
        run_id = generate_rand_int_from_bytes(RUN_ID_NUM_BYTES)
        task_id = generate_rand_int_from_bytes(TASK_ID_NUM_BYTES)
```

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
```diff
@@ -32,6 +38,126 @@
...
+
+    runs = (
+        bind.execute(
+            sa.text(
+                """
+            SELECT run_id, fab_hash, pending_at, starting_at, running_at, finished_at,
+                   sub_status, details, run_type
+            FROM run
+            WHERE primary_task_id IS NULL
+            ORDER BY run_id
```
panh99: Do we need it to be ordered?

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
panh99, line 1: I noticed that the migration will raise `RuntimeError`. I wonder what this means for DB migration. Does that mean that the migration will be aborted once any exception is raised? If so, should we avoid raising `RuntimeError` and find a milder way to handle them?

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
```diff
@@ -32,6 +38,126 @@
...
 
+def _primary_task_type_from_run_type(run_type: str) -> str:
+    """Return the primary task type for the given run type."""
+    if run_type == RunType.SIMULATION:
+        return TaskType.SIMULATION
+    if run_type == RunType.SERVER_APP:
+        return TaskType.SERVER_APP
+    raise RuntimeError(
+        f"Unsupported run_type while backfilling primary tasks: {run_type}"
+    )
```
panh99, line 50: Should we simply default to `flwr-serverapp` so the DB migration won't fail? Though based on rev files, the run_type can only be `RunType.SIMULATION` or `RunType.SERVER_APP`.
```suggestion
    return TaskType.SERVER_APP
```

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
```diff
@@ -32,6 +38,126 @@
...
+        .scalars()
+        .all()
+    )
+    if in_flight_run_ids:
+        run_id_examples = ", ".join(str(run_id) for run_id in in_flight_run_ids)
+        raise RuntimeError(
+            "Cannot backfill primary tasks while runs are STARTING or RUNNING. "
+            "Stop SuperLink and retry the migration. Example run_ids: "
+            f"{run_id_examples}"
+        )
```
panh99: Should we simply fail all these runs with a details saying the run failed due to DB migration? Btw, I think this part can go to a separate private function.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
```diff
@@ -58,6 +184,15 @@ def upgrade() -> None:
...
         )
 
+    _backfill_primary_tasks()
+
+    with op.batch_alter_table("run", schema=None) as batch_op:
+        batch_op.alter_column(
+            "primary_task_id",
+            existing_type=sa.BigInteger(),
+            nullable=False,
+        )
```
panh99, line 206: Does that mean, before this point, the `primary_task_id` remains nullable? That makes sense to me.

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +171,248 @@ def test_migrations_create_fab_table(self) -> None:
         finally:
             engine.dispose()
 
+    def test_primary_task_backfill_populates_historical_runs(self) -> None:
```
panh99, line 250: Can we simplify this unit test a bit? It feels a little too long, which obscures the actual goal of the test and may make it unclear what is being verified, though I feel like this unit test is correct.

For example, could we replace the run-row creation logic with a helper function? That might make the test easier to read and keep the focus on the behavior being tested.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
danielnugraha: Yes, before backfill, it is nullable. Then during migration we backfill existing runs. After that we enfore the `primary_task_id` to be non-nullable.

framework/py/flwr/supercore/state/alembic/utils_test.py
danielnugraha: I added the test to make sure that the migration works as intended and not corrupt the deployment database. I'll refactor this to make it more readable.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
danielnugraha: The validation fails the migration early before any changes to the db, so it's retry-safe. Milder way to handle bad data is not a good thing since it shouldn't be there in the first place.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
danielnugraha: Failing the DB migration here is not a bad thing. It gives us a chance to repair bad row. Given the current code, I don't think we need to be afraid of failing here.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
danielnugraha: Removed.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
danielnugraha: The DB migration can fail here safely and be retried at a later stage. I don't think we need to fail the runs due to DB migration.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
```diff
@@ -32,8 +38,124 @@
...
+                """
+            SELECT run_id, fab_hash, pending_at, starting_at, running_at, finished_at,
+                   sub_status, details, run_type
+            FROM run
+            """
+            )
+        )
+        .mappings()
+        .all()
+    )
```
panh99:
```suggestion
    query = """
        SELECT run_id, fab_hash, pending_at, starting_at, running_at, finished_at,
            sub_status, details, run_type
        FROM run
    """
    return list(bind.execute(sa.text(query)).mappings().all())
```

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
```diff
@@ -32,8 +38,117 @@
...
+        for run in runs
+        if not run["finished_at"] and (run["starting_at"] or run["running_at"])
+    )[:10]
+    if in_flight_run_ids:
+        run_id_examples = ", ".join(str(run_id) for run_id in in_flight_run_ids)
+        raise RuntimeError(
+            "Cannot backfill primary tasks while runs are STARTING or RUNNING. "
+            "Stop SuperLink and retry the migration. Example run_ids: "
+            f"{run_id_examples}"
+        )
```
panh99: Should we simply mark all starting/running runs as stopped? With this check, it means that we cannot upgrade the DB unless no one is using the SuperLink.

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -73,6 +76,90 @@ def create_engine(self, db_name: str = "state.db") -> Engine:
...
+    def build_run_row(  # pylint: disable=too-many-arguments
+        self,
+        run_id: int,
+        *,
+        fab_id: str,
+        fab_version: str,
+        fab_hash: str,
+        pending_at: str,
+        **overrides: object,
+    ) -> dict[str, object]:
```
panh99:
```suggestion
        **overrides: Any,
    ) -> dict[str, Any]:
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+                            "model_ref": "models/existing",
+                            "connector_ref": None,
+                            "token": None,
+                            "pending_at": "2026-04-27T12:01:00+00:00",
+                            "starting_at": None,
+                            "running_at": None,
+                            "finished_at": None,
+                        }
+                    ],
+                )
```
panh99: Why this? Before the migration, the DB should not contain any tasks, right?

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+                            pending_at="2026-04-27T12:00:00+00:00",
+                            finished_at="2026-04-27T12:05:00+00:00",
+                            sub_status="failed",
+                            details="boom",
+                            federation="fed-c",
+                            flwr_aid="aid-c",
+                            bytes_sent=1,
+                            bytes_recv=2,
+                            clientapp_runtime=3.0,
+                        ),
```
panh99, line 301: Could we set them as `run_row_1/2/3`, and then in the assertions we can directly use the `run_row_1/2/3`?

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+        """Ensure historical runs get backfilled primary tasks during migration."""
+        engine = self.create_engine("primary_task_backfill.db")
+        try:
+            self.upgrade_to_revision(engine, "dee9b802b5c9")
+            with engine.begin() as connection:
+                self.insert_runs(
+                    connection,
+                    [
+                        self.build_run_row(
+                            101,
```
panh99:
```suggestion
                            run_id=101,
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+                            fab_id="publisher/pending",
+                            fab_version="1.0.0",
+                            fab_hash="fab-pending",
+                            pending_at="2026-04-27T10:00:00+00:00",
+                            federation="fed-a",
+                            flwr_aid="aid-a",
+                        ),
+                        self.build_run_row(
+                            102,
```
panh99:
```suggestion
                            run_id=102,
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+                            federation="fed-b",
+                            federation_config="{}",
+                            run_type=RunType.SIMULATION,
+                            flwr_aid="aid-b",
+                            bytes_sent=7,
+                            bytes_recv=8,
+                            clientapp_runtime=9.5,
+                        ),
+                        self.build_run_row(
+                            103,
```
panh99:
```suggestion
                            run_id=103,
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+                                   pending_at, starting_at, running_at, finished_at,
+                                   sub_status, details
+                            FROM task
+                            ORDER BY task_id
+                            """
+                        )
+                    ).mappings()
+                }
+
+            pending_primary = tasks[runs[101]["primary_task_id"]]
```
panh99:
```suggestion
            # Assert: Primary task inserted for run 101
            pending_primary = tasks[runs[101]["primary_task_id"]]
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+            self.assertEqual(pending_primary["type"], TaskType.SERVER_APP)
+            self.assertEqual(pending_primary["run_id"], 101)
+            self.assertEqual(pending_primary["pending_at"], "2026-04-27T10:00:00+00:00")
+            self.assertIsNone(pending_primary["starting_at"])
+            self.assertIsNone(pending_primary["running_at"])
+            self.assertIsNone(pending_primary["finished_at"])
+
+            simulation_primary = tasks[runs[102]["primary_task_id"]]
```
panh99:
```suggestion
            # Assert: Primary task inserted for run 102
            simulation_primary = tasks[runs[102]["primary_task_id"]]
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +252,270 @@ def test_migrations_create_fab_table(self) -> None:
...
+            self.assertEqual(
+                simulation_primary["running_at"], "2026-04-27T11:02:00+00:00"
+            )
+            self.assertEqual(
+                simulation_primary["finished_at"], "2026-04-27T11:03:00+00:00"
+            )
+            self.assertEqual(simulation_primary["sub_status"], "completed")
+            self.assertEqual(simulation_primary["details"], "done")
+
+            failed_primary = tasks[runs[103]["primary_task_id"]]
```
panh99:
```suggestion
            # Assert: Primary task inserted for run 103
            failed_primary = tasks[runs[103]["primary_task_id"]]
```

framework/py/flwr/supercore/state/alembic/utils_test.py
panh99: Could we just use `task` instead of `pending/simulation/failed_primary`? The name `task` makes more sense to me cuz we are checking if a correct primary task is inserted for each run.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_04_27_add_primary_task_to_run_and_more.py
```diff
@@ -32,8 +39,139 @@
...
+    """Return True if the historical run was STARTING or RUNNING."""
+    return not run["finished_at"] and (run["starting_at"] or run["running_at"])
+
+
+def _backfilled_primary_task_status(
+    run: RowMapping, stopped_at: str
+) -> tuple[str | None, str, str]:
+    """Return the backfilled finished_at, sub_status, and details for a run."""
+    if _is_in_flight_run(run):
+        return stopped_at, SubStatus.STOPPED, ""
```
panh99:
```suggestion
        return stopped_at, SubStatus.STOPPED, "Run stopped during server upgrade."
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +233,192 @@ def test_migrations_create_fab_table(self) -> None:
...
+                    .one()
+                )
+
+            self.assertIsNotNone(primary_task["primary_task_id"])
+            self.assertEqual(primary_task["type"], TaskType.SERVER_APP)
+            self.assertEqual(primary_task["starting_at"], "2026-04-27T13:01:00+00:00")
+            self.assertEqual(primary_task["running_at"], "2026-04-27T13:02:00+00:00")
+            self.assertTrue(primary_task["finished_at"])
+            self.assertEqual(primary_task["sub_status"], SubStatus.STOPPED)
+            self.assertEqual(primary_task["details"], "")
```
panh99:
```suggestion
            self.assertEqual(primary_task["details"], "Run stopped during server upgrade.")
```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -165,6 +233,192 @@ def test_migrations_create_fab_table(self) -> None:
...
+                    .one()
+                )
+
+            self.assertIsNotNone(primary_task["primary_task_id"])
+            self.assertEqual(primary_task["type"], TaskType.SERVER_APP)
+            self.assertEqual(primary_task["starting_at"], "2026-04-27T13:01:00+00:00")
+            self.assertEqual(primary_task["running_at"], "2026-04-27T13:02:00+00:00")
+            self.assertTrue(primary_task["finished_at"])
+            self.assertEqual(primary_task["sub_status"], SubStatus.STOPPED)
+            self.assertEqual(primary_task["details"], "Run stopped during server upgrade.")
```
panh99:
```suggestion
            self.assertEqual(
                primary_task["details"], "Run stopped during server upgrade."
            )
```

panh99 (dismissed):
LGTM!

danieljanes (dismissed):

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -179,6 +247,194 @@ def test_migrations_create_fab_table(self) -> None:
...
+            self.assertTrue(primary_task["finished_at"])
+            self.assertEqual(primary_task["sub_status"], SubStatus.STOPPED)
+            self.assertEqual(
+                primary_task["details"], "Run stopped during server upgrade."
+            )
+            self.assertEqual(
+                primary_task["run_finished_at"], primary_task["finished_at"]
+            )
+            self.assertEqual(primary_task["run_sub_status"], SubStatus.STOPPED)
+            self.assertEqual(primary_task["run_details"], "")
```
panh99:
```suggestion
            self.assertEqual(
                primary_task["run_details"], "Run stopped during server upgrade."
            )
```

danieljanes (approved):

PR #7134, opened 2026-05-11
refactor(framework): Make `LinkState` node liveness SQL portable

chongshenng:
### Description

`SqlLinkState` still uses SQLite-specific SQL in node-liveness paths:
- `delete_node` uses `IIF(...)`
- `_check_and_tag_offline_nodes` uses SQLite `strftime(..., 'unixepoch')` to derive `last_deactivated_at`.

These expressions are not portable to other SQL dialects, even after `online_until` is stored consistently as epoch seconds.

### Proposal

- Replace `IIF(...)` with standard `CASE WHEN ... THEN ... ELSE ... END`.
- `_check_and_tag_offline_nodes`:
  - Keep Python-side formatting for `last_deactivated_at`.
  - Add optimistic guards to the final `UPDATE` where `online_until <= :current_time`.
  - This preserves `last_deactivated_at` as the expiry time while preventing stale reads from marking refreshed nodes offline.
  - Add test coverage to verify the final `UPDATE` includes the timeout guards.

Merge after
* #7136


framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -769,7 +776,28 @@ def _check_and_tag_offline_nodes(self, node_ids: list[int] | None = None) -> Non
...
-        self.query(query, params)
+
+        rows = self.query(query, params)
+        if not rows:
+            return
+
+        update_query = """
+            UPDATE node
+            SET status = :offline, last_deactivated_at = :last_deactivated_at
+            WHERE node_id = :node_id AND status = :online
```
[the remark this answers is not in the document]
chongshenng: This is resolved optimistically in [91c0620](https://github.com/flwrlabs/flower/pull/7134/commits/91c06208fd918de63552b0322a2edba7b4c0165e).

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -791,7 +798,37 @@ def _check_and_tag_offline_nodes(self, node_ids: list[int] | None = None) -> Non
...
+            return
+
+        update_query = """
+            UPDATE node
+            SET status = :offline, last_deactivated_at = :last_deactivated_at
+            WHERE node_id = :node_id
+              AND status = :online
+              AND online_until = :online_until
+              AND online_until <= :current_time
+        """
```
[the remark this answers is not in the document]
chongshenng: Addressed in [85083c2](https://github.com/flwrlabs/flower/pull/7134/commits/85083c2e1d80b61e0c6b924a25b6dfee98740a31).

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -2015,6 +2015,41 @@ def test_token_expiry_does_not_overwrite_finished_completed_run(self) -> None:
...
+        def fake_query(query: str, _params: Any = None) -> list[dict[str, Any]]:
+            captured.append(query)
+            return [{"node_id": 1}]
+
+        state.query = fake_query  # type: ignore[assignment,method-assign]
+        state.delete_node("owner", 1)
+
+        self.assertTrue(captured)
+        self.assertIn("CASE", captured[0])
+        self.assertNotIn("IIF", captured[0])
```
panh99: Do we need a dedicated unit test to check if we use `IIF` in `delete_node`?

chongshenng: I am on the fence about this. Preferably, the test should inspect for any non database-specific syntax for all `SqlLinkState` methods. Let me investigate how to improve this.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -2015,6 +2015,41 @@ def test_token_expiry_does_not_overwrite_finished_completed_run(self) -> None:
...
+
+    def test_check_and_tag_offline_nodes_does_not_use_strftime(self) -> None:
```
panh99: Same question as above. I feel like a dedicated unit test that essentially only does `assert xx not in query` is a bit narrow and unclear.

If the goal is to make sure we don’t use SQLite-specific functions such as `strftime` or `IIF`, I think it might be more useful to either:

1. add a unit test that scans the full `sql_linkstate.py` file for those functions, or
2. rely on code review for this and leave a comment in `sql_linkstate` or `sql_mixin` explaining that we should avoid SQLite-only functions like `strftime` and `IIF`.

Wdyt?

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -791,7 +798,35 @@ def _check_and_tag_offline_nodes(self, node_ids: list[int] | None = None) -> Non
...
+            WHERE node_id = :node_id
+              AND status = :online
+              AND online_until <= :current_time
+        """
+        update_data = [
+            {
+                "offline": NodeStatus.OFFLINE,
+                "last_deactivated_at": self._online_until_to_utc_iso(
+                    row["online_until"]
+                ),
```
panh99: Since `self._online_until_to_utc_iso` method is a thin wrapper of `datetime.fromtimestamp`, should we directly use that function? It remains clear, right?
```suggestion
                "last_deactivated_at": datetime.fromtimestamp(
                    row["online_until"], tz=timezone.utc
                ).isoformat(),
```

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
chongshenng: Good point. I'll remove the tests and also add an instruction into `AGENTS.md`, which will help with review.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
chongshenng: Refer to comment [here](https://github.com/flwrlabs/flower/pull/7134#discussion_r3226571544).

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -769,17 +773,19 @@ def get_nodes(self, run_id: int) -> set[int]:
         # Filter node IDs by federation
         return self.federation_manager.filter_nodes(node_ids, federation)
 
+    def _online_until_to_utc_iso(self, online_until: float) -> str:
+        """Convert epoch seconds to a UTC ISO-8601 string."""
+        return datetime.fromtimestamp(online_until, tz=timezone.utc).isoformat()
+
```
panh99: We don't need this now, right?
```suggestion
```

chongshenng: Done.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7136, opened 2026-05-11
refactor(framework): Make node `online_until` float

chongshenng:
## Description

`SqlLinkState` stores `node.online_until` as epoch seconds, but the schema declares the column as `TIMESTAMP`. SQLite tolerates this mismatch because it allows numeric values in timestamp-declared columns, but stricter SQL dialects require the declared column type to match the stored value.

## Proposal

- Change `node.online_until` from `TIMESTAMP` to `Float`.
- Add an Alembic migration to convert the column type.
- Update the schema documentation.
- Add a migration test to verify `online_until` is reflected as a floating-point column.

This makes the schema match the existing `LinkState` runtime behavior and avoids timestamp/float comparison issues in other SQL dialects.

Merge before
* #7134


panh99 (approved):
LGTM!

PR #7138, opened 2026-05-12
ci(*:skip): Fix `e2e-bare` test for REST API

chongshenng:
### Description

The REST SuperNode client was sending Fleet API requests to paths with a double slash, for example `//api/v0/fleet/register-node` (such as in [this CI job](https://github.com/flwrlabs/flower/actions/runs/25699763173/job/75457189577?pr=7136#step:13:78)). SuperLink registers REST routes with a single leading slash, such as `/api/v0/fleet/register-node`, so the registration request returned 404 and caused the SuperNode to fail with `Failed to register node`.

### Proposal

Use the existing convention that REST endpoint constants include a leading slash, and join URLs by concatenating `base_url` and `api_path` directly. This produces paths like `/api/v0/fleet/register-node` instead of `//api/v0/fleet/register-node`. Also make the heartbeat endpoint consistent with the other REST endpoint constants by adding its leading slash.


danielnugraha (dismissed):

framework/py/flwr/client/rest_client/connection.py
```diff
@@ -208,7 +208,7 @@ def _request(
         # Send the request
         def post() -> requests.Response:
             return requests.post(
-                f"{base_url}/{api_path}",
+                f"{base_url.rstrip('/')}/{api_path.lstrip('/')}",
```
panh99:
```suggestion
                f"{base_url}{api_path}",
```

framework/py/flwr/client/rest_client/connection.py
panh99:
```suggestion
PATH_SEND_NODE_HEARTBEAT: str = "/api/v0/fleet/send-node-heartbeat"
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7140, opened 2026-05-12
refactor(framework): Update `EntitlementError` definition

jafermarq:
Makes it more similar to `FlowerError` by making use of both a server-side _message_ and a client-side _message_


panh99 (approved):
LGTM!

PR #7146, opened 2026-05-12
fix(framework): Handle gRPC calls without callback support in runtime version interceptor

mohammadnaseri:


framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -88,7 +88,14 @@ def _handle_completion() -> None:
...
 
-        if not call.add_callback(_handle_completion):
+        try:
+            callback_added = call.add_callback(_handle_completion)
+        except NotImplementedError:
+            if isinstance(call, grpc.RpcError):
+                _handle_completion()
+            return call
+
+        if not callback_added:
```
panh99: Should we do this to unify the control flow?
```suggestion
        callback_added = False
        try:
            callback_added = call.add_callback(_handle_completion)
        except NotImplementedError:
            pass

        if not callback_added:
```

framework/py/flwr/supercore/interceptors/runtime_version_interceptor.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Runtime version metadata interceptors."""
 
-
 from __future__ import annotations
```
```suggestion

from __future__ import annotations
```

panh99 (dismissed):
LGTM!

panh99 (approved):

PR #7148, opened 2026-05-12
fix(framework): Fix SQLite in-memory database not initializing correctly 

chongshenng:
### Description

Running `flower-superlink --database :memory: --insecure` crashes shortly after startup with:
```
ERROR :     {'query': 'SELECT task_id, type, run_id, fab_hash, model_ref, connector_ref, pending_at, starting_at, running_at, finished_at, sub_status, details FROM task WHERE ((starting_at IS NULL AND finished_at IS NULL)) ORDER BY pending_at ASC', 'data': {}, 'exception': OperationalError('(sqlite3.OperationalError) no such table: task')}
```

**Root cause:** SQLite in-memory databases are per-connection: each new connection to `sqlite:///:memory:` creates a separate, empty database. The SuperLink's gRPC server uses a thread pool, so only the thread that runs `metadata.create_all()` has the tables. All other threads open new connections to fresh, empty databases and fail when querying.

### Proposal

Use `StaticPool` when creating the SQLAlchemy engine for `sqlite:///:memory:`, ensuring all connections share the same underlying database instance. This is the same pattern already used in `alembic/utils.py` for in-memory migration testing.


panh99 (approved):
LGTM!

PR #7153, opened 2026-05-13
feat(framework): Add task message table

danielnugraha:
Adds a new `CoreState` `task_message` table for run-local, task-routed messages. The table mirrors the existing message payload fields while using `src_task_id` and `dst_task_id` instead of node routing fields.


framework/py/flwr/supercore/state/alembic/versions/rev_2026_05_13_add_task_message_table.py
```diff
@@ -0,0 +1,85 @@
...
+        sa.Column("src_task_id", sa.BigInteger(), nullable=False),
+        sa.Column("dst_task_id", sa.BigInteger(), nullable=False),
+        sa.Column("reply_to_message_id", sa.String(), nullable=True),
+        sa.Column("created_at", sa.Float(), nullable=False),
+        sa.Column(
+            "delivered_at",
+            sa.String(),
+            server_default=sa.text("''"),
+            nullable=False,
+        ),
```
panh99: Could we avoid adding a `delivered_at` field for now? We could delete the message once it has been retrieved. Adding `delivered_at` to the message table seems to make message cleanup more complicated without providing clear benefits. Should we create the task message table without this column first and see whether that keeps the design simpler?

danielnugraha: Removed.

panh99 (approved):
LGTM!

PR #7155, opened 2026-05-13
feat(framework): Fail non-finished tasks when the run expires or finishes

panh99:


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -753,8 +754,26 @@ def finish_task(self, task_id: int, sub_status: str, details: str) -> bool:
 
     def _on_task_tokens_expired(self, tasks: list[Task]) -> None:
         """Report usage when an expired task is the primary task of its run."""
```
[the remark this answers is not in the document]
panh99: Addressed in [commit](https://github.com/flwrlabs/flower/pull/7155/commits/f55e84571ccac35e2fc574e87d6f1f6e373f7113).

danielnugraha (dismissed):

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -721,17 +722,58 @@ def get_run_status(self, run_ids: set[int]) -> dict[int, RunStatus]:
...
     def finish_task(self, task_id: int, sub_status: str, details: str) -> bool:
         """Move an unfinished task to finished."""
         result = super().finish_task(task_id, sub_status, details)
         if result and self._is_primary_task(task_id):
+            with self.lock_task_store:
+                task = self.task_store.get(task_id)
+                if task is not None:
+                    self._fail_sibling_tasks(
+                        [(task.run_id, task_id)],
+                        details="Task failed because the primary task finished",
```
danielnugraha: Can we change to `details="Task failed because the run finished"` for consistency?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -721,17 +722,58 @@ def get_run_status(self, run_ids: set[int]) -> dict[int, RunStatus]:
...
+            for task in self.task_store.values():
+                if task.run_id == run_id and task.status.status != Status.FINISHED:
+                    task.finished_at = finished_at
+                    task.status.status = Status.FINISHED
+                    task.status.sub_status = SubStatus.FAILED
+                    task.status.details = details
+                    if record := self.task_token_store.pop(task.task_id, None):
+                        self.task_token_to_task_id.pop(record.token, None)
+
     def finish_task(self, task_id: int, sub_status: str, details: str) -> bool:
```
danielnugraha, line 793: Should we use `sub_status: SubStatus` instead of `str`?

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -402,6 +402,71 @@ def test_run_failed_due_to_heartbeat(self, test_method: str) -> None:
...
+        tasks = {task.task_id: task for task in state.get_tasks(run_ids=[run_id])}
+        primary_task = tasks[primary_task_id]
+        extra_task = tasks[extra_task_id]
+        assert primary_task.status.status == Status.FINISHED
+        assert primary_task.status.sub_status == SubStatus.FAILED
+        assert extra_task.status.status == Status.FINISHED
+        assert extra_task.status.sub_status == SubStatus.FAILED
+        assert extra_task.status.details == "Task failed because the run expired"
+
+    def test_sibling_tasks_inherit_finished_at_on_expiry(self) -> None:
```
danielnugraha: This seems similar to `test_primary_task_expiry_fails_unfinished_run_tasks`. Should we merge them into one unit test?

framework/py/flwr/server/superlink/linkstate/utils.py
```diff
@@ -49,6 +50,27 @@
...
+        ``:pfx_0,:pfx_1,...``).
+
+    Returns
+    -------
+    tuple[str, dict[str, Any]]
+        A comma-separated placeholder string and the corresponding parameter dict.
+    """
+    placeholders = ",".join(f":{prefix}_{i}" for i in range(len(values)))
+    params: dict[str, Any] = {f"{prefix}_{i}": v for i, v in enumerate(values)}
+    return placeholders, params
```
danielnugraha, line 70: Should we try to not return `("", {})` and return `None` if the params are not valid?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: YES! Good catch.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: I see wym. The `SubStatus` is a `StrEnum`, which is also a `str`. And using `str` can avoid many mypy warnings. But yeah, I think we should check if the string is a valid `SubStatus`.

panh99: Will update the PR.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
panh99: Yeah, good point. I will merge it.

framework/py/flwr/server/superlink/linkstate/utils.py
panh99: I see wym. Returning `None` will require us to check the return values everywhere to make mypy happy. I think it's fine as long as we use the this function right. Without this function, we'd still need to check if the input `values` is an empty list.

danielnugraha (dismissed):

jafermarq (dismissed):

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -379,6 +400,23 @@ def get_message_ids_from_run_id(self, run_id: int) -> set[str]:
...
+        if run_record is None:
+            return False
+
+        # Stop the run's primary task, which will cascade to stop all its tasks
+        primary_task_id = cast(int, run_record.run.primary_task_id)
+        if not self.finish_task(primary_task_id, SubStatus.STOPPED, ""):
+            return False
+
+        # Clean up messages and their objects related to the run
+        self.delete_messages(self.get_message_ids_from_run_id(run_id))
```
danielnugraha, line 416: Should we also clean up task messages here?

panh99: Good point! Yes, we should. However, we don't yet have a `CoreState.delete_task_messages(run_id=xxx)`. Should we do that in a separate PR?

danielnugraha (dismissed):

danielnugraha (approved):

danieljanes (approved):

PR #7156, opened 2026-05-13
feat(framework): Add corestate methods for task to task communication

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -239,6 +240,49 @@ def get_task_by_token(self, token: str) -> Task | None:
             The task if the token is valid, otherwise None.
         """
 
+    @abstractmethod
+    def store_task_message(self, message: Message) -> str | None:
+        """Store one task-addressed Message.
+
+        The source and destination task IDs are read from
+        `message.metadata.src_node_id` and `message.metadata.dst_node_id`.
```
panh99: Since this is implementation details, let's remove them. In an ABC method docstring, I think it's better to describe the contract only.
```suggestion
```

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -239,6 +240,49 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+
+        Parameters
+        ----------
+        message : Message
+            The task-addressed message to store.
+
+        Returns
+        -------
+        str | None
+            The message ID if the message was stored, otherwise None.
```
panh99: We used to have this design b/c, by that time, the message ID is generated UUID, so we need to return it here. But now the Message ID is the object ID of the `Message` object (which is pre-calculated SHA256, not a generated one), we don't need to return the message ID. If it's to check whether the message is stored successfully, I think `bool` is better.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -1972,6 +1972,37 @@ def fake_query(query: str, data: Any = None) -> list[dict[str, Any]]:
...
         self.assertNotIn("rowid", captured[0])
 
+    def test_store_task_message_rechecks_destination_status_during_insert(
+        self,
+    ) -> None:
+        """Task Message insert should fail if the destination just finished."""
+        state = self.state_factory()
+        run_id = self.task_run_id(state)
+        src_task_id = state.create_task(task_type=TaskType.SERVER_APP, run_id=run_id)
+        dst_task_id = state.create_task(task_type=TaskType.CLIENT_APP, run_id=run_id)
```
panh99: This sounds a bit strange. Shall we use agent app and model for this unit test (or serverapp + connector)?

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -2127,6 +2158,35 @@ def test_get_message_res_claim_is_unique_across_replicas(self) -> None:
...
+                lambda state: state.get_task_message(
+                    dst_task_ids=[dst_task_id], limit=1
+                )
+            )
+            claimed = [msgs for msgs in results if msgs]
+
+            # Assert
+            assert len(claimed) == 1
+            assert len(claimed[0]) == 1
+
```
panh99: That's fine. We don't really need this check, and I don't think every `LinkState` method needs a high-availability test.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -1972,6 +1972,37 @@ def fake_query(query: str, data: Any = None) -> list[dict[str, Any]]:
         self.assertIn("ORDER BY created_at, message_id", captured[0])
         self.assertNotIn("rowid", captured[0])
 
+    def test_store_task_message_rechecks_destination_status_during_insert(
+        self,
+    ) -> None:
+        """Task Message insert should fail if the destination just finished."""
```
panh99: Is this function to check if the message won't be stored if the destination task is finished? If so, should we add this check to `StateTest` instead of the sqlite-specific test class?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -71,6 +72,23 @@ def _patch_task_log_datetime_now(self, *timestamps: datetime) -> ExitStack:
...
+        self,
+        src_task_id: int,
+        dst_task_id: int,
+        run_id: int,
+        ttl: float = 60.0,
+        created_at: float | None = None,
+    ) -> Message:
+        """Create a task-addressed Message for CoreState tests."""
+        message = Message(RecordDict(), dst_task_id, "query", ttl=ttl)
+        message.metadata.__dict__["_message_id"] = str(uuid4())
```
panh99: Maybe it's fine in the unit tests, but message ID is not UUID.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
danielnugraha: I'll remove this and add it when `flwr-agentapp` and `flwr-model` logic is there.

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -71,6 +72,40 @@ def _patch_task_log_datetime_now(self, *timestamps: datetime) -> ExitStack:
...
 
+    def _create_task_message(  # pylint: disable=too-many-arguments,too-many-positional-arguments
+        self,
+        src_task_id: int,
+        dst_task_id: int,
+        run_id: int,
+        ttl: float = 60.0,
+        created_at: float | None = None,
+        group_id: str = "",
+    ) -> Message:
```
panh99: I feel like we have this helper function for unit tests also in `framework/py/flwr/supercore/corestate/utils_test.py`, right?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -71,6 +72,40 @@ def _patch_task_log_datetime_now(self, *timestamps: datetime) -> ExitStack:
...
+        """Create a task-addressed Message for CoreState tests."""
+        src_node_id = 1
+        while src_node_id in (src_task_id, dst_task_id):
+            src_node_id += 1
+        dst_node_id = src_node_id + 1
+        while dst_node_id in (src_task_id, dst_task_id):
+            dst_node_id += 1
```
panh99: Why do we compare node ID to task ID here?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
 
         self.assertIsNone(state.get_task_by_token("missing-token"))
 
+    def test_store_and_get_task_message(self) -> None:
+        """Task Messages should round-trip and be delivered once."""
+        state = self.state_factory()
+        run_id = self.task_run_id(state)
+        src_task_id = state.create_task(task_type=TaskType.SERVER_APP, run_id=run_id)
+        dst_task_id = state.create_task(task_type=TaskType.CLIENT_APP, run_id=run_id)
```
panh99: Should we use AGENT_APP and MODEL here to make it closer to the real usage?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
...
+        assert src_task_id is not None and dst_task_id is not None
+
+        message = self._create_task_message(
+            src_task_id=src_task_id,
+            dst_task_id=dst_task_id,
+            run_id=run_id + 1,
```
panh99: Why we do this?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
...
+        missing_task_id = (
+            max(src_task_id, dst_task_id, other_run_task_id, finished_task_id) + 1
+        )
+        while state.get_tasks(task_ids=[missing_task_id]):
+            missing_task_id += 1
+        invalid_messages = [
+            self._create_task_message(missing_task_id, dst_task_id, run_id),
+            self._create_task_message(src_task_id, missing_task_id, run_id),
+            self._create_task_message(src_task_id, other_run_task_id, run_id),
+            self._create_task_message(src_task_id, finished_task_id, run_id),
```
panh99: Should we also test this? Just in case of the race condition where a task finishes after task token auth, before `.store_task_message` is called.
```suggestion
            self._create_task_message(src_task_id, finished_task_id, run_id),
            self._create_task_message(finished_task_id, dst_task_id, run_id),
```

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
...
+        run_id = self.task_run_id(state)
+        other_run_id = self.other_task_run_id(state)
+        src_task_id = state.create_task(task_type=TaskType.SERVER_APP, run_id=run_id)
+        dst_task_id = state.create_task(task_type=TaskType.CLIENT_APP, run_id=run_id)
+        other_run_task_id = state.create_task(
+            task_type=TaskType.CLIENT_APP, run_id=other_run_id
+        )
+        finished_task_id = state.create_task(
+            task_type=TaskType.CLIENT_APP, run_id=run_id
+        )
```
panh99: Same as above

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -239,6 +240,49 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+        bool
+            True if the message was stored, otherwise False.
+        """
+
+    @abstractmethod
+    def get_task_message(
+        self,
+        *,
+        dst_task_ids: Sequence[int] | None = None,
+        limit: int | None = None,
```
panh99, line 270: Should we add `order_by: Literal["created_at"] | None = None`? I noticed that we sort the messages by `created_at` silently.

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
...
+            self._create_task_message(src_task_id, other_run_task_id, run_id),
+            self._create_task_message(src_task_id, finished_task_id, run_id),
+        ]
+
+        for message in invalid_messages:
+            self.assertFalse(state.store_task_message(message))
+
+        self.assertEqual(state.get_task_message(dst_task_ids=[dst_task_id]), [])
+        self.assertEqual(state.get_task_message(dst_task_ids=[other_run_task_id]), [])
+        self.assertEqual(state.get_task_message(dst_task_ids=[finished_task_id]), [])
```
panh99: Should we get all messages to see if the state is empty? We don't expect any task message to be persisted, right?
```suggestion
        self.assertEqual(state.get_task_message(), [])
```

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
...
+        expected = self._create_task_message(src_task_id, dst_task_id, run_id)
+        expired = self._create_task_message(
+            src_task_id, dst_task_id, run_id, ttl=1.0, created_at=current - 2.0
+        )
+        other_destination = self._create_task_message(
+            src_task_id, other_dst_task_id, run_id
+        )
+
+        self.assertTrue(state.store_task_message(expected))
+        self.assertTrue(state.store_task_message(expired))
```
panh99, line 660: Should we disallow storing expired message? I thought that's checked by `validate_task_message`, right?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
...
+        self.assertTrue(state.store_task_message(other_destination))
+
+        pulled = state.get_task_message(dst_task_ids=[dst_task_id])
+
+        self.assertEqual(
+            [message.metadata.message_id for message in pulled],
+            [expected.metadata.message_id],
+        )
```
panh99: Should we do this?
```suggestion
        self.assertEqual(len(pulled), 1)
        self.assertEqual(pulled[0].object_id, expected.object_id)
```

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -489,6 +524,147 @@ def test_get_task_by_token_returns_none_for_unknown_token(self) -> None:
...
+        self.assertTrue(state.store_task_message(expired))
```
panh99: It feels wrong to allow storing expired messages. Should we call `validate_task_message` to enforce this? Also, we normally test method behavior for expiration by patching `datetime.datetime`.

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -23,18 +23,19 @@
...
+    SUPERLINK_NODE_ID,
     Status,
     SubStatus,
 )
-from flwr.common.message import make_message
 from flwr.proto.task_pb2 import TaskStatus  # pylint: disable=E0611
 from flwr.supercore.constant import TaskType
 
 from . import CoreState
+from .utils_test import create_task_message as _create_task_message
```
panh99: Should we use `create_task_message` directly?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -563,107 +556,119 @@ def test_store_task_message_validates_task_relationship(self) -> None:
...
-
-        self.assertTrue(state.store_task_message(expected))
+        replacement = _create_task_message(src_task_id, dst_task_id, run_id)
+        replacement.metadata.__dict__["_message_id"] = expired.metadata.message_id
         self.assertTrue(state.store_task_message(expired))
-        self.assertTrue(state.store_task_message(other_destination))
 
+        future = current + timedelta(seconds=61)
+        with patch("flwr.supercore.corestate.in_memory_corestate.now") as memory_now:
+            with patch("flwr.supercore.corestate.sql_corestate.now") as sql_now:
```
panh99: We can just patch `datetime.datetime` instead of patching `now` in multiple places.

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -563,107 +556,119 @@ def test_store_task_message_validates_task_relationship(self) -> None:
...
-            src_task_id, other_dst_task_id, run_id
-        )
-
-        self.assertTrue(state.store_task_message(expected))
+        replacement = _create_task_message(src_task_id, dst_task_id, run_id)
+        replacement.metadata.__dict__["_message_id"] = expired.metadata.message_id
         self.assertTrue(state.store_task_message(expired))
-        self.assertTrue(state.store_task_message(other_destination))
 
+        future = current + timedelta(seconds=61)
```
panh99: Should we set a variable `msg_ttl = 60` and then use it here `future = current + timedelta(seconds=msg_ttl + 1)`?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -563,107 +556,119 @@ def test_store_task_message_validates_task_relationship(self) -> None:
...
+            created_at=current.timestamp(),
+            ttl=60.0,
         )
-        other_destination = self._create_task_message(
-            src_task_id, other_dst_task_id, run_id
-        )
-
-        self.assertTrue(state.store_task_message(expected))
+        replacement = _create_task_message(src_task_id, dst_task_id, run_id)
+        replacement.metadata.__dict__["_message_id"] = expired.metadata.message_id
```
panh99: I might not fully understand what `replacement` means here. The message ID is the computed object ID, and technically messages with different metadata will have different message IDs.

Could we remove this `replacement`? I think testing that `CoreState` does not return an expired message should be enough. We cannot actually know whether the message has been removed, since that behavior is not exposed by the ABC.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -324,48 +324,50 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+                return False
             if src_task.run_id != dst_task.run_id:
-                return None
-            if dst_task.status.status == Status.FINISHED:
-                return None
+                return False
+            if Status.FINISHED in (
+                src_task.status.status,
+                dst_task.status.status,
+            ):
```
panh99: Should we allow messages from a finished task, as long as the destination task remains non-finished?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +324,85 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+            if src_task.run_id != dst_task.run_id:
+                return False
+            if Status.FINISHED in (
+                src_task.status.status,
+                dst_task.status.status,
+            ):
+                return False
+            run_id = src_task.run_id
+
+        self._cleanup_expired_task_messages()
```
panh99: We don't need to trigger clean up here. We could just do cleanup upon message retrieval (less overhead).
```suggestion
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +324,85 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+        message_copy = message_from_proto(message_to_proto(message))
+        message_copy.metadata.__dict__["_run_id"] = run_id
+        message_copy.metadata.__dict__["_group_id"] = ""
```
panh99: Setting or overriding message metadata here seems problematic. Message integrity is guaranteed by the object ID, so the infra layer should not modify the message.

Would it be better to validate that the destination task’s `run_id` matches the source task’s `run_id`, and return `False` if they do not match? We could also log an error in that case.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +324,85 @@ def get_task_by_token(self, token: str) -> Task | None:
...
 
+    def store_task_message(self, message: Message) -> bool:
+        """Store one task-addressed Message."""
+        message_id = message.metadata.message_id
+        src_task_id = message.metadata.src_task_id
+        dst_task_id = message.metadata.dst_task_id
+        if validate_task_message(message) or src_task_id is None or dst_task_id is None:
+            return False
+
+        with self.lock_task_store:
```
panh99: Should we lock both `self.lock_task_store` and `self.lock_task_message_store` to avoid race conditions?

I noticed that we first lock the task table, check the task status, then release the lock before locking the task message table. The task status could change in that gap.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
panh99: Btw, no need to make a copy of the message. It's totally fine to store the original object.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -352,6 +435,14 @@ def _cleanup_expired_task_tokens_locked(self) -> None:
         if expired_tasks:
             self._on_task_tokens_expired(expired_tasks)
 
+    def _cleanup_expired_task_messages(self) -> None:
+        """Remove expired task-addressed Messages."""
+        current = now().timestamp()
+        with self.lock_task_message_store:
+            for message_id, message in list(self.task_message_store.items()):
+                if message.metadata.created_at + message.metadata.ttl <= current:
+                    del self.task_message_store[message_id]
```
panh99: Should we also cleanup the messages whose `dst_task_id` is finished here, and rename the method to, e.g., `_cleanup_invalid_task_messages()`?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +324,85 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+            return []
+
+        with self.lock_task_store:
+            self._cleanup_expired_task_tokens_locked()
+        self._cleanup_expired_task_messages()
+
+        dst_task_id_set = set(dst_task_ids) if dst_task_ids is not None else None
+        selected_messages: list[Message] = []
+        current = now().timestamp()
+        with self.lock_task_message_store:
```
panh99: Let's acquire both locks to avoid race conditions, though unlikely to happen in practice.
```suggestion
        with self.lock_task_store, self.lock_task_message_store:
            self._cleanup_expired_task_tokens_locked()
            self._cleanup_expired_task_messages()

            dst_task_id_set = set(dst_task_ids) if dst_task_ids is not None else None
            selected_messages: list[Message] = []
            current = now().timestamp()
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +324,85 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+            message_ids = sorted(
+                self.task_message_store.keys(),
+                key=lambda msg_id: self.task_message_store[msg_id].metadata.created_at,
+            )
```
panh99: Why sorting the messages here? Should we only sort it when `order_by` is set? I think it might be costly to sort the whole message table whenever getting a message.

We could follow the SQL logic here:
1. Apply all query conditions.
2. Sort by the required `order_by`.
3. Return the first `limit` rows.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +324,85 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+            for message_id in message_ids:
+                message = self.task_message_store[message_id]
+                if dst_task_id_set is not None:
+                    if message.metadata.dst_task_id not in dst_task_id_set:
+                        continue
+                if message.metadata.created_at + message.metadata.ttl <= current:
+                    del self.task_message_store[message_id]
+                    continue
```
panh99: This check seems not necessary since the `self._cleanup_expired_task_messages()` call above.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -406,6 +415,118 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+                    RETURNING message_id
+                    """,
+                    message_dict,
+                )
+            except IntegrityError:
+                return False
+            if not inserted:
+                return False
+
+        return True
```
```suggestion
                )
                return bool(inserted)
            except IntegrityError:
                return False
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -406,6 +415,115 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+            return []
+        if dst_task_ids is not None and not dst_task_ids:
+            return []
+
+        with self.session():
+            self._cleanup_expired_task_tokens()
+            self._cleanup_expired_task_messages()
+            rows = self._claim_task_message_rows(dst_task_ids, limit)
+
+        rows.sort(key=lambda row: row["created_at"])
```
panh99: Should we use `ORDER BY` keyword instead of sorting it via Python? Plus, `ORDER BY` need to work with `LIMIT`, otherwise we may likely select wrong ones.

panh99: Btw, could you add `self.select_lock_sql` in the sql query when using `ORDER BY` + `LIMIT`, like [this](https://github.com/flwrlabs/flower/blob/main/framework/py/flwr/server/superlink/linkstate/sql_linkstate.py#L320)?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -502,3 +630,53 @@ def task_from_row(row: dict[str, Any]) -> Task:
...
+    if row["error"] is not None:
+        error = error_from_proto(ProtoError.FromString(row["error"]))
+
+    metadata = Metadata(
+        run_id=int64_to_uint64(row["run_id"]),
+        message_id=row["message_id"],
+        src_node_id=SUPERLINK_NODE_ID,
+        dst_node_id=SUPERLINK_NODE_ID,
+        reply_to_message_id=row["reply_to_message_id"] or "",
+        group_id="",
```
```suggestion
        group_id="",  # Task messages don't have this field for now
```

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -85,11 +85,6 @@
 class SqlLinkState(LinkState, SqlCoreState):  # pylint: disable=R0904
     """SQLAlchemy-based LinkState implementation."""
 
-    @property
-    def select_lock_sql(self) -> str:
-        """Return the SQL clause for row-locking, which is overridable by subclasses."""
-        return ""
-
```
panh99, line 93: Why removing this?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +325,93 @@ def get_task_by_token(self, token: str) -> Task | None:
...
 
+    def store_task_message(self, message: Message) -> bool:
+        """Store one task-addressed Message."""
+        message_id = message.metadata.message_id
+        if validate_task_message(message):
+            return False
+        src_task_id = message.metadata.src_task_id
+        dst_task_id = message.metadata.dst_task_id
+        assert src_task_id is not None
+        assert dst_task_id is not None
```
panh99: Should we avoid `assert` in the production code?
```suggestion
        src_task_id = cast(int, message.metadata.src_task_id)
        dst_task_id = cast(int, message.metadata.dst_task_id)
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +325,93 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+            self._cleanup_invalid_task_messages_locked(now().timestamp())
+            src_task = self.task_store.get(src_task_id)
+            dst_task = self.task_store.get(dst_task_id)
+            if src_task is None or dst_task is None:
+                return False
+            if src_task.run_id != dst_task.run_id:
+                log(
+                    ERROR,
+                    "Source task %d and destination task %d belong to "
+                    "different runs.",
```
```suggestion
                    "Cannot store message: source task %d and destination task %d "
                    "belong to different runs.",
```

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
danielnugraha: It's needed in corestate.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +325,93 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+                    src_task_id,
+                    dst_task_id,
+                )
+                return False
+            if message.metadata.run_id != src_task.run_id:
+                log(
+                    ERROR,
+                    "Task Message %s run ID %d does not match task run ID %d.",
```
panh99: For operator-facing logs, it will be better to log useful info that helps debug. Since we will drop the message here, logging the message ID may not work; instead, we can log the task ID so we know who's sending the wrong message.
```suggestion
                    "Cannot store message for task %s: message run ID %d does not match task run ID %d.",
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +325,91 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+        if limit == 0:
+            return []
+        if dst_task_ids is not None and not dst_task_ids:
+            return []
+
+        with self.lock_task_store, self.lock_task_message_store:
+            self._cleanup_expired_task_tokens_locked()
+            current = now().timestamp()
+            self._cleanup_invalid_task_messages_locked(current)
+            dst_task_id_set = set(dst_task_ids) if dst_task_ids is not None else None
```
panh99, line 404:
```suggestion

            # Filter by dst_task_id
            dst_task_id_set = set(dst_task_ids) if dst_task_ids is not None else None
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -320,6 +325,93 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+
+            # Filter by dst_task_id
+            dst_task_id_set = set(dst_task_ids) if dst_task_ids is not None else None
+            candidate_ids = [
+                message_id
+                for message_id, message in self.task_message_store.items()
+                if dst_task_id_set is None
+                or message.metadata.dst_task_id in dst_task_id_set
+            ]
+            if order_by == "created_at":
```
panh99, line 413:
```suggestion

            # Apply requested sort order
            if order_by == "created_at":
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -352,6 +444,20 @@ def _cleanup_expired_task_tokens_locked(self) -> None:
         if expired_tasks:
             self._on_task_tokens_expired(expired_tasks)
 
+    def _cleanup_invalid_task_messages_locked(self, current: float) -> None:
+        """Remove expired Messages and Messages for invalid destination tasks."""
+        for message_id, message in list(self.task_message_store.items()):
+            dst_task_id = message.metadata.dst_task_id
+            dst_task = (
+                self.task_store.get(dst_task_id) if dst_task_id is not None else None
+            )
```
```suggestion
            dst_task = self.task_store.get(message.metadata.dst_task_id)
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -406,6 +420,125 @@ def get_task_by_token(self, token: str) -> Task | None:
...
         return task_from_row(rows[0])
 
+    def store_task_message(self, message: Message) -> bool:
+        """Store one task-addressed Message."""
+        if validate_task_message(message):
+            return False
+
+        with self.session():
+            self._cleanup_expired_task_tokens()
+            self._cleanup_invalid_task_messages()
```
```suggestion
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -406,6 +420,124 @@ def get_task_by_token(self, token: str) -> Task | None:
...
+                    SELECT
+                        :message_id, :run_id, :src_task_id, :dst_task_id,
+                        :reply_to_message_id, :created_at, :ttl, :message_type,
+                        :content, :error
+                    FROM task AS src
+                    JOIN task AS dst
+                        ON dst.task_id = :dst_task_id
+                        AND dst.run_id = src.run_id
+                    WHERE src.task_id = :src_task_id
+                        AND :run_id = src.run_id
```
```suggestion
                    WHERE src.task_id = :src_task_id
                        AND src.run_id = :run_id
                        AND dst.run_id = :run_id
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -433,6 +558,21 @@ def _cleanup_expired_task_tokens(self) -> None:
...
+        """Remove expired Messages and Messages for invalid destination tasks."""
+        self.query(
+            """
+            DELETE FROM task_message
+            WHERE (created_at + ttl) <= :current
+                OR NOT EXISTS (
+                    SELECT 1 FROM task AS dst
+                    WHERE dst.task_id = task_message.dst_task_id
+                        AND dst.finished_at IS NULL
+                )
```
panh99: I think this cleanup query is too heavy for the value it adds.

IIUC, `get_task_message` is only callable by a non-finished task (enforced by auth), and it only pulls messages targeting that task. Given that, the destination-task-finished check here feels unnecessarily expensive, since it effectively scans task_message and rechecks against task for little practical benefit.

I’d prefer we keep the expiry cleanup and drop the finished-destination validation from this path. The auth boundary already guarantees the important part. Wdyt?

```suggestion
            WHERE (created_at + ttl) <= :current
```

panh99 (dismissed):
LGTM!

danieljanes (approved):

PR #7161, opened 2026-05-13
refactor(framework): Introduce select lock clause hook

chongshenng:
### Description

Limited message claiming can update more rows than intended when the `UPDATE` uses a same-table subquery with `LIMIT`.

### Proposal

Use a Common Table Expression (CTE) to select the limited candidate rows first, then update only those rows. This makes the candidate set stable before the `UPDATE` runs.

### DIFF
--- a/framework/py/flwr/server/superlink/linkstate/linkstate_test.py
+++ b/framework/py/flwr/server/superlink/linkstate/linkstate_test.py
@@ -1972,6 +1972,26 @@ class SqlInMemoryStateTest(StateTest, unittest.TestCase):
         self.assertIn("ORDER BY created_at, message_id", captured[0])
         self.assertNotIn("rowid", captured[0])
 
+    def test_message_ins_claim_can_append_select_lock_clause(self) -> None:
+        """Message claiming can append a subclass-provided row-locking clause."""
+        state = self.state_factory()
+        captured: list[str] = []
+
+        # pylint: disable-next=unused-argument
+        def fake_query(query: str, data: Any = None) -> list[dict[str, Any]]:
+            captured.append(query)
+            return []
+
+        state.query = fake_query  # type: ignore[method-assign]
+        state._claim_message_ins_rows(1, 3)  # pylint: disable=protected-access
+        self.assertNotIn("FOR TEST LOCK", captured[0])
+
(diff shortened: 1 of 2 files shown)


framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 """Tests all LinkState implemenations have to conform to."""
+
```
chongshenng:
```suggestion
```

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -324,21 +326,39 @@ def _claim_message_ins_rows(
...
+            # otherwise re-evaluate same-table subqueries while UPDATE scans rows.
+            candidate_cte = f"""
+                WITH candidate_message_ins AS (
                     SELECT message_id
                     FROM message_ins
                     WHERE {common_condition}
                     ORDER BY created_at, message_id
+                    {select_lock_sql}
                     LIMIT :limit
                 )
```
panh99: Is `_claim_message_ins_select_lock_clause` just `FOR UPDATE SKIP LOCKED`? If so, can we create a property `select_lock_clause` which return `""` in this class, while in the subclass, we can override it and return `"FOR UPDATE SKIP LOCKED"`? And then we can prolly do:

```python
class SqlLinkState(LinkState, SqlCoreState):

    @property
    def select_lock_sql(self) -> str:
        return ""

    ...

    def _claim_message_ins_rows(
        self, node_id: int, limit: int | None
    ) -> list[dict[str, Any]]:
            ...
            candidate_cte = f"""
                WITH candidate_message_ins AS (
                   SELECT message_id
                   FROM message_ins
                   WHERE {common_condition}
                   ORDER BY created_at, message_id
                    {self.select_lock_sql}
                   LIMIT :limit
               )
```

And in subclass:

```python
class DifferentSqlLinkState(SqlLinkState):

    @property
    def select_lock_sql(self) -> str:
        return "FOR UPDATE SKIP LOCKED"
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7169, opened 2026-05-14
fix(framework): Make SQL object store cleanup concurrency-safe

psfoley:
This PR should be merged after https://github.com/flwrlabs/flower/pull/7172


framework/py/flwr/supercore/state/alembic/versions/rev_2026_05_14_add_object_refcount_check.py
```diff
@@ -0,0 +1,68 @@
...
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# pylint: disable=no-member
+
+# revision identifiers, used by Alembic.
+revision: str = "93df222d77e1"
+down_revision: str | Sequence[str] | None = "154388986f5f"
```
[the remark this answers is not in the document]
psfoley: https://github.com/flwrlabs/flower/pull/7172 is a dependency for this reason

framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Flower SQLAlchemy-based ObjectStore implementation."""
```
chongshenng, line 16:
```suggestion


```

framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -90,6 +82,37 @@ def preregister(self, run_id: int, object_tree: ObjectTree) -> list[str]:
...
+                            "RETURNING child_id",
+                            {"parent_id": obj_id, "child_id": cid},
+                        )
+                        if edge_rows:
+                            self.query(
+                                "UPDATE objects SET ref_count = ("
+                                "SELECT COUNT(*) FROM object_children "
+                                "WHERE child_id = :object_id"
+                                ") WHERE object_id = :object_id",
+                                {"object_id": cid},
```
panh99: The transaction already guarantees that child-edge insertion and ref-count update commit or roll back together. Concurrent updates of `ref_count = ref_count + 1` are also safe because the increment happens inside the SQL engine, not in Python. Since object_children has a primary key and only the transaction that inserts a new parent inserts its child edges, duplicate edge insertion should already be prevented. Therefore, replacing the increment with a SELECT COUNT(*) subquery seems unnecessary. Wdyt?

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -202,11 +185,35 @@ def delete_objects_in_run(self, run_id: int) -> None:
                 # and every message object must have a `ref_count` of 0
                 if object_entry.ref_count == 0:
                     # Delete the message object and its unreferenced descendants
-                    self.delete(object_id)
+                    self._delete_unreferenced(object_id, respect_run_refs=True)
 
             # Remove the run from the mapping
             del self.run_objects_mapping[run_id]
 
+    def _delete_unreferenced(self, object_id: str, *, respect_run_refs: bool) -> None:
```
panh99: I’m not sure we should make deletion respect run refs here.

My understanding is that we don’t currently support root objects (only `Message` for now) being shared across multiple runs. A message includes `metadata.run_id`, and the metadata is part of the serialized object content used to derive the object ID. So two messages from different runs should naturally have different object IDs.

Shared non-root objects should already be protected by `ref_count`, because they are referenced through parent/child object relationships. In other words, `run_objects` should not be needed as an additional lifetime reference for the current model.

I’m also a bit concerned that adding `respect_run_refs` gives a misleading sense of safety. It only protects the `delete_objects_in_run` path. A direct `.delete(object_id)` still uses `respect_run_refs=False`, so it can still delete an object even if that object is registered to multiple runs. If we ever decide to support root objects shared by multiple runs, we probably need to define that as an explicit `ObjectStore` invariant and enforce it consistently across all deletion paths, not only during run cleanup.

For now, I’d suggest keeping deletion based on `ref_count` only, and not treating run refs as lifetime refs until we intentionally support shared root objects across runs.

panh99 (commented):
Thanks for working on this! I like the intent of making `ObjectStore` behavior more robust, especially around consistency checks and concurrent SQL access.

I left two comments on places where I’m not sure the added complexity is necessary for the current `ObjectStore` model. Happy to be corrected if there’s a concrete race/failure mode these changes are meant to cover.

framework/py/flwr/supercore/object_store/in_memory_object_store.py
psfoley: I tested this locally, and you're right. For actual `Message` objects, root object IDs differ across runs because `metadata.run_id` is part of the serialized message content. Shared non-root objects can still happen, but those are already protected by `ref_count`. The only case where `run_objects` changes lifetime behavior is a raw `ObjectStore` case where the exact same root object is preregistered in multiple runs, and we don’t currently support that consistently because direct `delete(object_id)` still ignores run refs. I'm pushing a change that removes the `respect_run_refs` path and the shared-root test.

framework/py/flwr/supercore/object_store/sql_object_store.py
psfoley: Thanks for pointing this out. Since only a newly inserted parent inserts child edges, and the insert/refcount update happen in one transaction, `ref_count = ref_count + 1` is enough and is safe inside SQL. I'll change this back from the `SELECT COUNT(*)` recomputation. I'll also remove `ON CONFLICT DO NOTHING` for child edges and add an explicit duplicate-child check so malformed trees don’t get silently collapsed.

framework/py/flwr/supercore/object_store/object_store_test.py
```diff
@@ -14,10 +14,11 @@
 # ==============================================================================
 """Tests for ObjectStore."""
 
-
 import tempfile
```
panh99, line 18:
```suggestion

import tempfile
```

framework/py/flwr/supercore/object_store/in_memory_object_store.py
```diff
@@ -202,11 +185,35 @@ def delete_objects_in_run(self, run_id: int) -> None:
                 # and every message object must have a `ref_count` of 0
                 if object_entry.ref_count == 0:
                     # Delete the message object and its unreferenced descendants
-                    self.delete(object_id)
+                    self._delete_unreferenced(object_id)
 
             # Remove the run from the mapping
             del self.run_objects_mapping[run_id]
 
+    def _delete_unreferenced(self, object_id: str) -> None:
```
panh99: Just curious, do we need to introduce this method? Is it better than directly calling `.delete` in the `.delete_objects_in_run` method?

framework/py/flwr/supercore/object_store/object_store_test.py
```diff
@@ -397,6 +399,23 @@ def test_in_memory_does_not_create_alembic_version(self) -> None:
         ).get_table_names()
         self.assertNotIn("alembic_version", table_names)
 
+    def test_preregister_rejects_new_children_for_existing_object_id(self) -> None:
```
panh99, line 402: This makes total sense to me. I wonder why we don't need to update the `InMemoryObjectStore` for this.

panh99 (approved):
LGTM!

PR #7172, opened 2026-05-14
fix(framework): Merge flwr Alembic heads

psfoley:


chongshenng (dismissed):

chongshenng:
This PR linearizes the Alembic migration graph instead of adding a new script to merge the heads. This is safe because the migration scripts are not part of the official `flwr` release.

panh99 (approved):
LGTM!

PR #7173, opened 2026-05-14
fix(framework): Resolve StopRun race condition

psfoley:
Behavior before PR:
image
Behavior after PR:
image


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """In-memory LinkState implementation."""
 
-
 import threading
```
panh99, line 18:
```suggestion

import threading
```

panh99 (commented):
Thanks a lot for working on this and for digging into the race condition. I agree the race is real, and the PR does address it, but I think the current shape introduces coupling/complexity that we should avoid. To be fair, we probably should have made this design principle explicit earlier.

The main concern is that it relies more heavily on `ObjectStore` and `LinkState` sharing the same DB. That is currently a temporary implementation detail, not a design contract. Until we are confident we want to merge `ObjectStore` into `LinkState`/`NodeState`, we should avoid making `ObjectStore` aware of `LinkState`-owned concepts like run/task status or locking `LinkState` tables from `SqlObjectStore`. The two operate relatively independently today, which unfortunately causes some race conditions. That said, I’d rather fix those in `CoreState`/`LinkState`, without making `ObjectStore` depend on `LinkState`-owned run/task tables.

The second concern is `LinkState`/`NodeState` divergence. This PR adds more LinkState-specific lifecycle logic and SQL-specific helpers, which makes the two state abstractions drift further apart. Where possible, I think we should move shared semantics into `CoreState` and keep `LinkState`/`NodeState` aligned.

A cleaner approach might be:
1. For `CreateTask`, pass `requesting_task_id` into `CoreState.create_task`, and atomically reject task creation if the requesting task is already finished. This fixes the parent-task race at the `CoreState` level and benefits both `LinkState` and `NodeState`.
2. For `store_message_ins` / `store_message_res`, check run status directly inside the `LinkState` transaction. If the run is already finished, reject storing the message and then trigger `ObjectStore` cleanup explicitly in the servicer (not in the `LinkState`).
3. Keep `ObjectStore` lifecycle-agnostic (we don't need to change it for this purpose, IMHO). It can still support explicit cleanup like `delete_objects_in_run(run_id)`, but it should not decide whether a run is writable by inspecting LinkState tables.

Separately, once finishing the primary task atomically finishes all tasks in the run (by [PR#7155](https://github.com/flwrlabs/flower/pull/7155), `StopRun` can simply finish the primary task and perform cleanup. Then the race is handled by the ordering:

- If `StopRun` wins first: primary task becomes finished; subsequent `store_message` sees finished run inside its transaction; it rejects and triggers cleanup.
- If `store_message` wins first: message/object registration succeeds while run is still writable; `StopRun` later finishes the primary task; `StopRun` cleanup removes run messages and objects.

So I think the fix should be more focused: encode the lifecycle invariants in `CoreState`, avoid strengthening `ObjectStore`-`LinkState` coupling, and avoid further `LinkState`/`NodeState` divergence.

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -208,6 +208,12 @@ def push_messages(
         objects_to_push |= set(store.preregister(run_id, object_tree))
     # Store Message in State
     message_id: str | None = state.store_message_res(message=msg)
+    if (
+        message_id is None
+        and state.get_run_status({run_id})[run_id].status == Status.FINISHED
+    ):
```
panh99, line 217: No need to check status here, I think. As long as the message is rejected, I need to clean up registered object IDs.
```suggestion
    if message_id is None:
```

panh99: Btw, will deleting preregistered objects here cause any new race conditions? Like, when two ServerApps try to push messages of the same content, but one fails? I guess not?

framework/py/flwr/server/superlink/fleet/message_handler/message_handler_test.py
```diff
@@ -119,3 +122,38 @@ def test_push_messages() -> None:
     state.store_message_res.assert_called_once()
     state.get_message_res.assert_not_called()
     state.store_traffic.assert_called_once()
+
+
+def test_push_messages_cleans_up_only_failed_reply_objects() -> None:
+    """Test push_messages cleanup is scoped to the failed reply."""
```
panh99: I think the wording is a bit misleading, b/c we previously didn’t clean up on failure at all. Wdyt?

```suggestion
def test_push_messages_cleans_up_failed_message_objects() -> None:
    """Test push_messages cleanup pregistered objects when it fails to store the message."""
```

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -124,18 +125,60 @@ def _is_primary_task(self, task_id: int) -> bool:
...
+                    )
+                if self._is_finished_run(run_id):
+                    raise RuntimeError(f"Run {run_id} is finished.")
+                if requesting_task_id is not None:
+                    requesting_task = self.task_store.get(requesting_task_id)
+                    if (
+                        requesting_task is None
+                        or requesting_task.status.status == Status.FINISHED
+                    ):
+                        return None
```
panh99: Could we move this check to `CoreState`? And remove the `.create_task` override here. It's better to avoid overriding `CoreState` methods if possible.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -768,9 +863,12 @@ def get_serverapp_context(self, run_id: int) -> Context | None:
...
         """Set the context for the specified `run_id`."""
-        if run_id not in self.run_ids:
-            raise ValueError(f"Run {run_id} not found")
-        self.contexts[run_id] = context
+        with self.lock:
+            if run_id not in self.run_ids:
+                raise ValueError(f"Run {run_id} not found")
+            if self._is_stopped_run(run_id):
+                raise ValueError(f"Run {run_id} not found or already stopped")
+            self.contexts[run_id] = context
```
panh99, line 811: Do we need to update this method? Is to avoid calling `set_serverapp_context` in `StartRun` when `StopRun` is also called? If so, I think it's fine to keep it as it is.
The method is only used in two places:
1. `ControlServicer.StartRun`: it's used to init the context. Even if the run is stopped, it's also fine to init the context.
2. `ServerAppIoServicer.PushTaskOutput`: it's already guarded by `.finish_task`, which is atomic.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -124,18 +125,60 @@ def _is_primary_task(self, task_id: int) -> bool:
             return False
         return self.run_ids[task.run_id].run.primary_task_id == task_id
 
+    def _is_finished_run(self, run_id: int) -> bool:
+        """Return True if the run has finished."""
+        return self._get_run(run_id).status.status == Status.FINISHED
+
+    def _is_stopped_run(self, run_id: int) -> bool:
+        """Return True if the run has been stopped."""
+        return self._get_run(run_id).status.sub_status == SubStatus.STOPPED
```
panh99: I think it's better not to treat finished/stopped run differently. "stopped" is a substatus, and a stopped run is also a finished run.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -378,6 +433,46 @@ def get_message_ids_from_run_id(self, run_id: int) -> set[str]:
...
+                    return False
+                primary_task_id = cast(int, self.run_ids[run_id].run.primary_task_id)
+                primary_task = self.task_store[primary_task_id]
+                if primary_task.status.sub_status == SubStatus.STOPPED:
+                    if not update_success:
+                        return False
+                else:
+                    primary_task.status.sub_status = SubStatus.STOPPED
+                    primary_task.status.details = ""
+                    update_success = True
```
panh99: Let's just call `.finish_task` here for now. Another PR will handle this.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -572,8 +667,8 @@ def create_run(
         """Create a new run."""
         task_type = primary_task_type_from_run_type(run_type)
 
-        with self.lock:
-            with self.lock_task_store:
+        with self.lock_task_store:
+            with self.lock:
```
panh99: Should we do this?
```suggestion
        with self.lock_task_store, self.lock:
```

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -176,6 +176,47 @@ def test_create_run_creates_primary_task(self) -> None:
...
 
+    def test_create_task_rejects_missing_run(self) -> None:
+        """Creating a task for an unknown run should fail."""
+        state = self.state_factory()
+
+        with self.assertRaisesRegex(
+            RuntimeError,
+            "Run 42 not found. create_task requires an existing run.",
+        ):
+            state.create_task(task_type="flwr-model", run_id=42)
```
panh99: Thanks for adding this check! Could we remove it, though? We used to have this check, but later decided to remove the run ID validation because the run ID is already guaranteed by the token authentication layer. Validating run ID would require `LinkState`/`NodeState` to override `CoreState.create_task`

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -160,6 +160,7 @@ def create_task(  # pylint: disable=too-many-arguments,too-many-positional-argum
         fab_hash: str | None = None,
         model_ref: str | None = None,
         connector_ref: str | None = None,
+        requesting_task_id: int | None = None,
```
panh99, line 163: Should we make it a required argument, and place it in the top?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -195,6 +196,20 @@ def create_task(  # pylint: disable=too-many-arguments,too-many-positional-argum
...
+                        UPDATE task
+                        SET task_id = task_id
+                        WHERE task_id = :requesting_task_id
+                        AND finished_at IS NULL
+                        RETURNING task_id
+                        """,
+                        {"requesting_task_id": uint64_to_int64(requesting_task_id)},
+                    )
+                    if not requesting_task_rows:
+                        return None
```
panh99: Wow, this is a good idea to start the exclusive transaction. I wonder if it's more efficient to make this a subquery of the `insert_query`? For example:

```sql

    INSERT INTO task
    (task_id, type, run_id, fab_hash, model_ref, connector_ref, token,
     active_until, pending_at, starting_at, running_at, finished_at,
     sub_status, details)
    SELECT
     :task_id, :type, :run_id, :fab_hash, :model_ref, :connector_ref, :token,
     :active_until, :pending_at, :starting_at, :running_at, :finished_at,
     :sub_status, :details
    WHERE EXISTS (
        SELECT 1
        FROM task
        WHERE task_id = :requesting_task_id AND finished_at IS NULL
    )
    RETURNING task_id;

```

framework/py/flwr/supercore/corestate/sql_corestate.py
panh99: Ah, nvrm. I noticed that there's no requesting task ID when creating the primary task.

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -147,6 +147,11 @@ def PushMessages(
             objects_to_push |= set(store.preregister(run_id, object_tree))
             # Store message
             message_id: str | None = state.store_message_ins(message=message)
+            if (
+                message_id is None
+                and state.get_run_status({run_id})[run_id].status == Status.FINISHED
+            ):
```
panh99, line 157: Same as above.
```suggestion
            if message_id is None:
```

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -117,6 +118,73 @@ def federation_manager(self) -> FederationManager:
...
+
+        run_rows = self.query(
+            """
+            SELECT run_id, federation, primary_task_id
+            FROM run
+            WHERE run_id = :run_id
+            """,
+            {"run_id": sint64_run_id},
+        )
+        return dict(run_rows[0]) if run_rows else None
```
panh99:
```suggestion
        condition = ""
        if require_unfinished:
            condition = """
            AND EXISTS (
                SELECT 1
                FROM task
                WHERE task.task_id = run.primary_task_id
                AND task.finished_at IS NULL
            )
            """
        rows = self.query(
            f"""
            UPDATE run
            SET run_id = run_id
            WHERE run_id = :run_id {condition}
            RETURNING run_id, federation, primary_task_id
            """,
            {"run_id": uint64_to_int64(run_id)},
        )
        return dict(rows[0]) if rows else None
```

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
psfoley: Done!

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
psfoley: @panh99 I tried simplifying to what you recommended, but it seems there are valid cases when `message_id is None` that preregistered objects should not be deleted when a run is still in progress. It allows for deletion in the case that the parent instruction message was missing ([this resulted in a failed CI test](https://github.com/flwrlabs/flower/actions/runs/26189007472/job/77052224848)). The other one is what you called out - if duplicate messages are sent, this could result in valid objects being deleted as a consequence.

panh99: I see. Just to clarify, I think that if we cannot store the message, we should always delete all pre-registered objects. The alternative would be to wait until the run is finished, but that feels a bit late.

The unit tests are failing because we did not set this up correctly. For example, `test_successful_push_messages_if_running` tries to store a reply message without patching `LinkState.store_message_res` or storing a valid instruction message first. This is incorrect, but the test still passes because it only checks whether the objects are registered.

I’m slightly unsure about duplicate messages, though. This could happen if a SuperNode pushes a message, does not receive a response, and then retries. Given that possibility, do you think it would be better to only perform the cleanup once the run has finished?

panh99 (dismissed):

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
psfoley: > I’m slightly unsure about duplicate messages, though. This could happen if a SuperNode pushes a message, does not receive a response, and then retries. Given that possibility, do you think it would be better to only perform the cleanup once the run has finished?

That's exactly the scenario I want to prevent from happening as that could lead to a premature deletion (if before the run has finished). I agree there's cases where we should delete earlier, but that could be added in a separate PR

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -208,6 +211,14 @@ def push_messages(
...
+        for object_tree in request.message_object_trees:
+            store.delete(object_tree.object_id)
+            for tree_node in iterate_object_tree(object_tree):
+                objects_to_push.discard(tree_node.object_id)
```
panh99:
```suggestion
    # This is temporary. We should consider a more robust cleanup
    # mechanism that protects duplicate messages from premature deletion.
    # Once that is in place, we can remove the run status check below.
    if (
        message_id is None
        and state.get_run_status({run_id})[run_id].status == Status.FINISHED
    ):
        # The request currently contains only one message and its object tree.
        store.delete(request.message_object_trees[0].object_id)
        objects_to_push.clear()
```

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -147,6 +148,13 @@ def PushMessages(
...
+                store.delete(object_tree.object_id)
+                for tree_node in iterate_object_tree(object_tree):
+                    objects_to_push.discard(tree_node.object_id)
```
panh99, line 160:
```suggestion
            # This is temporary. We should consider a more robust cleanup
            # mechanism that protects duplicate messages from premature deletion.
            # Once that is in place, we can remove the run status check below.
            if (
                message_id is None
                and state.get_run_status({run_id})[run_id].status == Status.FINISHED
            ):
                store.delete(object_tree.object_id)
                for tree_node in iterate_object_tree(object_tree):
                    objects_to_push.discard(tree_node.object_id)
```

panh99 (dismissed):

panh99 (dismissed):

danieljanes (dismissed):

danieljanes (approved):

PR #7178, opened 2026-05-15
feat(framework): Add `AppIo` task to task communication

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/servicers/appio_servicer.py
```diff
@@ -125,20 +126,52 @@ def PushTaskMessage(
...
+        task = get_authenticated_task()
+        if not request.HasField("message"):
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, "`message` is required.")
+
+        message_proto = request.message
+        if not message_proto.HasField("metadata"):
+            context.abort(
+                grpc.StatusCode.FAILED_PRECONDITION, "`Message.metadata` is required."
+            )
+
```
panh99: No need to have these validation.
```suggestion
```

framework/py/flwr/supercore/servicers/appio_servicer.py
```diff
@@ -125,20 +126,52 @@ def PushTaskMessage(
...
+        message = message_from_proto(request.message)
+        message.metadata.__dict__["_run_id"] = task.run_id
+        message.metadata.__dict__["_src_node_id"] = SUPERLINK_NODE_ID
+        message.metadata.__dict__["_dst_node_id"] = SUPERLINK_NODE_ID
+        message.metadata.src_task_id = task.task_id
+        message.metadata.dst_task_id = request.message.metadata.dst_task_id
+        if message.metadata.message_id == "":
+            message.metadata.__dict__["_message_id"] = message.object_id
```
panh99: I don’t think we should modify any message metadata on the server side. The server should verify the integrity of the message and drop invalid messages, but it should not mutate the message metadata.

More importantly, the message ID is the message’s object ID, which is computed as a SHA-256 hash based on the metadata and content. Overriding any metadata fields may break that relationship and compromise the integrity guarantees.

Normally, this metadata should be set on the client side by Flower code. For example, `GrpcGrid` already sets some of these fields. For task messages, I’d also prefer to keep metadata assignment on the client side rather than doing it in the server-side servicer.

panh99: Maybe we could introduce a process-local, write-once `TaskIdentity` class. The caller process would initialize it once after the first server request/handshake, and message construction could then use it to populate stable source-side metadata automatically.

framework/py/flwr/supercore/servicers/appio_servicer.py
```diff
@@ -125,20 +126,60 @@ def PushTaskMessage(
...
+                grpc.StatusCode.FAILED_PRECONDITION,
+                "`Message.metadata.src_task_id` does not match the authenticated task.",
+            )
+
+        if message_proto.HasField("content") == message_proto.HasField("error"):
+            context.abort(
+                grpc.StatusCode.FAILED_PRECONDITION,
+                "Either `Message.content` or `Message.error` must be set, "
+                "but not both.",
+            )
```
panh99: We've actually checked all these in the `.store_task_message` method. Checking it here does not add any value. Plus, checking it inside the `.store_task_message` can ensure atomic operation. Should we remove the checks here, and then update the docstring of the `CoreState.store_task_message` abstract method to mention that the method will check these?

framework/py/flwr/supercore/servicers/appio_servicer.py
```diff
@@ -125,20 +126,44 @@ def PushTaskMessage(
...
         log(DEBUG, "AppIoServicer.PushTaskMessage")
-        context.abort(
-            grpc.StatusCode.UNIMPLEMENTED, "PushTaskMessage is not implemented"
-        )
-        raise RuntimeError("This line should never be reached.")
+
+        message_proto = request.message
+        metadata_proto = message_proto.metadata
+        task = get_authenticated_task()
+        if metadata_proto.src_task_id != task.task_id:
```
panh99:
```suggestion
        task = get_authenticated_task()

        message_proto = request.message
        if message_proto.metadata.src_task_id != task.task_id:
```

framework/py/flwr/supercore/servicers/appio_servicer.py
```diff
@@ -125,20 +126,44 @@ def PushTaskMessage(
...
         log(DEBUG, "AppIoServicer.PushTaskMessage")
-        context.abort(
-            grpc.StatusCode.UNIMPLEMENTED, "PushTaskMessage is not implemented"
-        )
-        raise RuntimeError("This line should never be reached.")
+
+        task = get_authenticated_task()
+
+        message_proto = request.message
+        if message_proto.metadata.src_task_id != task.task_id:
```
panh99:
```suggestion
        if request.message.metadata.src_task_id != task.task_id:
```

framework/py/flwr/supercore/servicers/appio_servicer.py
```diff
@@ -125,20 +126,44 @@ def PushTaskMessage(
...
+            context.abort(
+                grpc.StatusCode.FAILED_PRECONDITION,
+                "`Message.metadata.src_task_id` does not match the authenticated task.",
+            )
+
+        message = message_from_proto(message_proto)
```
panh99:
```suggestion
        message = message_from_proto(request.message)
```

framework/py/flwr/supercore/servicers/appio_servicer.py
```diff
@@ -125,20 +126,60 @@ def PushTaskMessage(
...
         """Pull task messages."""
         log(DEBUG, "AppIoServicer.PullTaskMessage")
-        context.abort(
-            grpc.StatusCode.UNIMPLEMENTED, "PullTaskMessage is not implemented"
+
+        task = get_authenticated_task()
+        limit = request.limit if request.HasField("limit") else None
+        messages = self.state().get_task_message(
+            dst_task_ids=[task.task_id],
+            limit=limit,
```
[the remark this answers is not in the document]
panh99: I think this is probably over-defensive. Since omitting limit already allows fetching all available messages, a very large limit does not give the caller additional access or meaningful extra behavior. The only concrete issue I see is potential backend/driver conversion failure if a uint64 value exceeds the signed 64-bit range accepted by SQL bindings. Given the low practical impact, I’d prefer not to add extra validation here unless we have evidence this causes real failures or we decide to enforce a product-level maximum for all limit fields.

panh99 (approved):
LGTM!

PR #7204, opened 2026-05-19
refactor(framework): Simplify check for `ray` being installed

jafermarq:
Performing the check at import time isn't ideal when we want to rely on auto-install dependencies for an app.


chongshenng (approved):
LGTM.

panh99 (approved):

danieljanes (approved):

PR #7213, opened 2026-05-20
feat(framework): Add `ModelRequest` and `ModelResponse`

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+from flwr.supercore.date import now
+from flwr.supercore.inflatable.inflatable_object import InflatableObject
+
+JSONScalar: TypeAlias = bool | float | int | str | None
+JSONValue: TypeAlias = JSONScalar | dict[str, "JSONValue"] | list["JSONValue"]
+JSONObject: TypeAlias = dict[str, JSONValue]
+
+_PAYLOAD_RECORD_KEY = "payload"
+_PAYLOAD_JSON_KEY = "json"
+_DEFAULT_TASK_MESSAGE_TTL = 3600.0
```
panh99, line 34: Will this be used outside this module? If not, good to keep it just here.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+
+T = TypeVar("T", bound=Message)
+
+
+class ModelRequest(Message):
+    """Task-routed model request in OpenAI Responses create-request shape."""
+
+    MESSAGE_TYPE: ClassVar[str] = "query.model_request"
```
panh99: Should we just use `"query"` here? The message type `"some_type.some_name"` is designed to route the message to the `ClientApp` side function decorated by `@app.some_type("some_name")`. Prolly unnecessary here.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+
+from flwr.app.metadata import Metadata
+from flwr.common.message import Message, make_message
+from flwr.common.record import ConfigRecord, RecordDict
+from flwr.supercore.date import now
+from flwr.supercore.inflatable.inflatable_object import InflatableObject
+
+JSONScalar: TypeAlias = bool | float | int | str | None
+JSONValue: TypeAlias = JSONScalar | dict[str, "JSONValue"] | list["JSONValue"]
+JSONObject: TypeAlias = dict[str, JSONValue]
```
panh99: This is great if `mypy` can understand this expression. Could you move it to `supercore.typing`? Btw, should we add a `JSONArray: TypeAlias = Sequence[JSONValue]` since it's used in multiple places it seems?

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+from typing import ClassVar, TypeAlias, TypeVar, cast
+
+from flwr.app.metadata import Metadata
+from flwr.common.message import Message, make_message
+from flwr.common.record import ConfigRecord, RecordDict
+from flwr.supercore.date import now
+from flwr.supercore.inflatable.inflatable_object import InflatableObject
+
+JSONScalar: TypeAlias = bool | float | int | str | None
+JSONValue: TypeAlias = JSONScalar | dict[str, "JSONValue"] | list["JSONValue"]
```
panh99: Should we use `Sequence` here, so we don't need to do `cast` in the code below?
```suggestion
JSONValue: TypeAlias = JSONScalar | dict[str, "JSONValue"] | Sequence["JSONValue"]
```

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+    MESSAGE_TYPE: ClassVar[str] = "query.model_request"
+
+    def __init__(  # pylint: disable=too-many-arguments,too-many-positional-arguments,too-many-locals,super-init-not-called
+        self,
+        *,
+        dst_task_id: int,
+        input: list[JSONObject],  # pylint: disable=redefined-builtin
+        model: str,
+        stream: bool,
+        tools: list[JSONObject] | None = None,
```
panh99: Same as above.
```suggestion
        input: Sequence[JSONObject],  # pylint: disable=redefined-builtin
        model: str,
        stream: bool,
        tools: Sequence[JSONObject] | None = None,
```

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+        text: JSONObject | None = None,
+        reply_to_message_id: str = "",
+        ttl: float = _DEFAULT_TASK_MESSAGE_TTL,
+    ) -> None:
+        payload: JSONObject = {
+            "model": model,
+            "input": cast(JSONValue, input),
+            "stream": stream,
+        }
+        _set_optional(payload, "tools", cast(JSONValue | None, tools))
```
panh99: If we use `Sequence` for typing, we can then drop `cast` here.
```suggestion
            "input": input,
            "stream": stream,
        }
        _set_optional(payload, "tools", tools)
```

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+    if not isinstance(payload, dict):
+        raise ValueError("Payload JSON must be an object.")
+    return cast(JSONObject, payload)
+
+
+def _validate_message_type(message: Message, expected: str) -> None:
+    """Validate that the message type matches the expected value."""
+    actual = message.metadata.message_type
+    if actual != expected:
+        raise ValueError(f"Expected message type {expected}, got {actual}.")
```
panh99: This function is a thin wrapper around a simple equality check. We could consider removing it and writing the `if` check directly for clarity, as it might make the code easier to read at the call site.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+
+def _init_message_from_payload(  # pylint: disable=too-many-arguments
+    message: Message,
+    *,
+    dst_task_id: int,
+    message_type: str,
+    payload: JSONObject,
+    reply_to_message_id: str,
+    ttl: float,
+) -> None:
```
panh99: Could we change this function to return a `metadata` and a `RecordDict`, and on the caller site, we do:

```python
metadata, content = _build_metadata_and_content(
    dst_task_id, message_type, payload, reply_to_message_id, ttl
)
super().__init__(metadata=metadata, content=content)
```

Right now, the function reads more like a runtime hack than a clean helper for initialization. Returning the two components explicitly would make its purpose clearer and usage cleaner.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+    def from_message(cls, message: Message) -> ModelRequest:
+        """Parse a generic message into a model request."""
+        _validate_message_type(message, cls.MESSAGE_TYPE)
+        payload = _payload_from_message(message)
+        _validate_model_request_payload(payload)
+        return _copy_as(cls, message)
+
+    def deflate(self) -> bytes:
+        """Deflate as the underlying transport `Message`."""
+        return _deflate_as_message(self)
```
panh99: We don't need this method. No need to override it.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+        """Return this request's Responses create-request payload."""
+        return _payload_from_message(self)
+
+    @classmethod
+    def from_message(cls, message: Message) -> ModelRequest:
+        """Parse a generic message into a model request."""
+        _validate_message_type(message, cls.MESSAGE_TYPE)
+        payload = _payload_from_message(message)
+        _validate_model_request_payload(payload)
+        return _copy_as(cls, message)
```
panh99: We don't need a `_copy_as`. We only use it to interpret a `Message` as `ModelRequest`. Using `_copy_as` can introduce unnecessary overheads.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+        cls, object_content: bytes, children: dict[str, InflatableObject] | None = None
+    ) -> ModelRequest:
+        """Inflate a model request from bytes."""
+        return cls.from_message(Message.inflate(object_content, children))
+
+
+class ModelResponse(Message):
+    """Task-routed model response in OpenAI Responses object shape."""
+
+    MESSAGE_TYPE: ClassVar[str] = "query.model_response"
```
panh99: Same as above.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def __init__(  # pylint: disable=super-init-not-called
+        self,
+        *,
+        dst_task_id: int,
+        response: JSONObject,
+        reply_to_message_id: str = "",
```
panh99: Is this optional for responses?

framework/py/flwr/supercore/model_message.py
panh99: No need to override it.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+        encoded = json.dumps(payload, separators=(",", ":"), allow_nan=False)
+    except (TypeError, ValueError) as err:
+        raise ValueError("Payload must be JSON serializable.") from err
+    return ConfigRecord({_PAYLOAD_JSON_KEY: encoded})
+
+
+def _payload_from_message(message: Message) -> JSONObject:
+    """Parse a JSON object payload from a message."""
+    if message.has_error():
+        raise ValueError("Expected a message with content, got an error message.")
```
panh99: We don't need this check as we do `if not message.has_content():` below.
```suggestion
```

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,322 @@
...
+def _config_record_from_payload(payload: JSONObject) -> ConfigRecord:
+    """Serialize a JSON object payload into a ConfigRecord."""
+    try:
+        encoded = json.dumps(payload, separators=(",", ":"), allow_nan=False)
+    except (TypeError, ValueError) as err:
+        raise ValueError("Payload must be JSON serializable.") from err
+    return ConfigRecord({_PAYLOAD_JSON_KEY: encoded})
+
+
+def _payload_from_message(message: Message) -> JSONObject:
```
panh99: Should we make them symmetric? For example:

```python
def _payload_to_content(payload: JSONObject) -> RecordDict:
    ...

def _payload_from_content(content: RecordDict) -> JSONObject:
    ...
```

framework/py/flwr/supercore/model_message.py
danielnugraha: I don't think so. If we need it in future PR, I will refactor it and move it outside to constant.

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,291 @@
...
+
+    if not isinstance(payload, dict):
+        raise ValueError("Payload JSON must be an object.")
+    return cast(JSONObject, payload)
+
+
+def _validate_object_sequence_field(
+    payload: JSONObject, field: str, *, owner: str, required: bool = False
+) -> None:
+    """Validate that a payload field is a sequence of JSON objects."""
```
panh99: Is this normally referred to as `JSONArray`?

framework/py/flwr/supercore/model_message.py
danielnugraha: I added an arg and pass it here.

framework/py/flwr/supercore/model_message.py
danielnugraha: We want to check this with this function:

` {"input": [{"role": "user", "content": "hello"}]}`

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,291 @@
...
+        payload = json.loads(raw)
+    except json.JSONDecodeError as err:
+        raise ValueError("Payload JSON is malformed.") from err
+
+    if not isinstance(payload, dict):
+        raise ValueError("Payload JSON must be an object.")
+    return cast(JSONObject, payload)
+
+
+def _validate_object_sequence_field(
```
panh99:
```suggestion
def _validate_jsonobject_sequence_field(
```

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,306 @@
...
+    if isinstance(value, str):
+        return
+    if not isinstance(value, Sequence):
+        raise ValueError(
+            "ModelRequest payload field 'input' must be a string or sequence."
+        )
+    if not all(isinstance(item, dict) for item in value):
+        raise ValueError(
+            "ModelRequest payload field 'input' must be a string or sequence "
+            "of objects."
```
panh99:
```suggestion
            "of JSON objects."
```

framework/py/flwr/supercore/model_message.py
```diff
@@ -0,0 +1,306 @@
...
+    record = content.config_records.get(_PAYLOAD_RECORD_KEY)
+    if record is None:
+        raise ValueError("Expected a payload ConfigRecord.")
+
+    raw = record.get(_PAYLOAD_JSON_KEY)
+    if not isinstance(raw, str):
+        raise ValueError("Expected payload JSON to be a string.")
+
+    try:
+        payload = json.loads(raw, parse_constant=_reject_non_finite_json_number)
```
panh99, line 216: Do we need to special-case this here? Python does accept NaN/Infinity even though they’re not valid JSON, but we don’t seem to reject them elsewhere in the framework. Unless this path is meant to be stricter than the others, I’d keep this as a normal json.loads for consistency.

panh99 (approved):
LGTM!

PR #7216, opened 2026-05-20
feat(framework): Add AgentApp API

danielnugraha:
Adds a minimal public flwr.agentapp API skeleton.

  Includes AgentApp, AgentSession, AgentApp exceptions, DEFAULT_MODEL_NAME, and local JSON type aliases.
  AgentSession parses agent.start state from Context and exposes input_items, agent_ref, and
  conversation_id.


framework/py/flwr/agentapp/session.py
```diff
@@ -0,0 +1,131 @@
...
+from flwr.common import Context
+from flwr.common.typing import Run
+
+from .exceptions import AgentAppError
+from .typing import JSONObject, JSONValue
+
+_AGENT_START_RECORD_KEY = "agent.start"
+_AGENT_REF_KEY = "agent_ref"
+_AGENT_INPUT_JSON_KEY = "input_json"
+_AGENT_CONVERSATION_ID_KEY = "conversation_id"
```
jafermarq: Worth moving these and other constants in this PR to a new `agentapp/constants.py`?

framework/py/flwr/agentapp/session.py
```diff
@@ -0,0 +1,49 @@
...
+    """Runtime session passed to AgentApp main functions."""
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        *,
+        task_id: int,
+        run: Run,
+        agent_ref: str,
+        conversation_id: str,
```
panh99: Should we now call it `series_id`?

framework/py/flwr/agentapp/session_test.py
```diff
@@ -0,0 +1,42 @@
...
+        agent_ref="test-agent",
+        conversation_id="conv-1",
+        input_items=input_items,
+    )
+
+    assert session.task_id == 101
+    assert session.run is run
+    assert session.agent_ref == "test-agent"
+    assert session.conversation_id == "conv-1"
+    assert session.input_items == input_items
```
panh99: This test feels a little too simple to me. It only checks that constructor arguments are stored as attributes, so I’m not sure it adds much value. Do we have a specific public API contract in mind here? If not, I’d be inclined to remove it.

panh99:
@copilot Should we put the ABCs + the AgentAppCallable type alias in one `base.py`?

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7233, opened 2026-05-25
feat(framework): Add wait for capacity

msheller:
### Description

SuperExec executors need a small pre-claim capacity check before later PRs wire capacity waiting into the launch loop.

### Related issues/PRs

Stacked on `codex/superexec-launch-result-handling`.

### Explanation

Adds `Executor.wait_for_capacity() -> None` as a capacity check for one TaskExecutor launch. The subprocess executor implements it as a no-op, with focused tests for the protocol shape and no-op behavior.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Address LLM-reviewer comments, if applicable
- [x] Make CI checks pass
- [x] Ping maintainers on Slack


framework/py/flwr/supercore/superexec/executor/subprocess_executor.py
```diff
@@ -28,6 +28,10 @@
 class SubprocessExecutor:
     """Run TaskExecutor processes as local subprocesses."""
 
+    def wait_for_capacity(self) -> None:
+        """Return immediately because subprocess launches have no capacity gate."""
+        pass
```
panh99: Do we need a timeout here? I guess it's fine to start without a timeout, cuz I am less sure how we should handle a timeout.

msheller: Timeout will be exec impl specific for now. (we resolved in chat)

panh99: FTR, as discussed, the timeouts will be part of the specific implementation rather than a parameter set by the plugin.

panh99 (approved):
LGTM!

PR #7244, opened 2026-05-26
fix(framework): Make message delivery crash-safe

psfoley:
### Description

Message delivery currently uses `delivered_at` as a permanent marker as soon as a message is pulled from LinkState. If a SuperNode, ServerApp, or SuperLink replica crashes before the receiver confirms the message, the message can be left unavailable even though it was never durably received.

This matters for HA SuperLink deployments because unconfirmed work/results must be recoverable by another SuperLink replica when LinkState is backed by a durable DB such as Postgres.

Affected boundaries:

- `LinkState`: message claim/redelivery semantics and acknowledgement contract
- `SqlLinkState`: durable lease/ack state used by HA deployments
- `InMemoryLinkState`: parity with the shared `LinkState` contract for local/dev/tests, not an HA backend
- `Fleet` and `ServerAppIo` servicers: call acknowledgement on `ConfirmMessageReceived`
- State schema/Alembic migration: adds delivery lease/ack columns

### Related issues/PRs

Related to the SuperLink HA race-condition work.

### Explanation

This PR changes pulled messages from permanently delivered to leased-until-confirmed.

- Pulling an instruction or reply message claims it by setting `delivered_at` and `lease_expires_at`.
- A second pull will not return the same message while the lease is active.
- `ConfirmMessageReceived` acknowledges the message and performs the existing cleanup.
- If confirmation never happens, the lease expires and the message can be redelivered.

The lease is currently 60 seconds. This keeps crash recovery reasonably fast while avoiding very aggressive redelivery during normal object transfer. The PR intentionally does not change object-transfer protocol semantics; confirmation still means the receiver has pulled the message object tree contents.

Migration notes:

- Adds non-null `lease_expires_at` to `message_ins` and `message_res` with default `0` so existing rows are immediately eligible for delivery.
- Adds non-null `acknowledged_at` to `message_ins` with default empty string so existing instruction messages remain unacknowledged.
- No data rewrite or backfill job is required beyond the server defaults.
- The migration was generated with the Flower migration tooling and the schema README was updated to match generated ordering.

Review focus:

- Whether the lease/ack lifecycle belongs at `LinkState` level.
- Whether 60 seconds is an acceptable static lease for now.
- Whether keeping `ObjectStore` lifecycle-agnostic and cleanup on `ConfirmMessageReceived` matches the intended boundary.

### Checklist

- [x] Write tests
- [x] Update [documentation](https://flower.ai/docs/writing-documentation.html) / generated schema docs only; no user-facing docs needed
- [x] Make CI checks pass

### Any other comments?

Tests added/updated at the state and servicer layer:

- Redelivery after instruction lease expiry
- Redelivery after reply lease expiry
- Acknowledgement prevents instruction redelivery
- `ConfirmMessageReceived` performs reply cleanup

Latest GitHub checks are passing.


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -221,7 +233,8 @@ def get_message_ins(self, node_id: int, limit: int | None) -> list[Message]:
                 if (
                     (msg_ins := self.message_ins_store.get(msg_id))
                     and msg_ins.metadata.dst_node_id == node_id
-                    and msg_ins.metadata.delivered_at == ""
+                    and msg_id not in self.acknowledged_message_ins_ids
+                    and not self._has_active_delivery_lease(msg_ins)
```
msheller, line 237: @panh99 @psfoley , here is a point I think we really want our agents to understand and help us with. This if block encodes policy that is independent of the backing store, e.g. it defines the policy around lease and acknowledgment implicitly. This is a maintenance/correctness problem. In memory vs. SQL has nothing to do with this policy, so it should be encoded in a common place. The mechanisms for retrieving the values the policy needs are specific to in-memory vs. SQL, so they should be here, but the policy should not.

Additionally, I wonder if an if with 4 sub-conditions is a strong signal that we are likely conflating domain policy with persistence mechanics. The lease/acknowledgement policy should be an "invariant" that lives in one common place. SQL and in-memory implementations should only differ in how they fetch and update the facts needed to enforce it.

I am open to adding this as a follow-up PR, so long as we add it to our list before resolving this comment.

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -2224,6 +2224,110 @@ def test_get_message_res_claim_is_unique_across_replicas(self) -> None:
             # Assert
             assert sum(len(res) for res in results if res is not None) == 1
 
+    def test_get_message_ins_redelivers_after_lease_expires(self) -> None:
```
msheller: This test does more than the title suggests. It also checks that once acknowledged, the expired lease does not cause redelivery.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -429,8 +439,10 @@ def store_message_res(  # pylint: disable=too-many-return-statements
                 msg_dict, ["run_id", "src_node_id", "dst_node_id"]
             )
 
+            msg_dict["lease_expires_at"] = 0.0
+            column_names = ", ".join(msg_dict)
             columns = ", ".join([f":{key}" for key in msg_dict])
-            query = f"INSERT INTO message_res VALUES({columns})"
+            query = f"INSERT INTO message_res ({column_names}) VALUES({columns})"
```
msheller, line 447: Maybe worth a comment here along the lines of:

"acknowledged_at exists only for instruction messages because instruction acknowledgement is not terminal: the instruction row must remain until a reply is stored and delivered. Reply acknowledgement is terminal for the instruction/reply pair, so the SQL implementation deletes both rows instead of recording acknowledged_at on message_res."

But shorter :) I asked Codex to explain this to me, so I figure I'd call it out as a possible comment spot.

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -224,8 +216,9 @@ def PullMessages(  # pylint: disable=R0914
                 trees.append(obj_tree)
             except NoObjectInStoreError as e:
                 log(ERROR, e.message)
-                # Delete message ins from state
-                state.delete_messages(message_ins_ids={msg_object_id})
+                state.delete_messages(
```
msheller, line 221: Maybe a comment about "now we delete by message pairs (ins and res), so we need the ins id, specifically.

msheller (dismissed):
LGTM! I added some comments that I think should be captured in a future PR, so would appreciate of those could be captured in our roadmap before resolving, assuming you agree with my comments.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
msheller: I should add that a big reason why I think this is likely a follow-up PR is that this issue exists elsewhere in the surrounding code, not just in this PR.

msheller: In fact, I'll just toss it in our list for discussion in our next meeting :)

psfoley: > The mechanisms for retrieving the values the policy needs are specific to in-memory vs. SQL, so they should be here, but the policy should not.

Fully agree with this point. Delineating the policy from the specific implementation is important for maintainability. If @panh99 agrees, I would suggest this a follow on PR (to identify where policy is encoded more broadly across the in-memory / SQL implementations, and figure out the best way to extract this)

PR #7252, opened 2026-05-27
feat(framework): Add RunEvents RPC

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/proto/flwr/proto/appio.proto
```diff
@@ -58,6 +58,14 @@ message PullAppMessagesResponse {
 message PushTaskMessageRequest { Message message = 1; }
 message PushTaskMessageResponse { string message_id = 1; }
 
+// PushTaskEvents messages
+message TaskEvent {
+  string event = 1;
+  string data = 2;
+}
```
panh99: The task_event table contains the following fields:
- id
- timestamp
- run_id
- task_id
- event
- data

For the `TaskEvent` proto, I think it will be good to contain all these, mirroring the table structure. Wdyt?

framework/proto/flwr/proto/appio.proto
```diff
@@ -58,6 +58,14 @@ message PullAppMessagesResponse {
...
+message PushTaskEventsRequest { repeated TaskEvent events = 1; }
+message PushTaskEventsResponse { uint32 stored_count = 1; }
```
panh99: Since task events are either accepted as a batch or rejected at the RPC/auth layer, stored_count does not currently communicate meaningful information. An empty response keeps the API contract simpler. We can add explicit partial-success or deduplication fields later if those semantics are introduced.

framework/proto/flwr/proto/control.proto
```diff
@@ -244,11 +244,13 @@ message ConfigureSimulationFederationRequest {
...
-message StreamRunEventsRequest { uint64 run_id = 1; }
+message StreamRunEventsRequest {
+  uint64 run_id = 1;
+  optional uint64 after_id = 2;
+}
 message StreamRunEventsResponse {
   string event = 1;
   string data = 2;
+  uint64 id = 3;
+  uint64 task_id = 4;
```
panh99: Following the above comment about extending `TaskEvent`, should we just return a `TaskEvent` here?

framework/proto/flwr/proto/control.proto
```diff
@@ -244,11 +244,13 @@ message ConfigureSimulationFederationRequest {
 }
 message ConfigureSimulationFederationResponse {}
 
-// ##############
-// Unused for now
-// ##############
-message StreamRunEventsRequest { uint64 run_id = 1; }
+message StreamRunEventsRequest {
+  uint64 run_id = 1;
+  optional uint64 after_id = 2;
```
panh99: There are many ids. Prolly we can say `after_event_id` for clarity, or that's not necessary?

framework/proto/flwr/proto/appio.proto
danielnugraha: I think id, timestamp, run_id, task_id will be populated by the servicer. The task process only needs to send the event and data.

framework/proto/flwr/proto/control.proto
danielnugraha: For clarity it should be `after_task_event_id`. Is this okay and not too long?

panh99: That's okay. It's long, but clear.

framework/proto/flwr/proto/appio.proto
panh99: I see. Could we still include those fields in the `TaskEvent` proto to mirror the table structure?

When the task process pushes a `TaskEvent`, it can leave `id`, `timestamp`, `run_id`, and `task_id` unset, and the servicer can populate them. But having the complete structure in the proto might make things easier and more consistent on the retrieval side. Wdyt?

danielnugraha: I see. Should we move TaskEvent elsewhere then into task.proto?

What do you mean by unset? Make it optional?

panh99: By unset, I mean we can just send `TaskEvent(event="xx", data="yy")` without setting other fields. This is allowed by gRPC. We don't need to mark the fields as optional.

And yeah, moving it to `task.proto` makes sense to me.

danielnugraha: I moved the TaskEvent to task.proto

panh99 (dismissed):
LGTM!

danieljanes (approved):

PR #7253, opened 2026-05-27
feat(framework): Add RunSeries RPCs

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/proto/flwr/proto/message.proto
```diff
@@ -34,6 +34,7 @@ message Context {
   map<string, Scalar> node_config = 3;
   RecordDict state = 4;
   map<string, Scalar> run_config = 5;
+  optional uint64 series_id = 6;
```
panh99: Since `series_id` will be the primary identifier of a context, I think it's fine to remove the `optional` label of the field. After all, without the `optional` label, gRPC still allows us to leave the field unset.
```suggestion
  uint64 series_id = 6;
```

framework/proto/flwr/proto/run.proto
```diff
@@ -39,6 +39,7 @@ message Run {
   double clientapp_runtime = 15;
   string run_type = 16;
   optional uint64 primary_task_id = 17;
+  optional uint64 series_id = 18;
```
panh99: Same as above. This field will be non-optional. The `optional` modifier here only enables the `.HasField` check. AFAIK, we don’t rely on `.HasField` for `series_id`, right?

framework/proto/flwr/proto/runseries.proto
```diff
@@ -0,0 +1,31 @@
...
+syntax = "proto3";
+
+package flwr.proto;
+
+import "flwr/proto/run.proto";
+
+message RunSeries {
+  uint64 series_id = 1;
+  string federation = 2;
+  optional string description = 3;
```
panh99: Same as above. I think we don't need to make it optional.

framework/proto/flwr/proto/runseries.proto
```diff
@@ -0,0 +1,31 @@
...
+  string created_at = 4;
+  string updated_at = 5;
+  optional uint64 last_run_id = 6;
```
panh99: Do we still need this if we have a repeated uint64 run_ids field? Or you'd recommend to keep it for consistency with the `last_run_status`? If so, that also makes sense to me.

panh99 (dismissed):
LGTM!

danieljanes (approved):

PR #7256, opened 2026-05-28
feat(framework): Add TaskEvents in CoreState

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -291,6 +291,47 @@ def get_task_message(
...
+        bool
+            True if the events were stored, otherwise False.
+        """
+
+    @abstractmethod
+    def get_task_events(
+        self,
+        run_id: int,
+        after_task_event_id: int | None,
+    ) -> Sequence[TaskEvent]:
```
panh99, line 321: We usually design retrieval methods with optional keyword-only filters, similar to query parameters. Should we make this method more filter-oriented instead?

```suggestion
    def get_task_events(
        self,
        *,
        run_id: int | None = None,
        after_task_event_id: int | None = None,
    ) -> Sequence[TaskEvent]:
```

framework/py/flwr/supercore/corestate/corestate_test.py
panh99, line 1: Could you add minimal comments in the unit tests with `# Prepare/Execute/Assert: <details>` so the purpose of each block of code is clearer?

framework/py/flwr/supercore/utils.py
```diff
@@ -60,6 +61,33 @@ def __init__(self, key: str, error_type: MetadataLookupErrorType) -> None:
         super().__init__(message)
 
 
+def _reject_non_finite_strict_json_value(value: str) -> None:
+    """Reject JSON constants that are not valid JSON values."""
+    raise ValueError(f"Strict JSON value contains non-finite number {value}.")
+
+
+def strict_json_value_loads(raw: str | bytes | bytearray) -> JSONValue:
```
panh99: JSON value is any valid json, right? If so, should we just name it as

```suggestion
def strict_json_loads(raw: str | bytes | bytearray) -> JSONValue:
```

The same question for `strict_json_value_dumps`.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -423,6 +430,57 @@ def get_task_message(
...
+            for event in events:
+                validate_task_event_data(event.data)
+        except ValueError:
+            return False
+
+        with self.lock_task_store, self.lock_task_event_store:
+            for event in events:
+                task = self.task_store.get(event.task_id)
+                if task is None or task.run_id != event.run_id:
+                    return False
```
panh99: Do we need this check? IIUC, the task_id and the run_id fields are set on the servicer side (not the caller side, so it's in the same place as CoreState) with the input from the auth layer, which also reads the CoreState. Given that, prolly we don't need to check it again here for them?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -423,6 +430,57 @@ def get_task_message(
...
+                task_events.append(
+                    TaskEvent(
+                        id=self._next_task_event_id,
+                        timestamp=current,
+                        run_id=event.run_id,
+                        task_id=event.task_id,
+                        event=event.event,
+                        data=event.data,
+                    )
+                )
```
panh99: It's fine to just store the input event. This is a simple storage method and we won't continue use the event object after this.


```suggestion
                event.id = self._next_task_event_id
                event.timestamp = current
                task_events.append(event)
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -423,6 +430,57 @@ def get_task_message(
...
+    ) -> bool:
+        """Store task-produced run events."""
+        if not events:
+            return False
+
+        try:
+            for event in events:
+                validate_task_event_data(event.data)
+        except ValueError:
+            return False
```
panh99, line 445: Will we drop the full batch of events if any event data is invalid? Should we mention this in the docstring of the abc method?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -493,6 +498,87 @@ def get_task_message(
...
+                    FROM task
+                    WHERE task_id = :task_id AND run_id = :run_id
+                    """,
+                    {
+                        "task_id": uint64_to_int64(event.task_id),
+                        "run_id": uint64_to_int64(event.run_id),
+                    },
+                )
+                if not task_rows:
+                    return False
```
panh99: Same as in `in_memory_corestate`, we don't need to check this. These fields are automatically populated by the servicer based on auth input before calling this method.

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -291,6 +291,50 @@ def get_task_message(
...
+        """Return task-produced run events after the cursor.
+
+        Parameters
+        ----------
+        run_id : int
+            The run ID to retrieve events for.
+        after_task_event_id : Optional[int]
+            Return only events with an ID greater than this cursor. If set to
+            `None`, retrieve all events.
```
panh99: Could you update the docstring to align with the method signature?

danielnugraha: Wait. I think run_id is mandatory right?

panh99: Yes, we will always pass `run_id` when calling this method from `ControlServicer`. However, since `CoreState` methods are general internal API contracts and may be reused for different purposes, I think it may be better to keep this method filter-oriented.

In other words, while `run_id` is required for the `StreamRunEvents` RPC, we may not want that specific use case to drive the design of the underlying `CoreState` method too much.

danielnugraha: In other words, the only difference here is `get_task_events` will return all tasks when no `run_id` is given? Does it makes sense to return all tasks?

panh99: Yes, I think that is the expected behavior for a filter-oriented design. If no filter is provided, the method returns all matching records by default. This is similar to running `SELECT * FROM xxx` without a `WHERE` clause, which returns all rows.

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -291,6 +291,51 @@ def get_task_message(
...
+        after_task_event_id: int | None = None,
+    ) -> Sequence[TaskEvent]:
+        """Return task-produced run events matching the filters.
+
+        Parameters
+        ----------
+        run_id : Optional[int]
+            If set, return only events for this run. If set to `None`, return
+            events for all runs.
+        after_task_event_id : Optional[int]
```
```suggestion
        run_id : Optional[int] (default: None)
            If set, return only events for this run. If set to `None`, return
            events for all runs.
        after_task_event_id : Optional[int] (default: None)
```

panh99 (approved):
LGTM!

PR #7261, opened 2026-05-28
feat(framework): Add RunSeries in CoreState

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -17,15 +17,28 @@
...
+@dataclass(frozen=True)
+class RunSeries:
+    """RunSeries metadata stored in CoreState."""
+
+    series_id: int
+    federation: str
+    description: str
+    created_at: str
+    updated_at: str
+    last_run_id: int | None
```
panh99: Should we use `RunSeries` proto directly? Better to avoid introducing dataclass if there's a proto to reduce complexity.

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -42,6 +55,59 @@ def store_fab(self, fab: Fab) -> str:
     def get_fab(self, fab_hash: str) -> Fab | None:
         """Return the FAB for the given hash, if present."""
 
+    @abstractmethod
+    def get_run_series(
+        self,
+        *,
+        federation: str,
```
panh99: Does this suffice? I thought agentapp also need to get all run series of an account if it is a member of multiple federations, right? Or it's good enough for now?

framework/py/flwr/supercore/corestate/utils.py
```diff
@@ -18,8 +18,9 @@
 from datetime import datetime
 from os import urandom
 
-from flwr.common import Message
+from flwr.common import Context, Message, serde
```
panh99: I think `Message` and `Context` has been moved to `flwr.app`, no?

framework/py/flwr/supercore/corestate/corestate.py
danielnugraha: It is good enough for now.

framework/py/flwr/supercore/corestate/__init__.py
```diff
@@ -15,8 +15,11 @@
 """Flower CoreState."""
 
 
+from flwr.proto.runseries_pb2 import RunSeries  # pylint: disable=E0611
+
 from .corestate import CoreState
 
 __all__ = [
     "CoreState",
+    "RunSeries",
```
panh99: Why we export the RunSeries from CoreState? It doesn't seem necessary, no?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -42,6 +43,59 @@ def store_fab(self, fab: Fab) -> str:
...
+    ) -> Sequence[RunSeries]:
+        """Return RunSeries metadata for the specified federation.
+
+        Parameters
+        ----------
+        federation : str
+            Federation name used to filter RunSeries.
+        updated_before : str | None
+            If set, return only RunSeries updated before this ISO timestamp.
+        limit : int | None
```
```suggestion
        updated_before : str | None (default: None)
            If set, return only RunSeries updated before this ISO timestamp.
        limit : int | None (default: None)
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -107,6 +112,49 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+        if limit == 0:
+            return []
+
+        with self.lock_run_series_store:
+            run_series = [
+                record
+                for record in self.run_series_store.values()
+                if record.federation == federation
+                and (updated_before is None or record.updated_at < updated_before)
+            ]
```
panh99: Should we use a `for` loop here instead of an inline expression to improve readability?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +131,85 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+        """Return RunSeries metadata for the specified federation."""
+        if limit is not None and limit < 0:
+            raise AssertionError("`limit` must be >= 0")
+        if limit == 0:
+            return []
+
+        conditions = ["federation = :federation"]
+        params: dict[str, Any] = {"federation": federation}
+        if updated_before is not None:
+            conditions.append("updated_at < :updated_before")
```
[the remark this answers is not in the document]
panh99: This might be a valid concern. Though it's a rare case, but I wonder if we should use `<` or `<=`? Or you think there's no risk of losing conversations that are updated at the same time?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -665,6 +752,21 @@ def task_from_row(row: dict[str, Any]) -> Task:
...
+        [int64_to_uint64(row["last_run_id"])] if row["last_run_id"] is not None else []
+    )
+    return RunSeries(
+        series_id=int64_to_uint64(row["series_id"]),
+        federation=row["federation"],
+        description=row["description"] or "",
+        created_at=timestamp_to_iso(row["created_at"]),
+        updated_at=timestamp_to_iso(row["updated_at"]),
+        run_ids=run_ids,
+    )
```
panh99: Should we store all run IDs in the series? This code appears to silently treat the last run ID as if it were the full list of run IDs. That feels risky, especially since the behavior is undocumented.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +131,85 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+            query += " LIMIT :limit"
+            params["limit"] = limit
+
+        rows = self.query(query, params)
+        return [_run_series_from_row(row) for row in rows]
+
+    def get_run_series_context(self, series_id: int) -> Context | None:
+        """Return the shared Context for the specified RunSeries, if present."""
+        if series_id == 0:
+            return None
```
panh99: Fine to remove this check. It doesn't seem to have meaningful impact.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +131,85 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+            {"series_id": uint64_to_int64(series_id)},
+        )
+        if not rows or rows[0]["context"] is None:
+            return None
+        return context_from_bytes(rows[0]["context"])
+
+    def set_run_series_context(self, series_id: int, context: Context) -> None:
+        """Set the shared Context for the specified RunSeries."""
+        if series_id == 0:
+            raise ValueError("RunSeries ID must be non-zero")
```
panh99: Fine to remove this check. It doesn't seem to have meaningful impact. The run series are uint, and we don't require positive integers, right? Though practically it's hard to get a random uint that happens to be zero.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +131,85 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+            if not self.query(
+                """
+                SELECT series_id
+                FROM run_series
+                WHERE series_id = :series_id
+                """,
+                {"series_id": sint_series_id},
+            ):
+                raise ValueError(f"RunSeries {series_id} not found")
+
```
panh99: Should we remove this check? Based on previously discussion (SuperNode may not be able to have a valid run series table, and thus we should keep the run series context table just a simple storage with no checks), it's intended to avoid a foreign key constraint. Should we remove this bit that seems to restore a similar foreign key constraint?

framework/py/flwr/supercore/corestate/sql_corestate.py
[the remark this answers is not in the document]
danielnugraha: It's very rare that a conversation will be updated at the same time.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +131,73 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+        conditions: list[str] = []
+        params: dict[str, Any] = {}
+        if federation is not None:
+            conditions.append("federation = :federation")
+            params["federation"] = federation
+        if updated_before is not None:
+            conditions.append("updated_at < :updated_before")
+            params["updated_before"] = datetime.fromisoformat(
+                updated_before.replace("Z", "+00:00")
+            )
```
panh99: Why we need this? I think we can assume the input `updated_before` is a standard ISO format string. The `updated_at` fields are generated by `timestamp_to_iso` which just use `datetime.isoformat()`.

```suggestion
            params["updated_before"] = datetime.fromisoformat(updated_before)
```

panh99 (dismissed):
LGTM!

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -128,6 +133,89 @@ def get_fab(self, fab_hash: str) -> Fab | None:
             verifications=json.loads(row["verifications"]),
         )
 
+    def get_run_series(
```
panh99, line 138: Could you add minimal comments in the method to make the functionality of each block clearer?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -128,6 +133,89 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+            FROM run_series
+            {where_clause}
+            ORDER BY updated_at DESC
+        """
+        if limit is not None:
+            selected_run_series_query += " LIMIT :limit"
+            params["limit"] = limit
+
+        query = f"""
+            WITH selected_run_series AS (
```
panh99: Nit: Instead of selected_run_series_query, could we create a run_series_cte for consistency?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -128,6 +133,95 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+            limit_clause = "LIMIT :limit"
+            params["limit"] = limit
+
+        # Select the requested page before joining run IDs so limit applies to series.
+        run_series_cte = f"""
+            run_series_cte AS (
+                SELECT series_id, federation, description, created_at, updated_at
+                FROM run_series
+                {where_clause}
+                ORDER BY updated_at DESC
```
panh99, line 174:
```suggestion
                ORDER BY updated_at DESC
                {self.select_lock_sql}
```

panh99 (dismissed):
LGTM!

panh99 (dismissed):
LGTM!

framework/py/flwr/supercore/corestate/sql_corestate.py
panh99: We should update the `updated_at` field here, right? I think we can `UPDATE` the field where `series_id = :series_id AND federation = :federation`, and `RETURNING series_id`. Then log an error and return None when no row is returned.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -132,6 +135,96 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+        self,
+        *,
+        federation: str | None = None,
+        updated_before: str | None = None,
+        limit: int | None = None,
+    ) -> Sequence[RunSeries]:
+        """Return RunSeries metadata, optionally filtered by federation."""
+        # Validate limit before building the SQL query.
+        if limit is not None and limit < 0:
+            raise AssertionError("`limit` must be >= 0")
```
panh99: Let's use `ValueError` instead. `AssertionError` is less proper in production code.
```suggestion
            raise ValueError("`limit` must be >= 0")
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7264, opened 2026-05-28
feat(framework): Add RunSeries to CreateRun

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -107,6 +111,39 @@ def get_fab(self, fab_hash: str) -> Fab | None:
                 verifications=dict(fab.verifications),
             )
 
+    def ensure_run_series(self, federation: str, series_id: int | None = None) -> int:
+        """Ensure a run series exists and return its ID."""
+        if not federation:
+            raise ValueError("Federation must be set")
+        if series_id is not None and not 0 < series_id < (1 << 64):
+            raise ValueError("Series ID must be a positive uint64")
```
panh99: Could we remove these checks? They do not seem to add value in practice. This is an internal API, and `federation` should already be validated before this function is called. The `uint64` bounds check also feels a bit inconsistent with how we validate IDs elsewhere.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -107,6 +111,39 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+
+        with self.lock_run_series_store:
+            resolved_series_id = series_id
+            if resolved_series_id is None:
+                resolved_series_id = generate_rand_int_from_bytes(
+                    SERIES_ID_NUM_BYTES,
+                    exclude=set(self.run_series_store) | {0},
```
panh99: Is there a reason for requiring positive IDs? We don't require this for other IDs. If not, let's remove the constraint.
```suggestion
                    exclude=set(self.run_series_store),
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -107,6 +111,39 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+
+            timestamp = now().isoformat()
+            self.run_series_store[resolved_series_id] = RunSeries(
+                series_id=resolved_series_id,
+                federation=federation,
+                description="",
+                created_at=timestamp,
+                updated_at=timestamp,
+            )
+            return resolved_series_id
```
panh99: Should this be moved to the `if resolved_series_id is None:` branch? The control flow here is a bit confusing.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -107,6 +111,39 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+                    exclude=set(self.run_series_store) | {0},
+                )
+            elif resolved_series_id in self.run_series_store:
+                existing = self.run_series_store[resolved_series_id]
+                if existing.federation != federation:
+                    raise ValueError(
+                        f"Run series {resolved_series_id} belongs to federation "
+                        f"{existing.federation!r}, not {federation!r}"
+                    )
+                return resolved_series_id
```
panh99: After this bit, should we raise a ValueError that provided `series_id` does not exist? This is a possible error case, right?

framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -2170,6 +2210,20 @@ def _run(idx: int) -> None:
...
+        with tempfile.TemporaryDirectory() as tmpdir:
+            db_path = self._shared_sql_database(tmpdir)
+            self._create_shared_sql_states(db_path)
+
+            results = self._query_states_in_parallel(
+                lambda state: state.ensure_run_series("federation-a", series_id=123)
+            )
+
+            self.assertEqual(results, [123, 123])
+
```
panh99: We don't need a concurrency test for every new feature/method unless necessary.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -264,6 +268,14 @@ def StartRun(  # pylint: disable=too-many-locals, too-many-branches, too-many-st
                     "tmp_dir": self.artifact_provider.tmp_dir,
                 }
 
+            runs = state.get_run_info(run_ids=[run_id])
+            if not runs:
+                context.abort(
+                    grpc.StatusCode.INTERNAL,
+                    "Failed to retrieve the initialized run.",
+                )
```
panh99: We just created the run successfully and runs cannot be deleted for now. Should we remove this check?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -643,10 +646,18 @@ def create_run(
...
                 ),
                 federation_config=federation_config,
             )
             self.run_ids[run_id] = run_record
+            with self.lock_run_series_store:
+                run_series = self.run_series_store[resolved_series_id]
+                run_series.run_ids.append(run_id)
+                run_series.last_run_status.CopyFrom(
+                    run_status_to_proto(run_record.run.status)
+                )
```
panh99: No need to update it here. It won't be accurate. Let's set the `last_run_status` when retrieving the run series.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +124,76 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+                            "series_id": uint64_to_int64(candidate),
+                            "federation": federation,
+                            "description": None,
+                            "created_at": timestamp,
+                            "updated_at": timestamp,
+                            "last_run_id": None,
+                        },
+                    )
+                    if rows:
+                        return candidate
```
panh99: Let's not do a `while True` just for sampling IDs. We could simply sample one ID and proceed. No need to handle for this super rare case.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +124,76 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+
+            assert resolved_series_id is not None
```
panh99: Let's avoid `assert` in production code.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +124,76 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+                """,
+                {"series_id": uint64_to_int64(resolved_series_id)},
+            )
+            existing = rows[0]
+            if existing["federation"] != federation:
+                raise ValueError(
+                    f"Run series {resolved_series_id} belongs to federation "
+                    f"{existing['federation']!r}, not {federation!r}"
+                )
+            return resolved_series_id
```
panh99: The control flow is a bit unclear here. I think we should structure it as follows:

1. If `series_id` is not provided, sample a random ID and insert a new row into the `run_series` table. Return the sampled ID.
2. If `series_id` is provided, select the corresponding row and verify that its federation matches the provided `federation`. Raise a `FlowerError` if the `series_id` does not exist or if the federation does not match. Return the provided `series_id`.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -248,6 +251,7 @@ def StartRun(  # pylint: disable=too-many-locals, too-many-branches, too-many-st
                 resolved_federation_config,
                 flwr_aid,
                 run_type,
+                request_series_id,
```
panh99: We don't need a new local variable, I think. Let's pass the inline expression as the argument directly.
```suggestion
                request.series_id if request.HasField("series_id") else None,
```

framework/py/flwr/supercore/corestate/sql_corestate.py
panh99, line 1: The comments in framework/py/flwr/supercore/corestate/in_memory_corestate.py also applies here.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -993,6 +999,18 @@ def create_run(  # pylint: disable=R0913, R0914, R0917
...
                 )
+                self.query(
+                    """
+                    UPDATE run_series
+                    SET last_run_id = :run_id, updated_at = :updated_at
+                    WHERE series_id = :series_id
+                    """,
+                    {
+                        "run_id": uint64_to_int64(run_id),
+                        "updated_at": pending_at,
```
panh99: Should we rename this variable to `current` for clarity? Cuz the `current` timestamp is used both as the `pending_at` field of the newly created run, but also the `updated_at` field of the run series.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
panh99: Let's move it after `self.ensure_run_series(federation, series_id)`, otherwise the `updated_at` of the run series might be before its `created_at`.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -643,10 +648,15 @@ def create_run(
...
                     run_type=run_type,
+                    series_id=resolved_series_id,
                 ),
                 federation_config=federation_config,
             )
             self.run_ids[run_id] = run_record
+            with self.lock_run_series_store:
+                run_series = self.run_series_store[resolved_series_id]
+                run_series.run_ids.append(run_id)
+                run_series.updated_at = now().isoformat()
```
panh99: Should we set this the same as the pending_at of the run? There's a `pending_at` variable. Btw, we can rename it to `current`.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +124,61 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+                {"series_id": uint64_to_int64(series_id)},
+            )
+            if not rows:
+                raise ValueError(f"Run series {series_id} not found")
+            existing = rows[0]
+            if existing["federation"] != federation:
+                raise ValueError(
+                    f"Run series {series_id} belongs to federation "
+                    f"{existing['federation']!r}, not {federation!r}"
+                )
```
panh99: `ValueError` is raised here but we don't handle the error in the `.create_run` nor in the servicer (currently, request => servicer -> `LinkState.create_run` -> `CoreState.ensure_series_id`. That means the raw exception will be exposed to users.

Should we log the error and return `None` instead, or how you'd like to handle the ValueError?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -107,6 +111,39 @@ def get_fab(self, fab_hash: str) -> Fab | None:
                 verifications=dict(fab.verifications),
             )
 
+    def ensure_run_series(
+        self, federation: str, series_id: int | None = None
+    ) -> int | None:
+        """Ensure a run series exists and return its ID."""
+        if not federation:
+            raise ValueError("Federation must be set")
```
panh99: Let's remove this check. It does not seem to add much value in practice, does it?

After all, we already need to use `FederationManager` to validate that the federation exists before reaching this point. Given that, checking whether the federation is empty does not seem to matter much.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -123,6 +124,61 @@ def get_fab(self, fab_hash: str) -> Fab | None:
             verifications=json.loads(row["verifications"]),
         )
 
+    def ensure_run_series(
+        self, federation: str, series_id: int | None = None
+    ) -> int | None:
+        """Ensure a run series exists and return its ID."""
+        if not federation:
+            raise ValueError("Federation must be set")
```
panh99: Same as in the InMemoryCoreState.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -985,14 +991,26 @@ def create_run(  # pylint: disable=R0913, R0914, R0917
...
                         "running_at": None,
                         "finished_at": None,
                         "sub_status": "",
                         "details": "",
                     },
                 )
+                self.query(
+                    """
+                    UPDATE run_series
+                    SET last_run_id = :run_id, updated_at = :updated_at
```
panh99: We only store the last run ID here. Should we store the full list of run IDs that belong to the series? I think that may require a new table like `(id, series_id, run_id)`.

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -114,6 +118,56 @@ def get_fab(self, fab_hash: str) -> Fab | None:
                 verifications=dict(fab.verifications),
             )
 
+    def store_run_in_series(
+        self,
+        *,
+        run_id: int,
+        federation: str,
+        series_id: int | None = None,
```
panh99: We don't need to require keyword arguments here, I think. Plus, it would be better to require caller to specify `series_id`.
```suggestion
        run_id: int,
        federation: str,
        series_id: int | None,
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7265, opened 2026-05-28
fix(framework): Prevent ObjectStore put success after delete

psfoley:
### Description

`SqlObjectStore.put` first checked whether an object row existed and was unavailable, then updated the row in a later statement. If run cleanup deleted that preregistered object between the read and the write, the `UPDATE` affected zero rows but `put` still returned successfully.

This matters for HA SuperLink deployments because an object upload racing with run cleanup could be acknowledged as stored even though the object no longer exists.

### Explanation

This PR makes the write path authoritative by changing `put` to use `UPDATE ... RETURNING` with `WHERE object_id = :object_id AND is_available = 0`.

- If the row is updated, `put` succeeds.
- If no row is updated, `put` checks whether the object is already available and returns idempotently.
- If the object is missing, `put` raises `NoObjectInStoreError` instead of reporting success.

The ObjectStore remains lifecycle-agnostic; this only tightens the SQL write outcome.

### Checklist

- [x] Write focused regression test
- [x] No schema or user-facing docs changes needed

### Validation

- `uv run --python=3.11.14 python -m pytest py/flwr/supercore/object_store/object_store_test.py`
- `git diff --check`


framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -173,16 +184,11 @@ def put(self, object_id: str, object_content: bytes) -> None:
...
                 return
 
-            # Update the object entry in the store
-            self.query(
-                "UPDATE objects SET content = :content, is_available = 1 "
-                "WHERE object_id = :object_id",
-                {"content": object_content, "object_id": object_id},
+            raise NoObjectInStoreError(
+                f"Object with ID '{object_id}' was not pre-registered."
             )
```
panh99: I think we can remove this block after moving the `UPDATE` query to the top.

If the `UPDATE` query returns no row, then it can only mean one of two things: either the `object_id` does not exist, or `is_available` is already `1`. Given that, we only need to check whether the `object_id` exists and raise `NoObjectInStoreError` if it does not.

So I don’t think we need to keep this additional check. Does that sound right?

psfoley: Good catch! I've removed the additional check

framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -164,7 +164,18 @@ def put(self, object_id: str, object_content: bytes) -> None:
             validate_object_content(content=object_content)
 
         with self.session():
-            # Only allow adding the object if it has been preregistered
+            # The write is the authoritative preregistration check: if cleanup
+            # deleted the row concurrently, the UPDATE returns no rows and put
+            # must not report success.
+            rows = self.query(
```
msheller, line 169: I assume that the semantics in the comment will hold for any SQL DB implementation we use? If not, we need to capture that assumption somehow so we don't break by changing the implementation choice. I'm probably using the wrong terms, being not a SQL guy :)

msheller (dismissed):
No blocking comments, so LGTM!

panh99 (dismissed):
LGTM!

framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -164,7 +164,18 @@ def put(self, object_id: str, object_content: bytes) -> None:
             validate_object_content(content=object_content)
 
         with self.session():
-            # Only allow adding the object if it has been preregistered
+            # The write is the authoritative preregistration check: if cleanup
+            # deleted the row concurrently, the UPDATE returns no rows and put
+            # must not report success.
```
panh99:
```suggestion
            # UPDATE is the authoritative preregistration check: if cleanup
            # deleted the row concurrently, no row is updated and put must fail.
```

framework/py/flwr/supercore/object_store/sql_object_store.py
panh99: Yes, this should hold for both SQLite and PostgreSQL: if the UPDATE matches no rows, RETURNING produces no rows. MySQL might be different, but I think it's good for now.

panh99 (approved):

PR #7266, opened 2026-05-28
fix(framework): Prevent duplicate message replies

psfoley:
## Issue

`message_res.message_id` is unique, but `reply_to_message_id` was not. Concurrent or retried response pushes could therefore create multiple replies for one instruction, while downstream LinkState reads assume one reply per instruction.

## Proposal

This PR makes reply storage conflict on duplicates:

- Adds a unique index on `message_res.reply_to_message_id` and an Alembic migration.
- Keeps `SqlLinkState.store_message_res` returning `None` when a duplicate reply insert loses the race.
- Mirrors the same one-reply behavior in `InMemoryLinkState`.
- Adds shared LinkState coverage to verify the first reply wins and a second reply is rejected.

## Validation

- `uv run python -m pytest py/flwr/server/superlink/linkstate/linkstate_test.py -k duplicate_reply`
- `uv run python -m pytest py/flwr/supercore/state/alembic/utils_test.py -k migrated_schema_matches_metadata`


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -14,7 +14,6 @@
 # ==============================================================================
 """In-memory LinkState implementation."""
```
chongshenng, line 16:
```suggestion


```

framework/py/flwr/supercore/state/alembic/versions/rev_2026_05_28_unique_message_res_reply_to.py
```diff
@@ -0,0 +1,73 @@
...
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# pylint: disable=no-member
+
+# revision identifiers, used by Alembic.
+revision: str = "b2f7c9e5a4d1"
+down_revision: str | Sequence[str] | None = "c8f4f6e2c1ad"
```
chongshenng: We can safely set the `down_revision` to `f934b49300f8`.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_05_28_unique_message_res_reply_to.py
```diff
@@ -0,0 +1,73 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Ensure each instruction has at most one reply.
+
+Revision ID: b2f7c9e5a4d1
+Revises: c8f4f6e2c1ad
```
chongshenng:
```suggestion
Revises: f934b49300f8
```
Let's set it to the latest revision ID.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_05_28_unique_message_res_reply_to.py
psfoley: Updated!

framework/py/flwr/supercore/state/alembic/versions/rev_2026_05_28_unique_message_res_reply_to.py
psfoley: Updated!

chongshenng (approved):

msheller (approved):
LGTM!

Note, would be nice if the resolved bot comment about existing DBs with duplicate rows had a brief reply. I was about the add that comment myself when I saw the resolved comment. Would help me know why it was okay to resolve.

framework/py/flwr/supercore/state/alembic/versions/rev_2026_05_28_unique_message_res_reply_to.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.create_index(
+        "idx_message_res_reply_to_message_id_unique",
+        "message_res",
+        ["reply_to_message_id"],
+        unique=True,
+    )
```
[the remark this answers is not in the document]
psfoley: This is addressed by adding a preflight duplicate check before creating the unique index. If an existing DB already has multiple `message_res` rows for the same `reply_to_message_id`, the migration now raises a targeted `RuntimeError` listing the affected instruction IDs instead of failing with a generic index-creation error.

Fixing duplicates is a manual operation for now rather than deleting or merging rows in the migration because choosing which duplicate reply to preserve is a data-policy decision (@msheller this would likely be a good candidate to encode as a policy in your proposal)

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7269, opened 2026-05-28
feat(framework): Add K8s Launch implementation

msheller:
### Description

This PR adds the first Kubernetes SuperExec executor launch implementation on top of the mocked Kubernetes executor test PR.

### Related issues/PRs

Builds on #7235.

### Explanation

This upgrades `KubernetesExecutor` from object-construction tests into a production-shaped launch path for TaskExecutor Pods.

The executor now renders configurable TaskExecutor Pod metadata and placement fields, including resource requests/limits, stable labels, optional extra labels and annotations, resource-pool identity, node selectors, tolerations, affinity, priority class, optional service account, and optional Pod/container security contexts.

The PR also maps immediate Kubernetes submission outcomes into structured `LaunchResult` values: accepted, capacity rejected, failed, or unknown. Capacity waiting, cleanup, reconciliation, and production Helm/Kapitan wiring remain follow-up work.

### Checklist

- [x] Write tests
- [x] Update [documentation](https://flower.ai/docs/writing-documentation.html) (not needed; no user-facing docs changed)
- [x] Make CI checks pass


framework/py/flwr/supercore/superexec/executor/kubernetes_executor_test.py
```diff
@@ -300,31 +477,68 @@ def test_launch_submits_secret_before_pod_and_returns_accepted() -> None:
...
+    )
+
+    result = KubernetesExecutor(client=client, config=_executor_config()).launch(
+        _execution_spec()
+    )
+
+    assert result.status == LaunchResultStatus.UNKNOWN
+    assert result.message == "_KubernetesApiError: service unavailable"
     client.create_namespaced_secret.assert_called_once()
     client.create_namespaced_pod.assert_called_once()
```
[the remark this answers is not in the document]
msheller: Low value test. Not needed.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -227,11 +252,150 @@ def _credential_secret_name(spec: ExecutionSpec) -> str:
...
+    """Map immediate Kubernetes API exceptions to launch results."""
+    message = f"{type(exc).__name__}: {exc}"
+    status = _exception_status(exc)
+    lower_message = message.lower()
+
+    if isinstance(exc, (ConnectionError, TimeoutError)):
+        return LaunchResult.unknown(message)
+
+    if status == 429 or _is_capacity_message(lower_message):
+        return LaunchResult.capacity_rejected(message)
```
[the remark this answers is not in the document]
msheller: Retries are being added for runs that have heartbeat expiry while in the STARTING state, but the token is different in the retries. This means that we must not use the recommended solution. However, the issue holds, so we have added a different solution that tags secrets with an attempt hash so that stale secrets are not used (and we delete them when we can be certain it is safe to do so).

Good comment, just can't use the solution because of changes in other PRs.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -53,25 +62,39 @@ class KubernetesExecutorConfig:
...
+            "Resource pool", self.resource_pool, reject_outer_whitespace=True
+        )
+        _validate_optional_string(
+            "Priority class name",
+            self.priority_class_name,
+            reject_outer_whitespace=True,
+        )
+        _validate_optional_labels(self.labels)
+        _validate_optional_string_map("Kubernetes annotations", self.annotations)
+        _validate_optional_string_map("Node selector", self.node_selector)
```
chongshenng: There seem to be quite a lot of indirection here, which makes the logic slightly hard to read. Are these necessary? Can they be simplified?

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -53,25 +62,39 @@ class KubernetesExecutorConfig:
...
+    resource_pool: str | None = None
+    resources: JSONObject | None = None
+    node_selector: dict[str, str] | None = None
+    tolerations: list[JSONObject] | None = None
+    affinity: JSONObject | None = None
+    priority_class_name: str | None = None
+    pod_security_context: JSONObject | None = None
+    container_security_context: JSONObject | None = None
     # Optional Pod field only; service account policy/RBAC is decided elsewhere.
     service_account_name: str | None = None
```
chongshenng, line 108: Given that there are many new parameters, I think it would be useful to document the parameters in the `KubernetesExecutorConfig` docstring.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -135,6 +170,7 @@ def _build_taskexecutor_pod(
     spec: ExecutionSpec,
     config: KubernetesExecutorConfig,
     appio_root_certificates: str | None,
+    launch_attempt: str,
```
psfoley: @msheller what is the intent of `launch_attempt`? That name gives me the impression is the attempt at a launch of the application and I would expect something like sequential numbering were the launch to fail so that the admin knows how many launches have been reattempted. Looking at how this gets assigned, should I think of this more as a `launch_id`?

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -96,14 +119,27 @@ def launch(self, spec: ExecutionSpec) -> LaunchResult:
         """Submit the TaskExecutor Pod and credential Secret."""
         try:
             appio_root_certificates = _get_appio_root_certificates(spec, self._config)
+            launch_attempt = _new_launch_attempt_suffix()
```
chongshenng:
```suggestion
            launch_attempt_id = _new_launch_attempt_suffix()
```

I think using `launch_attempt_id` is clearer and consistent with how we use `uuid4` for `_id` variable names in our repo. Wdyt?

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -217,21 +269,199 @@ def _get_appio_root_certificates(
...
+    try:
+        client.delete_namespaced_secret(secret_name, namespace)
+    except Exception:  # pylint: disable=broad-exception-caught
+        LOGGER.warning(
+            "Failed to delete Kubernetes credential Secret %r in namespace %r",
+            secret_name,
+            namespace,
+            exc_info=True,
+        )
+
```
chongshenng, line 442: Since this is related to our Flower infrastructure, should we use our Flower logger here with `log(WARNING, ...)`?

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -29,6 +32,9 @@
 APPIO_CREDENTIALS_MOUNT_PATH = "/run/flwr/appio"
 APPIO_TOKEN_FILE_PATH = f"{APPIO_CREDENTIALS_MOUNT_PATH}/token"
 APPIO_ROOT_CERTIFICATES_FILE_PATH = f"{APPIO_CREDENTIALS_MOUNT_PATH}/ca.crt"
+LAUNCH_ATTEMPT_LABEL = "flower.ai/launch-attempt"
+
+LOGGER = logging.getLogger(__name__)
```
chongshenng: Should we use our Flower logger here (with `from flwr.common import log`) instead of redefining the logger?

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -29,6 +32,9 @@
 APPIO_CREDENTIALS_MOUNT_PATH = "/run/flwr/appio"
 APPIO_TOKEN_FILE_PATH = f"{APPIO_CREDENTIALS_MOUNT_PATH}/token"
 APPIO_ROOT_CERTIFICATES_FILE_PATH = f"{APPIO_CREDENTIALS_MOUNT_PATH}/ca.crt"
+LAUNCH_ATTEMPT_LABEL = "flower.ai/launch-attempt"
```
chongshenng, line 36: Given that we are introducing more constants here, should we move them to `flwr.supercore.constant`?

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: Will add and close when completed.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: The helper functions are not needed. I find them slightly easier to read/reason about than the previous version (you can see the diff for how this looked without the helpers, noting that the previous version had less to validate, so it would be about 2x what's there).

LMK what you think. I'm fine either way.

As far as necessity, __post_init validation is useful when we have values that are "invariant" and for which there are illegal values. In this case, I chose to have shallow __post_init checks to prevent malformed configs/schemas because I'm not sure yet to what degree we can just rely on the caller to get things right. I can imagine that the final version of this doesn't need validation, as the configs may be coming from sources that have their own validation/schema checks.

So I can see choosing to remove the validation entirely, but I do like having them in place until we have a full picture of how these configs are populated and passed.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: Good catch! Will adopt and close when pushed.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: Agree. Will adopt.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: Yes, good catch. This is definitely a mistake on my part.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
chongshenng: If we can simplify the validations to the absolute minimum, I think we should do that first to help with maintainability. I'd prefer to avoid introducing overly strict validation logic and then needing to simplify/refactor later.

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: My take is that if I don't see a case where it needs to be shared, I keep it local to where it is used. LMK if I should take a different approach, or if you think this is something that will be used by other code.

psfoley (commented):
@msheller I think this mostly looks good. I assume the wiring that allows the `KubernetesExecutor` to actually be used (`--executor kubernetes`) is reserved for a follow up PR?

As an aside, it appears there's also some potential for duplicate pods to be submitted for the same task id:

1. SuperExec claims task 123.
2. `KubernetesExecutor.launch()` creates Pod/Secret with a random launch_attempt suffix.
3. Kubernetes accepts the Pod, but the Pod stays Pending or starts slowly.
4. In PR #7303, if the task remains STARTING past heartbeat expiry, CoreState moves it back to PENDING and removes the old token.
5. SuperExec can claim task 123 again, gets a new token, and launches another Pod with a new launch_attempt.
6. The old Pod may still exist and may later start.

But I think this should be okay because the RPCs should invoke `get_task_by_token(token)`, and only the new pod should be considered valid. Do I have this right?

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: Accepted

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
msheller: I double-checked that the claim is atomic, so we shouldn't have a race condition issue.

psfoley (approved):
Approved after the latest fixes, and getting further confirmation on the scope offline!

framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -217,21 +297,172 @@ def _get_appio_root_certificates(
...
-def _pod_name(spec: ExecutionSpec) -> str:
+def _new_launch_attempt_id() -> str:
+    """Return a DNS-label-safe opaque identifier for one local launch call."""
+    return uuid4().hex[:12]
+
+
+def _pod_name(spec: ExecutionSpec, launch_attempt_id: str) -> str:
     """Return the TaskExecutor Pod name."""
-    return f"flwr-taskexecutor-{spec.task_id}"
+    return f"flwr-taskexecutor-{spec.task_id}-{launch_attempt_id}"
```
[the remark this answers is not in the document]
msheller: Not an issue as the expired token is rejected when reporting.

danieljanes (approved):

PR #7282, opened 2026-05-29
fix(framework): Prevent heartbeats reviving deleted nodes

psfoley:
## What
Make SqlLinkState.acknowledge_node_heartbeat a single conditional write. The heartbeat now updates the node only when the row still exists and is not UNREGISTERED, and returns False if a concurrent delete_node already made the node terminal.

## Why
The previous SQL path read node status first, then updated by node_id only. In a multi-SuperLink deployment, a heartbeat could read a registered/offline node, another replica could unregister it, and the heartbeat could then set it back to online from stale state.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #7284, opened 2026-05-29
fix(framework): Preserve PushMessages upload hints

psfoley:
### Description

`ServerAppIo.PushMessages` built one batch-wide `objects_to_push` set before knowing whether each message was accepted. If an earlier message was accepted and a later message was rejected because the run finished concurrently, rejection cleanup could remove upload hints for object IDs shared with the accepted message.

This matters for HA SuperLink deployments because an accepted message can remain in LinkState while the response no longer asks the ServerApp to upload all objects needed by that message.

### Explanation

This PR tracks upload hints per message and only adds them to the response after the message write is accepted. If a later batch item is rejected during run-finished cleanup, its cleanup no longer removes hints needed by earlier accepted messages.

### Checklist

- [x] Write tests
- [x] Run focused ServerAppIo tests


panh99 (approved):
LGTM!

danieljanes (approved):

PR #7285, opened 2026-05-29
fix(framework): Guard startup migrations on Postgres

psfoley:
## Issue

Multiple SuperLink replicas can start against the same shared Postgres database and all try to run Alembic state migrations at the same time. That can make startup contend on schema inspection or DDL even though only one process should mutate the schema.

Affected boundary: Flower state Alembic startup migrations used by `LinkState`, `CoreState`, and `ObjectStore` initialization.

## Proposal

This PR wraps `run_migrations` in a backend-specific migration guard:

- PostgreSQL acquires a blocking `pg_advisory_lock` before any Alembic inspection, stamping, or upgrade work.
- The locked PostgreSQL connection is reused for migration inspection and Alembic execution, so single-connection pools do not deadlock while holding the guard.
- The advisory lock is released in `finally`, including when migration execution fails.
- SQLite and in-memory startup paths keep the existing behavior and do not acquire a guard.

No schema migration, data backfill, or user-facing config change is included.


framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -14,14 +14,13 @@
 # ==============================================================================
 """Tests for Alembic migration helpers."""
```
chongshenng, line 16:
```suggestion


```

chongshenng (approved):
LGTM!

panh99 (approved):

PR #7288, opened 2026-05-30
refactor(framework): Move Server/ClientAppCallable and two mods

danieljanes:
This PR moves and/or deprecates several APIs.

Keep:
- `arrays_size_mod`, `message_size_mod`: `flwr.client.mod.comms_mods` → `flwr.clientapp.mod.comms_mods`  
  Public import: `flwr.clientapp.mod`. Compatibility re-export remains in `flwr.client.mod`.
- `ClientAppCallable`, `Mod`: `flwr.client.typing` → `flwr.clientapp.typing`
- `ServerAppCallable`: `flwr.server.typing` → `flwr.serverapp.typing`

Deprecate:
- `ClientFn`, `ClientFnExt`: `flwr.client.typing` → `flwr.compat.client.typing`  
  Compatibility re-export remains in `flwr.client`.
- `ServerFn`, `Workflow`: `flwr.server.typing` → `flwr.compat.server.typing`
- `ServerAppComponents`: `flwr.server` → `flwr.compat.server`


framework/py/flwr/__init__.py
```diff
@@ -31,14 +37,17 @@
 __version__ = _package_version
 
 
-# Lazy imports for legacy support
-_lazy_imports = {"simulation", "server", "client", "common"}
```
danielnugraha: Why do we change to `lazy_exports`, remove the other lazy imports and leave simulation?

framework/py/flwr/common/__init__.py
```diff
@@ -90,12 +90,13 @@
 
 def __getattr__(name: str) -> Any:
     """Lazily resolve compatibility exports that depend on app.message."""
-    if name not in _LAZY_EXPORTS:
-        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
-    module_name, attr_name = _LAZY_EXPORTS[name]
-    value = getattr(import_module(module_name), attr_name)
-    globals()[name] = value
-    return value
+    if name in _LAZY_EXPORTS:
```
danielnugraha, line 93: Why do we need to change the existing code here?

framework/py/flwr/client/__init__.py
```diff
@@ -15,14 +15,24 @@
...
+
+from . import mod
+from .client import Client
+from .numpy_client import NumPyClient
+
+if TYPE_CHECKING:
+    from flwr.clientapp.client_app import ClientApp
+    from flwr.compat.client.app import start_client, start_numpy_client
+
+_LAZY_EXPORTS: dict[str, tuple[str, str | None]] = {
```
danielnugraha, line 31: All entries have concrete str, is it possible to change `str | None` to `str` and update `__getattr__`?

framework/py/flwr/client/__init__.py
```diff
@@ -15,14 +15,24 @@
...
+    "ClientApp": ("flwr.clientapp.client_app", "ClientApp"),
+    "start_client": ("flwr.compat.client.app", "start_client"),
```
danielnugraha, line 34: Should we preserve the `# Deprecated` comment here?

framework/py/flwr/__init__.py
danieljanes: In a recent PR, we updated many internal imports. This made some of those lazy exports obsolete.

framework/py/flwr/common/__init__.py
danieljanes: It's not strictly necessary to change it, but part of what this PR does is to align the way lazy exports are implemented across all modules that use them. So it's a change to improve consistency.

framework/py/flwr/client/__init__.py
danieljanes: This type is the same across all modules that use lazy exports. It could be different here, but the `| None` option is necessary in other modules.

framework/py/flwr/client/__init__.py
danieljanes: Good idea, updated

danielnugraha (dismissed):

danielnugraha (approved):

tanertopal (approved):

PR #7291, opened 2026-05-31
feat(framework): Migrate to RunSeries Context

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/server/superlink/linkstate/linkstate_test.py
panh99, line 1: Let's remove the related unit tests in `linkstate_test.py`. The `get/set_run_series_context` methods are tested in the test class for `CoreState` already.

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -259,11 +259,13 @@ def PullTaskInput(
...
 
-        # Retrieve Context, Run and Fab for the run_id
-        serverapp_ctxt = state.get_serverapp_context(run_id)
+        # Retrieve Run, FAB, and shared RunSeries context for the run_id
         runs = state.get_run_info(run_ids=[run_id])
         run = runs[0] if runs else None
         fab = state.get_fab(run.fab_hash) if run and run.fab_hash else None
+        serverapp_ctxt = None
+        if run and run.series_id:
+            serverapp_ctxt = state.get_run_series_context(run.series_id)
```
panh99: Should we name it `series_context` or `series_ctxt`? This `PullTaskInput` RPC will also be used by other task processes, not just `flwr-serverapp`, right?

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -307,7 +309,13 @@ def PushTaskOutput(
...
             log(INFO, "Finished task %d of run %d", task.task_id, run_id)
             if request.HasField("context"):
-                state.set_serverapp_context(run_id, context_from_proto(request.context))
+                runs = state.get_run_info(run_ids=[run_id])
+                run = runs[0] if runs else None
+                if run and run.series_id:
+                    state.set_run_series_context(
+                        run.series_id,
+                        context_from_proto(request.context),
+                    )
```
panh99, line 318: As discussed, should we only allow the primary task to store the context? I.e., check if `run.primary_task_id == task.task_id`.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -281,8 +281,13 @@ def StartRun(  # pylint: disable=too-many-locals, too-many-branches, too-many-st
...
 
-            # Register the context at the LinkState
-            state.set_serverapp_context(run_id=run_id, context=context)
+            # Register the context for the run series. If the run joins an
+            # existing series, preserve its shared state while refreshing
+            # run-specific fields for the newly created run.
+            existing_context = state.get_run_series_context(series_id)
+            if existing_context is not None:
+                run_context.state = existing_context.state
+            state.set_run_series_context(series_id=series_id, context=run_context)
```
panh99: I wonder if we should initialise the series context in the `CoreState` method to avoid race conditions.

With the current logic, if the task process picks up the run after it is created at line 244 but before the initial series context is set, it may encounter unexpected errors because the series context is unavailable.

A better approach might be to handle this in `LinkState.create_run`: after calling `self.store_run_in_series`, we call `context = self.get_run_series_context(...)`. If `context` is `None`, we initialise it with `context = Context(...)`; otherwise, we simply refresh the `run_id` field. We then call `self.set_run_series_context(...)`.

This logic could be moved into a private helper method to keep `self.create_run` clean.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7303, opened 2026-06-01
feat(framework): Move `STARTING` tasks that miss heartbeats back to `PENDING`

panh99:


framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -583,13 +629,66 @@ def test_get_tasks_expires_stale_task_tokens(self) -> None:
...
+                statuses=[Status.PENDING],
+                order_by="pending_at",
+                ascending=True,
+            )
+
+        self.assertEqual(
+            [task.task_id for task in tasks], [revived_task_id, later_task_id]
+        )
+        self.assertEqual(tasks[0].pending_at, revived_pending_at)
+
```
panh99:
```suggestion
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -595,17 +595,21 @@ def _cleanup_expired_task_tokens_locked(self) -> None:
...
         """
         current = now()
         expired_tasks: list[Task] = []
         for task_id, record in list(self.task_token_store.items()):
             if record.active_until < current:
-                # The task is considered expired. Mark it as finished with a failed
-                # status if it's not already finished, and remove the token.
                 task = self.task_store.get(task_id)
-                if task and task.status.status != Status.FINISHED:
+                if task and task.status.status == Status.STARTING:
```
msheller, line 607: Unless migration in the DB a major headache, I would want to add something like "task.claimed_count", or track as internal state if editing the task object is bad. (It's a protobuf here, right?). Then the claim task implementation could increment that value as part of claiming.

msheller: If we track as internal state, then we get some new code for in_memory_corestate that looks sometime like:
def __init__(...):
  ...
  self.task_start_attempt_counts: dict[int, int] = {}

def claim_task(...):
  ...
  with self.lock_task_store:
    ...
    self.task_start_attempt_counts[task_id] = (
            self.task_start_attempt_counts.get(task_id, 0) + 1
        )
    ...

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -830,18 +831,31 @@ def _claim_task_message_rows(
...
-        Expired tasks are marked as finished with a failed status, and their tokens are
-        removed.
+        Expired starting tasks are moved back to pending. Expired running tasks
+        are marked as finished with a failed status. Tokens are removed in both
+        cases.
         """
         expired_at = now()
-        # Expired task claims are terminal failures and lose their token.
+        # Claims that never reached RUNNING are retryable launch failures.
+        self.query(
```
msheller, line 840: Similar to the comment for in_memory, I think we could have this query have an AND that checks the claim count. Would then have 3 queries here. One that looks for STARTING and claim count < MAX CLAIMS, and otherwise the same as we have here, one that looks for STARTING and claim count >= MAX CLAIMS and expires it as the query below this one, and then the query below this one for RUNNING tasks.

claim task would have an extra line to set claim count += 1 in the UPDATE SET section, I think.

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -528,20 +529,66 @@ def test_task_heartbeat_extends_token_expiration(self) -> None:
...
+        self.assertEqual(
+            tasks[0].status,
+            TaskStatus(status=Status.PENDING, sub_status="", details=""),
+        )
+        self.assertEqual(tasks[0].pending_at, pending_at)
+        self.assertEqual(tasks[0].starting_at, "")
+        self.assertEqual(tasks[0].running_at, "")
+        self.assertEqual(tasks[0].finished_at, "")
+        self.assertIsNone(state.get_task_by_token(token))
+        new_token = state.claim_task(task_id)
```
msheller, line 567: Worth asserting new_token != token here? Or does the assertIsNone above guarantee the old token is gone?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -566,12 +613,11 @@ def test_get_tasks_expires_stale_task_tokens(self) -> None:
...
         fixed_now = now()
-        active_until = fixed_now + timedelta(seconds=HEARTBEAT_DEFAULT_INTERVAL)
         run_id = self.task_run_id(state)
 
         with patch("datetime.datetime") as mock_dt:
             mock_dt.now.return_value = fixed_now
-            task_id = state.create_task(task_type="flwr-model", run_id=run_id)
+            task_id = state.create_task(task_type=TaskType.MODEL, run_id=run_id)
             assert task_id is not None
             assert state.claim_task(task_id) is not None
```
msheller: Should we save this token and check that get_task_by_token returns None later?

msheller (changes requested):
I think we should add the retry count limiting through counting how many times the task is claimed.

Otherwise, just a few minor comments on the tests. I don't think either change is critical, but could be worth the small updates (adds 3 lines I think).

framework/py/flwr/supercore/corestate/corestate_test.py
panh99: Yeah, that makes sense. In case the old token is revived accidentally. Good catch!

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -528,20 +529,66 @@ def test_task_heartbeat_extends_token_expiration(self) -> None:
...
+            TaskStatus(status=Status.PENDING, sub_status="", details=""),
+        )
+        self.assertEqual(tasks[0].pending_at, pending_at)
+        self.assertEqual(tasks[0].starting_at, "")
+        self.assertEqual(tasks[0].running_at, "")
+        self.assertEqual(tasks[0].finished_at, "")
+        self.assertIsNone(state.get_task_by_token(token))
+        new_token = state.claim_task(task_id)
+        self.assertIsNotNone(new_token)
+        assert new_token is not None
```
panh99:
```suggestion
        self.self.assertNotEqual(new_token, token)
        assert new_token is not None
```

framework/py/flwr/supercore/corestate/corestate_test.py
panh99: Addressed in https://github.com/flwrlabs/flower/pull/7303/changes/aebf1c78b4fb32b3952de1ce0eb25d9e29101345

framework/py/flwr/supercore/corestate/corestate_test.py
panh99:
```suggestion
            assert token := state.claim_task(task_id)
```

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -583,13 +629,32 @@ def test_get_tasks_expires_stale_task_tokens(self) -> None:
         self.assertEqual(len(tasks), 1)
```
panh99, line 630:
```suggestion
        self.assertIsNone(state.get_task_by_token(token))
        self.assertEqual(len(tasks), 1)
```

framework/py/flwr/supercore/corestate/corestate_test.py
panh99: 💯 , addressed in https://github.com/flwrlabs/flower/pull/7303/changes/5d41c7a428c0a0d2c01d3eec7c57a9aea6af4dec

framework/py/flwr/supercore/corestate/in_memory_corestate.py
panh99: As discussed, we will log the transition to `PENDING` and keep infinite retry.

framework/py/flwr/supercore/corestate/sql_corestate.py
panh99: As discussed, we will log the transition to `PENDING` and keep infinite retry.

danieljanes (approved):

msheller (approved):
Approved!

PR #7336, opened 2026-06-08
feat(framework): Add ConnectorRequest and ConnectorResponse

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/json_message/connector_message.py
```diff
@@ -0,0 +1,108 @@
...
+            if field not in payload:
+                raise ValueError(f"{cls.__name__} payload requires field '{field}'.")
+
+        error = payload["error"]
+        if error is not None and not isinstance(error, dict):
+            raise ValueError(
+                f"{cls.__name__} payload field 'error' must be a JSON object."
+            )
+        if error is not None and payload["output"] is not None:
+            raise ValueError(
```
panh99, line 105: [reviewer-bot] Important (API-001) framework/py/flwr/supercore/json_message/connector_message.py:105 confidence: medium

Can we reject the all-empty response state here? Right now `ConnectorResponse.from_message` accepts a payload with both `"output": null` and `"error": null`, so a caller cannot tell whether the connector succeeded with no result or failed to populate either branch. Since error responses already use `output=None`, the minimal fix would be to require a non-null `output` when `error` is `None`, or make a null success explicit in the schema.

framework/py/flwr/supercore/json_message/connector_message.py
```diff
@@ -0,0 +1,108 @@
...
+    """Task-routed connector request."""
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        *,
+        dst_task_id: int,
+        name: str,
+        call_id: str,
+        arguments: JSONObject,
+        ttl: float = DEFAULT_TASK_MESSAGE_TTL,
```
panh99: I wonder if we could use the same default TTL as `Message`, i.e., `from flwr.app.constants import DEFAULT_TTL`. Is there a reason for using a different ttl here?

panh99 (approved):
LGTM!

PR #7351, opened 2026-06-10
feat(framework): Add flwr-connector implementation

danielnugraha:
Adds the minimal standalone flwr-connector task process wiring.


framework/py/flwr/supercore/cli/flwr_connector_test.py
panh99, line 1: The unit tests look good to me. Just curious: most of the tests seem fairly simple and primarily verify that arguments are passed correctly. Do you think it would make sense to unify some of the unit tests across the `flwr-*` commands, or would that not provide much benefit? Just a thought—no need to make any changes in this PR.

framework/py/flwr/supercore/task_process/connector/__init__.py
```diff
@@ -12,13 +12,9 @@
...
 
-from collections.abc import Callable
+from .run_connector import run_connector
 
-from flwr.app import Context
-
-from .grid import Grid
-
-ServerAppCallable = Callable[[Grid, Context], None]
+__all__ = ["run_connector"]
```
panh99, line 20:
```suggestion
__all__ = [
    "run_connector"
]
```

framework/py/flwr/supercore/task_process/connector/run_connector.py
```diff
@@ -0,0 +1,187 @@
...
+from flwr.supercore.heartbeat import HeartbeatSender, make_task_heartbeat_fn_grpc
+from flwr.supercore.interceptors import (
+    AppIoTokenClientInterceptor,
+    RuntimeVersionClientInterceptor,
+)
+
+from .task import handle_task
+
+
+def run_connector(  # pylint: disable=R0912,R0913,R0914,R0915,R0917
```
panh99: Are all these pylint error codes needed here? I am curious about this b/c pylint won't warn about unnecessary `pylint: disable`.

framework/py/flwr/supercore/task_process/connector/run_connector.py
```diff
@@ -0,0 +1,187 @@
...
+
+    try:
+        heartbeat_sender = HeartbeatSender(make_task_heartbeat_fn_grpc(stub))
+        heartbeat_sender.start()
+
+        log(DEBUG, "[flwr-connector] Pull task input")
+        task_input: PullTaskInputResponse = stub.PullTaskInput(PullTaskInputRequest())
+
+        context = context_from_proto(task_input.context)
+        run = run_from_proto(task_input.run)
```
panh99: We don't need to convert the proto to dataclass as we only use the `run_id` field, right?

framework/py/flwr/supercore/task_process/connector/run_connector.py
```diff
@@ -0,0 +1,187 @@
...
+    AppIoTokenClientInterceptor,
+    RuntimeVersionClientInterceptor,
+)
+
+from .task import handle_task
+
+
+def run_connector(  # pylint: disable=R0912,R0913,R0914,R0915,R0917
+    serverappio_api_address: str,
+    log_queue: Queue[str | None] | None,
```
panh99: Let's make it non-optional since `flwr_connector` will always create a queue, right?

framework/py/flwr/supercore/task_process/connector/run_connector.py
```diff
@@ -0,0 +1,187 @@
...
+            heartbeat_sender.stop()
+
+        channel.close()
+
+    flwr_exit(
+        code=exit_code,
+        event_type=EventType.FLWR_CONNECTOR_RUN_LEAVE,
+        event_details={
+            "run-id-hash": hash_run_id,
+            "success": exit_code == ExitCode.SUCCESS,
```
panh99: I think it's fine to remove this key b/c `exit_code` is already in the telemetry event.

framework/py/flwr/supercore/cli/flwr_connector_test.py
danielnugraha: I think it would make more sense if we have a unified utility file that manages the `flwr-*` task process lifecycle. In that case we add the unit tests there and in each task process we only add unit tests about how they handle the tasks. Wdyt?

framework/pyproject.toml
```diff
@@ -83,6 +83,7 @@ flwr-serverapp = "flwr.server.serverapp:flwr_serverapp"
 flwr-clientapp = "flwr.supernode.cli:flwr_clientapp"
 flwr-agentapp = "flwr.supercore.cli:flwr_agentapp"
 flwr-model = "flwr.supercore.cli:flwr_model"
+flwr-connector = "flwr.supercore.cli:flwr_connector"
```
panh99, line 86: This adds the `flwr-connector` entry point, but SuperExec still cannot launch connector tasks because `TaskType.CONNECTOR` is not in the task-type launch mappings. `SubprocessExecutor.launch()` will still raise `KeyError` for a connector task. Could we add connector to both mappings and update the existing unsupported-task test?

framework/py/flwr/supercore/cli/flwr_connector_test.py
panh99: Yeah, that makes sense to me

panh99 (dismissed):
LGTM!

danieljanes (approved):

PR #7360, opened 2026-06-11
feat(framework): Add web fetch connector

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


framework/py/flwr/supercore/task_process/connector/web_fetch.py
```diff
@@ -0,0 +1,234 @@
...
+            formatted_detail = detail
+        else:
+            formatted_detail = json.dumps(detail, separators=(",", ":"))
+        if status_code is None:
+            super().__init__(f"Web fetch provider request failed: {formatted_detail}")
+        else:
+            super().__init__(
+                f"Web fetch provider request failed: {status_code} "
+                f"{formatted_detail}"
+            )
```
panh99: Nit: this could improve readability
```suggestion
        if status_code is not None:
            formatted_detail = f"{status_code} {formatted_detail}"

        super().__init__(f"Web fetch provider request failed: {formatted_detail}")
```

framework/py/flwr/supercore/task_process/connector/web_fetch.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class WebFetchProviderError(RuntimeError):
+    """Error returned by the web fetch provider."""
+
+    def __init__(
+        self,
+        *,
+        code: str,
+        detail: JSONValue,
```
panh99: May I know why we use `JSONValue` here? I checked the usage in this file and it seems that we always pass a `str`.

framework/py/flwr/supercore/task_process/connector/web_fetch.py
```diff
@@ -0,0 +1,234 @@
...
+    content = (
+        trafilatura.extract(
+            text,
+            url=final_url,
+            output_format="markdown",
+            include_comments=False,
+            include_tables=True,
+        )
+        or text
+    )
```
panh99: The inline `or` seems to make the code harder to reaad.
```suggestion
    content = trafilatura.extract(
        text,
        url=final_url,
        output_format="markdown",
        include_comments=False,
        include_tables=True,
    )
    if content is None:
        content = text
```

framework/py/flwr/supercore/task_process/connector/web_fetch.py
```diff
@@ -0,0 +1,234 @@
...
+from flwr.supercore.typing import JSONObject, JSONValue
+
+_MAX_RESPONSE_BYTES = 1024 * 1024
+_TIMEOUT = 30.0
+_MAX_REDIRECTS = 10
+_READ_CHUNK_SIZE = 64 * 1024
+_REDIRECT_STATUS_CODES = frozenset({301, 302, 303, 307, 308})
+
+
+class WebFetchProviderError(RuntimeError):
```
panh99, line 35: We raise this error many times. Will the caller `flwr-agenapp` know why the connector fails when the error is raised?

framework/py/flwr/supercore/task_process/connector/web_fetch.py
```diff
@@ -0,0 +1,234 @@
...
+            raise WebFetchProviderError(
+                code="too_many_redirects",
+                detail=f"Web fetch exceeded {_MAX_REDIRECTS} redirects.",
+            )
+        current_url = urljoin(response_url, location)
+
+    raise RuntimeError("This line should never be reached.")
+
+
+def _validate_url(url: str) -> str:
```
panh99, line 150: I might be missing something, but I don't quite understand the purpose of this validation. Why do we need this extensive validation before fetching the URL?

The function docstring doesn't seem to explain the rationale, and I couldn't find any comments describing the intent.

framework/py/flwr/supercore/task_process/connector/web_fetch.py
```diff
@@ -0,0 +1,234 @@
...
+        if status_code is None:
+            super().__init__(f"Web fetch provider request failed: {formatted_detail}")
+        else:
+            super().__init__(
+                f"Web fetch provider request failed: {status_code} "
+                f"{formatted_detail}"
+            )
+
+
+def invoke_web_fetch_provider(url: str) -> JSONObject:
```
panh99, line 56: Could we add some comments in the function?

framework/py/flwr/supercore/task_process/connector/web_fetch.py
danielnugraha: Good point. No we don't expose this to the caller. Should we have like a dedicated error code? One for agent, model, and connector? Right now we just have Error classes like this.

framework/py/flwr/supercore/task_process/connector/web_fetch.py
danielnugraha: We don't want the url to be endless redirects or not using https and blocked urls. I'll try to simplify and add comments to explain.

framework/py/flwr/supercore/task_process/connector/web_fetch.py
panh99: Yeah, I think it would be better to surface that to the caller in some way. No rush on this, of course. We don't really expect `AgentApp` to handle these errors for now, do we?

framework/py/flwr/supercore/task_process/connector/web_fetch.py
panh99: Yeah, I mean, endless redirects are already prevented in `_fetch_url`. What I'm curious about is the rationale for requiring HTTPS and blocking certain URLs. Those seem like fairly generic security restrictions rather than something specific to the agent workflow itself, no?

Also, it performs quite a few checks:

```python
ip_address.is_private
or ip_address.is_loopback
or ip_address.is_link_local
or ip_address.is_multicast
or ip_address.is_reserved
or ip_address.is_unspecified
```

So I'm wondering whether these protections are must-have at this layer.

framework/py/flwr/supercore/task_process/connector/web_fetch.py
panh99: As discussed, I will resolve this for now since no urgent need for surfacing the detailed error to the caller (`flwr-agentapp`).

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7363, opened 2026-06-11
refactor(framework): Move remaining types out of flwr.common.typing

danieljanes:
It moves the remaining types out of `flwr.common.typing` into new modules:
- `Run`, `RunStatus`, `RunNotRunningException`, `InvalidRunStatusException` -> `flwr.supercore.run`
- `Fab` -> `flwr.supercore.fab`
- `AccountAuthLoginDetails`, `AccountAuthCredentials`, `AccountInfo` -> `flwr.supercore.auth.typing`
- `Actor`, `Event`, `LogEntry` -> `flwr.supercore.event_log.typing`
- `Federation` -> `flwr.superlink.federation.typing`

PR #7364, opened 2026-06-12
refactor(examples): Update `UserConfig` imports in examples

jafermarq:
missing from https://github.com/flwrlabs/flower/pull/7363

### DIFF
--- a/examples/advanced-pytorch/pytorch_example/task.py
+++ b/examples/advanced-pytorch/pytorch_example/task.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from flwr.common.typing import UserConfig
+from flwr.app.user_config import UserConfig
 from flwr_datasets import FederatedDataset
 from flwr_datasets.partitioner import DirichletPartitioner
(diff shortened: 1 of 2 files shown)


examples/advanced-pytorch/pytorch_example/task.py
```diff
@@ -7,7 +7,7 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from flwr.common.typing import UserConfig
+from flwr.app.user_config import UserConfig
```
panh99: In examples, I think it might be easier to import these from `flwr.app` instead of the full path `flwr.app.user_config`, wdyt?

examples/whisper-federated-finetuning/whisper_example/server_app.py
```diff
@@ -6,8 +6,8 @@
 import torch
 from datasets import load_dataset
 from flwr.app import ArrayRecord, Context, Message, MetricRecord
+from flwr.app.user_config import UserConfig
```
panh99: Same as above. We may merge this into the `from flwr.app` above.

panh99 (approved):
LGTM!

PR #7370, opened 2026-06-12
fix(framework): Assign message IDs consistently across deployments

psfoley:
## What changed

This makes `message_id` assignment consistent across deployment modes by decoupling logical message identity from ObjectStore identity:

- `message_id` is now generated as a UUID for gRPC/runtime message creation paths.
- ObjectStore root IDs are tracked separately as `object_id` metadata for LinkState persistence and transport object retrieval.
- `GrpcGrid`, ServerAppIo, Fleet message handling, SuperNode runtime, and ClientAppIo now use the object tree root for ObjectStore fetch/delete/confirm operations instead of treating `message_id` as a content hash.
- LinkState SQL schema now includes nullable `object_id` columns for `message_ins` and `message_res`, with an Alembic migration backfilling existing rows from their historical `message_id` values.
- Regression coverage asserts that gRPC message IDs are UUIDs and distinct from ObjectStore roots, and that LinkState serialization round-trips both identifiers.

## Why

`GrpcGrid` had been using `msg.object_id` as `metadata.message_id`, while the in-memory path used random UUIDs. That made logical message identity deployment-dependent and allowed content-addressed IDs to leak into message semantics.

The fix gives the identifiers clear meanings regardless of deployment mode:

- `message_id`: random logical message identifier.
- `object_id`: content-addressed ObjectStore root identifier.


framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -161,7 +162,7 @@ def pull_messages(  # pylint: disable=too-many-locals
         except NoObjectInStoreError as e:
             log(ERROR, e.message)
             # Delete message ins from state
-            state.delete_messages(message_ins_ids={msg_object_id})
+            state.delete_messages(message_ins_ids={msg.metadata.message_id})
```
jafermarq, line 165: could we leave this line as it was before?

framework/py/flwr/supercore/message_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def assign_message_id(message: Message) -> None:
+    """Assign a fresh logical UUID message ID."""
+    set_message_id(message, str(uuid4()))
+
+
+def set_message_object_id(message: Message, object_id: str) -> None:
+    """Set the ObjectStore root ID associated with a Message."""
+    message.__dict__[MESSAGE_OBJECT_ID_KEY] = object_id
```
jafermarq, line 38: I wonder if it would better to introduce a new attribute in `Metadata` that helps us capture this (e.g. `messages.metadata.message_object_id`) ? (instead of injecting it like this). wdyt ?

jafermarq: in that way we probably won't need `get_message_object_id` ?

framework/py/flwr/server/superlink/serverappio/serverappio_servicer.py
```diff
@@ -146,6 +147,7 @@ def PushMessages(
             # Store objects
             message_objects_to_push = set(store.preregister(run_id, object_tree))
             # Store message
+            set_message_object_id(message, object_tree.object_id)
```
jafermarq, line 150: If we introduce a new `message_object_id` in `Metadata` (see my comments later in `message_utils.py`), we probably don't need this here? instead, we could set it before the `Message` is pushed (e.g. in the grid object)

framework/py/flwr/superlink/grid/grpc_grid.py
```diff
@@ -279,19 +280,20 @@ def push_messages(self, messages: Iterable[Message]) -> Iterable[str]:
         # Construct Messages
         run_id = cast(Run, self._run).run_id
         message_ids: list[str] = []
-        if not messages:
+        message_list = list(messages)
```
jafermarq, line 283: I was a bit concerned about this, but digging into the code made me realise that to `push_messages` we already pass a `list` of `Messages`, so there's no additional harm. cc @panh99, maybe we should update the signature of this and related methods?

PR #7371, opened 2026-06-13
fix(framework): Extend task heartbeat lease on activation

psfoley:
## Summary
- refresh task heartbeat lease when a claimed task transitions from STARTING to RUNNING
- keep STARTING task retry behavior unchanged
- add regression coverage for activation extending the task lease

## Why
A task can be activated while its first heartbeat is delayed by a transient SuperLink/backend disruption. Previously, the RUNNING task could still expire at the initial one-interval claim deadline before it received the normal heartbeat grace window.

## Validation
- uv run --project framework python -m pytest framework/py/flwr/supercore/corestate/corestate_test.py framework/py/flwr/server/superlink/linkstate/linkstate_test.py
- Remote HA validation: 100 runs, 50 SuperNodes, 10 federations, 5 SuperLinks, 1 SuperExec, deleted 3 SuperLink pods after submission; 100/100 finished:completed, 0 failures, 0 stuck runs


panh99 (approved):
LGTM!

danieljanes (approved):

PR #7372, opened 2026-06-13
refactor(framework): Enable installation of app dependencies in `SuperLink` running in subprocess mode

jafermarq:
- A `--disable-runtime-dependency-installation` flag is introduced to disable this behaviour
- App e2e test now rely on auto-install of dependencies


framework/py/flwr/supercore/cli/flower_superexec.py
```diff
@@ -46,6 +48,12 @@
 from flwr.supercore.utils import disable_process_dumping
 from flwr.supercore.version import package_version
 
+_SERVERAPP_PLUGIN_TYPES = {
```
jafermarq:
```suggestion
# Plugin types that install dependencies by default
_SERVERAPP_PLUGIN_TYPES = {
```

.github/workflows/framework-e2e.yml
```diff
@@ -336,13 +336,6 @@ jobs:
       - name: Bootstrap
```
panh99, line 336: [reviewer-bot] Important (OPS-001) .github/workflows/framework-e2e.yml:336 confidence: medium

Could we keep the CI-only dependency overrides in the app job even though the install now happens at runtime? `uv sync` will read the unmodified example `pyproject.toml` from the FAB, so this deletion drops the TensorFlow -> tensorflow-cpu rewrite and the PyTorch CPU wheel index that this job was using. That can make the quickstart app e2e resolve GPU/default-index wheels and burn the 10-minute job budget. A small pre-run step that rewrites the pyproject and/or sets the runtime index before `flwr run` would preserve the old CI behavior while still exercising runtime dependency installation.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7381, opened 2026-06-15
fix(framework): Make instruction message inserts idempotent

psfoley:
## Summary
- make duplicate instruction-message inserts idempotent when the existing row matches the incoming message
- ignore `delivered_at` during duplicate comparison because delivery state can change before a retried `PushMessages` call completes
- reject conflicting duplicate `message_id` values in both SQL and in-memory LinkState

## Why
A retried ServerAppIo `PushMessages` call can hit a `message_ins_message_id_key` duplicate after the first attempt already persisted the instruction message. This should be treated as a successful idempotent retry when the message is the same instead of failing the ServerApp task.


framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -164,45 +165,78 @@ def store_message_ins(self, message: Message) -> str | None:
...
 
-            # Only invalid run_id can trigger IntegrityError.
-            # This may need to be changed in the future version
-            # with more integrity checks.
-            self.query(query, data[0])
+        stored_message_id: str | None = message.metadata.message_id
+        existing_message = get_existing_message()
+        if existing_message:
+            stored_message_id = handle_duplicate_message(existing_message)
+        else:
```
jafermarq:
```suggestion
            return handle_duplicate_message(existing_message)
```

If we return early, we avoid having a `else` nesting. wdyt?

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -164,45 +165,78 @@ def store_message_ins(self, message: Message) -> str | None:
...
+            try:
+                with self.session():
```
jafermarq: I wonder if we should put the lines that call `get_existing_message()` also under the `self.session()`

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -147,37 +149,51 @@ def store_message_ins(self, message: Message) -> str | None:
...
-                log(ERROR, "Invalid run ID for Message: %s", message.metadata.run_id)
-                return None
-            if self._is_finished_run(message.metadata.run_id):
-                log(ERROR, "Invalid run ID for Message: %s", message.metadata.run_id)
-                return None
-            federation = self.run_ids[message.metadata.run_id].run.federation
+            existing = self.message_ins_store.get(message_id)
+            if existing:
+                if not is_same_message_ins(
+                    message_to_dict(existing), message_to_dict(message)
```
jafermarq: I'm a bit concern about the overheads of doing this, specially for simulations. However, is my understanding correct that in normal circumstances, we'll very rarely see two messages having the same `message_id` ? Because this requires messages having the exact same content and with almost all metadata fields matching (all except those set when the message is pushed -- i.e. `run_id`, `src_node_id`); cc @panh99

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -164,45 +165,78 @@ def store_message_ins(self, message: Message) -> str | None:
...
+                ERROR,
+                "Failed to store Message: conflicting duplicate message_id %s.",
+                message.metadata.message_id,
+            )
+            return None
 
-            # Only invalid run_id can trigger IntegrityError.
-            # This may need to be changed in the future version
-            # with more integrity checks.
-            self.query(query, data[0])
```
panh99, line 203: GitHub isn't handling the diff very well. Am I understanding correctly that this change:

1. Checks for duplicate messages at the beginning of the function, and
2. Handles duplicate messages again when an integrity error occurs,

while leaving the main body of the function unchanged?

If that's the case, could we simplify this by just wrapping `self.query(query, data[0])` in a `try-except` block and logging an error in the `except` branch when the integrity error is caused by a duplicate `message_id`? That would avoid the extra upfront check.

Since duplicate messages should be rare in production, the pre-check doesn't seem to add much value, and handling the error directly would keep the code simpler.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
panh99, line 142: Just curious, do we also need to make reply message insertion idempotent in a later PR?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: One question: since the message ID is derived from the message content, wouldn't a duplicate `message_id` already imply a duplicate message? If that's the case, I'm not sure the additional metadata comparison is offering us much. It seems like we could simply treat a duplicate `message_id` as a duplicate message and simply return the message ID to ensure idempotency, right?

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
psfoley: @panh99 yes, I agree this should be added later

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
psfoley: Agreed. I updated the PR to simplify this further. The SQL implementation no longer does the up-front duplicate lookup, so this nested branch is gone.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
psfoley: This path was removed in the latest update. There is no separate `get_existing_message()` call anymore; duplicate handling now happens only when the insert raises an `IntegrityError`.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
psfoley: Agreed - I've removed the metadata comparison in the latest push

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
psfoley: Done - I updated the PR to remove the up-front duplicate check and handle duplicate `message_id` only from the insert `IntegrityError`.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -148,13 +148,16 @@ def store_message_ins(self, message: Message) -> str | None:
 
         message_id = message.metadata.message_id
         with self.lock:
+            if message_id in self.message_ins_store:
```
panh99, line 151: [reviewer-bot] Blocking (DATA-001) framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py:151 confidence: high

This makes any duplicate `message_id` look successful before we check that the existing message is actually the same. If a caller reuses the ID with different content, run, or destination, the new instruction is silently dropped while `store_message_ins` returns success. Can we compare the stored message against the incoming one, ignoring only `delivered_at`, and return `None`/log an error when the duplicate conflicts?

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -164,43 +164,62 @@ def store_message_ins(self, message: Message) -> str | None:
...
+                        "Invalid destination node ID for Message: %s",
+                        message.metadata.dst_node_id,
+                    )
+                    return None
+
+                # Insert message
+                columns = ", ".join([f":{key}" for key in data[0]])
+                query = f"INSERT INTO message_ins VALUES({columns})"
+
+                self.query(query, data[0])
```
panh99, line 207: [reviewer-bot] Blocking (DATA-001) framework/py/flwr/server/superlink/linkstate/sql_linkstate.py:207 confidence: high

Catching the duplicate key here returns success even when the existing row does not match the incoming message. In that case the INSERT failed, but the caller gets the message id back and may assume a different instruction was scheduled. Can we load the existing `message_ins` row on duplicate, compare all persisted fields except `delivered_at`, and only treat it as idempotent when it matches?

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
panh99: Resolve this b/c message ID is computed based on its content, faking a message ID won't impact the messaging system as the receiver will not be able to pull anything.

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
panh99: Same as https://github.com/flwrlabs/flower/pull/7381#discussion_r3433949377

framework/py/flwr/server/superlink/linkstate/sql_linkstate.py
```diff
@@ -164,43 +164,62 @@ def store_message_ins(self, message: Message) -> str | None:
...
+                if constraint
+                else "message_ins.message_id" in str(orig)
+            )
+            if not is_duplicate_message_id:
+                raise
+            log(
+                ERROR,
+                "Message with message_id %s already exists.",
+                message.metadata.message_id,
+            )
```
panh99: For consistency with `InMemoryLinkState`. Plus, I think this is not an error; this is now expected behaviour of `store_message_ins`.
```suggestion
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7385, opened 2026-06-15
fix(datasets): Add shuffle option to IidPartitioner

### LINKED ISSUE #7327: [feature]: Support shuffling in IIDPartitioner or update documentation
### Describe the bug

When using IIDPartitioner on a custom local dataset, I discovered it partitions the data sequentially. While pre-shuffling the dataset resolves this, it would be great to add shuffle: bool = True and seed: int | None = 42 parameters to IIDPartitioner (similar to PathologicalPartitioner).

Alternatively, please update the documentation to explicitly mention the need for pre-shuffling.

### Steps/Code to Reproduce

import datasets
from flwr_datasets.partitioner import IidPartitioner
import collections

def test_iid_partitioner_with_sorted_data():
print("--- Reproduction Test: IidPartitioner with Sorted Data ---")

# 100 samples of Class 0, followed by 100 samples of Class 1
data = {
    "img_file": [f"path/to/img_{i}.png" for i in range(200)],
    "label": [0] * 100 + [1] * 100
}
ds = datasets.Dataset.from_dict(data)
print(f"Total dataset size: {len(ds)}")
print(f"First 5 labels: {ds['label'][:5]}")
print(f"Last 5 labels: {ds['label'][-5:]}")

num_nodes = 2
partitioner = IidPartitioner(num_partitions=num_nodes)
partitioner.dataset = ds

for i in range(num_nodes):
    partition = partitioner.load_partition(i)
    labels = partition["label"]
    counts = collections.Counter(labels)
    print(f"\nNode {i+1} Partition (Size {len(partition)}):")
    for label, count in sorted(counts.items()):
        print(f"  Class {label}: {count} images")
test_iid_partitioner_with_sorted_data()

### Expected Results

Should shuffle the ds or state in the description this is expected as a pre-condition

### Actual Results

Total dataset size: 200 │
│ First 5 labels: [0, 0, 0, 0, 0] │
│ Last 5 labels: [1, 1, 1, 1, 1] │
│ │
│ Node 1 Partition (Size 100): │
│ Class 0: 100 images │
│ │
│ Node 2 Partition (Size 100): │
│ Class 1: 100 images

WilliamLindskog:
## What changed

- Add optional `shuffle` and `seed` parameters to `IidPartitioner`
- Preserve the existing contiguous-slice behavior by default (`shuffle=False`)
- Cache the shuffled dataset per partitioner instance so repeated partition loads stay stable, including when `seed=None`
- Document the sorted-local-dataset case where shuffling before IID partitioning avoids label-skewed partitions

## Issue/PR mapping

- Fixes #7327
- Supersedes the focused shuffle behavior in #7329

## Validation

- `pytest`, `ruff`, `mypy`, and `black --check` on the touched IID partitioner files
- `git diff --check`


WilliamLindskog:
Update: I split this into the lower-friction review path.

- #7385 is now shuffle-only: `IidPartitioner(shuffle, seed)` plus docs/tests for sorted local datasets.
- The partition-skew metrics work moved to draft follow-up #7389, stacked on this branch so its diff is metrics-only.

Focused validation passed for the touched IID partitioner files: `pytest`, `ruff`, `mypy`, `black --check`, and `git diff --check`.

datasets/docs/source/index.rst
```diff
@@ -105,7 +105,7 @@ Thanks to using Hugging Face's ``datasets`` used under the hood, Flower Datasets
 Here are a few of the ``Partitioners`` that are available: (for a full list see `link <ref-api/flwr_datasets.partitioner.html#module-flwr_datasets.partitioner>`_ )
 
 * Partitioner (the abstract base class) ``Partitioner``
-* IID partitioning ``IidPartitioner(num_partitions)``
+* IID partitioning ``IidPartitioner(num_partitions, shuffle=False, seed=42)``
```
jafermarq: typically we don't include these arguments in this doc page

datasets/flwr_datasets/partitioner/iid_partitioner.py
```diff
@@ -20,12 +20,20 @@
 
 
 class IidPartitioner(Partitioner):
-    """Partitioner creates each partition sampled randomly from the dataset.
+    """Partitioner creates IID partitions from a dataset.
+
+    By default, partitions are contiguous shards of the dataset to preserve historical
+    behavior. Set ``shuffle=True`` to shuffle the dataset once before sharding. This is
+    useful for local datasets sorted by class or another target column.
```
jafermarq:
```suggestion
    By default, partitions are contiguous shards of the dataset. Set ``shuffle=True`` to shuffle the dataset once before sharding. This is
    useful for local datasets sorted by class or another target column.
```

I think we can drop those words from the docstrings

datasets/flwr_datasets/partitioner/iid_partitioner.py
```diff
@@ -56,11 +69,20 @@ def load_partition(self, partition_id: int) -> datasets.Dataset:
...
     def num_partitions(self) -> int:
         """Total number of partitions."""
         return self._num_partitions
+
+    def _dataset_to_partition(self) -> datasets.Dataset:
+        """Return the dataset used for sharding."""
+        if not self._shuffle:
+            return self.dataset
+        if self._shuffled_dataset is None:
+            self._shuffled_dataset = self.dataset.shuffle(seed=self._seed)
```
jafermarq, line 87: does this mean we have two copies of the dataset?

datasets/README.md
```diff
@@ -39,7 +39,7 @@ Thanks to using Hugging Face's `datasets` used under the hood, Flower Datasets i
 Create **custom partitioning schemes** or choose from the **implemented [partitioning schemes](https://flower.ai/docs/datasets/ref-api/flwr_datasets.partitioner.html#module-flwr_datasets.partitioner)**:
 
 * Partitioner (the abstract base class) `Partitioner`
-* IID partitioning `IidPartitioner(num_partitions)`
+* IID partitioning `IidPartitioner(num_partitions, shuffle=False, seed=42)`
```
jafermarq:
```suggestion
* IID partitioning `IidPartitioner(num_partitions)`
```

datasets/README.md
```diff
@@ -78,4 +78,8 @@ partition = fds.load_partition(0)
 centralized_data = fds.load_split("test")
 ```
 
+If a local dataset is sorted by label or another target column, use
+`IidPartitioner(num_partitions=100, shuffle=True, seed=42)` to shuffle the dataset
+once before sharding it.
+
```
jafermarq:
```suggestion
```

We can remove this from there. The idea of this readme is to keep things minimal

datasets/flwr_datasets/partitioner/iid_partitioner.py
WilliamLindskog: Good question. `Dataset.shuffle(...)` returns another Hugging Face `Dataset` object with shuffled indices/cache metadata rather than eagerly duplicating all row data. So this keeps a second dataset object around, but it should not be a full in-memory copy of the underlying dataset. The cache here is intentional so repeated `load_partition` calls use the same shuffled order, especially when `seed=None`.

PR #7386, opened 2026-06-15
docs(framework): Document how runtime installation of Flower Apps dependencies works

jafermarq:
Ideally merged after: https://github.com/flwrlabs/flower/pull/7372


framework/docs/source/how-to-install-app-dependencies-at-runtime.rst
```diff
@@ -0,0 +1,99 @@
...
+use this metadata to install app dependencies automatically before the app is executed.
+
+Runtime dependency installation is useful when different apps need different Python
+packages, or when the machine running Flower does not have every app dependency
+preinstalled. For production deployments where startup time, network access, or exact
+image contents matter, you can still preinstall dependencies and disable runtime
+installation.
+
+.. note::
+
```
panh99, line 19: [reviewer-bot] Important (PROD-001) framework/docs/source/how-to-install-app-dependencies-at-runtime.rst:19 confidence: High

This documents the post-#7372 behavior, but in the current target branch SuperLink still defaults runtime dependency installation to disabled and only exposes `--allow-runtime-dependency-installation`; there is no `--disable-runtime-dependency-installation` or `FLWR_DISABLE_RUNTIME_DEPENDENCY_INSTALLATION` handling. If this lands as-is before the implementation PR, users will get the opposite default plus unsupported flags/env vars. Can we either merge/rebase this after the implementation is in the base branch, or keep the wording to behavior that is already shipped?

framework/docs/source/how-to-install-app-dependencies-at-runtime.rst
```diff
@@ -0,0 +1,99 @@
...
+    ``ServerApp`` processes, but disabled by default in SuperNode for ``ClientApp``
+    processes. You may pass ``--disable-runtime-dependency-installation`` to your
+    SuperLink to disable dependency installation or set
+    ``FLWR_DISABLE_RUNTIME_DEPENDENCY_INSTALLATION=1`` before starting it, and
+    ``--allow-runtime-dependency-installation`` to enable it on SuperNodes.
+
+.. note::
+
+    In SuperGrid, automatic dependency installation is enabled. For :doc:`SuperNodes you
+    connect to SuperGrid <how-to-connect-supernodes-to-supergrid>`, runtime dependency
```
panh99:
```suggestion
    In SuperGrid, automatic dependency installation is enabled. For :doc:`SuperNodes
    connected to SuperGrid <how-to-connect-supernodes-to-supergrid>`, runtime dependency
```

framework/docs/source/how-to-install-app-dependencies-at-runtime.rst
```diff
@@ -0,0 +1,99 @@
...
+installation.
+
+.. note::
+
+    Automatic dependency installation is enabled by default in SuperLink for
+    ``ServerApp`` processes, but disabled by default in SuperNode for ``ClientApp``
+    processes. You may pass ``--disable-runtime-dependency-installation`` to your
+    SuperLink to disable dependency installation or set
+    ``FLWR_DISABLE_RUNTIME_DEPENDENCY_INSTALLATION=1`` before starting it, and
+    ``--allow-runtime-dependency-installation`` to enable it on SuperNodes.
```
panh99: The note reads slightly circular. Should we reword it like this?
```suggestion
    SuperLink enables automatic dependency installation for Flower Apps by
    default. SuperNode disables it by default.
    
    To disable this behavior in SuperLink, pass
    ``--disable-runtime-dependency-installation`` or set
    ``FLWR_DISABLE_RUNTIME_DEPENDENCY_INSTALLATION=1`` before starting it. To
    enable it in SuperNode, pass ``--allow-runtime-dependency-installation``.
```

framework/docs/source/how-to-install-app-dependencies-at-runtime.rst
```diff
@@ -0,0 +1,99 @@
...
+
+.. note::
+
+    In SuperGrid, automatic dependency installation is enabled. For :doc:`SuperNodes you
+    connect to SuperGrid <how-to-connect-supernodes-to-supergrid>`, runtime dependency
+    installation remains disabled by default, but you can decide whether to enable it.
```
panh99: I think the current wording could be read as saying that runtime dependency installation is disabled on SuperNodes because they are connected to SuperGrid. But the SuperNode default is independent of SuperGrid; SuperGrid enabling automatic dependency installation does not require connected SuperNodes to enable it. Maybe we can phrase this as two separate points: SuperGrid enables it by default, and each SuperNode operator decides whether to enable it on their side (or we just remove the bit about SuperNode in this note cuz it's covered in the note above already).

jafermarq: make sense to me. better to make it clear 👍

panh99 (approved):
LGTM!

PR #7388, opened 2026-06-16
fix(examples): Avoid torch dataset formatter

WilliamLindskog:
## What changed

- Remove `Dataset.with_format("torch")` from the PyTorch example paths that combine Hugging Face Datasets with torchvision transforms.
- Add a shared lazy tensor transform for the Whisper example so encoded `data` and `targets` are still returned as PyTorch tensors without using the global torch formatter.

## Why

This is a follow-up to #7330. The E2E blocker was fixed there, but a few examples still used `with_format("torch")`, which can trigger Hugging Face Datasets' torch formatter import path and the unsupported `torchvision.io.VideoReader` import under some dependency combinations.

This PR is intentionally examples-only and does not change package minimum versions.

## Validation

- `ruff`, `black --check`, and `py_compile` on touched example Python files
- `git diff --check`


WilliamLindskog:
Addressing the scope/dependency concern explicitly:

- I narrowed this PR to examples-only. It now only touches the PyTorch/custom-mods/Whisper example files; the docs and FedDebug baseline changes were removed.
- I do not think this needs a minimum torch/torchvision version bump. The PR removes `Dataset.with_format("torch")`; it does not rely on a newer torch/torchvision API. The replacement paths use existing example transforms and, for Whisper, `torch.as_tensor` to preserve tensor output for `data`/`targets`.
- The motivation is to avoid the Hugging Face Datasets torch formatter import path that can hit `torchvision.io.VideoReader` under some dependency combinations, while keeping the example behavior equivalent.

PR #7397, opened 2026-06-16
feat(framework): Allow serverapp-ephemeral to run flwr-connector

danielnugraha:
Add flwr-connector to supported task type in serverapp-ephemeral.


panh99 (approved):
LGTM!

PR #7402, opened 2026-06-16
feat(framework): Align examples docs with baselines

Julianrussmeyer:
## Summary

- Align the Examples docs Furo theme settings with the Baselines docs treatment.
- Add the Examples light/dark logo assets referenced by the theme configuration.
- Align Examples custom CSS with Baselines typography, sidebar active state, and community CTA styling.
- Update `devtool.update_html_themes` to merge generated light/dark CSS variables into existing `html_theme_options` dictionaries without appending duplicate keys or overwriting doc-specific values.

This PR intentionally leaves the generated Examples docs content structure and navigation unchanged.

## Verification

- `cd dev && /Users/julianrussmeyer/.pyenv/versions/flower-3.12.9/bin/uv run python -m devtool.build_example_docs`
  - Build succeeded with 22 existing docs warnings.
- Previewed the built `examples/docs/build/html/index.html` locally and verified the original Examples structure renders with the updated visual treatment.
- `PYENV_VERSION=flower-3.12.9 uv run --directory dev ./test-devtool.sh`
- Temp-copy check against `examples/docs/source/conf.py` confirmed the updater keeps the existing `#17222d` announcement background, does not introduce `#292f36`, and keeps one light/dark CSS-variable block.


dev/devtool/update_html_themes.py
```diff
@@ -65,36 +68,130 @@ def find_conf_files(root_dir: Path) -> list[Path]:
...
+    indent_match = re.match(r"^(\s*)}", line)
+    indent = indent_match.group(1) if indent_match else ""
+    new_fields_str = dict_to_fields_str(fields)
+    new_fields_indented = "\n".join(
+        indent + "    " + new_field_line
+        for new_field_line in new_fields_str.splitlines()
+    )
+    if new_fields_indented:
+        updated_content.insert(-1, new_fields_indented + ",")
+    return bool(new_fields_indented)
```
jafermarq, line 117: This return field is never read by the function calling `_append_indented_fields`

dev/devtool/update_html_themes.py
```diff
@@ -65,36 +68,130 @@ def find_conf_files(root_dir: Path) -> list[Path]:
...
+    merge_field: Optional[str] = None
     modified = False
 
-    for line in content:
+    for line in file_path.read_text(encoding="utf-8").splitlines():
+        if merge_field:
+            brace_depth, merge_field, line_modified = _process_merge_field_line(
+                updated_content, line, brace_depth, merge_field, merge_fields
+            )
+            modified |= line_modified
```
jafermarq, line 211: I think this is redundant (the third argument returned by `_process_merge_field_line`), because later `modified` will be always set to true. So we can remove this line and the third argument that `_process_merge_field_line` returns.

dev/devtool/update_html_themes.py
Julianrussmeyer: I updated the caller to use the returned boolean, so modified is only set when `_append_indented_fields` actually inserts something.

dev/devtool/update_html_themes.py
Julianrussmeyer: I removed the later unconditional `modified = True`, so this return value is no longer redundant. It now tracks whether merging actually added missing theme variables. This also avoids rewriting files when no changes are needed.
I also added a regression test for the no-change case to make sure the updater reports “No changes needed” instead of rewriting the file.

jafermarq (dismissed):

dev/devtool/update_html_themes.py
```diff
@@ -65,49 +68,145 @@ def find_conf_files(root_dir: Path) -> list[Path]:
...
+    new_fields_str = dict_to_fields_str(new_fields)
     if not new_fields_str.strip():
         print(f"Skipping {file_path} (no new fields to insert)")
         return
 
-    content = file_path.read_text(encoding="utf-8").splitlines()
-    updated_content = []
+    updated_content: list[str] = []
+    fields_to_append = _copy_fields(new_fields)
+    merge_fields = _merge_fields(new_fields)
```
smoroso, line 199: - What: Existing top-level generated theme options are not detected, so rerunning the updater can append duplicate keys.
- Why: The branch makes nested `light_css_variables` and `dark_css_variables` idempotent, but leaves non-merge fields such as `announcement` in `fields_to_append` even when they already exist in `html_theme_options`. If `announcement.enabled` is turned on and the script runs on docs that already have the banner, it writes a second `"announcement"` entry; Python accepts the duplicate key with last-one-wins semantics, which hides the stale value and keeps producing noisy rewrites.
- Fix: Track existing top-level `html_theme_options` keys while scanning the block and remove matching generated fields from `fields_to_append`, just like the merge fields are removed before appending.
- Solution Example:
```python
top_level_match = re.match(r'^\s*"(?P<key>[^"]+)"\s*:', line)
if inside_options and brace_depth == 1 and top_level_match:
    fields_to_append.pop(top_level_match.group("key"), None)
```

smoroso (approved):

yan-gao-GY (approved):

tanertopal (approved):

PR #7410, opened 2026-06-17
fix(framework): Disable early `ray` import that conflicts with runtime environment switch if app dependencies are installed

jafermarq:
Without this change, if the base environment has `ray` intalled, even though the installation of app dependencies at runtime will install `ray` in the runtime environment, the switch to use that `ray` binary won't be actioned. This is because `ray` is already loaded.

### DIFF
--- a/framework/py/flwr/simulation/__init__.py
+++ b/framework/py/flwr/simulation/__init__.py
@@ -16,26 +16,28 @@
 
 
 import importlib
+from typing import Any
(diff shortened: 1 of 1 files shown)


panh99 (dismissed):
LGTM!

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7418, opened 2026-06-18
feat(framework): Add connector to agentapp

danielnugraha:
Adds built-in connector tool support for AgentApp response creation.

  - Registers built-in web_search and web_fetch connectors with function-tool schemas.
  - Injects built-in connector tools into responses.create model requests.
  - Executes one batch of built-in connector function calls through TaskType.CONNECTOR.
  - Sends connector outputs back to the model as function_call_output.
  - Keeps connector-specific tool metadata and parsing outside session.py.
  - Adds focused tests for registry, connector tool extraction, connector task handling, and AgentApp
    connector flow.


framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -69,7 +78,55 @@ def __init__(  # pylint: disable=too-many-arguments
         self._task_id = task_id
 
     def create(self, request: JSONObject) -> JSONObject:
-        """Create a model response through a child model task."""
+        """Create a model response through child model and connector tasks."""
+        model_request = with_builtin_connector_tools(request)
+        response_payload = self._create_model_response(model_request)
+
+        tool_calls = extract_builtin_connector_tool_calls(response_payload)
+        if tool_calls:
```
panh99, line 87: [reviewer-bot] Important (API-001) framework/py/flwr/supercore/task_process/agent/session.py:86 confidence: high

Could we handle the case where the model returns connector calls together with other function calls? Right now any built-in connector call enters this branch, sends only the connector outputs back, and the original response is never returned to the AgentApp. If the same response also contains an app-owned function call, the app cannot satisfy it and the provider may reject the follow-up because one pending tool call has no output. A minimal fix would be to auto-follow-up only when all function calls in the response are built-in connectors, or otherwise return the mixed response to the AgentApp so it can handle the remaining calls.

framework/py/flwr/supercore/task_process/connector/tool_call.py
```diff
@@ -0,0 +1,102 @@
...
+    # Built-ins are appended once before the first model call.
+    builtin_tools = get_builtin_connector_tools()
+
+    if tools is None:
+        updated["tools"] = builtin_tools
+        return updated
+
+    if isinstance(tools, Sequence) and not isinstance(tools, str):
+        tool_list = list(tools)
+        if all(isinstance(tool, dict) for tool in tool_list):
```
panh99, line 51: [reviewer-bot] Important (API-001) framework/py/flwr/supercore/task_process/connector/tool_call.py:51 confidence: high

Can we avoid claiming names that the request already provided? Since built-ins are appended unconditionally and extraction later keys only on `name`, a user-supplied function tool named `web_search`/`web_fetch` would be silently treated as a Flower connector call instead of being returned to the AgentApp. The smallest fix is to detect a same-name function tool and either skip the built-in injection or raise a clear error before sending the request.

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -69,7 +78,55 @@ def __init__(  # pylint: disable=too-many-arguments
...
+                output = response_payload.get("output")
+                prior_output: list[JSONObject] = []
+                if _is_json_object_list(output):
+                    prior_output = cast(list[JSONObject], output)
+                followup_request["input"] = [*prior_output, *followup_input]
+            model_request = followup_request
+            response_payload = self._create_model_response(model_request)
+
+        output = response_payload.get("output")
+        if _is_json_object_list(output):
```
panh99, line 115: [reviewer-bot] Important (CORR-001) framework/py/flwr/supercore/task_process/agent/session.py:124 confidence: medium

In the no-`previous_response_id` fallback, this follow-up drops the original request input and sends only the model's prior output plus connector output. For an Open Responses-compatible provider that omits `id`, the second model call loses the user's actual instruction, so a call like “summarize this page for a beginner” can turn into just a `web_fetch(url)` result with no summarization goal. Could we include the original `model_request["input"]` in this fallback before the prior output/tool output items?

framework/py/flwr/supercore/task_process/connector/web_search/__init__.py
```diff
@@ -24,6 +24,22 @@
...
+        "properties": {
+            "query": {
+                "type": "string",
+                "description": "The search query.",
+            },
+        },
+        "required": ["query"],
+        "additionalProperties": False,
+    },
+}
```
panh99: Should we create the tool via a function instead of a constant? I think we need to pass it around in many places, and it might be slightly risky for a mutable object like `dict`. Maybe we can do this:

```python
def web_search_tool() -> JSONObject:
    return {
        "type": "function",
        "name": WEB_SEARCH_CONNECTOR_NAME,
        "description": "Search the web for current information.",
        "parameters": {
            "type": "object",
            "properties": {
                "query": {
                    "type": "string",
                    "description": "The search query.",
                },
            },
            "required": ["query"],
            "additionalProperties": False,
        },
    }
```

framework/py/flwr/supercore/task_process/agent/session.py
panh99: @danielnugraha Is this relevant? I guess we will only allow built-in connector tool calls for now, right?

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -69,7 +78,55 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        tool_calls = extract_builtin_connector_tool_calls(response_payload)
+        if tool_calls:
+            # Execute one connector batch; further tool-call loops stay in AgentApp.
+            followup_input: list[JSONObject] = []
+            for tool_call in tool_calls:
+                output = self._create_connector_response(
+                    name=tool_call.name,
+                    call_id=tool_call.call_id,
+                    arguments=tool_call.arguments,
+                )
```
panh99: I wonder if we should just make `_create_connector_response` accept a ToolCall dataclass object.

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -69,7 +78,55 @@ def __init__(  # pylint: disable=too-many-arguments
...
+                followup_input.append(
+                    {
+                        "type": "function_call_output",
+                        "call_id": tool_call.call_id,
+                        "output": (
+                            output
+                            if isinstance(output, str)
+                            else strict_json_dumps(output, compact=True)
+                        ),
```
panh99: Just curious, could we always use json here? Why we need special handling for `str`?
```suggestion
                        "output": strict_json_dumps(output, compact=True),
```

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -69,7 +78,55 @@ def __init__(  # pylint: disable=too-many-arguments
...
+                    }
+                )
+
+            # Keep the prepared model request, but replace user input with tool output.
+            followup_request = dict(model_request)
```
panh99: Why do we use `dict(model_request)` here? Is it to avoid mutating the existing `model_request` dictionary?

After reading the code, it seems that we eventually do `model_request = followup_request`. Would it be simpler to update `model_request` in place instead of creating a separate `followup_request` variable here?

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -69,7 +78,55 @@ def __init__(  # pylint: disable=too-many-arguments
...
+            previous_response_id = response_payload.get("id")
+            if isinstance(previous_response_id, str) and previous_response_id:
+                followup_request["input"] = followup_input
+                followup_request["previous_response_id"] = previous_response_id
+            else:
+                output = response_payload.get("output")
+                prior_output: list[JSONObject] = []
+                if _is_json_object_list(output):
+                    prior_output = cast(list[JSONObject], output)
+                followup_request["input"] = [*prior_output, *followup_input]
```
panh99: Just for my understanding, does this mean we will attach `previous_response_id` when it exists, or we will directly concatenate the previous output + followup input?

framework/py/flwr/supercore/task_process/connector/tool_call.py
panh99: @danielnugraha  Is this relevant? Do we allow user-supplied function tools?

framework/py/flwr/supercore/task_process/connector/web_fetch.py
```diff
@@ -25,6 +25,24 @@
...
+        "properties": {
+            "url": {
+                "type": "string",
+                "description": "The URL to fetch.",
+            },
+        },
+        "required": ["url"],
+        "additionalProperties": False,
+    },
+}
```
panh99: Same as above. Do we need to pass this around? If so, I think it might be safer to have a function. Lmkwyt.

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -69,7 +78,65 @@ def __init__(  # pylint: disable=too-many-arguments
...
+                )
+                followup_input.append(
+                    {
+                        "type": "function_call_output",
+                        "call_id": tool_call.call_id,
+                        "output": strict_json_dumps(output, compact=True),
+                    }
+                )
+
+            # Reuse the prepared model request, but replace input for continuation.
```
panh99, line 100: Do you think it’s worth mentioning that the original input is already available in `Context`, so we don’t need to preserve it here?

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -96,28 +96,38 @@ def create(self, request: JSONObject) -> JSONObject:
...
+                            "content": original_input,
+                        }
+                    ]
+                elif isinstance(original_input, Sequence):
+                    original_input_items = cast(list[JSONObject], list(original_input))
+                model_request["input"] = [
+                    *original_input_items,
+                    *prior_output,
+                    *followup_input,
+                ]
```
panh99: IIRC, you mentioned that the original input is already in the `Context` and we do not need to include it in here. May I know why we now decide to include it?

panh99 (dismissed):
LGTM!

panh99 (approved):

PR #7419, opened 2026-06-18
feat(framework): Allow SuperLink to run locally bundled agentapp FAB

danielnugraha:
Enable local AgentApp FAB execution via flwr run.

  SuperLink now derives the run type from the submitted FAB config. If [tool.flwr.app.components] contains
  agentapp, the run is created as RunType.AGENT_APP, which launches an flwr-agentapp task. This allows users
  to run arbitrary local AgentApp directories with flwr run ./my-agent-app ..., instead of only supporting
  the built-in @flwragent/flwr-agent special case.


panh99 (approved):
LGTM!

PR #7425, opened 2026-06-18
refactor(framework): Deprecate NDArray/Parameters conversion functions

danieljanes:


framework/py/flwr/common/__init__.py
```diff
@@ -24,6 +24,10 @@
 
 from ..app.error import Error as Error
 from ..app.metadata import Metadata as Metadata
+from ..compat.common.parameter import bytes_to_ndarray as bytes_to_ndarray
```
panh99, line 27: [reviewer-bot] Blocking (API-001) framework/py/flwr/common/__init__.py:27 confidence: high

This looks like a compatibility break rather than a deprecation: after the rename, user code doing `from flwr.common.parameter import ndarrays_to_parameters` will fail with `ModuleNotFoundError`, even though the public `flwr.common` re-export still works. Could we keep a small `flwr/common/parameter.py` shim that re-exports from `flwr.compat.common.parameter` (and optionally emits the deprecation warning) until the removal release?

panh99 (approved):
LGTM!

tanertopal (approved):

PR #7427, opened 2026-06-18
feat(framework): Allow switching `uv` indices for runtime dependencies installation

jafermarq:
Checks `UV_DEFAULT_INDEX` and the (new) `FLWR_UV_DEFAULT_INDEX`

### DIFF
--- a/framework/py/flwr/supercore/superexec/dependency_installer.py
+++ b/framework/py/flwr/supercore/superexec/dependency_installer.py
@@ -151,16 +151,16 @@ def install_app_dependencies(
     if uv_default_index_name == _DEFAULT_UV_INDEX:
         log(
             INFO,
-            "  Using uv package index (default): %s",
-            _redact_url(uv_default_index_url),
+            "  uv default index not explicitly set; uv will use its configured default.",
         )
     else:
         log(
             INFO,
-            "  Using uv package index from: %s",
+            "  Using uv package index from %s: %s",
+            uv_default_index_name,
             _redact_url(uv_default_index_url),
         )
-    _log_index_reachability(uv_default_index_url)
+        _log_index_reachability(uv_default_index_url)
     log(INFO, "  Starting uv sync for application dependencies.")
 
     installed_packages: set[str] = set()
--- a/framework/py/flwr/supercore/superexec/dependency_installer.py
+++ b/framework/py/flwr/supercore/superexec/dependency_installer.py
@@ -36,6 +36,8 @@ from flwr.supercore.utils import get_flwr_home
 _RUNTIME_ENV_DIR = "runtime-envs"
 _FLWR_UV_DEFAULT_INDEX = "FLWR_UV_DEFAULT_INDEX"
 _UV_DEFAULT_INDEX = "UV_DEFAULT_INDEX"
+_DEFAULT_UV_INDEX = "default"
+_DEFAULT_UV_INDEX_URL = "https://pypi.org/simple"
 _INDEX_REACHABILITY_TIMEOUT_SECONDS = 10.0
 RuntimeDependencyIndexContext = dict[str, str | int | None]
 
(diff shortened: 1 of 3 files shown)


framework/py/flwr/supercore/superexec/dependency_installer.py
```diff
@@ -25,12 +25,17 @@
...
 from logging import DEBUG, ERROR, INFO, WARNING
 from pathlib import Path
+from urllib.parse import urlsplit, urlunsplit
 
 from flwr.common.config import get_project_config
 from flwr.common.logger import log
 from flwr.supercore.utils import get_flwr_home
 
 _RUNTIME_ENV_DIR = "runtime-envs"
+_FLWR_UV_DEFAULT_INDEX = "FLWR_UV_DEFAULT_INDEX"
```
panh99, line 35: Should we move it `flwr.supercore.constant`?

framework/py/flwr/supercore/superexec/dependency_installer.py
```diff
@@ -172,6 +194,32 @@ def _get_project_dependencies(project_dir: Path) -> list[str]:
...
+    host = split_url.netloc.rsplit("@", maxsplit=1)[-1]
+    return urlunsplit(
+        (
+            split_url.scheme,
+            host,
+            split_url.path,
+            split_url.query,
+            split_url.fragment,
+        )
+    )
```
panh99, line 220: Should we move it to `supercore.utils`?

Btw, this seems to protect against credentials embedded in the index URL, which is more relevant for private/enterprise package indexes. For this feature, I think the intended use case is just a public PyPI mirror URL, right? That said, this util is useful and lightweight.

framework/py/flwr/supercore/superexec/dependency_installer.py
```diff
@@ -150,14 +172,14 @@ def install_app_dependencies(
         raise RuntimeError(f"uv sync failed: {sync_error}")
 
     if installed_packages:
-        log(INFO, "Installed: [%s]", ", ".join(sorted(installed_packages)))
+        log(INFO, "  Installed: [%s]", ", ".join(sorted(installed_packages)))
```
panh99, line 175: May I know why we apply indents now?

framework/py/flwr/supercore/superexec/dependency_installer.py
```diff
@@ -133,10 +139,26 @@ def install_app_dependencies(
 
     if dependency_index_url is not None:
         sync_cmd += ["--index-url", dependency_index_url]
```
panh99, line 141: Let's use `--default-index` here, see:
```shell
$ uv sync -h | grep -e '--index-url'
  -i, --index-url <INDEX_URL>                (Deprecated: use `--default-index` instead) The URL of the Python package index (by default: <https://pypi.org/simple>)
```

framework/py/flwr/supercore/superexec/dependency_installer.py
```diff
@@ -133,10 +139,26 @@ def install_app_dependencies(
...
+    if uv_default_index_name == _FLWR_UV_DEFAULT_INDEX:
+        sync_cmd += ["--default-index", uv_default_index_url]
```
panh99, line 143: Should we just log a warning saying `FLWR_UV_DEFAULT_INDEX` is ignored as `dependency_index_url` is provided? Btw, is setting uv index a EE feature?

framework/py/flwr/supercore/superexec/dependency_installer.py
panh99, line 1: I think we can simplify this quite a bit.

For the feature goal, Flower only needs to read `FLWR_UV_DEFAULT_INDEX` and, when it is non-empty, pass it to `uv sync` as `--default-index <value>`. I don’t think Flower needs to inspect `UV_DEFAULT_INDEX` or model uv’s default `https://pypi.org/simple`, because `sync_env = os.environ.copy()` already preserves native uv configuration and lets uv handle its own env vars/config files.

So I’d suggest keeping the implementation closer to:

- read `FLWR_UV_DEFAULT_INDEX`
- if set, append `["--default-index", value]` and log
- otherwise do nothing and let uv resolve its default/index config normally

framework/py/flwr/supercore/superexec/dependency_installer.py
jafermarq: I think it looks nicer visually. Everything related to dependencies installation is shifted by 2 spaces after the first log. I find it clearer in this way that the following 4 `log(INFO, ...)` are about the same topic (i.e. dependencies):

```
INFO :      Installing application dependencies...
```
It's subtle though. feel free to remove

PR #7434, opened 2026-06-19
refactor(framework): Align ClientAppIo with ServerAppIo

danieljanes:
Summary

- Align ClientAppIo with ServerAppIo at the gRPC/proto API surface:
  - Rename ClientAppIo.PushMessage / PullMessage to PushMessages / PullMessages.
  - Add ClientAppIo.GetNodes.
  - Make the service ClientAppIo and service ServerAppIo definitions match exactly apart from the service name.
- Move GetNodesRequest / GetNodesResponse into shared appio.proto, and update ServerAppIo/Grid imports to use appio_pb2.
- Regenerate Python protobuf/gRPC artifacts from the updated .proto files.
- Update ClientApp runtime calls to use stub.PullMessages / stub.PushMessages.
- Implement ClientAppIo.GetNodes as authenticated but UNIMPLEMENTED.
- Keep ClientAppIo message handling single-message for now, but make that explicit:
  - PullMessages aborts with NOT_FOUND instead of raising IndexError when no message is available.
  - PushMessages aborts with INVALID_ARGUMENT unless exactly one message and one object tree are provided.
- Unify ServerAppIo and ClientAppIo token-auth policy generation through shared _RUNTIME_METHOD_AUTH_POLICY and _build_runtime_method_auth_policy.
- Update tests for renamed RPCs, shared GetNodes messages, ClientAppIo GetNodes, message validation, and auth policy coverage.


framework/proto/flwr/proto/clientappio.proto
```diff
@@ -85,12 +85,15 @@ service ClientAppIo {
   rpc PushLogs(PushLogsRequest) returns (PushLogsResponse) {}
 
   /////////////////////////////////////////////////////////////////////////////
-  // Specific endpoints for ClientAppIo
+  // Message and node endpoints
   /////////////////////////////////////////////////////////////////////////////
 
-  // Push Message
-  rpc PushMessage(PushAppMessagesRequest) returns (PushAppMessagesResponse) {}
+  // Push messages
```
mohammadnaseri, line 91: Should we add a comment here that `ClientAppIo` rejects requests with anything other than one message?

framework/proto/flwr/proto/clientappio.proto
```diff
@@ -85,12 +85,15 @@ service ClientAppIo {
...
+  rpc PushMessages(PushAppMessagesRequest) returns (PushAppMessagesResponse) {}
 
-  // Pull Message
-  rpc PullMessage(PullAppMessagesRequest) returns (PullAppMessagesResponse) {}
+  // Pull messages
```
mohammadnaseri, line 94: and same comment here for pulling?

framework/proto/flwr/proto/clientappio.proto
danieljanes: These are aligned with `ServerAppIo` to reduce the diff between both.

framework/proto/flwr/proto/clientappio.proto
danieljanes: same as above

panh99 (approved):
LGTM!

tanertopal (approved):

tanertopal (approved):

PR #7437, opened 2026-06-19
feat(framework): Prepare creation of default federations

jafermarq:


framework/py/flwr/supercore/constant.py
```diff
@@ -116,6 +117,34 @@
     init_args_log_to_driver=True,
 )
 
+
+# Default federations
+DEFAULT_AGENT_RUNS_FEDERATION_NAME = "agentruns"
+DEFAULT_WORKSPACE_FEDERATION_NAME = "workspace"
```
panh99:
```suggestion
# Default federation names for every Flower account
DEFAULT_FEDERATION_NAME_SIMULATION = "workspace-simulation"
DEFAULT_FEDERATION_NAME_DEPLOYMENT = "workspace-deployment"
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -191,10 +191,13 @@ def StartRun(  # pylint: disable=too-many-locals, too-many-branches, too-many-st
             )
             return StartRunResponse()
 
-        flwr_aid = _get_flwr_aid(context)
+        account = _get_account(context)
+        flwr_aid = cast(str, account.flwr_aid)
```
panh99: Why this change? We only need the `flwr_aid`, no?

framework/py/flwr/supercore/constant.py
```diff
@@ -116,6 +117,34 @@
...
+
+
+@dataclass(frozen=True)
+class _DefaultFederation:
+    """Default federation metadata."""
+
+    name: str
+    description: str
+    simulation: bool
+
```
panh99: Should we use `Federation` from `federation.proto` instead of creating a new private dataclass?

jafermarq: I'd say no, because here by `name` we mean just what comes after `/` but in the `Federation` proto object it refers to it's full name: `@<account-name>/<federation-name>`

panh99: Ah, I see what you mean. Should we consider making the descriptions constants? The `_DefaultFederation` dataclass still feels a bit speculative to me.

Alternatively, we can move the descriptions into the default federation manager. We only need the names atm.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -657,12 +659,13 @@ def ListFederations(
 
         # Init link state
         state = self.linkstate_factory.state()
+        account = _get_account(context)
```
```suggestion
        flwr_aid = _get_flwr_aid(context)
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -657,12 +659,13 @@ def ListFederations(
...
 
         # Get federations the account is a member of
         with rpc_error_translator(context, rpc_name):
-            federations = state.federation_manager.get_federations(
-                _get_flwr_aid(context)
-            )
+            flwr_aid = cast(str, account.flwr_aid)
```
```suggestion
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -686,9 +688,11 @@ def ShowFederation(
         state = self.linkstate_factory.state()
 
         # Ensure flwr_aid is a member of the requested federation
-        flwr_aid = _get_flwr_aid(context)
+        account = _get_account(context)
+        flwr_aid = cast(str, account.flwr_aid)
         federation = request.federation_name
```
panh99, line 691:
```suggestion
        flwr_aid = _get_flwr_aid(context)
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -805,7 +810,9 @@ def AddNodeToFederation(
 
         with rpc_error_translator(context, rpc_name):
             # Validate federation, node ID, and ownership
-            flwr_aid = _get_flwr_aid(context)
+            account = _get_account(context)
+            flwr_aid = cast(str, account.flwr_aid)
```
```suggestion
            flwr_aid = _get_flwr_aid(context)
```

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -879,7 +886,9 @@ def CreateInvitation(
         state = self.linkstate_factory.state()
 
         with rpc_error_translator(context, rpc_name):
-            flwr_aid = _get_flwr_aid(context)
+            account = _get_account(context)
+            flwr_aid = cast(str, account.flwr_aid)
```
```suggestion
            flwr_aid = _get_flwr_aid(context)
```

framework/py/flwr/supercore/constant.py
```diff
@@ -116,6 +117,34 @@
...
+        name=DEFAULT_FEDERATION_SIMULATION,
+        description="Default simulation federation.",
+        simulation=True,
+    ),
+    _DefaultFederation(
+        name=DEFAULT_FEDERATION_DEPLOYMENT,
+        description="Default deployment federation.",
+        simulation=False,
+    ),
+)
```
jafermarq:
```suggestion
```

framework/py/flwr/superlink/servicer/control/control_servicer_test.py
```diff
@@ -1175,6 +1223,28 @@ def test_create_invitation_success(self) -> None:
...
+                        runtime=RunTime.DEPLOYMENT,
+                    ),
+                ),
+                call.create_invitation(
+                    flwr_aid=self.flwr_aid,
+                    federation="test-federation",
+                    invitee_account_name="invitee-aid",
+                ),
+            ]
+        )
```
panh99: Should we simply check whether `ensure_default_federations_exist` is called, or remove this part altogether? I think it is not very relevant to this specific test function, `test_create_invitation_success`. Also, `can_execute` and `create_invitation` are already checked in the code below.

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -733,15 +743,22 @@ def CreateFederation(
                     f"Invalid federation name: '{request.federation_name}'. {err_msg}",
                 )
 
+            if request.federation_name in _DEFAULT_FEDERATION_NAMES:
+                raise FlowerError(
+                    ApiErrorCode.FORBIDDEN_ACTION,
+                    f"Federation name '{request.federation_name}' is reserved.",
+                )
+
```
panh99: Could we remove this check? I am inclined to make the Control API endpoint behave as if every account already has the two default federations. In that case, if a user tries to create a federation with one of those names, we should treat it as a normal name conflict and return that the federation already exists, rather than saying the name is reserved.

I think this is clearer because we do not currently have the concept of reserved federation names. The default federation names should simply be treated as existing federation names.

panh99 (approved):
LGTM!

PR #7457, opened 2026-06-21
refactor(framework): Move exit package to flwr.supercore

danieljanes:


framework/py/flwr/common/exit/__init__.py
panh99, line 1: Since this is our internal API and, afaik, there are no references to the `flwr.common.exit` module, should we remove it directly instead of keeping it here and re-exporting things?

danieljanes: Yes, that would be better.

panh99 (approved):
LGTM!

tanertopal (approved):

PR #7467, opened 2026-06-23
ci(framework): Add e2e agentapp test

danielnugraha:
### Checklist

- [ ] Write tests
- [ ] Make CI checks pass


.github/workflows/framework-agentapp-e2e.yml
```diff
@@ -0,0 +1,29 @@
+name: Framework AgentApp E2E
+
+on:
+  workflow_dispatch:
```
panh99, line 4: [reviewer-bot] Important (TEST-001) .github/workflows/framework-agentapp-e2e.yml:4 confidence: medium

This new E2E workflow is only `workflow_dispatch`, so it won't run on PRs or pushes and won't protect AgentApp changes from regressing unless someone remembers to trigger it manually. Could we either wire this into the existing framework E2E workflow/checks, or add the intended automatic trigger with secret-safe gating?

.github/workflows/framework-agentapp-e2e.yml
```diff
@@ -0,0 +1,29 @@
...
+    defaults:
+      run:
+        working-directory: framework/e2e/e2e-agentapp
+    steps:
+      - uses: actions/checkout@v5
+      - name: Bootstrap
+        uses: ./.github/actions/bootstrap
+      - name: Install dependencies
+        run: python -m pip install --upgrade .
+      - name: Run AgentApp E2E
```
panh99, line 23: [reviewer-bot] Blocking (SEC-001) .github/workflows/framework-agentapp-e2e.yml:23 confidence: medium

This job passes model/search API keys into shell code from the checked-out ref. Since `workflow_dispatch` can run against an arbitrary branch/ref, a manual run can execute unreviewed changes with these credentials. Could we gate this with the existing `_repo-authorize-manual-trigger.yml` job, or otherwise restrict the workflow to a trusted ref before exposing the secrets?

framework/e2e/test_superlink.sh
panh99, line 1: I’m a bit concerned about the amount of duplication introduced here. Do we need to duplicate the main body of the test code in this file, or could we refactor it so the shared logic is reused?

Another option might be to move the shell script for the AgentApp E2E test into a dedicated test file. Right now, `test_superlink.sh` feels like it is serving two quite different purposes: it adds a large AgentApp E2E-specific block and then exits early. To me, that makes the structure a little surprising and harder to follow.

framework/e2e/e2e-agentapp/README.md
panh99, line 1: The README files for E2E tests are usually quite short, since we don’t typically expect people to run them manually. Is there a specific reason this README needs to be this detailed, for example, because the AgentApp setup is different from the other E2E tests?

If not, could we keep this README closer to the usual E2E test README style?

danielnugraha: The AgentApp setup requires adding env vars to the SuperLink and SuperExec. I think we can keep this README closer to the usual e2e README and move the detailed info to an example in a separate PR.

framework/e2e/test_superlink.sh
danielnugraha: I created a new test.sh in e2e-agentapp dir.

PR #7476, opened 2026-06-23
break(framework): Change number of default SuperNodes in simulation to two

jafermarq:


framework/docs/source/how-to-run-simulations.rst
```diff
@@ -157,8 +157,8 @@ The complete list of examples can be found in `the Flower GitHub
  Customize the Simulation Runtime
 **********************************
 
-By default, the Simulation Runtime simulates a cohort of 10 SuperNodes and assigns two
-CPU cores to each backend worker. This means that if your system has 12 CPU cores, six
+By default, the Simulation Runtime simulates a cohort of two SuperNodes and assigns two
+CPU cores to each backend worker. This means that if your system has four CPU cores, all
```
panh99: Should we say “four or more” here? The original sentence told the story that you could have six concurrent workers because each worker needed two cores, assuming 12 cores in total. Now we are telling a different story: all backend workers can run as long as the system has four or more cores, because we now only have two SuperNodes instead of ten.

```suggestion
CPU cores to each backend worker. This means that if your system has four or more CPU cores, all
```

framework/docs/source/how-to-run-simulations.rst
```diff
@@ -157,8 +157,8 @@ The complete list of examples can be found in `the Flower GitHub
  Customize the Simulation Runtime
 **********************************
 
-By default, the Simulation Runtime simulates a cohort of 10 SuperNodes and assigns two
-CPU cores to each backend worker. This means that if your system has 12 CPU cores, six
+By default, the Simulation Runtime simulates a cohort of two SuperNodes and assigns two
+CPU cores to each backend worker. This means that if your system has four or more CPU cores, all
 backend workers can be running in parallel, each executing a different ``ClientApp``
 instance.
```
panh99:
```suggestion
By default, the Simulation Runtime simulates a cohort of two SuperNodes and assigns two
CPU cores to each backend worker. This means that if your system has four or more CPU
cores, all backend workers can be running in parallel, each executing a different
``ClientApp`` instance.
```

panh99 (approved):
LGTM!

PR #7478, opened 2026-06-23
docs(framework): Update quickstart tutorials to reflect new number of default supernodes

jafermarq:


framework/docs/source/tutorial-quickstart-pytorch-lightning.rst
```diff
@@ -52,6 +51,8 @@ project, do:
 
 .. code-block:: shell
 
+    $ cd path/to/quickstart-pytorch-lightning
+
```
panh99: Should we keep one of the `cd ...` here? Btw, the tutorial here has a big `By default, this project uses ...` paragraph that doesn't exist in other tutorials. Is that intentional?

framework/docs/source/tutorial-quickstart-jax.rst
```diff
@@ -15,7 +15,7 @@ create a virtual environment and run everything within a :doc:`virtualenv
 <contributor-how-to-set-up-a-virtual-env>`.
 
 Let's use ``flwr new`` to create a complete Flower+JAX project. It will generate all the
-files needed to run a federation of 50 nodes using |fedavg|_. By default, the generated
+files needed to run a federation of two nodes using |fedavg|_. By default, the generated
```
panh99, line 18: Just noticed that JAX example uses 50 SuperNodes instead of the old default 10. Is there a reason for using 50 SuperNodes? Is it safe to set it to 2 nodes?

framework/docs/source/tutorial-quickstart-pytorch-lightning.rst
jafermarq: yes

framework/docs/source/tutorial-quickstart-jax.rst
jafermarq: it will run slower but it will run

framework/docs/source/tutorial-quickstart-pytorch-lightning.rst
jafermarq: this is so to keep it more aligned with other tutorials. A similar change was done for `quickstart-pytorch-lightning`

panh99 (approved):
LGTM!

PR #7479, opened 2026-06-23
refactor(examples): Use `fraction-train` and `fraction-evaluate` one 

jafermarq:
Because default number of supernodes has been changed to two: https://github.com/flwrlabs/flower/pull/7476


panh99 (approved):
LGTM!

PR #7486, opened 2026-06-24
refactor(framework): Update flower tutorial to use default federation for simulations

jafermarq:
- also updates content to adjust to the new 2-SuperNodes default
- removed `pip install -e .` and installation of `[simulation]` extra
- fixes some code blocks


panh99 (approved):
LGTM!

PR #7494, opened 2026-06-25
refactor(framework): Check if `ray` installation is quietly skipped on Windows

jafermarq:
`ray` (the _simulation_ extra) doesn't support py13 on windows. The dependency resolver can quietly skip its installation and through a related (701) but not that helpful message. This PR ensures error 608 is raised.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #7505, opened 2026-06-29
feat(framework): Add federation filter to ListRunSeries

danielnugraha:
Add federation filter to ListRunSeries.


framework/proto/flwr/proto/control.proto
```diff
@@ -159,6 +159,7 @@ message ListRunsResponse {
 message ListRunSeriesRequest {
   optional string updated_before = 1;
   optional uint32 limit = 2;
+  optional string federation = 3;
```
jafermarq:
```suggestion
  optional string federation_name = 3;
```

Should we rename it to `federation_name` (as we do with `ShowFederationRequest`)

jafermarq (approved):

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7517, opened 2026-07-01
feat(framework): Add task usage table

danielnugraha:
Adds task usage schema support.
  - Adds task_usage CoreState table metadata with one usage row per task
  - Tracks usage_id, run_id, task_id, token counters, created_at, and reported_at


panh99 (approved):
LGTM!

PR #7533, opened 2026-07-03
refactor(framework): Lower `ControlAPI` timeout from CLI

jafermarq:


psfoley (approved):
@jafermarq 30 seconds was likely too conservative; thanks for making this change

panh99 (approved):

PR #7537, opened 2026-07-03
feat(framework): Add task usage call in agentapp API

danielnugraha:
Embed task usage call in agentapp API in report when used manner.


framework/py/flwr/supercore/task_process/model/provider.py
```diff
@@ -100,22 +105,28 @@ def invoke_model_provider(
...
 
-def _invoke_provider_response(  # pylint: disable=too-many-locals,too-many-branches
+def _invoke_provider_response(  # pylint: disable=too-many-locals,too-many-branches,too-many-arguments,too-many-statements
     *,
     responses_url: str,
     headers: dict[str, str],
     timeout: float,
     request: JSONObject,
     on_stream_event: Callable[[JSONObject], None] | None,
+    usage_recorder: TaskUsageRecorder | None,
```
panh99: Can `usage_recorder` be `None`?

framework/py/flwr/supercore/task_process/model/provider.py
```diff
@@ -100,22 +105,28 @@ def invoke_model_provider(
     if api_key:
         headers["Authorization"] = f"Bearer {api_key}"
     payload = dict(request)
+    model = payload.get("model")
+    usage_type = model if isinstance(model, str) and model else "token"
```
panh99: `usage_type` seems quite important. I wonder if we should make the logic that determines it a bit clearer and more explicit. Also, is this the only place where `usage_type` is decided?

danielnugraha: I moved this into a helper function. Yes for model, this is where `usage_type` is decided.

panh99 (approved):
LGTM!

PR #7555, opened 2026-07-08
feat(framework): Add Automation table

danielnugraha:
Add Automation table in CoreState.


framework/py/flwr/supercore/state/schema/corestate_tables.py
```diff
@@ -98,6 +98,48 @@ def create_corestate_metadata() -> MetaData:
...
 
+    # --------------------------------------------------------------------------
+    #  Table: automation
+    # --------------------------------------------------------------------------
+    automation = Table(
+        "automation",
+        metadata,
+        Column(
+            "automation_id",
+            BigInteger().with_variant(Integer(), "sqlite"),
```
panh99: This is new. Should we just use `Integer` for consistency, or you'd prefer `BigInteger` in case of overflow?

danielnugraha: Changed back to Integer.

panh99 (approved):
LGTM!

PR #7558, opened 2026-07-08
feat(framework): Add Automation protos

danielnugraha:
Add Automation protos to control API:
  - StartAutomation
  - ListAutomations
  - StopAutomation


framework/proto/flwr/proto/control.proto
```diff
@@ -139,6 +150,36 @@ message StartRunResponse {
...
 }
+
+message StartAutomationRequest {
+  StartRunRequest start_run_request = 1;
+  optional uint64 start_at = 2;
+  optional uint64 fixed_interval = 3;
+  optional uint32 max_runs = 4;
+}
+message StartAutomationResponse {
+  uint64 automation_id = 1;
```
panh99, line 161: Can this be `uint64`? IIUC, this will be auto generated by SQL DB, using `Integer`, right? Is `Integer` a `int32`?

danielnugraha: According to codex it's okay:

> Integer in the SQLAlchemy table is there mainly because SQLite autoincrement requires the primary key column to be exactly INTEGER PRIMARY KEY. In SQLite, that maps to the rowid, which is a signed 64-bit integer, not a 32-bit int.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7559, opened 2026-07-08
refactor(framework): Move `get_account` outside `ControlAPI` handlers

jafermarq:
Now all handlers that require account information, they'll receive it as an input argument.


panh99 (approved):
LGTM!

PR #7561, opened 2026-07-08
feat(framework): Add `LinkState` dependency for `FastAPI`

jafermarq:


framework/py/flwr/superlink/routers/dependencies.py
```diff
@@ -0,0 +1,46 @@
...
+    state_factory: LinkStateFactory | None
+
+
+def get_linkstate(request: Request[State]) -> LinkState:
+    """Return the SuperLink LinkState for the current request."""
+    superlink_lifespan = cast(
+        _SuperLinkLifespanState | None,
+        getattr(request.app.state, "superlink_lifespan", None),
+    )
+    if superlink_lifespan is None or superlink_lifespan.state_factory is None:
```
panh99: Why do we use `Protocol` here? Is it because we can’t import `SuperLinkLifespan` in this module?

Also, would it be better to assign the `LinkStateFactory` directly to `app.state`, even though we can access it via `app.state.superlink_lifespan.state_factory`?

framework/py/flwr/superlink/cli/flower_superlink.py
```diff
@@ -801,14 +801,24 @@ def _run_superlink_http_api(lifespan_config: SuperLinkLifespanConfig) -> None:
...
     if start_legacy_grpc:
         superlink_lifespan = SuperLinkLifespan(lifespan_config)
+    else:
+        federation_manager = get_federation_manager(
+            is_simulation=lifespan_config.simulation
+        )
+        _, linkstate_factory = _get_objectstore_linkstate_factories(
+            lifespan_config.database, federation_manager
+        )
+        linkstate_factory.state()
```
panh99:
```suggestion
        # Initialize LinkState
        federation_manager = get_federation_manager(
            is_simulation=lifespan_config.simulation
        )
        _, linkstate_factory = _get_objectstore_linkstate_factories(
            lifespan_config.database, federation_manager
        )
        linkstate_factory.state()
```

framework/py/flwr/superlink/cli/flower_superlink.py
```diff
@@ -801,14 +801,24 @@ def _run_superlink_http_api(lifespan_config: SuperLinkLifespanConfig) -> None:
             "`--enable-http-api` cannot be combined with `--fleet-api-type rest`",
         )
     superlink_lifespan = None
+    linkstate_factory = None
     if start_legacy_grpc:
         superlink_lifespan = SuperLinkLifespan(lifespan_config)
```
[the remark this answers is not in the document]
panh99: @jafermarq I think this comment makes sense. IIUC, we’ll support different modes, including one where both FastAPI and gRPC are enabled. In that case, even when `start_legacy_grpc` is `True`, we would still always initialize the `LinkStateFactory` and store it in `app.state`, so that `get_linkstate` can work consistently. So from that perspective, I don’t think `linkstate_factory` is really optional.

framework/py/flwr/superlink/dependencies/linkstate.py
panh99, line 1: The implementation looks good to me! I checked with ChatGPT about a proper structure, given that we will need to migrate multiple interceptors to dependencies and middlewares. It suggested a structure like this. Should we make `dependencies` a module?

```
framework/py/flwr/superlink/
  cli/
    flower_superlink.py

  app/
    main.py
    lifespan.py
    state.py

  dependencies/
    __init__.py
    linkstate.py
    auth.py
    request_context.py

  middleware/
    __init__.py
    event_log.py
    account_auth.py

  routers/
    __init__.py
    fleet.py
    exec.py
    health.py
```

jafermarq: yep, i like the idea of introducing a `dependencies/` module

framework/py/flwr/superlink/main.py
```diff
@@ -57,13 +58,18 @@ def _merge_lifespan_state(
 def create_app(
     *,
     superlink_lifespan: SuperLinkLifespan | None = None,
+    linkstate_factory: LinkStateFactory | None = None,
```
panh99:
```suggestion
    *,
    linkstate_factory: LinkStateFactory,
    superlink_lifespan: SuperLinkLifespan | None = None,
```

framework/py/flwr/superlink/main.py
```diff
@@ -81,6 +86,12 @@ async def lifespan(fastapi_app: FastAPI) -> AsyncIterator[dict[str, object]]:
                 # Temporary compatibility path: start the existing gRPC APIs from
                 # FastAPI lifespan
                 superlink_lifespan.startup()
+                if superlink_lifespan.state_factory is None:
+                    raise RuntimeError(
+                        "SuperLink lifespan state has not been initialized."
+                    )
```
panh99: I think we don't need extra check here.
```suggestion
```

framework/py/flwr/superlink/main.py
```diff
@@ -141,6 +152,3 @@ def validate_unique_route_operation_ids(fastapi_app: FastAPI) -> None:
                     "Please ensure all route handler function names are unique."
                 )
             operation_ids.add(op_id)
-
-
-app = create_app()
```
panh99, line 146: May I know why removing this?

framework/py/flwr/superlink/cli/flower_superlink.py
```diff
@@ -262,13 +262,18 @@ class SuperLinkLifespan:  # pylint: disable=too-many-instance-attributes
     compatibility adapters.
     """
 
-    def __init__(self, config: SuperLinkLifespanConfig) -> None:
+    def __init__(
+        self,
+        config: SuperLinkLifespanConfig,
+        objectstore_factory: ObjectStoreFactory | None = None,
+        state_factory: LinkStateFactory | None = None,
```
panh99: Based on the current code, `state_factory` is also not optional, right?

framework/py/flwr/superlink/main.py
```diff
@@ -143,4 +164,4 @@ def validate_unique_route_operation_ids(fastapi_app: FastAPI) -> None:
             operation_ids.add(op_id)
 
 
-app = create_app()
+app = create_app(linkstate_factory=_create_default_linkstate_factory())
```
panh99, line 168:
```suggestion
# Temporary: we need a way to provision the FastAPI server
app = create_app(linkstate_factory=_create_default_linkstate_factory())
```

panh99 (dismissed):
LGTM!

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7564, opened 2026-07-09
feat(framework): Add and implement `CoreState.store_message_and_object_tree`

panh99:
## Summary

This PR introduces `CoreState.store_message_and_object_tree`.

This follows the direction of merging `ObjectStore` into `LinkState` and makes message storage an atomic state operation.

## Motivation

Previously, to avoid race conditions, we had to:

1. preregister the object tree,
2. store the message,
3. delete the preregistered object tree if message storage failed.

This flow is harder to maintain, especially as we move towards introducing sender sessions. It also leaves more room for unpredictable state if a SuperLink replica crashes while handling the RPC.

By moving this into a single `CoreState` method, we ensure the atomic message storage + object tree preregistration.


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -43,6 +44,12 @@ def store_fab(self, fab: Fab) -> str:
     def get_fab(self, fab_hash: str) -> Fab | None:
         """Return the FAB for the given hash, if present."""
 
+    @abstractmethod
+    def store_message_and_object_tree(
+        self, message: Message, object_tree: ObjectTree
+    ) -> tuple[bool, list[str]]:
+        """Store a Message and preregister its ObjectTree."""
+
```
jafermarq:
```suggestion
        """Store a Message and preregister its ObjectTree.

        Parameters
        ----------
        message : Message
            The Message to store.
        object_tree : ObjectTree
            The ObjectTree containing the IDs of objects to preregister.

        Returns
        -------
        tuple[bool, list[str]]
            A tuple containing a boolean indicating whether the Message was
            stored and a list of object IDs that still need to be pushed. If
            storing the Message fails, returns `(False, [])`.
        """

```

to keep it inline with docstrings in other methods

framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py
```diff
@@ -203,27 +203,15 @@ def push_messages(
...
-    # Once that is in place, we can remove the run status check below.
-    if (
-        message_id is None
-        and state.get_run_status({run_id})[run_id].status == Status.FINISHED
-    ):
-        # The request currently contains only one message and its object tree.
-        store.delete(request.message_object_trees[0].object_id)
-        objects_to_push.clear()
+    # Store Message in State and preregister its objects.
+    _, objects_to_push = state.store_message_and_object_tree(
```
jafermarq, line 207: i wonder if we should be making use of the first return arg when constructing the response

panh99: Yeah, that's a good point. I see what you mean. We haven't handled these failure cases in the Fleet API or ClientAppIo API before. We can address them in follow-up PRs, but that feels outside the scope of this PR.

jafermarq (approved):

danieljanes (approved):

PR #7565, opened 2026-07-09
feat(framework): Add Automation CoreState methods

danielnugraha:
Adds CoreState support for automations.

- Add automation metadata storage/list/stop methods to CoreState
- Implement automation persistence for in-memory and SQL CoreState
- Store automation run inputs inline using the existing StartRun-style fields
- Add automation statuses in flwr.supercore.constant
- Add shared CoreState tests covering create/list, due filtering, series reuse, and stop behavior


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -132,6 +136,103 @@ def store_run_in_series(
...
+        *,
+        federation_id: str,
+        flwr_aid: str,
+        fab_id: str | None,
+        fab_version: str | None,
+        fab_hash: str | None,
+        override_config: UserConfig,
+        federation_config: SimulationConfig | None,
+        primary_task_type: str,
+        next_run_at: datetime,
```
panh99: Is this also the `start_at` time for scheduled one-shot run?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -132,6 +136,103 @@ def store_run_in_series(
...
+        fixed_interval: int | None = None,
+        remaining_runs: int | None = None,
```
panh99: Should this be maximum number of runs? Will we update this `remaining_runs` after each execution?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -132,6 +136,103 @@ def store_run_in_series(
...
+        """
+
+    @abstractmethod
+    def list_automations(
+        self,
+        *,
+        federation: str | None = None,
+        statuses: Sequence[str] | None = None,
+        due_before: datetime | None = None,
+        limit: int | None = None,
```
panh99, line 225: Just curious, why the two filters? Is it b/c our UI will allow users to filter by statuses/due_before?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -132,6 +136,103 @@ def store_run_in_series(
...
+        fab_id: str | None,
+        fab_version: str | None,
+        fab_hash: str | None,
+        override_config: UserConfig,
+        federation_config: SimulationConfig | None,
+        primary_task_type: str,
+        next_run_at: datetime,
+        fixed_interval: int | None = None,
+        remaining_runs: int | None = None,
+        series_id: int | None = None,
```
panh99: May I know when this field will be set?

framework/py/flwr/supercore/corestate/corestate.py
panh99: Should we use `str` here? The field comes from gRPC request, which will carry string values, right?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -242,6 +262,147 @@ def store_run_in_series(
...
+                series_id=series_id,
+                flwr_aid=flwr_aid,
+                created_at=current.isoformat(),
+                updated_at=current.isoformat(),
+                next_run_at=next_run_at.isoformat(),
+            )
+            if fixed_interval is not None:
+                automation.fixed_interval = fixed_interval
+            if remaining_runs is not None:
+                automation.remaining_runs = remaining_runs
```
panh99: Since the two are `optional` fields in gRPC message `Automation`, we can just set them like this, which is equivalent to the `if` branches.
```suggestion
                fixed_interval=fixed_interval,
                remaining_runs=remaining_runs,
            )
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -242,6 +262,145 @@ def store_run_in_series(
...
+                    )
+                )
+            ]
+            if due_before is None:
+                automations.sort(
+                    key=lambda automation: automation.updated_at,
+                    reverse=True,
+                )
+            else:
+                automations.sort(key=lambda automation: automation.next_run_at)
```
panh99: This feels slightly inconsistent to me. If the ordering matters, should we add an `order_by` argument instead of silently applying different ordering depending on whether `due_before` is set?

framework/py/flwr/supercore/corestate/corestate.py
panh99: Based on the definition of `ListAutomationsRequest`, it seems that we only need `federation` filter, right?

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -242,6 +262,145 @@ def store_run_in_series(
...
+                if (federation is None or record.automation.federation == federation)
+                and (status_set is None or record.automation.status in status_set)
+                and (
+                    cutoff is None
+                    or (
+                        record.automation.HasField("next_run_at")
+                        and record.automation.next_run_at <= cutoff
+                    )
+                )
+            ]
```
panh99: Could we replace this with something like the following, or a similar structure? The Python one-line expression here has grown to 13 lines now, which makes it a bit hard to follow.

```python
# Apply xxx filter
if xxx is not None and xxx:
    continue

# Apply yyy filter
if yyy is not None and yyy:
    continue
```

framework/py/flwr/supercore/corestate/in_memory_corestate.py
```diff
@@ -242,6 +262,145 @@ def store_run_in_series(
...
+            if due_before is None:
+                automations.sort(
+                    key=lambda automation: automation.updated_at,
+                    reverse=True,
+                )
+            else:
+                automations.sort(key=lambda automation: automation.next_run_at)
+            if limit is not None:
+                automations = automations[:limit]
+            return list(automations)
```
```suggestion
            return automations
```

framework/py/flwr/supercore/constant.py
```diff
@@ -124,6 +124,10 @@
 # Default federation names for every Flower account
 DEFAULT_FEDERATION_SIMULATION = "workspace"
 
+# Constants for automations
+AUTOMATION_STATUS_ACTIVE = "active"
+AUTOMATION_STATUS_STOPPED = "stopped"
```
panh99: Could we use a `StrEnum` class for consistency?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,200 @@ def store_run_in_series(
...
+                        "created_at": current,
+                        "updated_at": current,
+                        "next_run_at": next_run_at,
+                        "fixed_interval": fixed_interval,
+                        "remaining_runs": remaining_runs,
+                        "stopped_at": None,
+                    },
+                )
+        except IntegrityError as exc:
+            raise ValueError("Could not store automation") from exc
```
```suggestion
            raise ValueError(f"Could not store automation: {exc}") from exc
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,200 @@ def store_run_in_series(
...
+                    """,
+                    {
+                        **_run_template_to_row(
+                            fab_id=fab_id,
+                            fab_version=fab_version,
+                            fab_hash=fab_hash,
+                            override_config=override_config,
+                            federation_config=federation_config,
+                            primary_task_type=primary_task_type,
+                        ),
```
panh99: Should we remove `_run_template_to_row`? It does not seem to improve readability or simplify the code much here, since we still need to concatenate it with extra fields to form the row. Or maybe I miss , could you clarify the purpose of this function?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -132,6 +136,103 @@ def store_run_in_series(
...
+        fab_hash : str | None
+            FAB hash used by future runs.
+        override_config : UserConfig
+            Run override config used by future runs.
+        federation_config : SimulationConfig | None
+            Federation config override used by future runs.
+        primary_task_type : str
+            Primary task type used by future runs.
+        next_run_at : datetime
+            Next due time.
```
panh99: Could extend the docstring to mention that this is the start_at for oneshot automation.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,200 @@ def store_run_in_series(
...
+        fab_id: str | None,
+        fab_version: str | None,
+        fab_hash: str | None,
+        override_config: UserConfig,
+        federation_config: SimulationConfig | None,
+        primary_task_type: str,
+        next_run_at: datetime,
+        fixed_interval: int | None = None,
+        remaining_runs: int | None = None,
+        series_id: int | None = None,
```
panh99: As discussed, we could make this non-optional, and always pop updates in the very conversation (run series) where the automation is created.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,200 @@ def store_run_in_series(
...
+
+        return _automation_from_row(rows[0])
+
+    def list_automations(  # pylint: disable=too-many-arguments
+        self,
+        *,
+        federation: str | None = None,
+        statuses: Sequence[str] | None = None,
+        due_before: datetime | None = None,
+        limit: int | None = None,
```
panh99, line 420: Should we add an `order_by` argument?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,266 @@ def store_run_in_series(
...
+        stopped_at = row["stopped_at"]
+        return Automation(
+            automation_id=row["automation_id"],
+            status=row["status"],
+            federation=row["federation_id"],
+            series_id=int64_to_uint64(row["series_id"]),
+            flwr_aid=row["flwr_aid"],
+            created_at=timestamp_to_iso(row["created_at"]),
+            updated_at=timestamp_to_iso(row["updated_at"]),
+            next_run_at=timestamp_to_iso(next_run_at) if next_run_at else None,
```
panh99: It seems that this field is not optional, right? Both one-shot automation and recurring automation will set it.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,266 @@ def store_run_in_series(
...
+            fixed_interval=row["fixed_interval"],
+            remaining_runs=row["remaining_runs"],
+            stopped_at=timestamp_to_iso(stopped_at) if stopped_at else None,
```
panh99: Shouldn't this be None? Should we just remove the line?
```suggestion
```

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,266 @@ def store_run_in_series(
...
+        fab_id: str | None,
+        fab_version: str | None,
+        fab_hash: str | None,
+        override_config: UserConfig,
+        federation_config: SimulationConfig | None,
+        primary_task_type: str,
+        series_id: int,
+        next_run_at: str,
+        fixed_interval: int | None = None,
+        remaining_runs: int | None = None,
```
panh99: I think `remaining_runs` is a valid column name for the table. However, when we first store an automation, this value represents the maximum number of runs, and the `remaining_runs` column is initialized to that value.

For that reason, I think `max_runs` would be a clearer argument name for `store_automation` than `remaining_runs`. Wdyt?

framework/py/flwr/supercore/state/alembic/versions/rev_2026_07_08_add_automations.py
```diff
@@ -55,7 +55,7 @@ def upgrade() -> None:
         sa.Column("primary_task_type", sa.String(), nullable=False),
         sa.Column("created_at", sa.TIMESTAMP(timezone=True), nullable=False),
         sa.Column("updated_at", sa.TIMESTAMP(timezone=True), nullable=False),
-        sa.Column("next_run_at", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column("next_run_at", sa.TIMESTAMP(timezone=True), nullable=False),
```
jafermarq:
```suggestion
        sa.Column("next_run_at", sa.TIMESTAMP(timezone=True), nullable=True),
```

Should we fully revert this?

jafermarq (dismissed):

panh99 (dismissed):
LGTM!

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -322,6 +330,179 @@ def store_run_in_series(
...
+        params: dict[str, Any] = {}
+        if federation is not None:
+            conditions.append("federation_id = :federation_id")
+            params["federation_id"] = federation
+        if statuses is not None:
+            placeholders = ",".join(f":status_{i}" for i in range(len(statuses)))
+            conditions.append(f"status IN ({placeholders})")
+            params.update({f"status_{i}": status for i, status in enumerate(statuses)})
+        if due_before is not None:
+            conditions.append("next_run_at IS NOT NULL")
```
```suggestion
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7567, opened 2026-07-09
feat(framework): Add Automation Control implementation

danielnugraha:
Add Control API implementation for `ListAutomations`, `StartAutomation`, `StopAutomation`.


framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -267,6 +274,127 @@ def start_run(  # pylint: disable=too-many-locals, too-many-statements
...
+) -> StopAutomationResponse:
+    """Stop an automation."""
+    log(INFO, "ControlServicer.StopAutomation")
+
+    for automation in state.list_automations(order_by="updated_at"):
+        if automation.automation_id == request.automation_id:
+            _validate_federation_membership_in_request(
+                state, account.flwr_aid, automation.federation
+            )
+            break
```
panh99: This isn’t ideal, since we’d have to scan the entire table to find a specific automation.

One option would be to rename `.list_automation` to `.get_automation` and add an `automation_id` filter.  This will be similar to `.get_run_info` we currently have. What do you think?

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -267,6 +274,127 @@ def start_run(  # pylint: disable=too-many-locals, too-many-statements
...
+            order_by="updated_at",
+        )
+    else:
+        federations = state.federation_manager.get_federations(flwr_aid)
+        federation_ids = {federation.id for federation in federations}
+        automations = [
+            automation
+            for automation in state.list_automations(order_by="updated_at")
+            if automation.federation in federation_ids
+        ]
```
panh99, line 758: This reads to me like we should extend the `federation` filter in `state.list_automations` to `federations`, allowing us to filter by multiple federations. Then the logic here could be simplified to:

```python
if request.federation:
    federations = [request.federation]
else:
    federations = [
        fed.id for fed in state.federation_manager.get_federations(flwr_aid)
    ]

return ListAutomationsResponse(
    automations=state.get_automations(
        federations=federations,
        order_by="updated_at",
    )
)
```

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -267,6 +274,127 @@ def start_run(  # pylint: disable=too-many-locals, too-many-statements
     )
 
 
+def start_automation(  # pylint: disable=unused-argument
```
panh99: Could you add some one-liner comments to this function?

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -267,6 +274,127 @@ def start_run(  # pylint: disable=too-many-locals, too-many-statements
...
+    if request.HasField("start_at"):
+        try:
+            next_run_at = datetime.fromisoformat(request.start_at).isoformat()
+        except ValueError as e:
+            raise FlowerError(
+                ApiErrorCode.INVALID_RUN_CONFIG,
+                f"Invalid automation start_at value: {request.start_at}",
+            ) from e
+    else:
+        next_run_at = now().isoformat()
```
panh99, line 624: Using `ApiErrorCode.INVALID_RUN_CONFIG` here isn’t ideal.

For context, a `FlowerError` is translated into a gRPC or HTTP error using the message defined in `framework/py/flwr/supercore/error/catalog.py`. The message provided here will not be visible to users; they will only see `"Invalid run configuration."`, which is not very informative in this case.

Should we add an `INVALID_AUTOMATION_ARG` error code, or something similar? If we want CLI or UI callers to receive more specific messages, we could either introduce multiple API error codes or include the additional context in the `public_details` field of `FlowerError`.

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -574,6 +580,137 @@ def start_run(  # pylint: disable=too-many-locals, too-many-statements
...
+    """Create automation."""
+    log(INFO, "ControlServicer.StartAutomation")
+
+    # Validate the run series shared by all runs in this automation.
+    start_run_request = request.start_run_request
+    if not start_run_request.HasField("series_id"):
+        raise FlowerError(
+            ApiErrorCode.INVALID_AUTOMATION_REQUEST,
+            "StartAutomation requires start_run_request.series_id.",
+            public_details="A run series ID is required to start an automation.",
```
jafermarq:
```suggestion
            public_details="The run `series_id` is required to start an automation.",
```

framework/py/flwr/superlink/servicer/control/control_servicer_test.py
jafermarq, line 1: would it make sense to move these tests to `control_handlers_tests.py`?

jafermarq (approved):

danieljanes (approved):

PR #7578, opened 2026-07-10
refactor(framework): Use dependency and protobuf router in control router

jafermarq:
This bring several additions for FastAPI together


framework/py/flwr/superlink/dependencies/linkstate.py
```diff
@@ -20,12 +20,11 @@
 from typing import cast
 
 from fastapi import HTTPException, Request, status
-from starlette.datastructures import State
 
 from flwr.server.superlink.linkstate import LinkState, LinkStateFactory
 
 
-def get_linkstate(request: Request[State]) -> LinkState:
+def get_linkstate(request: Request) -> LinkState:  # type: ignore[type-arg]
```
panh99: Curious why this change?

jafermarq: It's because `Request[State]` is fine to keep `mypy` happy, but FastAPI requires resolving the type of `Request` at runtime. I pushed a changed that keep both happy

panh99 (approved):
LGTM!

PR #7579, opened 2026-07-10
refactor(framework): Stop passing `ObjectStore` to `ControlAPI` handlers

jafermarq:
The `ObjectStore` is instead retrieved from the `LinkState`


panh99 (approved):
LGTM!

PR #7590, opened 2026-07-13
refactor(framework): Translate `FlowerError` in unary protobuf routes 

jafermarq:
`unary_stream` will need its own handling at a later date


jafermarq:
Will be solved via _middleware_ instead

PR #7600, opened 2026-07-14
feat(framework): Add Slack connector

charlesbvll:
Follow up to #7599 

Adds a new Slack connector


framework/py/flwr/supercore/task_process/connector/slack.py
```diff
@@ -0,0 +1,242 @@
...
+SLACK_LIST_CONVERSATIONS_TOOL = "slack_list_conversations"
+SLACK_GET_CONVERSATION_HISTORY_TOOL = "slack_get_conversation_history"
+SLACK_GET_THREAD_REPLIES_TOOL = "slack_get_thread_replies"
+
+SLACK_TOOL_NAMES = (
+    SLACK_SEARCH_MESSAGES_TOOL,
+    SLACK_LIST_CONVERSATIONS_TOOL,
+    SLACK_GET_CONVERSATION_HISTORY_TOOL,
+    SLACK_GET_THREAD_REPLIES_TOOL,
+)
```
panh99: Is this only used in unit tests? If so, should we move it there?

framework/py/flwr/supercore/task_process/connector/slack.py
```diff
@@ -0,0 +1,242 @@
...
+
+from .http import ConnectorApiError, request_json_object
+from .json_utils import optional_string, require_bool, require_int_range, require_string
+from .tool_schema import function_tool, integer_property, string_property
+
+SLACK_CONNECTOR_REF = "slack"
+SLACK_SEARCH_MESSAGES_TOOL = "slack_search_messages"
+SLACK_LIST_CONVERSATIONS_TOOL = "slack_list_conversations"
+SLACK_GET_CONVERSATION_HISTORY_TOOL = "slack_get_conversation_history"
+SLACK_GET_THREAD_REPLIES_TOOL = "slack_get_thread_replies"
```
panh99: Could we use a `class SlackTool(StrEnum)` (name TBD. or we can use `SlackToolName` or `SlackToolRef`) to group the tools? It will also make it easy to do things like `for tool in SlackTool` and `if tool in set(SlackTool)`.

framework/py/flwr/supercore/task_process/connector/slack.py
panh99, line 1: Just an idea for organizing the code: could we create a `connector.slack` submodule and move `slack.py` and `slack_oauth.py` into it? We could then rename them to `tools.py` and `oauth.py`, respectively, and define the Slack-specific constants in `slack/constants.py` instead of in the previous `slack.py`.

We could also define `ConnectorDefinition` in `slack/__init__.py` and export only that, since it seems to be the only thing needed by `registry.py`.

This would give us a clean, self-contained submodule for the Slack connector. Also, if we consistently add files such as `*.py`, `*_oauth.py`, and `*_test.py` for every OAuth connector, the top-level `connector` module could become a bit cluttered over time.

Wdyt?

framework/py/flwr/supercore/task_process/connector/slack/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Slack connector."""
+
+from .definition import CONNECTOR as CONNECTOR
+
+__all__ = ["CONNECTOR"]
```
panh99:
```suggestion
from .definition import CONNECTOR

__all__ = [
    "CONNECTOR",
]
```

panh99 (approved):
LGTM

PR #7596, opened 2026-07-14
feat(framework): Introduce `AuthnPluginDependency` and additional `unary` routers for `ControlAPI`

jafermarq:
The `AuthnPluginDependency` is needed for `get_login_details` and `get_auth_tokens`

PR #7601, opened 2026-07-14
refactor(framework): Raise `FlowerError` instead of `HTTPException`

jafermarq:
To be merged after https://github.com/flwrlabs/flower/pull/7596


framework/py/flwr/supercore/error/base.py
```diff
@@ -160,3 +166,6 @@ class ApiErrorCode(IntEnum):
     FLEET_GET_RUN_FAILED = 2008
     FLEET_GET_FAB_FAILED = 2009
     FLEET_OBJECT_CONTENT_INVALID = 2010
+
+    # Runtime API errors (3001-4000)
+    RUNTIME_API_NOT_IMPLEMENTED = 3001
```
panh99: I guess we don't need an API error code for this, but let me know if you think this is a more robust approach. 🤔

framework/py/flwr/superlink/main.py
```diff
@@ -110,6 +110,7 @@ async def lifespan(fastapi_app: FastAPI) -> AsyncIterator[dict[str, object]]:
 
             fastapi_app.state.linkstate_factory = linkstate_factory
             fastapi_app.state.account_access_dep = account_access_dep
+            fastapi_app.state.authn_plugin = authn_plugin
```
panh99:
```suggestion
```

framework/py/flwr/supercore/error/base.py
```diff
@@ -145,9 +145,15 @@ class ApiErrorCode(IntEnum):
     INVALID_APP_SPEC = 33
     FAB_DOWNLOAD_LINK_FAILURE = 34
     FAB_DOWNLOAD_FAILURE = 35
+    AUTHENTICATION_FAILED = 36
```
panh99: Should we name it like this? No need to change the error messages, just to avoid confusion with other types of authentication like node authentication.

```suggestion
    ACCOUNT_AUTHENTICATION_FAILED = 36
```

framework/py/flwr/supercore/error/base.py
```diff
@@ -145,9 +145,15 @@ class ApiErrorCode(IntEnum):
...
+    AUTHENTICATION_NOT_INITIALIZED = 37
```
panh99: Same as above.
```suggestion
    ACCOUNT_AUTHENTICATION_NOT_INITIALIZED = 37
```

framework/py/flwr/supercore/error/base.py
```diff
@@ -145,9 +145,15 @@ class ApiErrorCode(IntEnum):
...
+    LINKSTATE_NOT_INITIALIZED = 38
```
panh99: Would it be better to place it under "Common API errors"? This seems to be a common error shared by all APIs.

jafermarq: moved

framework/py/flwr/supernode/main.py
```diff
@@ -52,6 +53,8 @@ async def lifespan(_: FastAPI) -> AsyncIterator[None]:
     # SuperNode APIs
     fastapi_app.include_router(runtime.router)
 
+    fastapi_app.middleware("http")(http_error_translator)
```
panh99, line 57:
```suggestion
    # Apply the FlowerError translation layer last to make it outermost
    fastapi_app.middleware("http")(http_error_translator)
```

panh99 (approved):
LGTM!

PR #7652, opened 2026-07-20
fix(framework): Do not crash push_messages() on an empty messages_list

### LINKED ISSUE #7651: push_messages() raises IndexError on an empty messages_list instead of the no-op the servicer expects
### Describe the bug

`push_messages()` in `framework/py/flwr/server/superlink/fleet/message_handler/message_handler.py` indexes `request.messages_list[0]` unconditionally. An empty `messages_list` is a legitimate request: `Fleet.PushMessages` in `fleet_servicer.py` already has an explicit branch for it and logs `"No replies to push"`, but it still calls `push_messages()` regardless, which then raises `IndexError`.

### Steps/Code to Reproduce

```python
from flwr.proto.fleet_pb2 import PushMessagesRequest
from flwr.server.superlink.fleet.message_handler.message_handler import push_messages
from flwr.server.superlink.linkstate.linkstate_factory import LinkStateFactory
from flwr.supercore.constant import FLWR_IN_MEMORY_DB_NAME
from flwr.supercore.object_store import ObjectStoreFactory
from flwr.superlink.federation import NoOpFederationManager

state = LinkStateFactory(FLWR_IN_MEMORY_DB_NAME, NoOpFederationManager(), ObjectStoreFactory()).state()
push_messages(request=PushMessagesRequest(messages_list=[], message_object_trees=[]), state=state)
```

Output:

```
IndexError: list index out of range
```

### Expected Results

An empty `messages_list` should return a no-op `PushMessagesResponse`, matching what the servicer's own log message already implies happens.

### Actual Results

`IndexError`, unhandled by the servicer, so the gRPC call fails.

### Additional context

Have a fix + regression test ready to send as a PR.

Atishyy27:
### Issue

Fixes #7651

### Description

`push_messages()` indexed `request.messages_list[0]` unconditionally. An empty `messages_list` is a legitimate case: `Fleet.PushMessages` in `fleet_servicer.py` already has a branch that logs `"No replies to push"` for it, but still calls `push_messages()`, which then raised `IndexError` instead of returning cleanly.

This adds a guard that returns an empty `PushMessagesResponse` when the list is empty, matching the no-op behaviour the servicer's own log message already implies.

**before**

```
push_messages(request=PushMessagesRequest(messages_list=[], message_object_trees=[]), state=state)
IndexError: list index out of range
```

**after**

```
push_messages(...) -> PushMessagesResponse()   # no-op, state untouched
```

### Testing

Added `test_push_messages_empty_list`, asserting a no-op response and that `state.start_session` / `state.store_message_and_object_tree` are not called. It fails before the change and passes after.

```
pytest framework/py/flwr/server/superlink/fleet/message_handler/message_handler_test.py
4 passed
```


WilliamLindskog:
@panh99 seems okay, wdyt?

jafermarq:
Hi @Atishyy27 , thanks for opening a PR for your issue. I understand the proposal but I think for this to be useful (i.e. `FleetAPI.PushMessage` to ever receive an empty list) more changes are needed. This is because [`ClientApp`](https://github.com/flwrlabs/flower/blob/45edab61efac3fcdcb42a36023bd10de7e959c4b/framework/py/flwr/clientapp/client_app.py#L139) returns always a `Message` (it's not optional). How did you encounter this? 

We are looking into it. cc @panh99

Atishyy27:
Fair, I did not check the caller before opening this. If ClientApp always returns a Message the guard is unreachable and I would rather close it than leave dead defensive code. Is an empty list expected to become reachable later?

PR #7664, opened 2026-07-22
feat(framework): Deliver OAuth credentials to connector tasks

charlesbvll:
This PR securely delivers OAuth connector credentials to authorized connector tasks at runtime.

- Adds an authenticated `ServerAppIo.GetConnector` endpoint that validates the task, run binding, run owner, and connector reference
- Credential-backed connector tasks fetch credentials and configuration only when needed
- Credentials are never included in task messages, and provider failures are sanitized to prevent secret leakage
- Built-in credential-free connectors remain unchanged


framework/py/flwr/superlink/servicer/serverappio/serverappio_servicer.py
```diff
@@ -236,6 +238,49 @@ def GetRun(
...
+        task = get_authenticated_task()
+        if (
+            task.type != TaskType.CONNECTOR
+            or task.connector_ref != request.connector_ref
+        ):
+            context.abort(
+                grpc.StatusCode.PERMISSION_DENIED,
+                "Connector credentials are not available to this task.",
+            )
+
```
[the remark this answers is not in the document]
charlesbvll: `ServicerContext.abort()` always raises and terminates the RPC. Other authorization checks in this servicer rely on that behavior as well, so adding a second unreachable exception here would be redundant

framework/py/flwr/supercore/auth/policy.py
```diff
@@ -70,4 +70,8 @@ def _build_runtime_method_auth_policy(
 
 
 SERVERAPPIO_METHOD_AUTH_POLICY = _build_runtime_method_auth_policy("ServerAppIo")
+# GetConnector is ServerAppIo-only, so keep it out of the shared runtime policy.
```
danielnugraha: Can we make it part of the normal policy even though it is only implemented in serverappio for now?

framework/py/flwr/supercore/task_process/connector/task_test.py
```diff
@@ -0,0 +1,168 @@
+# Copyright 2026 Flower Labs GmbH. All Rights Reserved.
```
danielnugraha, line 1: Can we keep this tests only? Also use mocked provider instead of manually capturing arguments.

```
def test_handle_task_passes_credentials_to_matching_provider() -> None:

    ...

def test_handle_task_does_not_expose_credentials_in_provider_errors() -> None:

    ...
```

framework/py/flwr/superlink/servicer/serverappio/serverappio_servicer.py
```diff
@@ -236,6 +238,49 @@ def GetRun(
...
+        task = get_authenticated_task()
+        if (
+            task.type != TaskType.CONNECTOR
+            or task.connector_ref != request.connector_ref
+        ):
+            context.abort(
+                grpc.StatusCode.PERMISSION_DENIED,
+                "Connector credentials are not available to this task.",
+            )
```
danielnugraha: Do we need to have this check? This is already checked during `create_task` right?

framework/py/flwr/superlink/servicer/serverappio/serverappio_servicer.py
```diff
@@ -236,6 +238,49 @@ def GetRun(
...
+        run = runs[0] if runs else None
+        if run is None or not run.flwr_aid:
+            context.abort(grpc.StatusCode.NOT_FOUND, "Connector not found.")
+            raise RuntimeError("This line should never be reached.")
+
+        if request.connector_ref not in state.get_run_connector_refs(
+            run_id=task.run_id
+        ):
+            context.abort(grpc.StatusCode.NOT_FOUND, "Connector not found.")
+            raise RuntimeError("This line should never be reached.")
```
danielnugraha: Do we need to have this checks?

framework/py/flwr/superlink/servicer/serverappio/serverappio_connector_test.py
```diff
@@ -0,0 +1,190 @@
+# Copyright 2026 Flower Labs GmbH. All Rights Reserved.
```
danielnugraha: Should we avoid creating real runs/tasks for every tests to make this minimal? Also why do we create a new file?

framework/py/flwr/superlink/servicer/serverappio/serverappio_servicer.py
```diff
@@ -236,6 +238,49 @@ def GetRun(
 
         return GetRunResponse(run=run_to_proto(runs[0]))
 
+    def GetConnector(
```
danielnugraha, line 260: @panh99 qq: Should we have this in `appio_servicer` even though right now it's only used in serverappio?

framework/py/flwr/superlink/servicer/serverappio/serverappio_servicer.py
charlesbvll: Maybe I’m missing something, but doesn’t the interceptor only check that the token belongs to a valid task? If we remove this check, couldn’t another task use its token to call `GetConnector` with any `connector_ref`? Since the connector task already stores the ref when it is created, would it make more sense to make `GetConnectorRequest` empty and always use `task.connector_ref` instead?

danielnugraha: I think it makes sense.

framework/py/flwr/supercore/task_process/connector/registry.py
```diff
@@ -44,12 +47,35 @@ def invoke_connector(
...
+
+
+def requires_connector_credentials(name: str) -> bool:
+    """Return whether a connector uses account-scoped credentials."""
+    return name in _CREDENTIAL_CONNECTOR_HANDLERS
+
+
+def get_connector_ref(name: str) -> str:
+    """Return the connector reference associated with a tool name."""
+    return _CREDENTIAL_CONNECTOR_REFS.get(name, name)
```
panh99, line 80: May I know what's this function for? Does that mean we will replace the original connector ref with a new one if and only if the name is an OAuth connector?

framework/py/flwr/supercore/task_process/connector/task.py
```diff
@@ -61,23 +67,46 @@ def _push_connector_response(response: JSONObject) -> None:
...
             ),
             "error": None,
         }
-    except Exception as ex:
-        response = _make_error_response(ex)
-        raise
+    except Exception as ex:  # pylint: disable=broad-exception-caught
+        if uses_credentials:
+            response = _make_error_response(None)
+            credential_failure = True
```
panh99, line 96: Should we just `raise RuntimeError(...) from None`?

framework/py/flwr/superlink/servicer/serverappio/serverappio_servicer.py
panh99: The `GetConnector` implementation uses `LinkState.get_run_info`, so we cannot move it to `appio_servicer`. It's okay to keep it here for now.

framework/py/flwr/supercore/task_process/connector/task_test.py
panh99, line 1: The unit tests look good. Would it make sense to group them under a `TestCase` so we can move the common setup and patching logic into `setUp` instead of repeating it? For example, the patchers for `_pull_connector_request`, `_CREDENTIAL_CONNECTOR_HANDLERS` and etc.

framework/py/flwr/supercore/task_process/connector/registry.py
charlesbvll: @panh99 Yes, mostly, it actually maps the model-facing tool name to the OAuth connector ref used for task creation and credential lookup. For example, `slack_search_messages` maps to `slack`. If there is no mapping, we keep the original name. I added a small comment/docstring to make this clearer

framework/py/flwr/supercore/task_process/connector/task.py
charlesbvll: I thought about that, but `from None` only hides the original exception when the traceback is formatted, i.e the provider exception is still kept in `__context__`. Since that exception could contain credentials, the current code raises after leaving the except block so it is not attached at all. There is also a test checking that `__context__` is `None`, so I think we should keep this as is

framework/py/flwr/supercore/task_process/connector/task_test.py
charlesbvll: Good point! I grouped the tests under a `TestCase` and moved the common setup and patching into `setUp`

panh99 (approved):

danieljanes (approved):

PR #7688, opened 2026-07-27
refactor(framework): Add CoreState declarative metadata pattern

smoroso:
## Issue

The Framework state schema is currently defined with SQLAlchemy Core Table objects. That works, but it makes it difficult to incrementally move state tables toward declarative SQLAlchemy models and, later, ORM-based reads and writes.

This PR establishes the first small declarative-model pattern using the existing CoreState nonce_store table. It is intentionally a representation-only change: the database schema, SQL queries, LinkState runtime behavior, and transaction/session handling remain unchanged.

## What this PR adds

- FlwrBase, the declarative metadata base for future Framework state models.
- NonceStore, a declarative mapping for the existing nonce_store table.
- Metadata-parity tests that compare the declarative mapping with the existing SQLAlchemy Core definition.

NonceStore preserves the existing table exactly:

- table name: nonce_store;
- composite primary key: namespace + nonce;
- columns and SQLAlchemy types: namespace (String), nonce (String), expires_at (Float);
- non-nullability and primary-key flags;
- existing index: idx_nonce_store_expires_at on expires_at.

No migration is included because this PR does not change the physical schema.

## Why start with nonce_store?

nonce_store is a deliberately small and low-risk first example:

- it has only three columns;
- it has a real composite primary key;
- it has an existing index;
- it has no relationships or foreign keys;
- it is not being switched to ORM query code in this PR.

That keeps the first pattern easy to review while still covering the important schema details that future state-table conversions must preserve.

## How this relates to LinkState

LinkState is the runtime state abstraction used by the SuperLink. SqlLinkState combines the LinkState and CoreState responsibilities and uses the state database as the persistence boundary. Its database contains multiple groups of tables, including LinkState tables and CoreState tables such as nonce_store.

This PR does not change SqlLinkState or make NonceStore a runtime dependency. Existing LinkState/CoreState SQL paths continue to use the current definitions and behavior. The new declarative model is a second, checked representation of the same table so that we can validate the ORM metadata before changing any runtime access path.

The important invariant is that introducing a declarative model must not silently alter the database that LinkState operates on. If the declarative mapping changed a column type, nullability, primary-key definition, or index, Alembic could see a schema difference that was not intended. The parity test makes that risk visible immediately.

This separation is useful for the LinkState migration because it lets us migrate in stages:

1. establish declarative metadata for a table;
2. prove that it matches the existing Core definition;
3. include it in the combined state metadata used for schema comparison;
4. only then migrate individual LinkState/CoreState query paths to ORM operations;
5. keep the existing LinkState session, engine, transaction, locking, and atomicity behavior unchanged while each path is converted.

In other words, this PR prepares the schema/model layer without mixing it with the much more sensitive runtime LinkState conversion. It gives later ORM work a stable contract to build on.

The metadata is also kept separate from runtime query migration on purpose. A declarative model can exist and be validated without requiring every caller to switch from SQLAlchemy Core or raw SQL at the same time. That makes the conversion incremental and allows each LinkState path to be reviewed for its own locking, transaction, and concurrency requirements.

## Why the parity test matters

The test compares the declarative table with the existing Core metadata rather than merely checking that the model can be imported.

It verifies the schema-relevant details:

- table and column identity;
- SQLAlchemy column types;
- nullability;
- primary-key and uniqueness flags;
- server defaults and foreign-key details where present;
- primary-key column order;
- index names, columns, and uniqueness.

This protects against accidental schema drift while the model representation changes. It also provides a reusable test pattern for the next CoreState and LinkState table conversions.

## Scope and follow-up

This PR intentionally does not:

- change the database schema;
- add an Alembic migration;
- replace LinkState or CoreState SQL queries;
- change ORM sessions, engines, transactions, locking, or runtime behavior;
- convert all state tables at once.

Follow-up PRs can add more declarative models using the same parity approach. Once the relevant tables have equivalent metadata, runtime consumers can be migrated one path at a time with the LinkState transaction and concurrency guarantees preserved.

## Validation

The following checks were run from framework/:

- declarative model tests;
- Alembic metadata/schema parity tests;
- Ruff;
- mypy;
- Black;
- PR-title validation.

All checks passed locally.


tanertopal (approved):
Lgtm! I tested alembic revision generation locally and an empty revision was generated as planned.

PR #7726, opened 2026-07-30
feat(framework): Allow StartRun to reuse fab from previous run

danielnugraha:
Extend `StartRun` with `source_run_id` field to allow reusing fab from previous run.


framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
```diff
@@ -102,6 +103,18 @@ def federation_manager(self) -> FederationManager:
...
+        if federation_id is not None:
+            with self.lock:
+                if not any(
+                    record.run.fab_hash == fab_hash
+                    and record.run.federation_id == federation_id
+                    for record in self.run_ids.values()
+                ):
+                    return None
+        return InMemoryCoreState.get_fab(self, fab_hash)
+
```
panh99: Searching the entire run history table just to determine the FAB’s federation affiliation does not seem very efficient, especially since the FAB is not currently associated with a federation in the first place.

Since the FAB hash is derived from its content, I think it is generally fine to resolve the hash provided by the requester directly. As long as the hash is not publicly exposed, that should be safe enough for now.

The main point is to avoid introducing a new federation-level FAB ownership model in a PR whose scope is simply resolving a FAB hash to its content. We can certainly address ownership in a separate PR, but I do not think it is urgent. Wdyt?

framework/proto/flwr/proto/fab.proto
```diff
@@ -20,9 +20,9 @@ package flwr.proto;
 import "flwr/proto/node.proto";
 
 message Fab {
-  // This field is the hash of the data field. It is used to identify the data.
-  // The hash is calculated using the SHA-256 algorithm and is represented as a
-  // hex string (sha256hex).
+  // This field identifies FAB content using its SHA-256 hex digest. When
+  // `content` is empty in StartRunRequest, SuperLink uses this hash to resolve
+  // an existing FAB from the target federation.
```
panh99: IIUC, the SuperLink will only try to resolve an existing FAB from LinkState, but not from FlowerHub, right? Plus, a FAB is not currently associated with a federation in the LinkState. Prolly we'd better say that the SuperLink will try to resolve the FAB using the hash from stored FABs.

danielnugraha: The comment is removed.

framework/py/flwr/server/superlink/linkstate/in_memory_linkstate.py
danielnugraha: Added index in 7746.

framework/py/flwr/superlink/servicer/control/control_handlers_test.py
```diff
@@ -14,8 +14,9 @@
 # ==============================================================================
 """Tests for Control API handler functions."""
```
panh99, line 16:
```suggestion


```

framework/proto/flwr/proto/fab.proto
panh99: Should we say "from LinkState" instead of "from the target federation" since it's not yet implemented?

panh99 (approved):

tanertopal (approved):

PR #7756, opened 2026-08-04
feat(framework): Add OAuth connector provider utilities

danielnugraha:
Add OAuth connector provider utilities:
  - Shared JSON HTTP handling
  - JSON validators
  - OAuth base utilities
  - Minimal infrastructure test


framework/py/flwr/supercore/task_process/connector/oauth.py
```diff
@@ -47,3 +52,127 @@ def exchange_code(
...
+    names = (client_id_env, client_secret_env, redirect_uri_env)
+    values = {name: os.getenv(name, "").strip() for name in names}
+    if not any(values.values()):
+        return None
+    missing = [name for name, value in values.items() if not value]
+    if missing:
+        raise RuntimeError(
+            f"{provider_name} OAuth configuration requires environment variables: "
+            f"{', '.join(missing)}."
+        )
```
panh99: This is already checked in the constructor of `BaseOAuthProvider`, no? Do we need to double check here?

panh99: Prolly we can just do:

```python
    return provider_type(
        client_id=os.getenv(client_id_env, ""),
        client_secret=os.getenv(client_secret_env, ""),
        redirect_uri=os.getenv(redirect_uri_env, ""),
    )

```

panh99 (approved):
LGTM!

PR #7758, opened 2026-08-04
refactor(framework): Use ORM for object push sessions

smoroso:
### Description

Object-push session bookkeeping in `SqlCoreState` still used manual SQL strings even though declarative models exist for the affected tables.

### Related issues/PRs

Follow-up to the CoreState ORM migration series.

### Explanation

Use SQLAlchemy model/Core statements for object-push session bookkeeping, including session creation, run-scoped deletion, object-tree preregistration, pending object claims, object retrieval cleanup, and push-session cleanup. The change preserves the existing object-store side effects, pending-count synchronization, rollback behavior when object storage fails, and expired-session cleanup semantics.

No schema or Alembic changes are included.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Address LLM-reviewer comments, if applicable
- [x] Make CI checks pass
- [x] Ping maintainers on Slack

### Any other comments?

Existing object-push coverage exercises the changed behavior.

Validation:

```bash
cd framework
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m pytest py/flwr/server/superlink/linkstate/linkstate_test.py -k 'object'
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m pytest py/flwr/server/superlink/linkstate/linkstate_test.py
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m pytest py/flwr/supercore/state/schema/corestate_models_test.py
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m pytest py/flwr/supercore/state/alembic/utils_test.py -k "test_migrated_schema_matches_metadata"
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m ruff check py/flwr/supercore/corestate/sql_corestate.py
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m mypy py/flwr/supercore/corestate/sql_corestate.py
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m black --check py/flwr/supercore/corestate/sql_corestate.py
PYENV_VERSION=3.11 pyenv exec uv run --no-sync python -m isort --check-only py/flwr/supercore/corestate/sql_corestate.py
git diff --check
```


framework/py/flwr/supercore/sql_mixin.py
```diff
@@ -195,6 +196,15 @@ def session(self) -> Iterator[Session]:
             _current_sessions.reset(token)
             session.close()
 
+    def dialect_insert(self, table: Any) -> Any:
```
panh99:
```suggestion
    def dialect_insert(self, table: Any) -> Insert:
```

framework/py/flwr/supercore/sql_mixin.py
```diff
@@ -25,6 +25,7 @@
 from typing import Any
 
 from sqlalchemy import Engine, MetaData, create_engine, event, inspect, make_url, text
+from sqlalchemy.dialects.sqlite import insert as sqlite_insert
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session, sessionmaker
 from sqlalchemy.pool import StaticPool
```
panh99, line 31:
```suggestion
from sqlalchemy.pool import StaticPool
from sqlalchemy.sql.dml import Insert
```

framework/py/flwr/supercore/corestate/sql_corestate.py
panh99, line 1: Could you take another look at the timestamp handling and see whether there might be a cleaner, more dialect-independent approach?

The previous raw SQL passed an untyped `datetime` parameter to the SQLite driver, which preserved the UTC offset. With the ORM/Core statements, SQLAlchemy now applies the mapped `TIMESTAMP(timezone=True)` type. Since SQLite does not natively support timezone-aware timestamps, this changes how the value is stored and returned.

The new `_timestamp_bind`, `_timestamp_to_sqlite_text`, and `_datetime_assuming_utc` helpers address that SQLite compatibility issue. I wonder, though, whether we could centralize this behavior in the database/type layer instead of handling it at individual query call sites. That could keep the ORM code simpler and make future PostgreSQL support easier.

Would it be worth exploring one of the following?

- A dialect-aware SQLAlchemy type or model-level adapter.
- Native timezone-aware timestamp handling on PostgreSQL, with SQLite-specific behavior contained in the database layer.
- A one-time Alembic data revision to normalize existing SQLite timestamps, if that would simplify the runtime code. We would just need to consider rolling-upgrade compatibility.

Our production deployment uses PostgreSQL, where timezone-aware timestamps are supported natively. Ideally, the shared code could remain dialect-independent while keeping the SQLite compatibility logic as small and centralized as possible.

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -462,6 +463,102 @@ def test_get_object_returns_object_store_result_without_cleanup(self) -> None:
         self.assertEqual(load_object.call_count, 3)
         cleanup_session.assert_not_called()
 
+    def test_get_object_preserves_old_format_unexpired_session(self) -> None:
```
panh99: Just curious, are these unit tests essential? If not, would it be okay to remove them? This seems to be covered by `framework/py/flwr/supercore/state/schema/types_test.py` already.

This PR is intended to migrate the code to use the ORM without introducing any behavioural changes, so the existing unit tests should already provide sufficient coverage.

The new tests also rely on raw SQL queries, which exposes implementation details. Tests in `corestate_test.py` should ideally be written against the `CoreState` ABC rather than assuming the internal implementation of `SqlCoreState`.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -122,109 +131,84 @@ def object_store(self) -> ObjectStore:
...
             if missing_objects:
-                self.query(
-                    """
-                    INSERT INTO object_push_session_pending (session_id, object_id)
-                    VALUES (:session_id, :object_id)
-                    ON CONFLICT(session_id, object_id) DO NOTHING
-                    """,
+                stmt = cast(
+                    Any, self.dialect_insert(ObjectPushSessionPendingModel)
+                ).values(
```
panh99: Is it possible to annotate the `.dialect_insert` and drop `cast` here?

panh99: One idea is to override `.dialect_insert` in `SqlCoreState` and annotate it as the return type of `sqlite_select`, then we can let `mypy` check the API usage. Wdyt?

framework/py/flwr/supercore/corestate/corestate_test.py
```diff
@@ -462,6 +462,60 @@ def test_get_object_returns_object_store_result_without_cleanup(self) -> None:
...
+
+            mock_datetime.now.return_value = store_at
+            with patch.object(
+                session, "scalar", side_effect=scalar_with_concurrent_refresh
+            ):
+                stored = state.store_object(run_id, session_id, object_id, b"content")
+
+        self.assertTrue(stored)
+        put_object.assert_called_once_with(object_id, b"content")
+
```
panh99:
```suggestion
```

panh99 (approved):

framework/py/flwr/supercore/corestate/corestate_test.py
smoroso: this one covers refreshed expiry semantics rather than type conversion. Happy to simplify it if we can keep equivalent coverage

PR #7773, opened 2026-08-06
feat(framework): Fix unsupported SQL query for Postgres

danielnugraha:
Fix unsupported automation SQL query for Postgres.


panh99 (approved):
LGTM!

PR #7786, opened 2026-08-07
feat(framework): Unify OAuth configuration

danielnugraha:
## Summary

  Introduces declarative OAuth configuration for connectors and migrates Slack to use it.

  - Adds shared authorization-code exchange, PKCE, nested token, and scope validation
  - Moves Slack OAuth configuration into definition.py
  - Removes Slack’s provider-specific oauth.py
  - Preserves Slack’s existing user-token behavior


framework/py/flwr/supercore/task_process/connector/registry.py
```diff
@@ -123,11 +121,11 @@ def get_builtin_connector_tool(name: str) -> JSONObject:
...
-def get_oauth_connector_provider(connector_ref: str) -> OAuthConnectorProvider:
-    """Return the OAuth provider registered for a connector reference."""
-    for provider in OAUTH_CONNECTOR_PROVIDERS:
-        if provider.connector_ref == connector_ref:
-            return provider
+def get_oauth_flow(connector_ref: str) -> OAuthFlow:
+    """Return the OAuth flow registered for a connector reference."""
+    for flow in OAUTH_FLOWS:
+        if flow.connector_ref == connector_ref:
+            return flow
```
panh99: Do we use `OAUTH_FLOWS` anywhere else? Would it be better if we make it a `dict[str, OAuthFlow]` instead of a tuple?

panh99 (approved):
LGTM!

PR #7787, opened 2026-08-07
feat(framework): Add agent selector in flwr chat CLI

danielnugraha:
Adds an @ agent selector to flwr chat. Agents are fetched from /v1/user/agents, displayed in a dropdown, and used for subsequent chat runs. The default remains @flwrlabs/flwr-agent.


framework/py/flwr/cli/chat_app.py
```diff
@@ -282,12 +342,31 @@ def _submit_prompt(self, event: KeyPressEvent) -> None:
         if self._handle_command(event, stripped_prompt):
             return
 
+        selected_agent, prompt = _extract_agent_selection(prompt)
+        if selected_agent is not None:
+            agent = _find_agent(selected_agent, self.completer.agents)
```
panh99: **P1 — `fab_hash` can be lost before agent discovery completes**

`/v1/user/agents` includes custom unpublished AgentApps from the selected federation together with their exact `fab_hash`. However, if a user submits `@foo/bar ...` before the threaded completer has populated `self.completer.agents`, `_find_agent()` returns `None` and we fall back to `app_spec` with `fab_hash=None`.

For an unpublished custom AgentApp, that changes the run from using the exact discovered FAB to app-spec resolution, so a valid selectable agent can fail to run.

Could we distinguish “agents not loaded yet” from “agent not found”? For example, ensure discovery has completed before resolving a leading `@...`, while still allowing an actually unknown app spec to fall back to normal app-spec resolution.

framework/py/flwr/cli/chat.py
```diff
@@ -17,23 +17,57 @@
...
+    """Resolve the federation used for agent discovery and chat runs."""
+    if configured_federation is not None:
+        return configured_federation
+
+    active_federations = [
+        federation for federation in federations if not federation.archived
+    ]
+    default_suffix = f"/{DEFAULT_FEDERATION_SIMULATION}"
+    for federation in active_federations:
+        if federation.name.endswith(default_suffix):
```
panh99: **P1 — Default federation selection can pick another workspace**

`_resolve_chat_federation()` currently picks the first active federation whose name ends with `/workspace`.

`ListFederations` can return all federations the account is a member of, so if the user belongs to multiple workspaces, this heuristic can select another account/team's workspace instead of the user's default one. That federation is then used both for `/v1/user/agents` discovery and subsequent chat runs.

Could we avoid reimplementing default-federation resolution client-side here? Ideally the server/API would resolve the authenticated user's default federation when none is provided. If the CLI has to resolve it, I think we need a reliable identifier rather than matching the `/workspace` suffix.

framework/py/flwr/cli/chat_app.py
```diff
@@ -94,30 +106,69 @@ class _DetailsBlock:
...
+                    )
+            return
+
+        if not text.startswith("@"):
+            return
+        if self.agents is None:
+            try:
+                self.agents = fetch_chat_agents(self.auth_plugin, self.federation)
+            except (click.ClickException, requests.RequestException, ValueError):
+                self.agents = []
```
panh99: **P2 — A transient discovery failure is cached permanently**

If `fetch_chat_agents()` fails once, `get_completions()` sets:

```python
self.agents = []
```

After that we never retry because `self.agents is None` is false.

This means a temporary timeout, 5xx, expired-token race, etc. can disable agent discovery for the entire chat session without any indication to the user.

Could we cache only successful discovery results and leave transient failures retryable? If useful, permanent cases such as “feature unavailable for this user” could still be cached separately.

framework/py/flwr/cli/chat_app.py
```diff
@@ -576,6 +663,91 @@ def parse_task_event(task_event: TaskEvent) -> tuple[str, JSONObject]:
...
+        raise click.ClickException("Failed to load available agents.") from exc
+
+    return _parse_agents(payload)
+
+
+def _parse_agents(payload: Any) -> list[_Agent]:
+    """Parse the user agents API response into completion entries."""
+    raw_agents = payload
+    if isinstance(payload, dict):
+        raw_agents = payload.get("agents", payload.get("apps", payload.get("data", [])))
```
panh99: **Simplification — `_parse_agents()` seems to support speculative API shapes**

The `/v1/user/agents` response has a defined schema (`agents`, `app_id`, `display_name`, `description`, `fab_hash`), but `_parse_agents()` currently accepts several alternative shapes:

- `agents` / `apps` / `data`
- `app_id` / `app_spec` / `fab_id` / `id`
- `display_name` / `name`
- `description` / `summary`
- app IDs with or without `@`

I don't think we currently have producers for these alternative formats. Supporting them preemptively adds quite a bit of defensive code and can also hide an API contract regression by silently accepting or dropping unexpected data.

Could we parse the actual `/v1/user/agents` contract directly and fail narrowly if that contract changes? This should make this function significantly shorter and easier to reason about.

framework/py/flwr/cli/chat_app.py
```diff
@@ -94,30 +106,69 @@ class _DetailsBlock:
...
+                        selected_style="#ffffff bg:#dc8400 noreverse",
+                    )
+            return
+
+        if not text.startswith("@"):
+            return
+        if self.agents is None:
+            try:
+                self.agents = fetch_chat_agents(self.auth_plugin, self.federation)
+            except (click.ClickException, requests.RequestException, ValueError):
```
panh99: **Simplification — The outer exception list looks redundant**

`fetch_chat_agents()` already converts `requests.RequestException` and JSON `ValueError` into `click.ClickException`.

Because of that, this:

```python
except (click.ClickException, requests.RequestException, ValueError):
```

in the completer appears to defend against exceptions that `fetch_chat_agents()` no longer exposes.

Could this just catch `click.ClickException`? Keeping the exception translation in one layer would make the failure behavior easier to follow.

framework/py/flwr/cli/chat_app.py
```diff
@@ -576,6 +663,91 @@ def parse_task_event(task_event: TaskEvent) -> tuple[str, JSONObject]:
     return event_type, payload
 
 
+def fetch_chat_agents(
+    auth_plugin: CliAuthPlugin, federation: str | None
+) -> list[_Agent]:
+    """Fetch agents available to the authenticated Flower account."""
+    metadata = auth_plugin.write_tokens_to_metadata([])
```
panh99: **Simplification — We can access the OIDC token directly instead of round-tripping through gRPC metadata**

To call `/v1/user/agents`, we currently obtain the bearer token by calling `write_tokens_to_metadata([])` and then scanning the resulting gRPC metadata for `ACCESS_TOKEN_KEY`.

Since this is specifically a SuperGrid REST endpoint, and the existing SuperGrid REST CLI code already uses `OidcCliPlugin.access_token` directly, I think we can make this quite a bit simpler and avoid coupling the REST request to the gRPC metadata representation.

For example, instead of:

```python
metadata = auth_plugin.write_tokens_to_metadata([])
access_token = next(
    (
        value.decode("utf-8") if isinstance(value, bytes) else value
        for key, value in metadata
        if key == ACCESS_TOKEN_KEY
    ),
    None,
)
if access_token is None:
    raise click.ClickException(
        "Missing authentication tokens. Please login first."
    )
```

could we follow the existing REST pattern and do something along these lines?

```python
if not isinstance(auth_plugin, OidcCliPlugin) or not auth_plugin.access_token:
    raise click.ClickException(
        "Missing authentication tokens. Please login first."
    )

access_token = auth_plugin.access_token
```

Then the request remains simply:

```python
headers = {"Authorization": f"Bearer {access_token}"}
```

This removes the metadata scan, the `bytes` handling, and the dependency on `ACCESS_TOKEN_KEY`, and it matches how the CLI already authenticates other SuperGrid REST requests.

panh99 (commented):
Inline review comments below.

panh99 (approved):
LGTM!

PR #7792, opened 2026-08-08
feat(framework): Add OAuth connector registry

danielnugraha:
## Summary

  Adds generated connector registration and scaffolding.

  - Loads connectors from a generated static registry instead of manual imports.
  - Adds a command to scaffold actions.py, definition.py, and executors.py.
  - Validates connector identifiers and registry freshness in CI.
  - Keeps registry generation separate from formatting.


framework/dev/generate_connector_test.py
panh99, line 1: It's slightly unusual to have a unit test file for a dev script, but I'm okay with it if you think the script contains enough non-trivial logic to justify the tests. I'm less sure, however, whether these tests will actually be covered by CI. The current `./framework/dev/test.sh` only runs tests under `framework/py/flwr`.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7796, opened 2026-08-09
refactor(framework): Standardize Control HTTP bearer authentication

danieljanes:


framework/py/flwr/supercore/error/http.py
```diff
@@ -57,9 +57,17 @@ async def http_error_translator(
...
+        # this 401 response from other API errors and retry with a Bearer token.
+        headers = (
+            {"WWW-Authenticate": "Bearer"}
+            if err.code == ApiErrorCode.ACCOUNT_AUTHENTICATION_FAILED
+            else None
+        )
         return Response(
             status_code=http_status,
             content=err.to_json(public_message),
+            headers=headers,
```
panh99: We'd prefer to avoid adding error-specific logic to the shared `http_error_translator` infrastructure. Could we handle this differently?

framework/py/flwr/superlink/dependencies/account.py
```diff
@@ -46,30 +50,42 @@ def __init__(
...
             )
-
-        tokens, account = self.authn_plugin.refresh_tokens(metadata)
-        if tokens is None:
+            if scheme.lower() != "bearer" or not access_token:
+                raise FlowerError(
+                    ApiErrorCode.ACCOUNT_AUTHENTICATION_FAILED,
+                    "Authorization header does not contain a Bearer token.",
+                )
+            metadata = [(ACCESS_TOKEN_KEY, access_token)]
```
panh99: Why replace `metadata = request.headers.items()` with this? IIUC, the purpose of introducing `NoOpControlAuthnPlugin` is to avoid `if-else` logic like this. If needed, we can also revisit the design of the `AuthnPlugin` ABC.

framework/py/flwr/superlink/routers/control/router_test.py
```diff
@@ -147,27 +150,33 @@ def test_list_runs_returns_runs_from_linkstate() -> None:
...
-    )
     app.dependency_overrides[get_linkstate] = lambda: linkstate
     response = TestClient(app).post(
         "/v1/control/list-runs",
         content=ListRunsRequest().SerializeToString(),
-        headers={"content-type": PROTOBUF_MEDIA_TYPE},
+        headers={
+            "authorization": "Bearer invalid-token",
+            "content-type": PROTOBUF_MEDIA_TYPE,
+        },
```
panh99, line 163: Why are we changing how the tokens are carried in this PR, which is supposed to focus on removing the token refresh mechanism? This change seems potentially risky and outside the intended scope of the PR.

danieljanes: Do you mean why we're changing from `flwr-oidc-access-token: <access-token>` to `Authorization: Bearer <token>`?

danieljanes: The change from `flwr-oidc-access-token: <access-token>` to `Authorization: Bearer <token>` is intentional. Combined with the removed token refresh mechanism it standardizes the auth approach of these endpoints and brings them closer to standard practices. The header name change is also motivated by the fact that CORS only allows a very specific set of headers, with `Authorization` being one of them.

panh99: I see. That also works for me!

framework/py/flwr/superlink/dependencies/account.py
danieljanes: Good catch. I changed to logic here to delegate auth header checks to the auth plugin: https://github.com/flwrlabs/flower/pull/7796/changes/e2392d61e6b5548a05a7033ca52fe22bccbbf7c5

framework/py/flwr/supercore/error/http.py
danieljanes: Wdyt about this? https://github.com/flwrlabs/flower/pull/7796/changes/ed9fe3208305b3ac8c125790c08a9b60d77f9f12

panh99: Looks good!

panh99 (approved):
LGTM!

PR #7798, opened 2026-08-09
refactor(framework): Migrate event log from middleware to dependency

danieljanes:
## Summary

This PR converts Control API event logging from an application-wide middleware into a function-scoped FastAPI dependency attached directly to the Control router.

It:

- Replaces `ControlEventLogMiddleware` with `log_control_event`.
- Applies the dependency to every Control route through the router.
- Preserves before-event, after-event, and exception logging.
- Continues running synchronous plugin operations in a thread pool.
- Keeps authentication, licensing, and protobuf translation unchanged.
- Moves event-log tests next to the dependency implementation.

## Why this is better

Event logging is endpoint policy, not application-wide transport behavior. Expressing it as a router dependency makes that ownership explicit and ensures it applies only to Control endpoints.

This removes one layer from the global middleware stack and avoids relying on implicit middleware placement to decide which endpoints are logged. The dependency explicitly consumes the parsed protobuf request and executes around the Control handler.

The dependency uses `scope="function"` so its cleanup runs while the raw protobuf handler result is still available. This preserves the existing sequence:

1. Protobuf middleware parses the request.
2. Event logging writes the before-event.
3. The Control handler executes.
4. Event logging writes the response or exception.
5. Protobuf middleware serializes the response.

This is also a preparatory refactoring for moving protobuf translation out of middleware later. With event logging already represented as route-owned policy, a subsequent PR can move parsing and serialization closer to `ProtobufRoute` without needing to preserve an event-log middleware between those operations.

## Behavior

There are no intended changes to HTTP behavior or event-log contents.

The dependency preserves:

- Optional event logging when no plugin is configured.
- Authenticated account information where available.
- Before- and after-event ordering.
- Handler exception logging.
- Existing unary response handling.
- The existing limitation that streaming response event logging is not implemented.

`ProtobufTranslationMiddleware`, authentication, licensing, response serialization, and protobuf route handling are unchanged.


panh99 (approved):
LGTM!

PR #7800, opened 2026-08-10
refactor(framework): Move `SuperLinkRuntimeServicer` functionality into a shared module

jafermarq:
To support both `gRPC` and `FastAPI` implementations


framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
```diff
@@ -0,0 +1,420 @@
...
+        message_ids=message_ids,
+        objects_to_push=objects_to_push,
+        session_id=session_id,
+    )
+
+
+def pull_messages(  # pylint: disable=R0914
+    request: PullAppMessagesRequest,
+    state: LinkState,
+    store: ObjectStore,
```
panh99: Could we remove this argument and access the object store via `state.object_store`, similar to how we handle it in `control_handlers.py`?

panh99 (approved):
LGTM!

PR #7802, opened 2026-08-10
fix(framework): Skip exit-code docs check when docs are unavailable

tanertopal:
## Summary

- Skip the exit-code documentation URL test when the docs tree is unavailable.
- Keep the documentation validation enabled for complete Flower source trees.

## Why

Labs intentionally syncs the framework without `docs/`. The shared test previously failed there and required a Labs-only patch. This makes the test compatible with reduced source trees so Labs can consume the Flower framework without modifying it.

## Validation

- `python -m pytest py/flwr/supercore/exit/exit_code_test.py` — 2 passed
- Ruff check passed


chongshenng (approved):
LGTM!

panh99 (approved):

PR #7804, opened 2026-08-10
refactor(framework): Introduce error codes for `SuperLink` `RuntimeAPI`

jafermarq:
To be used in a subsequent PR to replace `gRPC`'s `context.abort()` calls with `raise FlowerError(...)`


panh99 (approved):
LGTM!

PR #7807, opened 2026-08-10
feat(framework): Introduce `get-task` dependency

jafermarq:
This dependency returns the authenticated `Task` and mirrors the behaviour of `gRPC`'s `get_authenticated_task`


panh99 (approved):
LGTM!

PR #7809, opened 2026-08-10
feat(framework): Introduce `SuperExecAuthDependency`

jafermarq:
The `SuperExecAuthDependency` bring the functionality of `SuperExecAuthServerInterceptor` to `FastAPI`. A dependency (instead of a middleware) is used because only three endpoints require this: `PullPendingTasks`, `ClaimTask` and `GetRun`


framework/py/flwr/supercore/auth/superexec.py
```diff
@@ -78,3 +91,53 @@ def verify_superexec_signature(
...
+
+def verify_superexec_request(  # pylint: disable=too-many-arguments
+    *,
+    request: ProtoMessage,
+    method: str,
+    auth_secret: bytes,
+    timestamp_raw: str | None,
+    nonce: str | None,
+    body_sha256_header: str | None,
+    signature: str | None,
```
panh99: Based on the first line of the function body, it seems that all the four arguments here cannot be `None`, right? Should we remove `| None`?

jafermarq: yes, we can do this change and introduce some validation in the body of `SuperExecAuthDependency.__call__` before calling this function

panh99 (dismissed):
LGTM!

panh99 (approved):

PR #7812, opened 2026-08-11
refactor(framework): Add LinkState declarative models

tanertopal:
### Description

LinkState table metadata is still represented only through SQLAlchemy Core tables. Runtime query migration needs declarative mappings whose physical metadata is proven equivalent to the existing schema first.

### Explanation

Add declarative mappings for the existing `node`, `run`, `context`, `message_ins`, and `message_res` tables without changing runtime behavior or the physical database schema.

Keep `logs` as an unmapped SQLAlchemy Core table because its nullable composite uniqueness constraint is not a safe ORM identity. Metadata parity tests compare columns, defaults, indexes, foreign keys, and unique constraints against the existing LinkState metadata.

No Alembic revision is included because this is a representation-only change.

### Validation

- `python -m pytest py/flwr/supercore/state/schema/linkstate_models_test.py py/flwr/server/superlink/linkstate/linkstate_test.py -q`
- `python -m ruff check py/flwr/supercore/state/schema/linkstate_models.py py/flwr/supercore/state/schema/linkstate_models_test.py`
- `python -m mypy py/flwr/supercore/state/schema/linkstate_models.py py/flwr/supercore/state/schema/linkstate_models_test.py`
- `python -m pylint py/flwr/supercore/state/schema/linkstate_models.py py/flwr/supercore/state/schema/linkstate_models_test.py`
- `./dev/check-migrations.sh` (`No new upgrade operations detected.`)


panh99 (approved):
LGTM!

PR #7817, opened 2026-08-11
refactor(framework): Migrate ObjectStore reads to ORM

smoroso:
## Issue

ObjectStore read paths still use manual SQL even though its declarative models are now available.

## Proposal

Migrate object-tree lookup, object retrieval, membership checks, and object counting to SQLAlchemy ORM/Core expressions while preserving existing behavior and shared-session semantics. The root lookup refreshes the ORM identity map so reads reflect database changes made earlier in the same transaction.

## Validation

- ObjectStore and metadata tests pass.
- Alembic reports no new upgrade operations.
- isort, Black, Ruff, mypy, and `git diff --check` pass.


framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -134,23 +135,24 @@ def preregister(self, run_id: int, object_tree: ObjectTree) -> list[str]:
...
         """Get the object tree for a given object ID."""
-        with self.session():
-            rows = self.query(
-                "SELECT object_id FROM objects WHERE object_id = :object_id",
-                {"object_id": object_id},
+        with self.session() as session:
+            object_exists = session.scalar(
+                select(StoredObject.object_id)
+                .where(StoredObject.object_id == object_id)
+                .execution_options(populate_existing=True)
```
tanertopal: **Simplicity/readability suggestions**
The populate_existing=True options in [sql_object_store.py (line 138)](/private/tmp/flower-pr7817-review-019ff11b/framework/py/flwr/supercore/object_store/sql_object_store.py:138) are redundant for scalar column projections, which always query the database and do not hydrate identity-map entities. Removing them would make the freshness mechanism clearer, but this is non-blocking.

framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -288,12 +290,24 @@ def _lock_objectstore_mutation(self) -> None:
...
-        return len(rows) > 0
+        with self.session() as session:
+            return (
+                session.scalar(
+                    select(StoredObject.object_id).where(
+                        StoredObject.object_id == object_id
+                    )
+                )
+                is not None
+            )
```
panh99:
```suggestion
            stmt = select(StoredObject.object_id).where(
                StoredObject.object_id == object_id
            )
            return session.scalar(stmt) is not None
```

framework/py/flwr/supercore/object_store/sql_object_store.py
```diff
@@ -288,12 +290,24 @@ def _lock_objectstore_mutation(self) -> None:
...
-        return int(rows[0]["cnt"])
+        with self.session() as session:
+            return int(
+                session.scalar(
+                    select(func.count()).select_from(  # pylint: disable=not-callable
+                        StoredObject
+                    )
+                )
+                or 0
+            )
```
panh99:
```suggestion
            # pylint: disable-next=not-callable
            cnt = session.scalar(select(func.count()).select_from(StoredObject))
            return cast(int, cnt)
```

panh99 (dismissed):
LGTM!

tanertopal (approved):

PR #7818, opened 2026-08-11
refactor(framework): Migrate ObjectStore writes to ORM

smoroso:
## Issue

ObjectStore mutation, cleanup, and lock paths still use manual SQL while the ObjectStore declarative models are available.

## Proposal

Migrate ObjectStore registration, storage, deletion, run cleanup, clearing, and mutation-lock operations to SQLAlchemy statements. Preserve conflict-safe registration, atomic `RETURNING` updates/deletes, transaction-scoped locking, recursive cleanup, and existing behavior. The test harness now verifies the lock row and deletion behavior through the public ObjectStore contract.

## Validation

- ObjectStore and metadata tests pass.
- Alembic reports no new upgrade operations.
- isort, Black, Ruff, mypy, and `git diff --check` pass.


panh99 (approved):
LGTM!

PR #7820, opened 2026-08-11
refactor(framework): Amend `get-task` dependency so it is wrappable with `Security`

jafermarq:


framework/py/flwr/superlink/dependencies/task.py
```diff
@@ -26,17 +27,31 @@
...
+    request: Request,
+    token: TaskTokenDependency,
+    state: LinkStateDependency,
+) -> Task:
     """Return the task authenticated by the Runtime task-token header."""
     tokens = request.headers.getlist(TASK_TOKEN_HEADER)
     # Match gRPC metadata validation and reject ambiguous credentials.
-    token = tokens[0] if len(tokens) == 1 and tokens[0] else None
-    task = state.get_task_by_token(token) if token is not None else None
+    token_is_valid = len(tokens) == 1 and bool(token) and tokens[0] == token
```
panh99: Do we still need this if we make the token retrieval as a `TaskTokenDependency`?

panh99 (approved):

PR #7830, opened 2026-08-11
refactor(framework): Define and use `FlowerError` in `SuperCore` `RuntimeAPI` handlers

jafermarq:
This makes the shared handlers transport neutral


panh99 (dismissed):
LGTM!

panh99 (approved):

PR #7831, opened 2026-08-11
refactor(framework): Use ORM for LinkState message paths

tanertopal:
## What this PR does

Migrates LinkState message persistence, validation, loading, and delivery-claim paths from raw SQL strings to SQLAlchemy ORM statements.

The change keeps message claims atomic, preserves deterministic ordering, and retains the existing message serialization shape through a small model-to-dictionary adapter. It also updates the focused tests to verify observable ordering and locking behavior.

## Why `_message_model_to_dict`?

The existing message utilities consume dictionaries. Refactoring all message utilities to consume ORM models or typed domain objects would be cleaner long-term, but is much larger in scope. This small adapter keeps the current serialization and validation interfaces stable while the storage layer moves to ORM.

## Validation

- Full `linkstate_test.py` suite
- Ruff
- mypy

This is the first chunk of the LinkState ORM lifecycle migration.

### DIFF
--- a/framework/py/flwr/server/superlink/linkstate/linkstate_test.py
+++ b/framework/py/flwr/server/superlink/linkstate/linkstate_test.py
@@ -34,6 +34,7 @@ from uuid import uuid4
 
 from google.protobuf.message import DecodeError
 from parameterized import parameterized
+from sqlalchemy import insert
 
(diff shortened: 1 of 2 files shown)


smoroso (approved):

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7835, opened 2026-08-11
ci(framework:skip): Optimize E2E dependency installation caching

tanertopal:
## Summary

- Namespace the E2E Python environment cache by the `frameworks` job and matrix directory.
- Invalidate the cache when the E2E manifest, framework manifest, or framework lockfile changes.
- Save the dependency environment only after all E2E tests pass, reusing the restore primary key because the E2E scripts mutate `pyproject.toml`.
- Keep the manual cache cleanup workflow aligned with the key format and isolate each matrix directory from sibling directory names.
- Reuse the cache restore primary key and explicitly save on cache misses.

## Benchmark

Measured locally with the exact pandas E2E install command:

`python -m pip install --upgrade . --extra-index-url https://download.pytorch.org/whl/cpu`

| Scenario | Install time |
| --- | ---: |
| Fresh virtual environment | 46.88s |
| Repeated install in the populated environment | 4.41s |
| Reduction | 42.47s (90.6%) |

Matrix-level impact:

- The `frameworks` E2E matrix contains 11 independent jobs, each with its own cache namespace.
- The representative warm-cache improvement is 42.47s less dependency-install work per leg.
- If the same reduction applies across all 11 legs, that is 467.17s (7m47s) of aggregate runner-time savings. This is not a wall-clock claim: the matrix runs in parallel, so wall-clock improvement is bounded by the slowest leg.
- A recent main-branch pandas leg spent 39s in dependency installation; the first PR attempt was cold and superseded before completion, so no misleading CI before/after wall-clock comparison is reported here.

The benchmark measures dependency installation only. Cache publication now happens after the E2E tests pass for cache safety, and its time is not included in the install-time comparison. The repeated-install result is a local proxy for the GitHub Actions cache-hit path; actual CI time depends on runner and cache transfer performance.

## Validation

- `uv run --no-sync --python=3.11.14 ./dev/test.sh false` — passed.
- Full pandas E2E matrix leg — passed:
  - legacy edge-client test
  - virtual-client simulation
  - SuperLink/driver test
- Workflow YAML, cache ordering, key reuse, cleanup-prefix isolation, and matrix validation — passed for all 11 framework matrix entries.
- `git diff --check` — passed.

The SuperLink test was run with GNU `sed` to match the Ubuntu GitHub runner; no repository files were changed by the test.


.github/workflows/framework-cache-cleanup.yml
```diff
@@ -39,12 +39,16 @@ jobs:
...
           gh extension install actions/gh-actions-cache
-          
+
           REPO=${{ github.repository }}
-          LATEST_KEY=pythonloc-${{ matrix.directory }}-${{ env.pythonLocation }}-${{ hashFiles(format('./e2e/{0}/pyproject.toml', matrix.directory)) }}
+          # The producer job is named `frameworks`; keep this namespace aligned
+          # with github.job in framework-e2e.yml.
+          CACHE_JOB_NAME=frameworks
+          CACHE_PREFIX="pythonloc-${{ runner.os }}-${CACHE_JOB_NAME}-${{ matrix.directory }}-${{ env.pythonLocation }}-"
+          LATEST_KEY="${CACHE_PREFIX}${{ env.pythonLocation }}-${{ hashFiles(format('./framework/e2e/{0}/pyproject.toml', matrix.directory), 'framework/pyproject.toml', 'framework/uv.lock') }}"
```
chongshenng, line 48: `pythonLocation` is included twice in `LATEST_KEY`. Is this intended? The producer key includes it only once. If someone manually runs the cleanup workflow, it will consider the current cache obsolete and delete it.

.github/workflows/framework-e2e.yml
```diff
@@ -218,7 +218,10 @@ jobs:
...
         with:
           path: ${{ env.pythonLocation }}
-          key: pythonloc-${{ runner.os }}-${{ matrix.directory }}-${{ env.pythonLocation }}-${{ hashFiles(format('./framework/e2e/{0}/pyproject.toml', matrix.directory)) }}
+          # Cache entries are shared across workflow jobs. Namespace this one
+          # by job and matrix directory so another E2E job cannot restore a
+          # different Python environment by accident.
+          key: pythonloc-${{ runner.os }}-${{ github.job }}-${{ matrix.directory }}-${{ env.pythonLocation }}-${{ hashFiles(format('./framework/e2e/{0}/pyproject.toml', matrix.directory), 'framework/pyproject.toml', 'framework/uv.lock') }}
       - name: Install dependencies
         if: ${{ needs.changes.outputs.framework == 'true' }}
         run: python -m pip install --upgrade . --extra-index-url https://download.pytorch.org/whl/cpu
```
chongshenng, line 227: This step still runs unconditionally even after a cache hit. So, the comment in the PR where it reports time saving from a `python -m pip install --upgrade . --extra-index-url https://download.pytorch.org/whl/cpu` doesn't seem correct. The changes in this PR would cache the Python environment but not bypass this installation command.

PR #7836, opened 2026-08-11
feat(framework): Add interactive conversation history to flwr chat CLI

danielnugraha:
## Summary

Adds interactive conversation history to flwr chat.

  - /history lists conversations from the default flower-agent-execution federation.
  - Up/Down navigates, Enter continues, and Esc/Ctrl-C cancels.
  - History is chronological, auto-scrolls with selection, and uses compact single-line rows.
  - /federation lists the federation the user belongs to.
Screenshot 2026-08-14 at 10 51 03
Screenshot 2026-08-14 at 10 51 17


framework/py/flwr/cli/chat_app.py
```diff
@@ -576,6 +747,21 @@ def parse_task_event(task_event: TaskEvent) -> tuple[str, JSONObject]:
...
+    """Resolve the account-scoped default federation used by Flower Chat."""
+    execution_suffix = f"/{CHAT_DEFAULT_FEDERATION_NAME}"
+    for federation in federations:
+        if federation.name.endswith(execution_suffix):
+            return federation.name
+
+    workspace_suffix = f"/{DEFAULT_FEDERATION_SIMULATION}"
+    for federation in federations:
+        if federation.name.endswith(workspace_suffix):
+            account_name = federation.name[: -len(workspace_suffix)]
```
panh99: Can we use simulation federation for agents?

framework/py/flwr/cli/chat_app.py
```diff
@@ -576,6 +747,21 @@ def parse_task_event(task_event: TaskEvent) -> tuple[str, JSONObject]:
     return event_type, payload
 
 
+def _resolve_default_chat_federation(federations: list[Federation]) -> str | None:
```
panh99: QQ: since SuperGrid now automatically creates default federations for each account, why we don't just return the default agent federation?

framework/py/flwr/cli/chat_app.py
```diff
@@ -306,10 +347,139 @@ def _handle_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
+        if self.federation is None:
+            self._append_transcript(
+                "class:error", "The default chat federation is unavailable.\n\n"
+            )
+            return
+        try:
+            with flwr_cli_grpc_exc_handler():
+                response = self.stub.ListRunSeries(
+                    ListRunSeriesRequest(federation_id=self.federation)
+                )
```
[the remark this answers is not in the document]
panh99: This seems relevant. Could you check whether the UI might get stuck if, for example, `ListRunSeries` takes a few seconds or longer? By "stuck", I mean whether the whole UI is not responding, ignoring keyboard events etc.

framework/py/flwr/cli/chat_app.py
```diff
@@ -306,10 +346,99 @@ def _handle_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
+        """Show conversation history for the default chat federation."""
+        if self.federation is None:
+            self._append_transcript(
+                "class:error", "The default chat federation is unavailable.\n\n"
+            )
+            return
+        try:
+            with flwr_cli_grpc_exc_handler():
+                response = self.stub.ListRunSeries(
+                    ListRunSeriesRequest(federation_id=self.federation)
```
panh99: Should we set a limit here, say 50? I think `ListRunSeriesRequest` supports `limit`. Or you think that's acceptable for now?

framework/py/flwr/cli/chat_app.py
```diff
@@ -131,14 +145,19 @@ def _get_menu_width(self, max_width: int, _complete_state: CompletionState) -> i
...
 
-    def __init__(self, stub: ControlStub, federation: str | None) -> None:
+    def __init__(
+        self,
+        stub: ControlStub,
+        federations: list[Federation],
+    ) -> None:
         self.stub = stub
-        self.federation = federation
+        self.federation = _resolve_default_chat_federation(federations)
```
panh99: What’s the purpose of accepting a list of federations? Is this mainly for a future use case where we might allow users to switch between federations?

It doesn’t feel necessary at the moment, since the code here still only uses a single federation. Just curious, could we hardcode `flower-agent-execution` for now, or users can now specify which federation they’d like to use via CLI?

framework/py/flwr/cli/chat_app.py
```diff
@@ -306,10 +346,99 @@ def _handle_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
+            self._show_history()
+            return True
+        if command == CHAT_FEDERATION_COMMAND or command.startswith(
+            f"{CHAT_FEDERATION_COMMAND} "
+        ):
+            self._append_transcript(
+                "class:notice",
+                f"{CHAT_FEDERATION_COMMAND} is currently disabled.\n\n",
+            )
+            return True
```
panh99: Do we ever support this command? This code looks a bit sus.

framework/py/flwr/cli/chat_app_test.py
```diff
@@ -0,0 +1,53 @@
...
+    history = list(completer.get_completions(Document("/hist"), CompleteEvent()))
+    federation = list(
+        completer.get_completions(Document("/federation"), CompleteEvent())
+    )
+
+    assert [completion.text for completion in history] == ["/history"]
+    assert federation == []
+
+
+def test_chat_always_uses_default_federation() -> None:
```
panh99: Is this testing an impossible case? IIUC, every account on SuperGrid will have the default `flower-agent-execution` federation, given that we call `FederationManager.ensure_default_federation` in Control APIs.

framework/py/flwr/cli/chat_app.py
```diff
@@ -102,13 +118,11 @@ def get_completions(
...
-        if (
-            document.text_after_cursor
-            or not text.startswith("/")
-            or any(char.isspace() for char in text)
-        ):
+        if document.text_after_cursor or not text.startswith("/"):
             return
 
+        if any(char.isspace() for char in text):
+            return
```
panh99: Do we need this change? Seems equivalent.

framework/py/flwr/cli/chat_app.py
danielnugraha: Yes we can.

framework/py/flwr/cli/chat_app.py
danielnugraha: The thinking is once we have federation agents merged, we can enable /federation command and remove the hardcoded default federation.

framework/py/flwr/cli/chat_app.py
danielnugraha: Since we need to scope the conversation to the federation and federation agent is coming soon after 2 PRs are merged, I implemented this PR with having /federation in mind, but hardcoded to use flower-agent-execution for now. Once we have those 2 PRs merged, we have the basic foundation to allow running in different federations, we just need to lift the hardcode and not reimplement everything from scratch.

framework/py/flwr/cli/chat_app.py
danielnugraha: We don't have a limit in the webui for now. I'd like to keep the behavior uniform between flwr chat and webui if possible.

framework/py/flwr/cli/chat/chat_history.py
panh99, line 1: Since we have a `flwr.cli.chat` module now, should we rename this file to `command_history` or similar? It might be cleaner to put the helper functions for one command in one file. Plus the module docstring is `"""History command helpers for Flower Chat."""`, which conflicts with the file name `chat_commands.py`.

framework/py/flwr/cli/chat/chat_commands.py
```diff
@@ -0,0 +1,156 @@
...
+
+    federation: str
+    entries: list[RunSeries]
+    selected_index: int = 0
+
+
+def load_history(stub: ControlStub, federation: str) -> HistoryBlock | None:
+    """Load chronological conversation history for a federation."""
+    with flwr_cli_grpc_exc_handler():
+        response = stub.ListRunSeries(ListRunSeriesRequest(federation_id=federation))
```
panh99: Should we filter by agent run series in case it includes FL runs?

```suggestion
        response = stub.ListRunSeries(ListRunSeriesRequest(federation_id=federation, is_agent=true))
```

framework/py/flwr/cli/chat/chat_history.py
```diff
@@ -0,0 +1,156 @@
...
+    federation: str
+    entries: list[RunSeries]
+    selected_index: int = 0
+
+
+def load_history(stub: ControlStub, federation: str) -> HistoryBlock | None:
+    """Load chronological conversation history for a federation."""
+    with flwr_cli_grpc_exc_handler():
+        response = stub.ListRunSeries(ListRunSeriesRequest(federation_id=federation))
+    entries = list(reversed(response.entries))
```
panh99, line 53:
```suggestion
    # Reverse entries to list them from newest to oldest
    entries = list(reversed(response.entries))
```

framework/py/flwr/cli/chat/chat_app.py
```diff
@@ -472,14 +500,73 @@ def _handle_federation_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
+        if self.history_block is None:
+            return
+        self.history_block.selected_index = (
+            self.history_block.selected_index + offset
+        ) % len(self.history_block.entries)
+        self.transcript_revision += 1
+        self.application.invalidate()
+
+    def _confirm_history_selection(self) -> None:
+        """Restore and continue the highlighted conversation."""
```
panh99: It looks like we will only restore the conversation, but not the agent app that was used in the conversation. This makes sense to me, but I wonder if we should mention it in the docstring or as a comment.

framework/py/flwr/cli/chat/chat_commands.py
panh99: I was learnt that the synchronous gRPC call might block prompt_toolkit UI, so it cannot handle keyboard/redraw, etc. Is that true?

framework/py/flwr/cli/chat/chat_history.py
```diff
@@ -0,0 +1,156 @@
...
+def load_conversation(
+    stub: ControlStub, entry: RunSeries, federation: str
+) -> list[tuple[str, str]]:
+    """Load displayable messages for one conversation."""
+    with flwr_cli_grpc_exc_handler():
+        response = stub.GetRunSeries(GetRunSeriesRequest(series_id=entry.series_id))
+    if response.series.federation != federation:
+        raise click.ClickException(
+            f"Conversation {entry.series_id} does not belong to {federation}."
+        )
```
panh99, line 68: Just curious, is `ClickException` compatible with `prompt_toolkit`?

framework/py/flwr/cli/chat/chat_history.py
dstripelis: Makes sense! Renamed!

framework/py/flwr/cli/chat/chat_commands.py
dstripelis: Yes good catch! Fixed.

framework/py/flwr/cli/chat/chat_app.py
dstripelis: Yes we should. Updated docstring as follows:
```"""Restore the conversation without changing the selected agent."""```

File: framework/py/flwr/cli/chat/chat_commands.py (hunk 9 above)
Comment (dstripelis) [reply framework/py/flwr/cli/chat/chat_commands.py]: Good catch. These synchronous RPCs currently run from prompt-toolkit key handlers and may block the UI event loop. We could offload both `ListRunSeries` and `GetRunSeries` using `asyncio.to_thread()` through prompt-toolkit background tasks, consistent with the existing chat-run implementation. 

wdyt?

Comment (panh99) [reply framework/py/flwr/cli/chat/chat_commands.py]: Yeah, moving to a thread makes sense to me!

File: framework/py/flwr/cli/chat/chat_history.py (hunk 12 above)
Comment (dstripelis) [reply framework/py/flwr/cli/chat/chat_history.py:68]: `ClickException` is used here as a user-facing error container; `prompt-toolkit` does not handle it directly. Related to the comment above, if we apply the non-blocking RPC fix above, then we only need to ensure the `try/except click.ClickException` wraps the `await asyncio.to_thread(...)` call.

File: framework/py/flwr/cli/chat/chat_app.py [hunk 13]
```diff
@@ -472,14 +500,75 @@ def _handle_federation_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
         self.agent_name = CHAT_AGENT_NAME
         self.series_id = None
-        self.transcript.clear()
+        self._clear_transcript()
+        return True
+
+    async def _show_history(self) -> None:
+        """Show conversation history for the active federation."""
+        try:
+            block = await asyncio.to_thread(load_history, self.stub, self.federation)
```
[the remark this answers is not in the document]
dstripelis: resolved!

framework/py/flwr/cli/chat/chat_app.py
```diff
@@ -472,14 +500,75 @@ def _handle_federation_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
+
+    async def _confirm_history_selection(self) -> None:
+        """Restore the conversation without changing the selected agent."""
+        if self.history_block is None:
+            return
+        entry = self.history_block.entries[self.history_block.selected_index]
+        try:
+            messages = await asyncio.to_thread(
+                load_conversation, self.stub, entry, self.federation
+            )
```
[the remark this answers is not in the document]
dstripelis: Suggestion applied!

framework/py/flwr/cli/chat/chat_app.py
```diff
@@ -472,14 +505,80 @@ def _handle_federation_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
+        self.transcript_revision += 1
+        self.application.invalidate()
+
+    def _move_history_selection(self, offset: int) -> None:
+        """Move the highlighted conversation history row."""
+        if self.history_block is None:
+            return
+        self.history_block.selected_index = (
+            self.history_block.selected_index + offset
+        ) % len(self.history_block.entries)
```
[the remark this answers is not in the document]
dstripelis: Applied change so that history navigation now ignores Up/Down while a conversation is loading.

framework/py/flwr/cli/chat/chat_app.py
panh99, line 168: (Not a blocker) QQ: would it be simpler to extend this method to handle history conversation selection, similar to `/federation` command?

dstripelis: We can revisit this in a follow up PR, this is not mandatory for now.

panh99 (approved):
LGTM!

PR #7844, opened 2026-08-12
refactor(framework): Use models as LinkState metadata source

smoroso:
### Description

LinkState metadata is still duplicated between declarative models and the legacy table wrapper.

### Explanation

Use the declarative LinkState model registry when building compatibility metadata. This removes the duplicate table definitions without changing the physical schema or runtime behavior.

Validation: LinkState/schema tests, Alembic metadata no-diff test, Ruff, Black, mypy, Pylint, and diff checks pass.


tanertopal (commented):
Lgtm!

tanertopal (approved):

PR #7849, opened 2026-08-12
feat(framework): Introduce `SuperExecAuthHttpInterceptor` and `RuntimeVersionHttpInterceptor` for HTTP client

jafermarq:
Refactors slightly the existing gRPC interceptors to avoid duplicated code.


framework/py/flwr/supercore/interceptors/http/runtime_version.py
panh99, line 1: Should we still put each interceptor in a separate file? Putting them all in a file named `http.py` seems a bit unusual.

jafermarq: fair point; the best way seems to move them inside a `http/` submodule to avoid collision with existing (grpc) ones name-wise.

panh99 (approved):
LGTM!

PR #7854, opened 2026-08-13
feat(framework): Introduce `RuntimeHttpStub`

jafermarq:
Introduced two `RuntimeHttpStub`: one under `supercore/` for shared functionality that contains the methods needed for `SuperExec <--> SuperLink` communication; and one under `superlink/` which contains the methods used from `flwr-serverapp <--> SuperLink`.


framework/py/flwr/supercore/runtime/runtime_http_stub_test.py
panh99, line 1: Could we simplify the test here? Probably just hardcode a list of unary-unary RPC names, then have a single unit test iterate over them and check that each RPC calls the expected function/method, passing expected arguments like the correct response type. For the latter, we could use reflection, or simply check that `response_type.__name__ == f"{rpc_name}Response"`.

framework/py/flwr/superlink/runtime/runtime_http_stub_test.py
panh99, line 1: Same as above.

panh99 (approved):
LGTM!

PR #7864, opened 2026-08-13
refactor(framework): Move `RuntimeAPI` functionality in `SuperNode` to a shared module

jafermarq:
This is so it can be temporarily be used by both `gRPC` and its ongoing HTTP counterpart


framework/py/flwr/supernode/servicer/runtime/runtime_handlers.py
panh99, line 1: Could we only pass `NodeState` and access object store via `state.object_store`?

jafermarq: ah, yes!

panh99 (approved):
LGTM!

PR #7867, opened 2026-08-13
refactor(framework): Move shared dependencies to `supercore/`

jafermarq:
Move `Runtime` authentication logic from `SuperLink` into `SuperCore` so it can later be reused by SuperNode


panh99 (approved):
LGTM!

PR #7871, opened 2026-08-14
feat(framework): Add federation agent table

danielnugraha:
## Summary

  Add a federation_agent table to persist agents associated with federations.

  - Store agent, federation, app, FAB hash, creator, and creation timestamp.
  - Require fab_hash.
  - Enforce one association per federation and app.
  - Add an index for efficient federation agent listing.


framework/py/flwr/supercore/state/alembic/versions/rev_2026_08_14_add_federation_app_table.py
```diff
@@ -0,0 +1,65 @@
...
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "federation_app",
+        sa.Column("federation_id", sa.String(), nullable=False),
+        sa.Column("app_id", sa.String(), nullable=False),
+        sa.Column("fab_hash", sa.String(), nullable=False),
+        sa.Column("app_type", sa.String(), nullable=False),
+        sa.Column("created_by", sa.String(), nullable=False),
+        sa.Column("created_at", sa.TIMESTAMP(timezone=True), nullable=False),
```
panh99: Should we use `UTCDateTime` for better dialect support?

panh99: Should we name it as `added_at/by` since the app itself is not created here? They should be created via `flwr new`.

panh99 (approved):
LGTM!

PR #7874, opened 2026-08-14
feat(framework): Add federation agent corestate methods

danielnugraha:
## Summary

  Add CoreState and LinkState methods for managing federation apps.

  - Add upsert_app, list_apps, and delete_app.
  - Implement in-memory and SQL backends.
  - Support federation filtering, limits, and newest-first ordering.
  - Add shared state conformance tests.


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -117,6 +121,27 @@ def store_fab(self, fab: Fab) -> str:
...
+    @abstractmethod
+    def upsert_app(  # pylint: disable=too-many-arguments,too-many-positional-arguments
+        self,
+        federation_id: str,
+        app_id: str,
+        fab_hash: str,
+        app_type: str,
+        created_by: str,
+    ) -> bool:
+        """Create or update an app associated with a federation."""
```
panh99: Should we merge `store_fab` and `upsert_app` to make it easier to ensure atomicity here? Currently, `Control.StartRun` is the only caller of `store_fab`, and it already has access to information such as the federation and user account.

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -117,6 +121,27 @@ def store_fab(self, fab: Fab) -> str:
...
         """Return the FAB for the given hash, if present."""
 
+    @abstractmethod
+    def upsert_app(  # pylint: disable=too-many-arguments,too-many-positional-arguments
+        self,
+        federation_id: str,
+        app_id: str,
+        fab_hash: str,
+        app_type: str,
+        created_by: str,
```
panh99: Should we set this to the `now()` automatically in the method instead of accepting an external timestamp?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -117,6 +121,27 @@ def store_fab(self, fab: Fab) -> str:
...
+
+    @abstractmethod
+    def list_apps(
+        self, federation_id: str, limit: int | None = None
+    ) -> Sequence[AppInfo]:
+        """List apps associated with a federation, newest first."""
+
+    @abstractmethod
+    def delete_app(self, federation_id: str, app_id: str) -> bool:
+        """Delete an app association from a federation."""
```
panh99: Should we add a sentence saying this only removes the bookkeeping, while the fab remains in the state?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -113,10 +117,31 @@ def _on_push_session_expired(self, message_object_ids: set[str]) -> None:
...
+    @abstractmethod
+    def store_app(  # pylint: disable=too-many-arguments,too-many-positional-arguments
+        self,
+        fab: Fab,
+        federation_id: str,
+        app_id: str,
+        app_type: str,
+        added_by: str,
+    ) -> str:
+        """Atomically store a FAB and associate its app with a federation."""
```
panh99: Could we add `Parameters` section and `Returns` section to the docstring?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -113,10 +117,31 @@ def _on_push_session_expired(self, message_object_ids: set[str]) -> None:
...
 
+    @abstractmethod
+    def list_apps(
+        self, federation_id: str, limit: int | None = None
+    ) -> Sequence[AppInfo]:
+        """List apps associated with a federation, newest first."""
+
+    @abstractmethod
+    def delete_app(self, federation_id: str, app_id: str) -> bool:
+        """Delete an app association; the referenced FAB remains in state."""
```
panh99: Just curious, do we need this method now?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -113,10 +117,31 @@ def _on_push_session_expired(self, message_object_ids: set[str]) -> None:
...
         """Return the FAB for the given hash, if present."""
 
+    @abstractmethod
+    def list_apps(
+        self, federation_id: str, limit: int | None = None
+    ) -> Sequence[AppInfo]:
+        """List apps associated with a federation, newest first."""
+
+    @abstractmethod
+    def delete_app(self, federation_id: str, app_id: str) -> bool:
```
panh99, line 165: Will all apps of the same ID be deleted? The app ID doesn't seem to be the unique identifier, unlike fab hash.

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -409,6 +416,61 @@ def store_fab(self, fab: Fab) -> str:
...
+    ) -> str:
+        """Atomically store a FAB and associate its app with a federation."""
+        if not all((federation_id, app_id, app_type, added_by)):
+            raise ValueError(
+                "Federation ID, app ID, app type, and added by are required"
+            )
+        fab_hash = hashlib.sha256(fab.content).hexdigest()
+        if fab.hash_str and fab.hash_str != fab_hash:
+            raise ValueError(
+                f"FAB hash mismatch: provided {fab.hash_str}, computed {fab_hash}"
```
panh99, line 435: Should we define an raise a FlowerError?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -409,6 +416,61 @@ def store_fab(self, fab: Fab) -> str:
...
+        fab: Fab,
+        federation_id: str,
+        app_id: str,
+        app_type: str,
+        added_by: str,
+    ) -> str:
+        """Atomically store a FAB and associate its app with a federation."""
+        if not all((federation_id, app_id, app_type, added_by)):
+            raise ValueError(
+                "Federation ID, app ID, app type, and added by are required"
```
panh99, line 430: Should we raise a FlowerError with a new error code here?

framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -423,6 +485,52 @@ def get_fab(self, fab_hash: str) -> Fab | None:
...
+            .order_by(
+                FederationAppModel.added_at.desc(),
+                FederationAppModel.app_id.desc(),
+            )
+        )
+        if limit is not None:
+            query = query.limit(limit)
+        with self.session() as session:
+            apps = session.scalars(
+                query.execution_options(populate_existing=True)
```
panh99: It seems that this is not very relevant. Wdyt?
```suggestion
                query
```

framework/py/flwr/supercore/corestate/corestate.py
danielnugraha: We just save the latest `fab_hash` in this table. So the app id would be a unique identifier.

framework/py/flwr/supercore/corestate/corestate.py
danielnugraha: Yes I will add a separate PR for this.

framework/py/flwr/supercore/corestate/corestate.py
danielnugraha: Added.

framework/py/flwr/supercore/corestate/sql_corestate.py
danielnugraha: Should we throw FlowerError in corestate? I don't see it elsewhere. It will be thrown as FlowerError in StartRun anyway.

framework/py/flwr/supercore/corestate/corestate.py
panh99: I see, thanks for clarifying.

framework/py/flwr/supercore/corestate/sql_corestate.py
panh99: Yes, we can put it in `CoreState`. It’s more commonly used in the federation manager. That said, I don’t have a strong opinion. It might be simpler to raise `FlowerError` here; otherwise, we’ll still need a `try-except` at the call site to convert it into a `FlowerError`.

But I’m also fine with keeping it as is, since this is an extremely rare case and doesn’t pose any security risk.

panh99 (dismissed):
LGTM!

panh99 (approved):
LGTM!

PR #7875, opened 2026-08-14
feat(framework): Add federation agent endpoint implementation

danielnugraha:
## Summary

Implemented the ListApps CoreState wiring:

  - Persists apps when a run starts, preserving AgentApp vs ServerApp.
  - Implements the shared list_apps handler with federation membership validation and optional limit.
  - Wires both gRPC and HTTP endpoints.
  - Registers HTTP protobuf translation.
  - Adds handler, servicer, and router tests.


framework/py/flwr/superlink/routers/control/router_test.py
panh99, line 1: Should we remove the test here? Since the `router.py` is only a thin wrapper around `control_handlers.py`, the new test added in `control_handlers_test.py` is sufficient.

framework/py/flwr/superlink/servicer/control/control_servicer_test.py
panh99, line 1: Same as above.

panh99 (approved):
LGTM!

PR #7884, opened 2026-08-14
feat(framework): Add RDP privacy accounting core

mohammadnaseri:
This PR adds the RDP privacy accounting foundation for computing and composing ε and δ across private releases.
Please note that strategy integration will follow in a separate PR.


framework/py/flwr/supercore/rdp_accountant.py
```diff
@@ -0,0 +1,338 @@
...
+            )
+        projected.compose(backend_event, count)
+        return bool(projected.get_epsilon(self._config.target_delta) > epsilon_limit)
+
+    def state_dict(self) -> dict[str, object]:
+        """Return JSON-serializable accountant state."""
+        return {
+            "schema_version": _STATE_SCHEMA_VERSION,
+            "backend": _ACCOUNTING_METHOD,
+            "backend_version": metadata.version("dp-accounting"),
```
msheller: It isn't clear to me how this is used. I don't see any kind of check during the load. Is this just for logging/provenance purporses?

framework/py/flwr/supercore/rdp_accountant.py
```diff
@@ -0,0 +1,338 @@
...
+                event.sample_size / event.population_size,
+                gaussian_event,
+            )
+        if self._config.sampling_method is SamplingMethod.WITHOUT_REPLACEMENT:
+            return dp_event.SampledWithoutReplacementDpEvent(
+                event.population_size,
+                event.sample_size,
+                gaussian_event,
+            )
+        return gaussian_event
```
msheller: Should we make this "fail-closed" in case a new method is added but not implemented? E.g.

elif self._config.sampling_method is SamplingMethod.NO_AMPLIFICATION:
    return gaussian_event

raise ValueError(f"Unsupported sampling method: {self._config.sampling_method}")

framework/py/flwr/supercore/rdp_accountant.py
```diff
@@ -0,0 +1,338 @@
...
+                    event_value["population_size"], "population_size"
+                ),
+            )
+            count = _as_int(composition["count"], "count")
+            _validate_count(count)
+            compositions.append((event, count))
+    except (KeyError, TypeError, ValueError) as exc:
+        raise ValueError("Invalid composition in accountant state.") from exc
+    return compositions
+
```
msheller, line 302: Minor nit: these are checkers, not converters, so I might change "_as_<type>" to "_require_<type>" or similar

framework/pyproject.toml
```diff
@@ -43,6 +43,7 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=1.26.0,<3.0.0",
+    "dp-accounting>=0.6.0,<0.7.0",
```
msheller: This is probably my only meaningful concern. dp-accounting also brings in scipy, so this isn't a totally trivial dependency.

msheller (commented):
I left a few minor comments to consider. The only comment I think needs a clean resolution is the addition of scipy to the dependencies. If Pan/Daniel are okay with that addition, then I approve the PR whether you address the other minor comments or not.

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7886, opened 2026-08-14
feat(framework): Implement remaining endpoints for `RuntimeAPI` in `SuperNode`

jafermarq:


framework/py/flwr/supernode/routers/runtime/router.py
```diff
@@ -14,16 +14,284 @@
...
+    Depends(SuperExecAuthDependency("/flwr.proto.Runtime/PullPendingTasks")),
+]
+ClaimTaskAuthDependency = Annotated[
+    None,
+    Depends(SuperExecAuthDependency("/flwr.proto.Runtime/ClaimTask")),
+]
+GetRunAuthDependency = Annotated[
+    None,
+    Depends(SuperExecAuthDependency("/flwr.proto.Runtime/GetRun")),
+]
```
panh99, line 97: Could we make `SuperExecAuthDependency` take a list of methods? This seems repetitive.

framework/py/flwr/supernode/routers/runtime/router.py
```diff
@@ -14,16 +14,284 @@
...
+) -> ClaimTaskResponse:
+    """Claim a pending task."""
+    return core_runtime_handlers.claim_task(request, state)
+
+
+@router.post("/get-run")
+def get_run(
+    request: Annotated[GetRunRequest, PROTOBUF_REQUEST],
+    state: NodeStateDependency,
+    _auth: GetRunAuthDependency,
```
panh99, line 124: Should we move the auth dependency to the decorator function since we don't need its return? Or you'd like to keep it as an argument for consistency?

framework/py/flwr/supernode/routers/runtime/router.py
```diff
@@ -14,16 +14,284 @@
...
 
-@router.post("/messages")
-def pull_messages() -> Response:
-    """Pull messages for the ClientApp."""
-    raise HTTPException(
-        status_code=status.HTTP_501_NOT_IMPLEMENTED,
-        detail="Not implemented",
-    )
+NodeStateDependency = Annotated[NodeState, Depends(get_nodestate)]
+PROTOBUF_REQUEST = Depends(get_protobuf_request)
```
panh99, line 84: Just curious, would it make sense to create an constant like this in `flwr.supercore.protobuf.translation`?

framework/py/flwr/supernode/routers/runtime/router.py
jafermarq: let's leave this for a refactoring later on

framework/py/flwr/supernode/routers/runtime/router.py
jafermarq: done in next PR: https://github.com/flwrlabs/flower/pull/7894

framework/py/flwr/supernode/routers/runtime/router.py
jafermarq: after considering alternatives, it seemed this design is the simplest and most explicit, at the cost of some repetition.

panh99 (approved):
LGTM!

PR #7890, opened 2026-08-16
refactor(framework): Use JSON for all HTTP API errors

danieljanes:
## What does this PR do?

This PR establishes one consistent JSON error contract across Flower’s HTTP APIs, including the Control and Runtime protobuf-over-HTTP APIs.

Successful Control and Runtime requests continue to send and receive serialized protobuf messages. Their HTTP failures are returned as JSON, consistent with ordinary FastAPI endpoints.

### Canonical HTTP error contract

Every HTTP error contains `detail`:

```json
{
  "detail": "A client-visible error message."
}
```

Mapped `FlowerError`s additionally contain a numeric Flower error code:

```json
{
  "detail": "Invalid automation request.",
  "code": 42
}
```

When a `FlowerError` contains client-safe public details, they are returned separately as `extra`:

```json
{
  "detail": "Invalid automation request.",
  "code": 42,
  "extra": "Invalid automation schedule."
}
```

Subtype-specific public fields are preserved. For example, `EntitlementError` includes its application-level entitlement code:

```json
{
  "detail": "Entitlement error.",
  "code": 13,
  "extra": "The current plan does not allow this action.",
  "entitlement_code": 101
}
```

For mapped `FlowerError`s:

- `detail` is the catalog-defined public message.
- `code` is the numeric `ApiErrorCode`.
- `extra` is included when `public_details` is not `None`.
- Subtypes may add public, machine-readable fields such as `entitlement_code`.
- Internal diagnostic messages are logged but never returned.

Ordinary `HTTPException`, FastAPI validation, authentication, and unexpected-error responses retain their existing FastAPI `detail` contract.

### Make the HTTP representation canonical

- Change `FlowerError.to_json()` to return the canonical HTTP error dictionary.
- Use this representation directly in the shared HTTP error translator.
- Preserve catalog-defined HTTP statuses and headers, including `WWW-Authenticate: Bearer`.
- Return a sanitized detail-only `500` response for unmapped error codes.
- Preserve `EntitlementError.entitlement_code` in HTTP responses.

### Preserve production Control gRPC compatibility

- Move legacy gRPC serialization into `flwr.supercore.error.grpc`.
- Continue sending the established gRPC error payload:

  ```json
  {
    "code": 42,
    "public_message": "Invalid automation request.",
    "public_details": "Invalid automation schedule."
  }
  ```

- Continue including `entitlement_code` for gRPC `EntitlementError`s.
- Keep existing gRPC statuses and client-visible behavior unchanged.
- Allow `FlowerError.from_json()` to parse both the canonical HTTP representation and the legacy gRPC representation.

### Preserve rendered JSON errors on protobuf routes

- Let completed JSON error responses bypass successful-response protobuf serialization.
- Preserve FastAPI-rendered `HTTPException` and validation responses instead of replacing them with `INVALID_PROTOBUF_RESPONSE`.
- Only pass through responses with a JSON media type.
- Reject non-JSON error responses from protobuf routes as invalid protobuf responses.
- Keep error responses JSON even when the request specifies `Accept: application/protobuf`.

### Cover Control and Runtime HTTP APIs

The shared behavior applies to both `/v1/control` and `/v1/runtime`:

- Successful responses remain serialized protobuf.
- Mapped `FlowerError`s use `detail`, `code`, optional `extra`, and any public subtype fields.
- Ordinary FastAPI errors remain detail-only JSON.
- Invalid protobuf requests and responses use the same JSON error contract.
- Runtime-version incompatibility responses retain their specific compatibility details in `extra`.

## Compatibility

- Production Control gRPC serialization remains backward-compatible.
- Successful Control and Runtime HTTP responses remain protobuf.
- Existing ordinary FastAPI `HTTPException` response bodies, statuses, and headers remain unchanged.
- Unmapped and unexpected exceptions remain sanitized.
- New clients can parse both legacy gRPC and canonical HTTP Flower error representations.

This intentionally changes the HTTP representation of mapped `FlowerError`s, including Runtime HTTP responses.


framework/py/flwr/supercore/error/base.py
```diff
@@ -87,8 +88,12 @@ def from_json(value: str | None) -> FlowerError | None:
...
 
         code = payload.get("code")
-        public_message = payload.get("public_message")
-        public_details = payload.get("public_details")
+        if "public_message" in payload:
+            public_message = payload.get("public_message")
+            public_details = payload.get("public_details")
+        else:
+            public_message = payload.get("detail")
+            public_details = payload.get("extra")
```
jafermarq, line 96: if we were to change the `to_json` method to make use of `detail/extra` instead, could we then remove the `if` part of the statement?

framework/py/flwr/supercore/protobuf/translation.py
```diff
@@ -164,6 +164,18 @@ async def dispatch(
...
+            content_type = response.headers.get("content-type", "")
+            media_type = content_type.partition(";")[0].strip().lower()
+            if media_type == "application/json" or media_type.endswith("+json"):
+                return response
+            raise FlowerError(
+                ApiErrorCode.INVALID_PROTOBUF_RESPONSE,
+                "Protobuf route returned a non-JSON HTTP error response with "
+                f"Content-Type {content_type!r}.",
+            )
+
```
panh99, line 178: Why is this change? The JSON format error for HTTP doesn't seem to affect translation.

framework/py/flwr/supercore/error/http.py
```diff
@@ -32,6 +32,19 @@
 NOT_AUTHENTICATED_DETAIL = "Not authenticated"
 
 
+def _flower_error_content(
+    err: FlowerError, public_message: str
+) -> dict[str, str | int]:
```
panh99: Not a fan of this glue function in this module. Could we instead make `FlowerError.to_json()` return the HTTP JSON error format, and move the compatibility layer into `flwr.supercore.error.grpc` to convert that HTTP JSON representation into the existing gRPC format?

Since we're moving toward HTTP, I'd prefer to keep backward-compatibility and glue code in the gRPC-specific modules that will eventually be removed, rather than introducing it into the new HTTP code.

framework/py/flwr/supercore/protobuf/translation.py
danieljanes: FastAPI renders HTTPException and validation failures into JSON responses before this middleware regains control. Those paths do not set request.state.protobuf_response, so the previous logic replaced the original 409/422 with INVALID_PROTOBUF_RESPONSE. This check preserves completed JSON errors while still rejecting non-JSON error responses and missing successful protobuf responses. The new router tests cover both sides.

framework/py/flwr/supercore/error/http.py
danieljanes: I agree that the legacy gRPC representation should eventually live in the gRPC module. However, changing FlowerError.to_json() here would expand this PR into a migration of production gRPC serialization, EntitlementError, CLI parsing, and the shared runtime-version parser. The HTTP helper is intentionally the transport-boundary adapter, while the existing to_json() behavior remains untouched for gRPC compatibility. I’d prefer a follow-up that moves serialization out of FlowerError altogether and gives HTTP and gRPC explicit transport-specific codecs.

framework/py/flwr/supercore/error/base.py
danieljanes: Not just by changing to_json(). from_json() also parses the existing production gRPC payload, which still uses public_message and public_details. The runtime-version compatibility code is shared by HTTP and gRPC, so this branch intentionally accepts both wire formats. We could remove it after introducing transport-specific decoders and migrating the gRPC callers, but I’d keep that larger refactor outside this compatibility-focused PR.

panh99 (dismissed):
LGTM!

panh99 (approved):
LGTM!

PR #7911, opened 2026-08-18
feat(framework): Add shared `ProtobufClient` factory as a class method

jafermarq:
Adds `create_runtime_http_stub`, a shared factory for constructing protobuf-over-HTTP Runtime API clients.
The factory:
- Selects HTTP or HTTPS based on the security configuration.
- Supports default TLS verification, certificate paths, and in-memory CA certificates.
- Accepts any `ProtobufClient`-based Runtime stub and HTTP interceptors.


framework/py/flwr/supercore/runtime/client.py
```diff
@@ -0,0 +1,53 @@
...
+
+RuntimeHttpStubT = TypeVar("RuntimeHttpStubT", bound=ProtobufClient)
+
+
+def create_runtime_http_stub(
+    stub_class: type[RuntimeHttpStubT],
+    runtime_api_address: str,
+    insecure: bool,
+    root_certificates: bytes | str | None,
+    interceptors: Sequence[ProtobufClientInterceptor],
```
panh99: This seems like a generic factory function. Should we change the type of `stub_class` to any `ProtobufClient` and rename `runtime_api_address` to `server_address`? And then all http clients can reuse this.

framework/py/flwr/supercore/runtime/client.py
```diff
@@ -0,0 +1,53 @@
...
+from collections.abc import Sequence
+from typing import TypeVar
+
+from flwr.supercore.protobuf.client import ProtobufClient, ProtobufClientInterceptor
+
+RuntimeHttpStubT = TypeVar("RuntimeHttpStubT", bound=ProtobufClient)
+
+
+def create_runtime_http_stub(
+    stub_class: type[RuntimeHttpStubT],
```
panh99: Should we call it `client_class` instead of `stub_class`? IIUC, `stub` is a word commonly used in gRPC, but seems very rare in HTTP. Since we are switching to HTTP, should we simply call it `client`?

framework/py/flwr/supercore/runtime/client.py
```diff
@@ -0,0 +1,53 @@
...
+    runtime_api_address: str,
+    insecure: bool,
+    root_certificates: bytes | str | None,
```
panh99: Just curious, what's the type of `root_certificates` supported natively by `httpx`?

framework/py/flwr/supercore/runtime/client.py
panh99: We can also move it to `protobuf.client` module or even make it a class method of `ProtobufClient`.

framework/py/flwr/supercore/runtime/client.py
jafermarq: the standard way is to construct the `SSLContext` via bytes from the certificates and pass it to the `verify` arg. One could also pass the path to such certificates (but that's deprecated)

framework/py/flwr/supercore/runtime/runtime_http_stub_test.py
```diff
@@ -40,6 +40,19 @@
...
+        client = RuntimeHttpStub.from_server_address(
+            server_address="runtime.example:443",
+            insecure=False,
+            root_certificates=None,
+            interceptors=[],
+        )
+
+    assert isinstance(client, RuntimeHttpStub)
+
+
```
panh99:
```suggestion
```

panh99 (approved):
LGTM!

PR #7914, opened 2026-08-18
fix(framework): Make exit handling bounded and reentrant

tanertopal:
## What changed

- preserve the two-phase exit flow introduced in [#7938](https://github.com/flwrlabs/flower/pull/7938)
- make the exit-handler registry lock reentrant for synchronous signal re-entry
- make the signal-handler exit guard reentrant for the same reason
- document that pre-force handlers must be short and non-blocking
- add regression tests for lock re-entry, at-most-once handler execution, and test isolation

## Why

Python signal handlers execute synchronously on the main thread and can interrupt a short registry or signal-guard critical section. If the same signal path re-enters while a regular lock is held, the main thread can wait for itself.

The exit registry already snapshots and removes handlers under the lock, then runs callbacks without holding it. This PR adds the remaining reentrancy protection for the short registry and signal-guard sections.

The pre-force phase is intentionally bounded by contract: it may only contain short, local, non-blocking operations. Network calls, uploads, unbounded joins, and other potentially blocking work must remain in the regular phase, after the force-exit watchdog has started.

~~~mermaid
flowchart LR
    Signal["Exit requested"] --> Guard["Reentrant signal guard"]
    Guard --> Pre["Pre-force handlers<br/>short and non-blocking"]
    Pre --> Watchdog["Start force-exit watchdog"]
    Watchdog --> Regular["Regular handlers<br/>registry lock released"]
    Regular --> Exit["Normal process exit"]
    Watchdog -. "handlers stall" .-> Forced["Forced process exit"]
~~~

## Stack

Umbrella/reference: [#7895 — Prevent task executor shutdown deadlock](https://github.com/flwrlabs/flower/pull/7895)

- Foundation: [#7938 — Add two-phase exit handler execution](https://github.com/flwrlabs/flower/pull/7938) (merged)
- This PR: **#7914 — Make exit handling bounded and reentrant**
- Next: [#7915 — Make retry waits cancellable](https://github.com/flwrlabs/flower/pull/7915)

This PR now targets main and is rebased on the merged #7938 implementation.

## Validation

- pytest py/flwr/supercore/exit (15 passed)
- Ruff
- Black
- mypy
- Pylint
- git diff --check


framework/py/flwr/supercore/exit/signal_handler_test.py
```diff
@@ -65,3 +67,27 @@ def test_add_exit_handler(self) -> None:
...
+                with guard:
+                    graceful_exit_handler(signal.SIGTERM, Mock())
+                thread_finished.set()
+
+            thread = threading.Thread(target=invoke_while_locked, daemon=True)
+            thread.start()
+            thread.join(timeout=1.0)
+
+        self.assertTrue(thread_finished.is_set())
+        self.assertFalse(thread.is_alive())
```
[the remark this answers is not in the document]
tanertopal: Fixed in `1aa34cf6c`. The test now waits up to five seconds on the explicit completion event with a diagnostic failure message, then joins the already-completed thread without a timeout.

framework/py/flwr/supercore/exit/signal_handler_test.py
```diff
@@ -65,3 +67,27 @@ def test_add_exit_handler(self) -> None:
...
+
+    def test_signal_handler_reenters_exiting_guard(self) -> None:
+        """A nested signal inside the guard must not deadlock shutdown."""
+        with (
+            patch("flwr.supercore.exit.signal_handler.signal.signal") as mock_signal,
+            patch("flwr.supercore.exit.signal_handler.flwr_exit") as mock_flwr_exit,
+        ):
+            register_signal_handlers(EventType.PING)
+            graceful_exit_handler = mock_signal.call_args_list[0].args[1]
+            guard = inspect.getclosurevars(graceful_exit_handler).nonlocals["lock"]
```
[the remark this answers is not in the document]
tanertopal: Fixed in `1aa34cf6c`. The test now patches `signal_handler.RLock` to return a known lock instance, so it exercises same-thread re-entry without depending on closure variable names or `inspect.getclosurevars`.

framework/py/flwr/supercore/exit/exit_handler_test.py
```diff
@@ -15,7 +15,7 @@
 """Tests for exit handler functions."""
 
 
-from .exit_handler import add_exit_handler, trigger_exit_handlers
+from .exit_handler import _lock_handlers, add_exit_handler, trigger_exit_handlers
```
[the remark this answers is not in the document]
tanertopal: I'm keeping this invariant test intentionally. The public nested-handler test does not exercise registry-lock re-entry because callbacks are copied and then invoked after the registry lock has been released. This direct assertion protects the specific requirement that a synchronous Python signal can reacquire the registry lock if it interrupts the short add/snapshot critical section. Exposing a production test hook would be a larger coupling than this focused private invariant check.

framework/py/flwr/supercore/exit/exit.py
```diff
@@ -96,16 +96,16 @@ def flwr_exit(
...
     # Start a daemon thread to force exit if graceful exit fails
     def force_exit() -> None:
         time.sleep(FORCE_EXIT_TIMEOUT_SECONDS)
         os._exit(sys_exit_code)
 
     threading.Thread(target=force_exit, daemon=True).start()
 
+    # Trigger exit handlers
+    trigger_exit_handlers()
+
```
panh99: Moving `trigger_exit_handlers` before starting the forced-shutdown timer is intentional because the task process may need some time to upload final context or messages before exiting. Starting the timer before running the exit handlers could help address the hanging issue, but it would make it harder to guarantee that critical cleanup operations—such as uploading final messages—have enough time to complete before the process is forcibly terminated.

framework/py/flwr/supercore/exit/signal_handler.py
```diff
@@ -65,7 +65,10 @@ def register_signal_handlers(
     """
     default_handlers: dict[int, Callable[[int, FrameType], None]] = {}
     is_exiting = False
-    lock = Lock()
+    # Python runs signal handlers on the main thread between bytecode instructions.
+    # A second signal can therefore re-enter this handler while its first invocation
+    # still holds the guard; an RLock prevents that same thread from deadlocking.
+    lock = RLock()
```
panh99, line 71: Is the same-thread deadlock you're referring to caused by the signal handler being re-entered while a previous invocation is still inside `with lock:`? Since the following is the only place where lock is used, I don't think this can cause deadlock.

```python
        with lock:
            if is_exiting:
                return
            is_exiting = True
```

framework/py/flwr/supercore/exit/exit_handler.py
panh99, line 1: Is this trying to resolve the issue where a signal is received while we’re still registering the exit handlers?

framework/py/flwr/supercore/exit/signal_handler.py
tanertopal: We could not do it but here is an explainer which seems valid to me with an example attached. Do you prefer not to change it?

---
That does not prevent the deadlock. The same lock site is entered twice by two nested invocations of the same handler:

```
First handler invocation:
    enters `with lock:` and owns the lock

Second signal interrupts it:
    calls the same handler again
    reaches the same `with lock:`
    waits for the lock owned by the interrupted invocation
```

The first invocation cannot release the lock until the second invocation returns, so the main thread waits for itself.

This is specifically a same-site, same-thread re-entrancy deadlock - the exact case RLock supports.

It only occurs if the second signal lands during the very small `with lock:` window. The primary production deadlock was the exit-handler registry lock; **this local RLock change is additional defensive hardening.**

Example:
[deadlock.py](https://github.com/user-attachments/files/31189276/deadlock.py)

framework/py/flwr/supercore/exit/exit_handler.py
tanertopal: Here is another codex explainer which seems reasonable to me.

---
Previously, `trigger_exit_handlers()` held the registry lock while executing every callback:

```python
with _lock_handlers:
    for handler in reversed(registered_exit_handlers):
        handler()
```

The observed deadlock was:

```
trigger_exit_handlers acquires _lock_handlers
    ↓
An exit callback performs shutdown work
    ↓
A background authentication failure sends SIGINT
    ↓
The signal handler re-enters flwr_exit
    ↓
Nested trigger_exit_handlers tries to acquire _lock_handlers
    ↓
The main thread waits for a lock it already owns
```

The new implementation separates registry bookkeeping from callback execution:

```python
with _lock_handlers:
    handlers = list(reversed(registered_exit_handlers))
    registered_exit_handlers.clear()

for handler in handlers:
    handler()
```

This does three things:
1. Releases the lock before running potentially slow callbacks, fixing the observed deadlock.
1. Clears the callbacks first, ensuring nested shutdown does not execute them twice.
1. Allows callbacks to register new handlers without blocking on the registry lock.

The `RLock` covers the smaller remaining case: if a signal interrupts the main thread while it is registering, snapshotting, or clearing handlers, nested shutdown can reacquire the registry lock instead of deadlocking.

So:
1. Running callbacks outside the lock fixes the actual incident.
2. Using RLock protects registration and other short registry operations from signal re-entry.

framework/py/flwr/supercore/exit/exit.py
tanertopal: That is an intentional tradeoff. With the old ordering, an exit handler could deadlock before the watchdog even existed, leaving the process stuck forever. The later PRs give cleanup 2.5 seconds, final output the remaining time until 4.5 seconds, and reserve 0.5 seconds for forced exit.
So the intended contract becomes: final output is critical and gets most of the window, but process termination is ultimately guaranteed. **If five seconds is insufficient, the timeout should be increased** - not made ineffective by starting it after potentially blocking callbacks.

framework/py/flwr/supercore/exit/signal_handler.py
panh99: Yeah, my ChatGPT also said something similar until I point out the bit. I guess it might be confused. It's wrong b/c our code is far from the deadlock example; we just do:

```python
        with lock:
            if is_exiting:
                return
            is_exiting = True
```

And the exit handlers are executed outside `with lock:`, not inside like the `deadlock.py` example.

framework/py/flwr/supercore/exit/exit_handler.py
panh99: Thanks for adding the explainer. Codex is wrong here:

```
trigger_exit_handlers acquires _lock_handlers
    ↓
An exit callback performs shutdown work
    ↓
A background authentication failure sends SIGINT
    ↓
The signal handler re-enters flwr_exit  <-- This is wrong
    ↓
Nested trigger_exit_handlers tries to acquire _lock_handlers
    ↓
The main thread waits for a lock it already owns
```
Due to the lock we discussed above, i.e.:

```python
        with lock:
            if is_exiting:
                return
            is_exiting = True
```
The second signal won't re-enter `flwr_exit`; instead, it will just return. That was a bug before we introducing the `with lock:`, but it has been fixed by the first time we tried to resolve the pods issue.

framework/py/flwr/supercore/exit/signal_handler.py
tanertopal: Yes, the guard is intentionally reentrant. The first invocation releases this lock before calling `flwr_exit`, so a later signal normally observes `is_exiting=True` and returns. The remaining edge case is a second signal arriving while the first invocation is still inside this short `with lock:` section, before it has released the lock. Python can synchronously run the signal handler on the same main thread at that point; a regular `Lock` would make that invocation wait for itself. `RLock` allows it to reacquire the guard, observe the exit state, and return. The added test exercises exactly that same-thread re-entry. This lock does not cover or serialize the exit callbacks.

framework/py/flwr/supercore/exit/exit_handler.py
tanertopal: It protects the short registry critical sections, including registration and the snapshot/remove operation. Exit callbacks are intentionally invoked after the registry lock is released; that is what prevents a slow callback from blocking a later registration or shutdown path. The `RLock` is defense in depth for a signal that interrupts one of those short sections and re-enters registry code on the same main thread. The regression test covers same-thread lock re-entry directly.

framework/py/flwr/supercore/exit/exit.py
tanertopal: This concern is now addressed by the merged two-phase flow in #7938. The pre-force phase is reserved for short, local, non-blocking operations; potentially blocking cleanup and final uploads remain in the regular phase after the watchdog starts. This PR preserves that ordering and documents the contract, so the original line is outdated.

framework/py/flwr/supercore/exit/signal_handler.py
panh99: I think this is exactly the behavior I do not want. The guard is intended to prevent a nested invocation from entering the critical section, not to make same-thread re-entry possible.

Also, with `RLock`, a second signal arriving after the `if is_exiting:` check but before `is_exiting = True` can reacquire the lock while is_exiting is still False and proceed through the handler as well. So reentrancy does not seem to provide the protection we want here.

Based on how the `lock` is used here, I don't currently see a scenario where it would lead to a deadlock. Could you point me to a concrete case where that could happen?

PR #7923, opened 2026-08-18
refactor(framework): Consolidate the Runtime HTTP client in `SuperCore`

jafermarq:
Description:
- Rename RuntimeHttpStub to RuntimeHttpClient to use HTTP-appropriate terminology.
- Move the remaining Runtime HTTP methods from the SuperLink client into the shared supercore client.
- Keep the SuperLink client as a thin compatibility subclass.
- Consolidate endpoint tests under supercore.


panh99 (approved):
LGTM!

PR #7934, opened 2026-08-19
feat(framework): Integrate privacy accounting with fixed clipping

mohammadnaseri:


framework/py/flwr/serverapp/strategy/dp_fixed_clipping.py
```diff
@@ -80,9 +96,122 @@ def __init__(
...
+        )
+        if (
+            self.accountant.config.max_epsilon is not None
+            and self.accountant.would_exceed(event)
+        ):
+            spent = self.accountant.get_privacy_spent()
+            raise PrivacyBudgetExhausted(
+                "The next private release would exceed max_epsilon "
+                f"(current epsilon: {spent.epsilon}, delta: {spent.delta})."
+            )
```
panh99, line 163: `AppExitException` (the superclass of `PrivacyBudgetExhausted`) is designed for severe, irrecoverable exceptions that occur at the application level. For example, a strategy might receive client model updates based on incompatible model architectures.

This type of exception is different from something like `ValueError`, which users are generally expected to catch and handle.

I’m less familiar with the `PrivacyAccountant`, but I wanted to double-check whether this behavior matches what you expect.

mohammadnaseri: I see, that is a fair point, the intent is to treat `max_epsilon` as a hard safety limit. Once the next release would exceed it, the strategy cannot continue safely, so it terminates before sending messages. So I would say let's keep it like that.

panh99: Makes sense!

danieljanes (dismissed):

danieljanes (approved):

PR #7935, opened 2026-08-19
ci(framework): Fix Docker release tag promotion

panh99:
## What changed

- Promote only variant-specific `unstable-*` tags to versioned release tags.
- Add the bare release version and `latest` aliases explicitly to the OCI indexes selected by `DOCKER_TAG_GROUPS`.
- Derive the image repository name from the final component of the fully qualified repository path.

## Why

The finalize workflow previously replaced the `unstable` prefix mechanically. Because the Python 3.13 Ubuntu build also carries the bare `unstable` tag, this created `base:<version>`, pointed the SuperLink and SuperNode release aliases at Ubuntu instead of Alpine, and did not update `latest`.

The corrected promotion keeps the existing public Docker tag policy and continues promoting the already-built OCI indexes from `digests.json` without rebuilding images.

## Validation

- Exercised the workflow's promotion script against the complete generated 13-index unstable build matrix.
- Compared all generated promotion targets directly with `DOCKER_TAG_GROUPS` for release `1.34.0`.
- Verified repository detection with a registry hostname containing a port and a nested namespace.
- Parsed the workflow YAML and checked the embedded Bash syntax.
- Ran `git diff --check`.

No Docker README changes are needed because the documented tag policy is unchanged.

### DIFF
--- a/.github/workflows/framework-release-finalize.yml
+++ b/.github/workflows/framework-release-finalize.yml
@@ -151,7 +151,7 @@ jobs:
               tag_suffix="${unstable_tag#"${repository}:unstable"}"
               tag_args+=(-t "${repository}:${VERSION}${tag_suffix}")
 
-              # Add stable aliases to the variants defined by DOCKER_TAG_GROUPS.
+              # Add stable aliases explicitly for the current release variants.
               case "${repository_name}:${tag_suffix}" in
                 "superlink:-py3.13-alpine3.22"|"supernode:-py3.13-alpine3.22")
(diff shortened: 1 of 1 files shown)


chongshenng (approved):
LGTM. One note: These alias rules duplicate `DOCKER_TAG_GROUPS`, so a future Python, distro, or tag-policy change must update both places. We should add a consistency test or move the mapping to one source of truth later.

danieljanes (approved):

PR #7938, opened 2026-08-19
feat(framework): Add two-phase exit handler execution

panh99:
## Summary

- add a `run_before_force_exit` registration flag for exit handlers
- run pre-force handlers before starting the force-shutdown timer
- run all remaining handlers after the timer starts
- preserve per-phase LIFO ordering, at-most-once execution, and exception isolation

## Motivation

Some cleanup must complete before the force-shutdown countdown begins. Previously, all handlers ran in a single phase before the timer, so callers could not opt into the new sequencing while retaining the existing default behavior.

Existing `add_exit_handler(...)` callers remain unchanged and continue to run in the regular post-timer phase by default.

## Validation

- `pytest -q py/flwr/supercore/exit` (13 passed)
- Ruff
- Black
- mypy
- pylint


framework/py/flwr/supercore/exit/exit_test.py
```diff
@@ -39,3 +39,29 @@ def test_get_code_url(version: str, subdir: str) -> None:
...
+        operations.attach_mock(thread, "thread")
+
+        flwr_exit(0)
+
+    assert operations.mock_calls == [
+        call.trigger_handlers(run_before_force_exit=True),
+        call.thread(target=ANY, daemon=True),
+        call.thread().start(),
+        call.trigger_handlers(run_before_force_exit=False),
+    ]
```
panh99:
```suggestion
```

framework/py/flwr/supercore/exit/exit_test.py
```diff
@@ -15,11 +15,11 @@
 """Tests for the exit function."""
 
 
-from unittest.mock import patch
+from unittest.mock import ANY, Mock, call, patch
```
panh99:
```suggestion
from unittest.mock import patch
```

framework/py/flwr/supercore/exit/exit_test.py
```diff
@@ -15,11 +15,11 @@
...
 
 from parameterized import parameterized
 
-from .exit import _get_code_url
+from .exit import _get_code_url, flwr_exit
```
panh99:
```suggestion
from .exit import _get_code_url
```

tanertopal (approved):

PR #7936, opened 2026-08-19
refactor(framework): Switch `RuntimeAPI` to `HTTP` from `gRPC`

jafermarq:
witch the `Runtime API` from gRPC to HTTP across `SuperLink`, `SuperNode`, `SuperExec`, and `flwr-*` processes. Fleet and Control APIs remain on gRPC, and the existing Runtime gRPC implementation is retained for removal in a follow-up cleanup PR.

> [!NOTE]
> Run the below after navigating to `flower/framework`

### Test (1xSuperLink + 2xSuperNode)

```shell
# SuperLink
uv run --no-sync --python=3.11.14 flower-superlink --insecure

# SuperNode 1
uv run --no-sync --python=3.11.14 flower-supernode --insecure --port 8004

# SuperNode 2
uv run --no-sync --python=3.11.14 flower-supernode --insecure --port 8005

# Submit a run
uvx flwr run @flwrlabs/demo <your-local-deployment-federation> --stream
```

### Test (1xSuperLink + 2xSuperNode all with `--isolation=process`)

```shell
# SuperLink + SuperExec
uv run --no-sync --python=3.11.14 flower-superlink --insecure --isolation=process
uv run --no-sync --python=3.11.14 flower-superexec \
  --insecure \
  --appio-api-address 127.0.0.1:8000 \
  --plugin-type serverapp-ephemeral

# SuperNode 1 + SuperExec
uv run --no-sync --python=3.11.14 flower-supernode --insecure --port 8004 --isolation=process
uv run --no-sync --python=3.11.14 flower-superexec \
  --insecure \
  --appio-api-address 127.0.0.1:8004 \
  --plugin-type clientapp

# SuperNode 2 + SuperExec
uv run --no-sync --python=3.11.14 flower-supernode --insecure --port 8005 --isolation=process
uv run --no-sync --python=3.11.14 flower-superexec \
  --insecure \
  --appio-api-address 127.0.0.1:8005 \
  --plugin-type clientapp

# Submit a run
uvx flwr run @flwrlabs/demo <your-local-deployment-federation> --stream
```

PR #7941, opened 2026-08-19
feat(framework): Add Open Responses compatible `/responses` endpoint to Runtime API

charlesbvll:
## New Flow

```
  AgentApp
    │ OpenAI SDK: POST /v1/runtime/responses
    │ Authorization: Bearer <AgentApp task token>
    ▼
  Flower Responses endpoint
    │ creates child MODEL task + stores ModelRequest
    ▼
  flwr-model
    │ calls configured model provider
    │ publishes Open Responses events
    ▼
  Flower endpoint
    │ relays events as SSE
    ▼
  OpenAI SDK stream iterator
```

## Decisions made

- Expose `POST /v1/runtime/responses` through a regular FastAPI JSON/SSE router because the existing Runtime router uses `ProtobufRoute`
- Authenticate with `Authorization: Bearer <task-token>` so standard Open Responses-compatible SDKs can call the endpoint directly
- Accept only AgentApp task tokens
- Reuse the existing model execution path by creating a child `flwr-model` task and sending it the existing `ModelRequest`
- Keep provider routing, task relationships, run attribution, and usage accounting unchanged
- Support both non-streaming JSON responses and SSE streaming in the initial implementation
- Initially support only the fields already represented by `ModelRequest`
  - `model`
  - `input`
  - `stream`
  - `tools`
  - `tool_choice`
  - `reasoning`
  - `previous_response_id`
  - `instructions`
  - `max_output_tokens`
  - `metadata`
  - `text`
- Reject unsupported fields explicitly instead of silently ignoring them
- Correlate replies atomically through `reply_to_message_id`
  - Model replies for concurrent requests share the same AgentApp destination
  - Filtering must happen inside the destructive CoreState claim so one request cannot consume another request's reply
- Filter stream events by the child model task so unrelated events from the same run cannot enter the SDK stream
- Consume the final task reply before exposing a terminal SSE event, preventing it from remaining in the AgentApp mailbox
- Mark unfinished child model tasks as `STOPPED` when the request times out or the client disconnects
- Inject `FLWR_RUNTIME_BASE_URL` and `FLWR_RUNTIME_API_KEY` before loading the AgentApp
- Keep the existing `agent.responses.create(...)` path unchanged
- Do not add the OpenAI SDK as a framework runtime dependency
- No database migration is required because reply correlation and task IDs are already stored

## Open questions

- [x] Is the initial request-field subset sufficient for the first release -> Yes we keep just the used subset
- [ ] Which additional Open Responses event types, if any, need explicit support
- [ ] Should HTTP errors follow the exact OpenAI error contract, including status codes and error fields, or is compatibility at the SDK level sufficient
- [ ] Is marking the child task as `STOPPED` enough on disconnect, or should we also add executor-level cancellation to terminate an in-flight provider request immediately
- [x] Should `agent.responses.create(...)` eventually become a wrapper around this endpoint or remain on its current lower-level Runtime path -> We leave it unchanged for now
- [ ] Should the OpenAI SDK end-to-end example be included in this PR or added separately

## Deps

- This branch is stacked on the Runtime HTTP work in #7936 and should be reconciled with its final merged form
- Documentation work should be coordinated with #7930
- This implements the supported Flower subset of the Open Responses API and does not claim complete API compatibility


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -879,6 +879,7 @@ def get_task_message(
         self,
         *,
         dst_task_ids: Sequence[int] | None = None,
+        reply_to_message_ids: Sequence[str] | None = None,
```
danielnugraha: Could we avoid adding `reply_to_message_ids` to `CoreState` and use `model_task_id` instead?

framework/py/flwr/superlink/routers/runtime/responses_test.py
```diff
@@ -0,0 +1,257 @@
...
+    return TaskEvent(
+        id=event_id,
+        run_id=789,
+        task_id=456,
+        event=event,
+        data=data or f'{{"type":"{event}"}}',
+    )
+
+
+def _capture_requests(state: Mock) -> list[ModelRequest]:
```
danielnugraha: Could we remove _capture_requests since LinkState is mocked?

framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -0,0 +1,419 @@
...
+        "metadata",
+        "text",
+    }
+)
+_TERMINAL_EVENTS = frozenset(
+    {"error", "response.completed", "response.failed", "response.incomplete"}
+)
+_REPLY_TIMEOUT = 300.0
+_POLL_INTERVAL = 0.25
+_TERMINAL_EVENT_GRACE = 1.0
```
danielnugraha: Why do we need this?

charlesbvll: This is needed because we poll the task events and the final `ModelResponse` separately. We can poll the events just before `response.completed` is stored, then observe the final response before polling the events again. Without a short grace period, we would close the stream and report a missing event even though it is available on the next poll. The `1s` value itself is somewhat arbitrary though ^^ I added a small comment to make this clearer

Note that this is what the test at `responses_test.py:160` reproduces

framework/py/flwr/supercore/corestate/corestate.py
charlesbvll: I switched to use task_ids

framework/py/flwr/superlink/routers/runtime/responses_test.py
charlesbvll: done

framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -0,0 +1,422 @@
...
+    """Extract a safe message from a structured failed response."""
+    error = response.get("error")
+    if isinstance(error, dict) and isinstance(error.get("message"), str):
+        return cast(str, error["message"])
+    return "Model request failed."
+
+
+def _stop_model_task(state: LinkState, exchange: _Exchange, details: str) -> None:
+    """Stop an unfinished model task and drain an already-arrived reply."""
+    state.finish_task(exchange.model_task_id, SubStatus.STOPPED, details)
```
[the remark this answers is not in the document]
charlesbvll: I think this is valid, but perhaps better handled in a follow-up PR. This PR uses the existing Runtime stop semantics, which mark the child task as stopped, but `SubprocessExecutor` does not currently retain a task-to-process handle, so it cannot interrupt a provider request that is already in flight. Properly fixing that seems broader than `/responses`

framework/proto/flwr/proto/runtime.proto
```diff
@@ -152,7 +152,10 @@ message PushTaskEventsRequest { repeated TaskEvent events = 1; }
 message PushTaskEventsResponse {}
 
 // PullTaskMessage messages
-message PullTaskMessageRequest { optional uint64 limit = 1; }
+message PullTaskMessageRequest {
+  optional uint64 limit = 1;
+  optional uint64 src_task_id = 2;
```
danielnugraha, line 158: Why do we need this?

framework/py/flwr/superlink/routers/runtime/responses.py
danielnugraha: I see, I think we don't need this since the flwr-model ordering is synchronous. Can you remove this?

framework/py/flwr/supercore/task_process/agent/run_agentapp.py
```diff
@@ -270,3 +284,60 @@ def on_exit() -> None:
...
+    """Expose the Open Responses-compatible Runtime endpoint to the AgentApp."""
+    scheme = "http" if insecure else "https"
+    address = runtime_api_address.rstrip("/")
+    os.environ[_RUNTIME_BASE_URL_ENV] = f"{scheme}://{address}/v1/runtime"
+    os.environ[_RUNTIME_API_KEY_ENV] = token
+
+    if certificates is None:
+        return None
+
+    # OpenAI/httpx reads custom CAs from SSL_CERT_FILE, which requires a path.
```
danielnugraha, line 305: @panh99 I'm not sure about this bit of code it seems different to the certificates that we have in the codebase. Can you check this?

framework/proto/flwr/proto/runtime.proto
panh99, line 158: Just in case this field gets confused with the authenticated task ID of the requester. Should we add a comment like this?

```suggestion
  optional uint64 src_task_id = 2;  // Pull message by sender task ID
```

**Btw, do we actually need this field? I don't seem to find any usage of it.**

framework/py/flwr/superlink/routers/runtime/__init__.py
```diff
@@ -15,6 +15,7 @@
 """Runtime API router."""
 
 
+from .responses import router as responses_router
 from .router import router
 
-__all__ = ["router"]
+__all__ = ["responses_router", "router"]
```
panh99, line 21:
```suggestion
__all__ = [
    "responses_router",
    "router"
]
```

framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -0,0 +1,411 @@
+# Copyright 2026 Flower Labs GmbH. All Rights Reserved.
```
danielnugraha, line 1: It seems like this code is the same as session.py in agent. Should we try unifying it?

framework/proto/flwr/proto/runtime.proto
danielnugraha: It seems like this is needed if we support concurrency. Do we have concurrency here?

charlesbvll: But OpenAI SDK and FastAPI allow concurrent calls, so we can’t assume there will only be one outstanding request per AgentApp, right? So it’s probably safer to keep this safeguard in place

framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -0,0 +1,437 @@
...
```
danielnugraha, line 1: Can we unify the runtime/responses and agent/session.py? Right now we have the duplication
RuntimeAgentResponses.create()
    → CreateTask
    → PushTaskMessage
    → poll PullTaskMessage

POST /v1/runtime/responses
    → CreateTask
    → store_task_message
    → poll LinkState

The CreateTask > Push/StoreTaskMessage > Poll can be unified.

framework/proto/flwr/proto/runtime.proto
danielnugraha: Okay, got it.

danielnugraha (commented):
Can we split this PR to make it more manageable?

1. proto and corestate changes
2. create a shared file that both POST /v1/runtime/responses and RuntimeAgentResponses.create calls
3. POST /v1/runtime/responses implementation

PR #7942, opened 2026-08-19
fix(framework): Restore SuperLink HTTP logging configuration

mohammadnaseri:
This PR restores the SuperLink HTTP logging configuration. Successful `/health` access logs are filtered, Uvicorn logs use UTC timestamps, and noisy `httpx` and `httpcore` logs are suppressed. Note that the configuration is applied explicitly when starting Uvicorn without modifying root logging at import time.


tanertopal (approved):
Lgtm!

PR #7946, opened 2026-08-20
fix(framework): Validate FedTrimmedAvg beta

tandede:
### Description

Both the ServerApp and legacy `FedTrimmedAvg` implementations accept trimming fractions outside the valid range. With `beta=0.5` and an even number of client updates, both tails remove every value. The existing `lowercut > uppercut` check misses the equal-boundary case, so NumPy computes the mean of an empty slice and returns `NaN` model parameters. Negative values also fail only later during partitioning.

### Related issues/PRs

None.

### Explanation

Validate that `beta` and the lower-level cut fraction are in `[0, 0.5)` before aggregation. Apply the same contract to the ServerApp and legacy strategy implementations, document the accepted range, and add regression coverage for both the negative and half-cut boundaries.

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Make CI checks pass
- [ ] Ping maintainers on Slack (channel `#contributions`)

### Any other comments?

Validation completed locally:

- `pytest -q py/flwr/server/strategy/aggregate_test.py py/flwr/server/strategy/fedtrimmedavg_test.py py/flwr/serverapp/strategy/fedtrimmedavg_test.py` (18 passed)
- `uv run --no-sync --python=3.11.14 ./dev/test.sh false` (all Framework checks passed)


jafermarq (approved):
Hey @tandede ! Many thanks for helping us setting the right beta limits for `FedTrimmedAvg`! I've just approved your PR and we'll merge it shortly 🚀

danieljanes (approved):

PR #7949, opened 2026-08-20
fix(framework): Configure logging for direct Uvicorn startup

mohammadnaseri:
This PR applies the shared SuperLink HTTP logging configuration when the API is launched directly with Uvicorn.


tanertopal (approved):
Lgtm!

PR #7951, opened 2026-08-20
feat(framework): Add StoreApp state methods

danielnugraha:
## Summary

  Add state-layer support for adding apps to a federation.
  - store_app requires a Fab.
  - federation_app.fab_hash is non-nullable.
  - get_app to retrieve app from LinkState.


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -147,14 +148,20 @@ def store_app(  # pylint: disable=too-many-arguments,too-many-positional-argumen
 
         Returns
         -------
-        str
-            Canonical SHA-256 hash of the stored FAB.
+        Optional[str]
+            Canonical SHA-256 hash of the stored FAB, or None if unpinned.
```
panh99: Do we need to return anything here? I don't think the FAB hash is actually used, right? The `Fab` object already contains the hash anyway. Would removing the return value make the logic simpler?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -147,14 +148,20 @@ def store_app(  # pylint: disable=too-many-arguments,too-many-positional-argumen
...
         """
 
     @abstractmethod
     def get_fab(self, fab_hash: str) -> Fab | None:
         """Return the FAB for the given hash, if present."""
 
+    @abstractmethod
+    def get_app_fab(
```
panh99: Should we name is as `get_app`?

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -147,14 +148,20 @@ def store_app(  # pylint: disable=too-many-arguments,too-many-positional-argumen
...
+        self, federation_id: str, app_id: str, fab_hash: str
```
panh99: Since we make `fab` optional in `store_app`, how will we get the app that does not have a `fab`?

danielnugraha: We always fetch from hub for apps without fab. This is to ensure that we always run the latest version of the app from hub.

panh99 (dismissed):
LGTM!

panh99 (approved):
LGTM!

PR #7955, opened 2026-08-20
refactor(framework): Consolidate SuperLink and SuperNode Uvicorn defaults

jafermarq:
Note moving forward we drop the API name in the constant. This is because all APIs, once moved to HTTP, will operate of the same port. Below the default ports for `SuperLink` and `SuperNode`

```python
SUPERLINK_UVICORN_DEFAULT_PORT = 8000
SUPERNODE_UVICORN_DEFAULT_PORT = 9094
```


panh99 (approved):
LGTM!

danieljanes (approved):

PR #7960, opened 2026-08-21
refactor(framework): Move Runtime tests from gRPC servicers to HTTP handlers

jafermarq:
Moves relevant Runtime tests from the obsolete gRPC servicers to HTTP routers and shared handlers. The gRPC servicers will be removed in a follow-up PR.


panh99 (approved):
LGTM!

danieljanes (approved):

PR #7961, opened 2026-08-21
fix(framework): Fix run once automations

charlesbvll:
Automations that run once should not have a `fixed_interval`


danielnugraha (approved):

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -836,6 +836,10 @@ def start_automation(  # pylint: disable=too-many-locals
             ),
         )
 
+    # One-run automations do not recur, so an interval is not meaningful.
+    if max_runs == 1:
+        fixed_interval = None
```
panh99, line 841: Can we raise an error instead of changing the request?

charlesbvll: It is not an error, it is in case the automation is something like `Remind me to do ... tomorrow at 9`.
In which case, it doesn't make sense to have an `interval`

panh99 (approved):

PR #7962, opened 2026-08-21
feat(framework): Pin flwr-agent to ListApps

danielnugraha:
Always populate ListApps with flwr-agent.


framework/py/flwr/cli/constant.py
```diff
@@ -47,7 +47,7 @@
 )
 
 # Constants for `flwr chat`
-CHAT_FLOWER_AGENT_APP_SPEC = "@flwrlabs/flwr-agent"
+CHAT_FLOWER_AGENT_APP_SPEC = FLOWER_AGENT_APP_ID
```
panh99: Just curious, why do we need both `CHAT_FLOWER_AGENT_APP_SPEC` and `FLOWER_AGENT_APP_ID`? Can we set `FLOWER_AGENT_APP_ID` as default directly?

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -1352,7 +1354,15 @@ def list_apps(
     federation_id = request.federation_id
     _validate_federation_membership_in_request(state, account.flwr_aid, federation_id)
     limit = request.limit if request.HasField("limit") else None
-    return ListAppsResponse(apps=state.list_apps(federation_id, limit))
+    apps = list(state.list_apps(federation_id, limit))
+    if limit != 0 and not any(app.app_id == FLOWER_AGENT_APP_ID for app in apps):
```
panh99: Should we run `list_apps(..., limit=limit-1)` and add `FLOWER_AGENT_APP_ID`? Alternatively, we can always add `FLOWER_AGENT_APP_ID` and return `limit+1` apps. I think `limit=0` is not really a valid case, no?

panh99 (approved):
LGTM!

PR #7967, opened 2026-08-21
fix(framework): Reuse Runtime CA for AgentApp SDKs

panh99:
## Summary

- expose the explicit Runtime CA file through `SSL_CERT_FILE` for OpenAI/httpx SDK clients
- keep `HttpGrid` using the same validated certificate bytes as before
- remove temporary CA bundle reconstruction and inherited trust-store merging
- keep `_set_runtime_environment` focused on the Runtime base URL and task token

### DIFF
--- a/framework/py/flwr/supercore/cli/flwr_agentapp_test.py
+++ b/framework/py/flwr/supercore/cli/flwr_agentapp_test.py
@@ -87,11 +87,11 @@ def test_flwr_agentapp_parses_args_before_mirroring_output() -> None:
     mirror_output_to_queue.assert_not_called()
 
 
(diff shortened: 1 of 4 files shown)


framework/py/flwr/supercore/cli/flwr_agentapp_test.py
```diff
@@ -87,6 +89,7 @@ def parse_args(self) -> SimpleNamespace:
 
 def test_flwr_agentapp_forwards_cli_args() -> None:
     """The AgentApp CLI should forward parsed args to the runtime."""
+    monkeypatch.delenv("SSL_CERT_FILE", raising=False)
```
panh99:
```suggestion
```

PR #7971, opened 2026-08-22
feat(framework): Add source-task message filters

charlesbvll:
Adds an optional source task filter to `PullTaskMessage`.

This lets an AgentApp wait for the response from a specific child task instead of consuming whichever response arrives first. It prevents concurrent model requests from receiving each other’s responses.


danielnugraha (approved):

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7983, opened 2026-08-23
feat(framework): Add read-only streaming to agentapp

charlesbvll:
Extends `POST /v1/runtime/responses` with `stream=True` support.

- Relays Open Responses events from the request’s child model task over SSE
- Filters events by task so concurrent model calls cannot receive each other’s events
- Waits for the final model reply before ending the stream
- Stops the model task when the stream is cancelled or ends early

Non-streaming behavior remains unchanged. AgentApp control over which events reach the frontend is handled separately.


framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -239,6 +247,101 @@ async def _wait_for_response(
...
+    response_deadline = started_at + _DEFAULT_MODEL_RESPONSE_TIMEOUT
+    cursor: int | None = None
+    complete = False
+    response: JSONObject | None = None
+    terminal_event_deadline: float | None = None
+    try:
+        while True:
+            events = await run_in_threadpool(
+                state.get_task_events,
+                run_id=exchange.run_id,
```
panh99: Since the `task_id` can already uniquely identify the event emitter, do we still need the `run_id=exchange.run_id` filter here?

```suggestion
```

framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -71,6 +77,7 @@ class _Exchange:
 
     agent_task_id: int
     model_task_id: int
+    run_id: int
```
panh99: This might not be necessary, see https://github.com/flwrlabs/flower/pull/7983/changes#r3842322029

framework/py/flwr/superlink/routers/runtime/responses.py
panh99, line 1: QQ: the producer `flwr-model` still batches 16 events before pushing. Do we need to update there, or it's good enough? No need to make any change for this comment, just curious about the future plan.

framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -339,6 +442,29 @@ def _stop_model_task(state: LinkState, exchange: _Exchange, details: str) -> Non
...
+    return f"event: {event.event}\n{data_fields}\n"
+
+
+def _stream_error(message: str, code: str) -> str:
+    """Encode a terminal Responses error event."""
+    data: JSONObject = {
+        "type": "error",
+        "error": {"type": "server_error", "code": code, "message": message},
+    }
+    return f"event: error\ndata: {strict_json_dumps(data, compact=True)}\n\n"
```
panh99: I am happy with a simple event structure like this. Just curious, this structure is intentional to be different from Open Responses format, right?

framework/py/flwr/superlink/routers/runtime/responses.py
```diff
@@ -239,6 +247,101 @@ async def _wait_for_response(
...
+                    state,
+                    exchange,
+                    launch_deadline,
+                    response_deadline,
+                )
+                if response is not None:
+                    terminal_event_deadline = min(
+                        response_deadline,
+                        time.monotonic() + _TERMINAL_EVENT_GRACE,
+                    )
```
panh99: IIUC, this is essentially saying: “I’ve received the final response, but I’ll wait another 1s in case the final response was pushed before the terminal event.” Is that correct?

Based on the event producer logic [here](https://github.com/flwrlabs/flower/blob/15046960aebdb0ef705a7a3f0a17a36c5989d5bd/framework/py/flwr/supercore/task_process/model/task.py#L91-L96), all events are pushed before the final response. Could we remove the `TERMINAL_EVENT_GRACE` mechanism entirely? I think that would make `_stream_response` easier to follow. Wdyt?

framework/py/flwr/superlink/routers/runtime/responses.py
charlesbvll: Good catch, I forgot to update it, it was not intentional ^^

framework/py/flwr/superlink/routers/runtime/responses.py
charlesbvll: Yes, done

framework/py/flwr/superlink/routers/runtime/responses.py
charlesbvll: It’s probably good enough for this first version, although a short stream could remain buffered until it completes. In a follow up we could also include some kind of timeout, for example, flush at 16 events or after a small maximum delay

panh99: Sounds good! Agreed that this is sufficient for the first version. No need to spend much time on it now.

panh99 (approved):
LGTM!

PR #7985, opened 2026-08-24
feat(framework): Control streaming events from agentapp

charlesbvll:
Give AgentApps control over which model streaming events are exposed to the frontend.

- Keep `flwr-model` events private to the AgentApp’s `/responses` stream
- Filter model-task events from `Control.StreamRunEvents` for AgentApp runs
- Let AgentApp code inspect events and selectively republish them with `agent.events.emit(...)`
- Publish selected events asynchronously and flush them before the AgentApp exits
- Keep event streaming unchanged for non-AgentApp runs


framework/py/flwr/supercore/task_process/agent/run_agentapp.py
```diff
@@ -261,7 +249,37 @@ def on_exit() -> None:
...
+            agent_app(agent=agent, context=context)
+        except Exception:
+            try:
+                agent_events.close()
+            except Exception as close_ex:  # pylint: disable=broad-exception-caught
+                log(
+                    ERROR, "Failed to close AgentApp event publisher", exc_info=close_ex
+                )
+            raise
+        agent_events.close()
```
panh99, line 273: No need to make this change specifically. Could we revert the changes in this block? The cleanup logic can stay inside `on_exit`, as you’ve already done. This extra `try-except` adds branches and nesting without much benefit. For context, we can assume that `on_exit` is guaranteed to be called on both success and failure paths.

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -59,6 +61,108 @@
...
+        if self._closed:
+            self._raise_worker_error()
+            return
+        try:
+            self.flush()
+        finally:
+            self._closed = True
+            self._queue.put(_EVENT_PUBLISH_STOP)
+            self._worker.join()
+        self._raise_worker_error()
```
panh99: Could we add a timeout here?
```suggestion
    def close(self, timeout: float | None = None) -> None:
        """Flush pending events and stop the background publisher."""
        if self._closed:
            self._raise_worker_error()
            return
        try:
            self.flush()
        finally:
            self._closed = True
            self._queue.put(_EVENT_PUBLISH_STOP)
            self._worker.join(timeout)
        self._raise_worker_error()
```

framework/py/flwr/supercore/task_process/agent/run_agentapp.py
```diff
@@ -109,13 +109,20 @@ def run_agentapp(  # pylint: disable=R0912, R0913, R0914, R0915, R0917, W0212
...
     exit_code = ExitCode.SUCCESS
 
     def on_exit() -> None:
         log(DEBUG, "[flwr-agentapp] Will push AgentApp task output")
 
         grid._retry_invoker.max_tries = 1
 
+        if agent_events is not None:
+            try:
+                agent_events.close()
```
panh99: Should we consider add a timeout, say 1s?
```suggestion
                agent_events.close(timeout=...)
```

framework/py/flwr/supercore/task_process/agent/session.py
charlesbvll: @panh99 I’m not sure this timeout would help much here. `flush()` waits on `Queue.join()` before we reach `worker.join(timeout)`, so if event publishing is stuck, the timeout is never reached. It would only limit the final thread shutdown after all events have already been handled.
A meaningful timeout would need to cover the whole flush, but then we would need to decide what happens to events that have not been persisted yet, and I would prefer not to silently drop them. Since `flwr_exit` already has a five-second hard fallback, would keeping the lossless wait here be acceptable?

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -737,9 +744,20 @@ def stream_run_events(
             run_id=run_id,
             after_task_event_id=after_task_event_id,
         )
+        visible_task_ids: set[int] | None = None
+        if agentapp_controls_events and events:
+            # Model events remain private to the AgentApp's response stream. Events
+            # selected for frontend exposure are republished under the AgentApp task.
+            event_tasks = state.get_tasks(
+                task_ids=list(dict.fromkeys(event.task_id for event in events))
```
panh99:
```suggestion
                task_ids=list({event.task_id for event in events})
```

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -737,9 +744,20 @@ def stream_run_events(
             run_id=run_id,
             after_task_event_id=after_task_event_id,
         )
```
panh99, line 746: Am I understanding correctly that the change in `stream_run_events` is only intended to hide task events originating from `flwr-model`? If so, I think a cleaner design would be to add a `from_primary_task: bool` filter to `state.get_task_events` and return only events from primary tasks. That seems semantically more straightforward and should achieve the same result.

panh99: Alternatively, we can add a `task_types` filter.

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -737,9 +744,20 @@ def stream_run_events(
             run_id=run_id,
```
panh99, line 743:
```suggestion
            task_ids=[run.primary_task_id],
```

framework/py/flwr/supercore/task_process/agent/run_agentapp.py
panh99:
```suggestion
                agent_events.close(1)
```
Could we add a 1s timeout?

panh99 (approved):
LGTM!

danieljanes (approved):

PR #7987, opened 2026-08-24
feat(framework): Identify Hub federation apps

smoroso:
## What

Persist whether a federation app originated from Flower Hub and expose that information through the control API's `ListApps` response.

The protobuf field is presence-aware:
- `true`: the app is known to come from Flower Hub.
- `false`: the app is known to be a locally supplied/custom FAB.
- absent: the association predates provenance tracking.

## Why

The Labs website needs to distinguish Hub-published apps from custom apps. Hub apps should resolve the current published FAB on each run, while custom apps should continue using their explicitly associated FAB hash.

Existing federation-app rows cannot be classified reliably from stored FAB data alone. Preserving them as unknown avoids permanently misclassifying old Hub apps as custom apps.

## How to manually test

1. Start a SuperLink with a database created before this migration and add an existing `federation_app` row.
2. Apply the migration and call `ListApps`; verify the legacy app has no `is_hub_app` field.
3. Add or run a Hub-published agent and verify `ListApps` returns `is_hub_app: true`.
4. Add or run a custom FAB and verify `ListApps` returns `is_hub_app: false`.
5. Confirm app registration, `StartRun`, protobuf generation, and migration downgrade continue to work.


framework/py/flwr/supercore/corestate/sql_corestate.py
```diff
@@ -540,10 +544,10 @@ def list_apps(
                     app_id=app.app_id,
                     fab_hash=app.fab_hash,
                     app_type=app.app_type,
+                    is_hub_app=app.is_hub_app,
                 )
                 for app in apps
             ]
```
panh99, line 550:
```suggestion
            ]

```

framework/py/flwr/supercore/state/alembic/utils_test.py
```diff
@@ -301,6 +301,56 @@ def test_migrated_schema_matches_metadata(self) -> None:
         finally:
             engine.dispose()
 
+    def test_hub_origin_migration_preserves_unknown_legacy_rows(self) -> None:
```
panh99, line 304: I guess we don't need a unit test for adding a nullable column.

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -606,6 +606,7 @@ def start_run(  # pylint: disable=too-many-branches,too-many-locals,too-many-sta
                 app_id=app_id,
                 app_type=app_type,
                 added_by=flwr_aid,
+                is_hub_app=bool(request.app_spec),
```
panh99: This is actually very accurate, but I wonder if we could add a `is_remote_app` or `is_hub_app` variable and set it to true after line 532? It might be easier to follow.

panh99 (approved):
LGTM!

tanertopal (approved):

PR #7990, opened 2026-08-24
refactor(framework): Simplify agent event publisher lifecycle

panh99:
### Description

RuntimeAgentEvents uses Queue.join/task_done bookkeeping and a separate public flush operation to drain its background publisher. The queue stop sentinel already provides the FIFO ordering needed for shutdown.

### Related issues/PRs

This PR targets the runtime-responses-agentapp-event-control development branch.

### Explanation

- Remove the public flush method and queue completion bookkeeping.
- Drain pending events by enqueueing the stop sentinel and joining the worker.
- Extract batch publication and first-error storage into a private helper.
- Keep batching state local to the worker and flush a partial batch before handling the stop sentinel.
- Preserve bounded, non-blocking event enqueueing and publication error propagation.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Address LLM-reviewer comments, if applicable
- [ ] Make CI checks pass
- [ ] Ping maintainers on Slack

### Any other comments?

Verified with the focused RuntimeAgentEvents test module, Ruff, and mypy.

### DIFF
--- a/framework/py/flwr/supercore/task_process/agent/session.py
+++ b/framework/py/flwr/supercore/task_process/agent/session.py
@@ -103,59 +103,52 @@ class RuntimeAgentEvents(AgentEvents):
             raise RuntimeError("Agent event publish queue is full.") from exc
         self._raise_worker_error()
 
-    def flush(self) -> None:
-        """Wait until all queued events have been persisted."""
-        self._queue.join()
-        self._raise_worker_error()
-
-    def close(self) -> None:
-        """Flush pending events and stop the background publisher."""
+    def close(self, timeout: float | None = None) -> None:
+        """Publish pending events and stop the background publisher."""
         if self._closed:
             self._raise_worker_error()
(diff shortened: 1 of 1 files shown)


framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -97,62 +97,58 @@ def emit(self, event: JSONObject) -> None:
...
             return
-        try:
-            self.flush()
-        finally:
-            self._closed = True
-            self._queue.put(_EVENT_PUBLISH_STOP)
-            self._worker.join()
+
+        self._closed = True
+        self._queue.put_nowait(_EVENT_PUBLISH_STOP)
```
charlesbvll, line 110: Why use `put_nowait()` instead of `put()` here? Shouldn’t we block while closing to ensure the pending events are drained?

PR #7998, opened 2026-08-25
feat(framework): Propagate run source attribution

tanertopal:
## Summary

Follow-up to #7970, which introduced the run-started extension hook.

This PR propagates best-effort run-source attribution through the `StartRun` entry points:

- HTTP requests default to `unknown`; callers may provide the `x-flwr-run-source` header with an explicit source.
- Direct gRPC requests default to `unknown`; callers may provide the `x-flwr-run-source` metadata with an explicit source.
- Automation continues to use `automation`.
- Invalid source values normalize to `unknown`.

The companion Labs changes are in [flwrlabs/labs#780](https://github.com/flwrlabs/labs/pull/780), stacked on #730.

## Attribution contract

This value is for analytics only and is never used for authentication or authorization. A user can only affect the analytics label for their own request, so we assume users will not intentionally corrupt their own analytics and keep this deliberately best effort.

## Validation

- Focused Flower pytest suite
- SuperGrid analytics pytest suite
- Ruff, Black, and isort checks
- TypeScript tests and type-check in the companion Labs changes


framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -127,8 +128,28 @@ def StartRun(
...
+            if isinstance(metadata, (tuple, list))
+            else None
+        )
         return control_handlers.start_run(
-            request, _get_account(), self.linkstate_factory.state(), self.fleet_api_type
+            request,
+            _get_account(),
+            self.linkstate_factory.state(),
+            self.fleet_api_type,
+            source=extensions.resolve_run_start_source(run_source, default="unknown"),
```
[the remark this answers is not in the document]
tanertopal: Unknown is fine until we update the CLI

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -127,8 +128,28 @@ def StartRun(
...
+                (
+                    value
+                    for key, value in metadata
+                    if key == extensions.RUN_SOURCE_METADATA_KEY
+                ),
+                None,
+            )
+            if isinstance(metadata, (tuple, list))
+            else None
+        )
```
```suggestion
        run_source = dict(metadata).get(extensions.RUN_SOURCE_METADATA_KEY)
```

framework/py/flwr/superlink/extensions.py
```diff
@@ -33,9 +33,33 @@
 ]
 RunStartSource = Literal["cli", "web_ui", "automation", "unknown"]
 ResultDeliveryChannel = Literal["logs", "chat"]
+RUN_SOURCE_METADATA_KEY = "x-flwr-run-source"
+_RUN_START_SOURCES = frozenset(get_args(RunStartSource))
 _SGXT_MODULE = "flwr.ee.superlink.extensions"
 
 
+def resolve_run_start_source(
+    value: str | bytes | None, *, default: RunStartSource
```
panh99: Should we default to `"unknown"` here, or remove the `default` argument entirely? Based on the two call sites and the function itself, it seems like resolving a missing value to `"unknown"` is part of the semantics of `resolve_run_start_source`. In that case, we probably don't want callers deciding which source to use when the value isn't provided, right?

```suggestion
    value: str | bytes | None, *, default: RunStartSource = "unknown"
```

framework/py/flwr/superlink/routers/control/router.py
panh99, line 1: Could we create a dependency for `run_source` instead of calling `extensions.resolve_run_start_source(run_source, default="unknown")` directly in this layer? It would be nice to avoid any business logic in this thin wrapper `router.py`.

framework/py/flwr/superlink/servicer/control/control_servicer.py
panh99, line 1: Same as above, could we create a server-side interceptor for this? That said, since we are migrating to HTTP from gRPC, it's prolly fine to keep the current change in `control_servicer.py`.

smoroso (dismissed):

framework/py/flwr/superlink/servicer/control/control_servicer.py
```suggestion
        # Best-effort analytics attribution only; 
        # trust caller-provided labels and default missing sources to unknown.
        metadata = context.invocation_metadata()
        run_source = get_metadata_str(metadata, extensions.RUN_SOURCE_METADATA_KEY)

```

framework/py/flwr/superlink/extensions.py
```diff
@@ -33,9 +33,31 @@
 ]
 RunStartSource = Literal["cli", "web_ui", "automation", "unknown"]
 ResultDeliveryChannel = Literal["logs", "chat"]
+RUN_SOURCE_METADATA_KEY = "x-flwr-run-source"
+_RUN_START_SOURCES = frozenset(get_args(RunStartSource))
 _SGXT_MODULE = "flwr.ee.superlink.extensions"
 
 
+def resolve_run_start_source(value: str | bytes | None) -> RunStartSource:
```
danieljanes: Why does this function need to be in `extensions`?

framework/py/flwr/superlink/servicer/control/control_servicer.py
```diff
@@ -97,6 +97,7 @@
 from flwr.supercore.auth.typing import AccountInfo
 from flwr.supercore.error import ApiErrorCode, FlowerError
 from flwr.supercore.object_store import ObjectStoreFactory
```
panh99, line 99:
```suggestion
from flwr.supercore.object_store import ObjectStoreFactory
from flwr.supercore.utils import get_metadata_str
```

framework/py/flwr/superlink/extensions.py
```diff
@@ -33,9 +33,31 @@
...
+    trusted and invalid values fall back to ``unknown``. This value is not a
+    security or authorization signal.
+    """
+    if value is None:
+        return "unknown"
+    if isinstance(value, bytes):
+        try:
+            value = value.decode("ascii")
+        except UnicodeDecodeError:
+            return "unknown"
```
panh99: I understand that we want source attribution to be best-effort. One thing I’m wondering about is why `value` needs to support `bytes` if we only ever use strings. Do you foresee a case where the string value we set in the header gets converted to ASCII bytes, or where a user might manually craft a request with a bytes value for the `x-flwr-run-source` header?

framework/py/flwr/superlink/extensions.py
```diff
@@ -33,9 +33,31 @@
...
+_RUN_START_SOURCES = frozenset(get_args(RunStartSource))
 _SGXT_MODULE = "flwr.ee.superlink.extensions"
 
 
+def resolve_run_start_source(value: str | bytes | None) -> RunStartSource:
+    """Normalize a caller-provided source label for analytics.
+
+    Source attribution is intentionally best effort. Callers can only affect
+    the analytics label for their own request, so recognized values are
+    trusted and invalid values fall back to ``unknown``. This value is not a
```
panh99: `extensions.py` seems to contain only glue code that uses extensions, while this function doesn’t appear to depend on any extension. Do you think we should place it somewhere else?

framework/py/flwr/superlink/dependencies/run_source.py
panh99, line 1: Nice and clean!

framework/py/flwr/superlink/servicer/control/control_servicer.py
tanertopal: yes - lets keep it as its gone soon..

framework/py/flwr/superlink/routers/control/router.py
tanertopal: done

framework/py/flwr/superlink/run_source.py
```diff
@@ -0,0 +1,43 @@
...
+    tolerant of callers passing raw gRPC metadata values; only ASCII labels
+    are accepted.
+    """
+    if value is None:
+        return "unknown"
+    if isinstance(value, bytes):
+        try:
+            value = value.decode("ascii")
+        except UnicodeDecodeError:
+            return "unknown"
```
panh99: Given that `x-flwr-run-source` is a non-binary gRPC metadata key (gRPC requires `-bin` suffix to use bytes), and FastAPI also exposes HTTP header values as strings, is there a concrete case where this function would actually receive bytes?

Since all producers are/will be Flower-owned code and set this value as a string, I think receiving bytes would more likely indicate that something went wrong in the caller/transport handling. Silently decoding it here could mask that issue and also make the function’s contract less clear. I’d be inclined to keep the input as str | None unless we have an actual caller that can pass bytes.

panh99 (approved):
LGTM!

PR #8005, opened 2026-08-26
feat(framework): Add streaming endpoints

danieljanes:


framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -672,6 +672,16 @@ def stream_logs(
...
+    yield from prepare_stream_logs(request, account, state, is_active)
+
+
+def prepare_stream_logs(
+    request: StreamLogsRequest,
+    account: AccountInfo,
+    state: LinkState,
+    is_active: Callable[[], bool] | None = None,
+) -> Generator[StreamLogsResponse, None, None]:
+    """Validate a log stream eagerly and return its response generator."""
```
panh99: We don't need an extra function for eager validation.
```suggestion
```

The gist is:
```python
def my_rpc(...):
    # Doing validation here
    ...

    # Return a generator instead of using yield
    return _my_rpc(...)

# And the private function can use `yield` as normal
def _my_rpc(...):
    yield res_1
    yield res_2
    ...
```

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -720,6 +741,16 @@ def stream_run_events(
...
+    yield from prepare_stream_run_events(request, account, state, is_active)
+
+
+def prepare_stream_run_events(
+    request: StreamRunEventsRequest,
+    account: AccountInfo,
+    state: LinkState,
+    is_active: Callable[[], bool] | None = None,
+) -> Generator[StreamRunEventsResponse, None, None]:
+    """Validate a run-event stream eagerly and return its response generator."""
```
panh99: Same as above.
```suggestion
```

framework/py/flwr/superlink/dependencies/artifact_provider.py
```diff
@@ -0,0 +1,29 @@
...
+
+from flwr.superlink.artifact_provider import ArtifactProvider
+
+
+def get_artifact_provider(request: Request) -> ArtifactProvider | None:
+    """Return the configured artifact provider, if one is available."""
+    return cast(
+        ArtifactProvider | None,
+        getattr(request.app.state, "artifact_provider", None),
+    )
```
panh99, line 29: Do we need to add `artifact_provider` here? It isn't currently used, so I wonder if we should leave it unimplemented until we have a clearer plan for how to support it properly. Also, this change seems unrelated to the PR title, "Add streaming endpoints."

framework/py/flwr/supercore/protobuf/translation.py
```diff
@@ -217,7 +227,7 @@ def _parse_request(body: bytes, request_type: type[Message]) -> Message:
         return message
 
     @staticmethod
-    def _response_for(result: object) -> Response:
+    def _response_for(result: object, request: Request) -> Response:
```
panh99: Since what we actually need is the `state` object provided by FastAPI, which is available via `request.state`, should we pass `state` directly instead of the entire request? Passing the full request might make the contract here broader and more confusing than necessary.

framework/py/flwr/superlink/servicer/control/control_handlers.py
panh99, line 1: The changes in this file are absolutely necessary. Great catch on this Python behavior.

framework/py/flwr/supercore/protobuf/translation.py
```diff
@@ -229,10 +239,16 @@ def _response_for(result: object) -> Response:
...
+            content = (
+                frame_message(message) for message in cast(Iterable[Message], result)
             )
+            if stream_context := peek_protobuf_stream_context(request):
+                return CancellableProtobufStreamingResponse(
+                    content,
+                    stream_context,
+                    PROTOBUF_STREAM_MEDIA_TYPE,
+                )
+            return StreamingResponse(content, media_type=PROTOBUF_STREAM_MEDIA_TYPE)
```
panh99: Could we always do this? Although not all `unary-stream` RPCs will use `stream_context`, I don’t see a need to maintain two separate paths.

```python
return CancellableProtobufStreamingResponse(
    content,
    get_protobuf_stream_context(request),
    PROTOBUF_STREAM_MEDIA_TYPE,
)
```

panh99 (dismissed):
LGTM!

framework/py/flwr/superlink/dependencies/artifact_provider.py
danieljanes: You're right, moved to https://github.com/flwrlabs/flower/pull/8008

panh99 (approved):
LGTM!

PR #8009, opened 2026-08-26
fix(framework): Pass proper fleet_api_type

danieljanes:
## Description

Propagate the configured Fleet API type through the Control HTTP API.

Previously, the `start-run` and `add-app` HTTP routes passed an empty string to their Control handlers. This differed from the gRPC Control API, which forwards the configured Fleet API type, and could bypass transport-specific FAB validation.

This PR:

- exposes `fleet_api_type` through the FastAPI application state;
- adds a typed FastAPI dependency for resolving it;
- forwards it from the `start-run` and `add-app` HTTP routes;
- defaults to `None` when `flwr.superlink.main:app` is created without a lifespan configuration;
- adds focused tests for state exposure, dependency resolution, and handler forwarding.


framework/py/flwr/superlink/dependencies/fleet_api.py
```diff
@@ -0,0 +1,27 @@
...
+"""FastAPI dependency for the configured Fleet API transport type."""
+
+from typing import Annotated, cast
+
+from fastapi import Depends, Request
+
+
+def get_fleet_api_type(request: Request) -> str | None:
+    """Return the configured Fleet API transport type, if available."""
+    return cast(str | None, getattr(request.app.state, "fleet_api_type", None))
```
jafermarq: wdyt about making the return argument `str` instead of `str | None`?, specially if the plan is to start specifying a non-empty `fleet_api_type`, would it make sense to raise an error if it's `None`?

This is similar to what we already do to, for example, `get_linkstate`: https://github.com/flwrlabs/flower/blob/6b12402780fab5890979fb05c7a558e6e84f2770/framework/py/flwr/superlink/dependencies/linkstate.py#L28

jafermarq (approved):

tanertopal (approved):

PR #8017, opened 2026-08-27
feat(framework): Add Control HTTP auth endpoints

danieljanes:
## Issue

The Control HTTP API did not expose the authentication operations required to migrate the `flwr` CLI from the Control gRPC API.

In particular, HTTP clients need endpoints to initiate the device login flow, poll for credentials, and explicitly refresh expired credentials. Unlike the gRPC protocol, protected HTTP requests carry only a Bearer access token and do not refresh implicitly through response metadata.

## Proposal

Add three protobuf-over-HTTP Control endpoints:

- `POST /v1/control/get-login-details`
- `POST /v1/control/get-auth-tokens`
- `POST /v1/control/refresh-auth-tokens`

The first two reuse the existing Control authentication handlers. The refresh endpoint adapts its request to the existing `ControlAuthnPlugin.refresh_tokens` contract and returns the rotated access and refresh tokens.

Additional behavior:

- Authentication is bypassed only for the exact three POST routes.
- License checks still apply to all three routes.
- Authentication request and response payloads are excluded from Control event logging to prevent credential leakage.
- Successful and error responses include `Cache-Control: no-store` and `Pragma: no-cache`.
- Protected Control endpoints remain unchanged and continue to accept only `Authorization: Bearer <access-token>`.
- Refresh is HTTP-only; the Control gRPC service and its implicit metadata-based refresh flow remain unchanged.

This PR intentionally does not add connector endpoints, migrate the CLI, or refactor the OIDC plugins.

Validation includes focused tests for:

- Device login details and token polling
- Pending and completed polling responses
- Successful token rotation
- Missing, malformed, duplicate, and incomplete refresh results
- Sanitized authentication failures
- Exact method/path authentication exemptions
- License enforcement
- Event-log exclusion
- No-cache headers on successful and error responses
- Existing Control HTTP and gRPC authentication regressions


framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -1238,6 +1243,40 @@ def get_auth_tokens(
...
+
+    tokens, account = authn_plugin.refresh_tokens(
+        [(REFRESH_TOKEN_KEY, request.refresh_token)]
+    )
+    access_token = get_metadata_str(tokens, ACCESS_TOKEN_KEY)
+    refresh_token = get_metadata_str(tokens, REFRESH_TOKEN_KEY)
+    if access_token is None or refresh_token is None or account is None:
+        raise FlowerError(
+            ApiErrorCode.ACCOUNT_AUTHENTICATION_FAILED,
+            "Authentication plugin failed to refresh account tokens.",
```
jafermarq, line 1271: Would it be more informative if we log here the `account`? (this are internal logs to the SuperLink operator)

framework/py/flwr/superlink/routers/control/middlewares.py
```diff
@@ -31,6 +31,33 @@
...
+    }
+)
+
+
+def _is_control_auth_route(request: Request) -> bool:
+    """Return whether the request targets a Control authentication endpoint."""
+    return (request.method, request.url.path) in CONTROL_AUTH_ROUTE_KEYS
+
+
+class ControlAuthResponseMiddleware(BaseHTTPMiddleware):
```
jafermarq, line 48: what's the role of this middleware?

framework/py/flwr/superlink/main.py
```diff
@@ -94,6 +95,7 @@ def _get_middleware() -> list[Middleware]:
     """Return middleware in request execution order, outermost first."""
     return [
         *extensions.get_middleware(),
+        Middleware(ControlAuthResponseMiddleware),
```
jafermarq, line 98: Do we need this after the middlware that does the `http_error_tranlator`? or we don't see that middleware ever raising a `FlowerError`?

framework/py/flwr/superlink/routers/control/middlewares.py
danieljanes: `ControlAuthResponseMiddleware` only adds cache control headers to prevent caching on certain sensitive routes.

framework/py/flwr/superlink/main.py
danieljanes: In the current implementation, `ControlAuthResponseMiddleware` can't raise a `FlowerError`. Are you asking if we see a potential future where it would do that?

PR #8041, opened 2026-08-31
feat(framework): Add configurable SuperExec task polling interval

tanertopal:
## Summary
- Set the default SuperExec pending-task polling interval to 1 second.
- Allow operators to override it with `FLWR_SUPEREXEC_TASK_POLL_INTERVAL`.
- Accept configured intervals from 0.01 through 60 seconds and document the Helm environment configuration.

## Configuration design
`FLWR_SUPEREXEC_TASK_POLL_INTERVAL` is read once at SuperExec startup and converted to a validated float. A dedicated typed `SuperExecSettings` object would be the cleaner boundary if more SuperExec options are added; for this single setting, keeping the parsing local avoids introducing an unnecessary abstraction. The value intentionally does not belong in `ExecutorConfig`, because that config describes TaskExecutor pods rather than the SuperExec process.

If more SuperExec configuration is introduced, the natural next step is to parse the environment at the CLI boundary into `SuperExecSettings` and pass that typed object into `run_superexec`.

## Validation
- `uv run --no-sync --python=3.11.14 ./dev/test.sh false`
- `uv run --no-sync --python=3.11.14 mdformat --check --number docs/source`

All package checks and GitHub CI checks pass.

### DIFF
--- a/framework/py/flwr/supercore/superexec/run_superexec.py
+++ b/framework/py/flwr/supercore/superexec/run_superexec.py
@@ -53,7 +53,7 @@ from .plugin.base_ephemeral_exec_plugin import BaseEphemeralExecPlugin
 _TASK_POLL_INTERVAL_ENV = "FLWR_SUPEREXEC_TASK_POLL_INTERVAL"
 _MIN_TASK_POLL_INTERVAL_SECONDS = 0.01
 _MAX_TASK_POLL_INTERVAL_SECONDS = 60.0
-_DEFAULT_TASK_POLL_INTERVAL_SECONDS = 0.1
+_DEFAULT_TASK_POLL_INTERVAL_SECONDS = 1.0
 
 
 def _get_task_poll_interval() -> float:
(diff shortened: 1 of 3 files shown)


framework/py/flwr/supercore/superexec/run_superexec.py
```diff
@@ -48,6 +50,40 @@
 from .plugin import ExecPlugin
 from .plugin.base_ephemeral_exec_plugin import BaseEphemeralExecPlugin
 
+_TASK_POLL_INTERVAL_ENV = "FLWR_SUPEREXEC_TASK_POLL_INTERVAL"
+_MIN_TASK_POLL_INTERVAL_SECONDS = 0.01
+_MAX_TASK_POLL_INTERVAL_SECONDS = 60.0
+_DEFAULT_TASK_POLL_INTERVAL_SECONDS = 0.1
```
danieljanes:
```suggestion
_DEFAULT_TASK_POLL_INTERVAL_SECONDS = 1.0
```

danieljanes (dismissed):

danieljanes (approved):

PR #8047, opened 2026-09-01
refactor(framework): Refresh expired CLI credentials for HTTP requests

jafermarq:
Extend the CLI HTTP authentication interceptor to refresh expired credentials after a `401 Unauthorized`, store the updated token pair, and retry the original request once. The interceptor remains unused by CLI commands, so current gRPC behavior is unchanged.


framework/py/flwr/cli/cli_account_auth_interceptor.py
```diff
@@ -118,19 +121,46 @@ def intercept_unary_stream(
...
                 {"Authorization": f"Bearer {access_token}"},
             )
+
+        response = call_next(context)
+        refresh_token = get_metadata_str(metadata, REFRESH_TOKEN_KEY)
+
+        # Refresh only after an authentication failure. Refresh requests themselves
+        # are excluded to prevent a failed refresh from starting a recursive loop.
+        if (
+            response.status_code != httpx.codes.UNAUTHORIZED
```
danieljanes, line 154: Not all unauthorized status codes mean that the token is expired. There's an additional header that gets returned in the case where the token is expired.

danieljanes (approved):

PR #8058, opened 2026-09-01
feat(framework): Add input and raw tool calls into task events

danielnugraha:
StartRun records agent.input as a primary-task message, while connector and automation calls use standard function_call/function_call_output items with JSON-encoded arguments and outputs.

### DIFF
--- a/framework/py/flwr/supercore/task_process/agent/session.py
+++ b/framework/py/flwr/supercore/task_process/agent/session.py
@@ -22,7 +22,7 @@ import time
 from collections.abc import Sequence
 from queue import Empty, Queue
 from threading import Lock, Thread
-from typing import Literal, cast
+from typing import cast
 
 from google.protobuf.json_format import ParseDict
 
@@ -51,9 +51,7 @@ from flwr.supercore.task_process.connector.automation import START_AUTOMATION_TO
 from flwr.supercore.task_process.connector.registry import (
     get_connector_ref,
     get_connector_tools,
-    has_builtin_connector,
 )
-from flwr.supercore.task_process.connector.web_fetch import WEB_FETCH_CONNECTOR_NAME
 from flwr.supercore.typing import JSONObject, JSONValue
 from flwr.supercore.utils import strict_json_dumps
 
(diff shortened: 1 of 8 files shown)


jafermarq (approved):

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -314,86 +312,58 @@ def call_connector_with_events(
...
-            if status == "completed":
-                event["output"] = output
-            elif status == "failed" and message is not None:
-                event["error"] = {"code": "connector_error", "message": message}
-
-            return [event]
-
-        self.append_and_push_run_events(connector_event("started"))
+        name = name.strip().lower()
+        function_call: JSONObject = {
```
smoroso, line 316: - What: [P1] Keep emitting the tool activity events consumed by Flower Chat.
- Problem Example:
  ```python
  function_call = {"type": "function_call", ...}
  self.append_and_push_run_events([function_call])
  ```
- Why: Production reproduction: **Yes.** Connector calls now stream only raw `function_call` and `function_call_output` items. The production CLI chat consumer in `framework/py/flwr/cli/chat/chat_app.py` dispatches tool details only for `response.tool_call.started` and `response.tool_call.completed`, and its web-search handler requires `connector_ref`, `tool_call_id`, and an object-valued `output`. It ignores the new items, so web-search progress and results disappear from `flwr chat` even though the connector still runs. This is introduced by this PR, is not test-only, and is not identified as stacked work.
- Fix: Update Flower Chat in this PR to consume the same raw function-call task events as the website and correlate them by `call_id`. This keeps one event contract for both clients and avoids emitting duplicate legacy and raw UI events.
- Solution Example:
  ```python
  if event_type == "function_call":
      tool_call_id = cast(str, payload["call_id"])
  elif event_type == "function_call_output":
      tool_call_id = cast(str, payload["call_id"])

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -314,86 +312,58 @@ def call_connector_with_events(
         self, *, name: str, call_id: str, arguments: JSONObject
```
smoroso, line 312: - What: [P1] Do not append the model's function call to context a second time.
- Problem Example:
  ```python
  append_items(self._context, cast(list[JSONObject], output))
  # Later, connectors.call(tool_call):
  self.append_and_push_run_events([function_call])
  ```
- Why: Production reproduction: **Yes.** `agent.responses.create()` appends every model output item, including its `function_call`, to the run context. The normal subsequent `agent.connectors.call(tool_call)` path reconstructs and appends that same call again before adding its output. The persisted Open Responses history therefore contains two function calls with the same `call_id` but only one output. Sending that history into the next model request can make the conversation invalid or cause the model to reason over a duplicated tool invocation. This is introduced by this PR, is not a fixture issue, and is not identified as stacked work.
- Fix: Stream the already-persisted function call without appending it again; append only the newly produced `function_call_output`. If callers can invoke connectors without a preceding model response, make persistence explicit at the caller boundary rather than unconditionally duplicating the common path.
- Solution Example:
  ```python
  self.push_run_events([function_call])
  output = self.create_connector_response(
      name=name, call_id=call_id, arguments=arguments
  )
  self.append_and_push_run_events([output_item])

danielnugraha: We are not using Context for event log anymore. So this will be removed in the cleanup PR.

framework/py/flwr/supercore/task_process/agent/session.py
danielnugraha: We will update the flwr chat in the following PR.

smoroso (approved):

danieljanes (approved):

PR #8068, opened 2026-09-02
feat(framework): Ensure hub apps use latest version

danielnugraha:
Implemented improvement to differentiate hub apps and ensure it always uses latest version.

  - Hub apps retain is_hub_app=true but no federation-level fab_hash.
  - Custom ServerApps and AgentApps remain pinned to their stored FAB hashes.
  - AddApp stores only Hub metadata.
  - Every Hub run fetches the latest FAB; each run still retains its exact execution snapshot.
  - Added an Alembic migration making federation_app.fab_hash nullable and clearing existing Hub hashes.
  - Updated schema documentation and coverage for both app types.


framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -472,6 +472,31 @@ def validate_run_connector_refs(
...
+
+
+def _get_start_run_app_id(request: StartRunRequest) -> str | None:
+    """Return the app ID declared by a StartRun request, if identifiable."""
+    if request.app_spec:
+        try:
+            app_id, _ = parse_app_spec(request.app_spec)
+            return app_id
+        except ValueError:
+            return None
```
jafermarq, line 483: I think only this `if` statement would be evaluate for apps in the hub. If so, would it be better to move this directly into `_is_hub_app` and remove the `if request.fab.content` below?

jafermarq (approved):

danieljanes (approved):

PR #8071, opened 2026-09-03
fix(framework): Preserve nested config keys

PLRpower:
### Description

`fuse_dicts` is intended to recursively merge configuration overrides.

When an override contains a nested dictionary, the function computes the recursive merge but immediately overwrites that result with the override dictionary. As a result, a partial nested override discards sibling keys from the base configuration.

### Related issues/PRs

None.

### Explanation

Previously, `fuse_dicts` attempted to recursively merge nested dictionary overrides:

```python
fused_dict[key] = fuse_dicts(main_dict[key], value)
```

However, the next line immediately overwrote that recursive result:

```python
fused_dict[key] = value
```

This caused partial nested overrides to discard sibling keys from the base configuration.

For example, merging:

```python
{"d": {"e": 4, "f": 5}}
```

with:

```python
{"d": {"e": 6}}
```

previously returned:

```python
{"d": {"e": 6}}
```

The fix assigns `value` only when it is not a dictionary. Nested dictionaries now retain the recursive merge result:

```python
{"d": {"e": 6, "f": 5}}
```

A regression test covers this behavior.

### Checklist

- [x] Write tests
- [x] Update documentation (not needed; no public API change)
- [x] Make CI checks pass
- [x] Ping maintainers on Slack (channel `#contributions`)

### Any other comments?

Tested locally:

```bash
uv run --no-sync --python=3.11.14 python -m pytest py/flwr/common/config_test.py -k fuse_dicts
```

**Result:** `2 passed`


jafermarq (approved):
Hey @PLRpower , many thanks for this fix!! 🚀

tanertopal (approved):

PR #8082, opened 2026-09-04
feat(framework): Add federation apps cache

danielnugraha:
## Summary

This PR adds caching for Flower Hub apps in `CoreState`, so SuperLink can reuse a downloaded FAB instead of fetching it from Flower Hub every time an unversioned Hub app is started.

When a Hub app is added or fetched, its FAB and verification metadata are stored locally and the federation-app association keeps the corresponding `fab_hash`. `StartRun` can then use the cached FAB directly for unversioned Hub app requests. If the cached entry is older than five minutes, the run still starts from the cached FAB and a background refresh fetches the latest version from Hub.

The refresh path validates the downloaded FAB and updates the federation-app association only if it still points to the FAB that triggered the refresh, avoiding an older refresh overwriting a newer cache entry.

This also:

- adds `get_hub_app` and `update_hub_app` to `CoreState`, with both in-memory and SQL implementations;
- makes `store_app` always store a FAB for federation apps;
- tracks `updated_at` for federation-app associations and makes `fab_hash` non-null in the schema;
- adds the corresponding Alembic migration and updates the schema documentation;
- updates and extends tests for storing, reusing, and refreshing cached Hub apps.

Overall, this avoids repeated Flower Hub fetches on the run path while still allowing cached unversioned Hub apps to be refreshed periodically.


framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -719,6 +768,26 @@ def start_run(  # pylint: disable=too-many-branches,too-many-locals,too-many-sta
...
+                            state,
+                            federation_id,
+                            app_id,
+                            cached_fab.hash_str,
+                            fleet_api_type,
+                        ),
+                        daemon=True,
+                    )
+                    _hub_app_refresh_threads[refresh_key] = refresh_thread
+                    refresh_thread.start()
```
panh99: Could we move this logic inside `_refresh_hub_app` and rename the function to `_refresh_hub_app_in_thread`?

panh99 (approved):
LGTM!

PR #8096, opened 2026-09-07
feat(framework): Add stdin token handoff

charlesbvll:
Add the ability to pass a token to an agentapp process (for warm executor)


framework/py/flwr/supercore/cli/flwr_agentapp.py
```diff
@@ -16,21 +16,33 @@
...
 
 from flwr.common.args import add_args_flwr_app_common, try_obtain_flwr_app_token
+from flwr.common.constant import FLWR_TASK_TOKEN_LENGTH
 from flwr.supercore import log
 from flwr.supercore.constant import SUPERLINK_DEFAULT_CLIENT_ADDRESS
 from flwr.supercore.logger import mirror_output_to_queue, restore_output
 from flwr.supercore.task_process import run_agentapp
 
+_TOKEN_STDIN_ACKNOWLEDGEMENT = "FLWR_AGENTAPP_TOKEN_ACCEPTED"
+_TASK_TOKEN_PATTERN = re.compile(rf"[0-9a-f]{{{FLWR_TASK_TOKEN_LENGTH * 2}}}")
```
danielnugraha, line 35: Just curious: Why do we expects a task token to be like this?

charlesbvll: I made the stdin-only path strict so malformed or repeated handoff data fails closed (the tokens are generated with `secrets.token_hex(FLWR_TASK_TOKEN_LENGTH)`, which is at most `FLWR_TASK_TOKEN_LENGTH * 2` characters)

charlesbvll: it does mean that if we update the token format, this will also have to be updated. Do you want to make it less strict?

danielnugraha: This looks good. We don't need to change anything.

danielnugraha (dismissed):

framework/py/flwr/supercore/cli/flwr_agentapp.py
```diff
@@ -77,5 +89,30 @@ def _parse_args_run_flwr_agentapp() -> argparse.ArgumentParser:
...
+
+def _try_obtain_agentapp_token(args: argparse.Namespace) -> str:
+    """Return the AgentApp token from an existing source or private stdin mode."""
+    if not cast(bool, getattr(args, "token_stdin", False)):
+        return try_obtain_flwr_app_token(args)
+
+    try:
+        token_input = (
+            cast(BufferedReader, sys.stdin.buffer)
+            .read1(_TOKEN_STDIN_READ_LIMIT)
```
tanertopal: [P2] Accumulate stdin reads until the token is complete

`read1()` returns available bytes without waiting for the complete token. If the sender delivers a valid token across multiple writes, this code validates the first fragment and closes stdin prematurely.

I reproduced this with a real pipe on Python 3.11.14: writing the first 128 characters, then the remaining 128 characters plus a newline after a short delay, causes the receiver to exit with the invalid-token error and the sender to receive `BrokenPipeError`. The existing 22 targeted tests pass, but the successful handoff test writes the entire token before invoking the reader.

Please accumulate bounded reads until the token frame is complete, without requiring EOF, and add a split-write regression test.

framework/py/flwr/supercore/cli/flwr_agentapp.py
```diff
@@ -77,5 +89,33 @@ def _parse_args_run_flwr_agentapp() -> argparse.ArgumentParser:
...
+        token_input = token_bytes.decode("ascii")
+    except (AttributeError, OSError, UnicodeError):
+        sys.exit("Standard input does not contain exactly one valid task token.")
+    finally:
+        try:
+            sys.stdin.close()
+        except OSError:
+            pass
+
+    token = token_input.removesuffix("\n")
```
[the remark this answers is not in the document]
charlesbvll: This stdin mode is private to the Kubernetes warm-executor handoff. The dispatcher sends an explicit `\n` through `write_stdin`, so the payload is LF-delimited even if the dispatcher itself runs on Windows

tanertopal (approved):

tanertopal (approved):

PR #8098, opened 2026-09-07
feat(framework): Dispatch AgentApp tasks to warm executors

charlesbvll:
Add the first active warm-executor dispatch path for AgentApp tasks.

SuperExec now owns static pools of compatible, idle TaskExecutor Pods and, for a matching AgentApp task, starts a fresh one-task `flwr-agentapp` child through `pods/exec`.

- Share AgentApp warm pools across FABs using the configured TaskExecutor image and Pod configuration
- The fresh child fetches and installs its FAB after token handoff
- During reconciliation, SuperExec retires and replaces warm Pods whose configuration SHA-256 does not match the current configuration
- Send the short-lived task token only through child stdin and require token-free acknowledgement
- Fall back to the existing cold Kubernetes path only before token delivery is attempted
- Return an unknown launch outcome after ambiguous delivery rather than risking duplicate execution
- Retire consumed Pods, replenish idle capacity, and clean up idle owned Pods on shutdown

The pool requires an explicit owner identity. Deployments must run one active SuperExec replica per owner unless leader election is added.

## Scope

This PR is AgentApp-only. It does not add Model or Connector pools, persistent child processes, dynamic scaling, SuperLink pull endpoints, Runtime API changes, chart/RBAC changes, or deployment enablement.

The normal cold TaskExecutor Pod and credential-Secret path remains unchanged for non-matching or unavailable tasks. Warm dispatch also falls back cold when explicit Runtime CA material cannot be delivered safely.


framework/py/flwr/supercore/superexec/executor/kubernetes_executor.py
```diff
@@ -16,8 +16,11 @@
```
danielnugraha, line 16: Can we move the logic of warm agentapp or executor out? Ideally we only keep the integration code here so it's readable.

danielnugraha (approved):

tanertopal (approved):

PR #8101, opened 2026-09-07
feat(framework): Add run info in GetRunSeries

danielnugraha:
## Summary
Add run details to the GetRunSeries response so callers can access lifecycle timestamps for every run in a series.

  - Add repeated Run runs to GetRunSeriesResponse
  - Retrieve runs through the existing get_run_info API
  - Preserve the ordering defined by RunSeries.run_ids
  - Serialize runs using the existing run_to_proto helper
  - Regenerate Python protobuf bindings
  - Extend Control servicer test coverage


framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -1205,9 +1205,19 @@ def get_run_series(
...
     series_context = state.get_run_series_context(request.series_id)
+    run_series = series_matches[0]
+    runs_by_id = {
+        run.run_id: run for run in state.get_run_info(run_ids=run_series.run_ids)
+    }
+    runs = [
+        run_to_proto(run)
+        for run_id in run_series.run_ids
+        if (run := runs_by_id.get(run_id)) is not None
+    ]
```
panh99: Could we simply do this instead, or would it introduce any risks? The extra steps here seem mainly intended to guard against `state.get_run_info` returning unrelated run IDs. I think we can reasonably treat `state.get_run_info` as a trusted source, tbh.

```suggestion
    runs = [
        run_to_proto(run)
        for run in state.get_run_info(run_ids=run_series.run_ids)
    ]
```

panh99 (approved):
LGTM!

danieljanes (approved):

PR #8104, opened 2026-09-07
feat(framework): Remove appending to context automatically

danielnugraha:
Removed all automatic AgentApp context appends:

  - Initial agent.input is validated but no longer appended.
  - responses.create() no longer appends model output.
  - Connector calls and outputs are emitted as run events without modifying context.


panh99 (dismissed):
LGTM!

panh99 (approved):
LGTM!

danieljanes (approved):

PR #8106, opened 2026-09-07
break(framework): Make all APIs share SSL settings by default

jafermarq:
This PR makes ``--appio-ssl-certfile``, ``--appio-ssl-keyfile``, and ``--appio-ssl-ca-certfile`` obsolete and launching a `SuperLink` with them will terminate it immediately.

### DIFF
--- a/framework/py/flwr/superlink/cli/flower_superlink.py
+++ b/framework/py/flwr/superlink/cli/flower_superlink.py
@@ -23,7 +23,6 @@ import sys
 import threading
 from collections.abc import Sequence
 from logging import INFO, WARN
-from pathlib import Path
 from time import sleep
 from typing import cast
 
@@ -84,7 +83,6 @@ from flwr.supercore.object_store import ObjectStoreFactory
 from flwr.supercore.telemetry import EventType, event
 from flwr.supercore.tls import (
     get_client_tls_args,
-    try_obtain_optional_runtime_server_certificates,
 )
 from flwr.supercore.update_check import warn_if_flwr_update_available
 from flwr.supercore.utils import get_popen_detach_kwargs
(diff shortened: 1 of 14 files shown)


framework/py/flwr/superlink/cli/flower_superlink.py
```diff
@@ -441,15 +437,19 @@ def _parse_superlink_lifespan_config() -> SuperLinkLifespanConfig:
...
+            str(Path(args.ssl_certfile).expanduser())
+            if certificates is not None
             else None
         ),
         runtime_ssl_keyfile=(
-            str(Path(args.runtime_ssl_keyfile).expanduser())
-            if runtime_certificates is not None
+            str(Path(args.ssl_keyfile).expanduser())
+            if certificates is not None
             else None
```
panh99: Do we still need to keep these attributes of `SuperLinkLifespanConfig`?

framework/py/flwr/superlink/cli/flower_superlink_test.py
```diff
@@ -144,22 +201,19 @@ def test_parse_superlink_log_rotation_args_custom_values() -> None:
...
+@pytest.mark.parametrize(
+    "flag",
+    ["--appio-ssl-certfile", "--appio-ssl-keyfile", "--appio-ssl-ca-certfile"],
+)
+def test_parse_superlink_appio_tls_args_rejected(
+    flag: str, capsys: pytest.CaptureFixture[str]
+) -> None:
+    """SuperLink should explain how to replace removed AppIO TLS args."""
+    with pytest.raises(SystemExit) as exc_info:
+        _parse_args_run_superlink().parse_args([flag, "certificate.pem"])
```
panh99: I wonder if it would be better to remove this unit test. My main concern is avoiding unit tests gradually turning into a kind of changelog that captures every API change.

Since we're essentially deprecating these arguments, I don't think we necessarily need a dedicated test to verify that they're rejected. Unless there's a specific compatibility guarantee we want to preserve here, this feels like something we could safely remove. Wdyt?

panh99 (approved):
LGTM!

danieljanes (approved):

PR #8108, opened 2026-09-07
feat(framework:skip): Add timing to uv sync

charlesbvll:
Add a timing log that says `uv sync took 3.1 seconds.` after `uv sync` step


tanertopal (approved):
LGTM!

PR #8110, opened 2026-09-08
feat(framework): Add get_trace to agentapp API

danielnugraha:
Add agent.events.get_trace() to retrieve persisted events from all runs in the current AgentApp run series.

  - Adds a task-authenticated GetRunSeriesEvents Runtime API endpoint.
  - Derives the run series from the authenticated task instead of accepting a caller-provided series ID.
  - Restricts access to the primary AgentApp task.
  - Returns events belonging to the primary task of each run in the series.
  - Exposes the events through agent.events.get_trace().
  - Parses each event’s JSON payload before returning it to the AgentApp.
  - Rejects this AgentApp-only endpoint on SuperNode.
  - Adds tests for authorization, event filtering, Runtime routing/client behavior, and AgentApp event conversion.


framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
```diff
@@ -81,6 +83,43 @@
 )
 
 
+def get_run_series_events(
+    request: GetRunSeriesEventsRequest,
+    state: LinkState,
+    task: Task,
+) -> GetRunSeriesEventsResponse:
```
panh99, line 90: I wonder if this implementation is a bit overly defensive. Is there any downside to simply returning the events for the run series that the authenticated task belongs to, without checking the task type or whether it is the primary task?

We do have authorization checks for more sensitive RPCs, such as task creation, but these additional restrictions here seem unnecessary to me.

framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
```diff
@@ -81,6 +83,27 @@
...
+    run = state.get_run_info(run_ids=[task.run_id])[0]
+    series = state.get_run_series(series_ids=[run.series_id])[0]
+    series_runs = state.get_run_info(run_ids=series.run_ids)
+    primary_task_ids = [
+        run.primary_task_id for run in series_runs if run.primary_task_id is not None
+    ]
+    events = state.get_task_events(
+        run_ids=series.run_ids,
+        task_ids=primary_task_ids,
+    )
```
panh99: May I know why we exclude events from non-primary tasks?

panh99: It seems that `task_ids` is a complete filter for this, right?
```suggestion
    events = state.get_task_events(task_ids=primary_task_ids)
```

FANTOME-PAN (approved):

panh99 (approved):

PR #8115, opened 2026-09-08
refactor(framework): Share Runtime API router between SuperLink and SuperNode

jafermarq:
Use one Runtime API router from `flwr.supercore` for both SuperLink and SuperNode.
- Configure component-specific handlers and version validation when mounting it.
- Remove the duplicated SuperNode router.
- Preserve existing ClientApp and ServerApp behavior.
Builds on #8116, which unified Runtime state and authentication dependencies.

> [!NOTE]
> There are 4 endpoints with with component-specific behavior receive the Runtime handlers configured by SuperLink or SuperNode; shared behavior calls supercore handlers directly.The casts provide static return types because the configured handler module is resolved dynamically


panh99 (approved):
LGTM!

PR #8120, opened 2026-09-09
refactor(framework): Handle Type Runtime API dependency

jafermarq:
Add a shared `RuntimeHandlers` protocol so `mypy` validates the SuperLink and SuperNode handler implementations. Remove the casts previously required in the shared Runtime API router.


panh99 (approved):
LGTM!

PR #8123, opened 2026-09-09
feat(framework): Generate conversation title with llm

danielnugraha:
## Summary

  Generate concise RunSeries descriptions for AgentApp conversations using a model request.

  Title generation:

  - Runs concurrently with the first AgentApp run using a background Future
  - Uses openai/gpt-5-nano
  - Falls back to the first four prompt words if generation fails
  - Normalizes and limits titles to the existing description length
  - Does not run for subsequent conversations in the same RunSeries


framework/proto/flwr/proto/runtime.proto
```diff
@@ -188,6 +189,7 @@ message PushTaskOutputRequest {
   string details = 3;
   // Simulation Runtime usage only.
   optional double clientapp_runtime = 4;
+  optional string series_description = 5;
```
panh99: Should we also put the description in a stream event? Otherwise, the user won't be able to see the conversation title until the full response is generated, no?

framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
```diff
@@ -257,12 +257,21 @@ def pull_task_input(
         series_context = state.get_run_series_context(run.series_id)
     if run and fab and series_context and state.activate_task(task.task_id):
         log(INFO, "Started task %d of run %d", task.task_id, run_id)
+        series = state.get_run_series(series_ids=[run.series_id])
+        should_generate_series_description = bool(
+            task.type == TaskType.AGENT_APP
+            and run.primary_task_id == task.task_id
+            and series
+            and series[0].run_ids
+            and series[0].run_ids[0] == run_id
```
panh99: QQ: Is it possible to have a run without a run series, or a run series with no runs? IIRC, one invariant is that the run series is created together with the run, right?

If so, could we simplify this to:

```suggestion
            and series[0].run_ids[0] == run_id
```

framework/py/flwr/supercore/task_process/agent/run_agentapp.py
```diff
@@ -258,8 +265,28 @@ def on_exit() -> None:
...
+                if title_future.done():
+                    try:
+                        series_description = title_future.result()
+                    except Exception as ex:  # pylint: disable=broad-exception-caught
+                        log(ERROR, "Failed to resolve RunSeries description: %s", ex)
+                else:
+                    title_future.cancel()
+            if title_executor is not None:
+                title_executor.shutdown(wait=False, cancel_futures=True)
+            agent_events.close()
```
panh99: Could we hide this concurrency complexity behind a small utility instead of adding it directly to `run_agentapp`?

Also, do we need the cancellation/defensive handling here at all? Since we need the conversation title eventually anyway, I think we could simply let the title-generation thread complete and collect its result after `agent_app` finishes. In practice, the title generation should usually finish well before the AgentApp, and even if it doesn't, waiting for it at the end shouldn't noticeably affect the UX: the AgentApp response is streamed, so the user isn't really waiting on this final task completion.

For example, could we make the call site roughly:

```python
title_future = generate_series_description_async(...)
agent_app(agent=agent, context=context)
series_description = title_future.result()
```

with `generate_series_description_async(...)` owning the executor and all error handling? Its contract could be that the returned future never fails: if title generation errors, it returns a fallback title such as `"Run series xxx"` instead. That way, we don't need the `done()` / `cancel()` / `shutdown(wait=False)` branching or another `try/except` in the AgentApp main loop.

In particular, `cancel()` won't stop the future once the worker has already started running, so the current cancellation path seems to add complexity without buying us much.

framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
```diff
@@ -287,14 +296,27 @@ def push_task_output(
...
+                    and request.HasField("series_description")
+                    and not state.set_run_series_description(
+                        run.series_id, request.series_description
+                    )
+                ):
+                    log(
+                        ERROR,
+                        "Failed to set description for RunSeries %d.",
+                        run.series_id,
+                    )
```
panh99: Could we avoid extending this condition with `or request.HasField("series_description")`? It makes the flow a bit harder to follow, since `context` and `series_description` are two independent outputs with different semantics.

I think it'd be cleaner to resolve/validate the primary run once and then handle the fields independently, roughly:

```python
runs = state.get_run_info(run_ids=[run_id])
run = runs[0] if runs else None

if run and run.series_id and run.primary_task_id == task.task_id:
    if request.HasField("context"):
        ...

    if task.type == TaskType.AGENT_APP and request.HasField("series_description"):
        ...
```

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -698,6 +700,11 @@ def start_run(  # pylint: disable=too-many-branches,too-many-locals,too-many-sta
 
         run = state.get_run_info(run_ids=[run_id])[0]
         series_id = run.series_id
+        if series_description and isinstance(agent_input, str) and series_id:
+            try:
+                start_title_generation(state, series_id, agent_input)
+            except Exception as ex:  # pylint: disable=broad-exception-caught
+                log(ERROR, "Failed to start RunSeries title generation: %s", ex)
```
panh99: Can we move the `try-except` logic into `start_title_generation`, so we can simply call the function here?

```suggestion
            start_title_generation(state, series_id, agent_input)
```

panh99 (dismissed):
LGTM!

framework/py/flwr/superlink/servicer/runtime/runtime_handlers_test.py
```diff
@@ -501,12 +501,14 @@ def test_push_task_output_stores_simulation_runtime(self) -> None:
         )
 
         # Execute
-        response = runtime_handlers.push_task_output(
-            request, self.state, self._auth_task
-        )
+        with patch.object(self.state, "get_run_info") as get_run_info:
+            response = runtime_handlers.push_task_output(
+                request, self.state, self._auth_task
+            )
```
panh99:
```suggestion
        response = runtime_handlers.push_task_output(
            request, self.state, self._auth_task
        )
```

framework/py/flwr/superlink/servicer/runtime/runtime_handlers_test.py
```diff
@@ -501,12 +501,14 @@ def test_push_task_output_stores_simulation_runtime(self) -> None:
...
 
         # Assert
         assert isinstance(response, PushTaskOutputResponse)
+        get_run_info.assert_not_called()
```
panh99:
```suggestion
```

panh99 (approved):
LGTM!

PR #8128, opened 2026-09-10
feat(framework): Add RunSeries description state method

danielnugraha:
## Summary

  Adds CoreState.set_run_series_description method for updating a RunSeries description.

  - Implements the method for SQL and in-memory state.
  - Trims surrounding whitespace.
  - Rejects empty descriptions and unknown series IDs.
  - Updates the RunSeries updated_at timestamp.
  - Adds shared CoreState tests for the new behavior.


framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -383,6 +384,23 @@ def get_run_series(  # pylint: disable=too-many-arguments
...
+        series_id : int
+            The ID of the RunSeries to update.
+        description : str
+            The non-empty description to store.
+
+        Returns
+        -------
+        bool
+            Whether the description was stored.
+        """
```
panh99: Should we make this function simply return `None`, similar to `set_run_series_context`? In practice, we don’t seem to rely on its return value to check whether the run series exists or whether the description is empty. It would help to keep this function thin.

framework/py/flwr/supercore/corestate/corestate.py
```diff
@@ -383,6 +384,18 @@ def get_run_series(  # pylint: disable=too-many-arguments
             RunSeries records ordered by `updated_at` descending.
         """
 
+    @abstractmethod
+    def set_run_series_description(self, series_id: int, description: str) -> None:
+        """Set the description of an existing RunSeries.
+
```
panh99, line 390:
```suggestion
        """Set the description of an existing RunSeries.

        Empty descriptions are ignored and do not update the RunSeries.

```

panh99 (approved):
LGTM!

PR #8143, opened 2026-09-14
feat(framework): Add Grid messaging tools to AgentApp

jafermarq:
Expose `get_nodes`, `push_messages`, and `pull_messages` through `AgentSession.grid`, enabling agents to discover SuperNodes and exchange JSON messages. Add AgentGrid payload constants and authorize AgentApp tasks to use the required Grid runtime APIs.


framework/py/flwr/supercore/task_process/agent/grid.py
```diff
@@ -0,0 +1,232 @@
...
+                    "minimum": 1,
+                    "description": "Optional maximum number of SuperNodes to return.",
+                }
+            },
+        ),
+        function_tool(
+            "push_messages",
+            (
+                "Send messages to SuperNodes and return one result per message in "
+                "the same order. Pass accepted message IDs to pull_messages if "
```
panh99, line 66: The `push_messages` tool call won't return reply messages, no? What does "return one result per message" mean here?

jafermarq: because the return of the tool is a `JSONObject`, by "one result per message" means one entry in the `"results"` dictionary that `JSONObject` carries. We could return a list of `JSONObjects` if you think that's better

framework/py/flwr/supercore/task_process/agent/grid.py
```diff
@@ -0,0 +1,232 @@
...
+
+_GRID_TOOL_NAMES = {"get_nodes", "push_messages", "pull_messages"}
+
+
+def _grid_tools() -> list[JSONObject]:
+    """Return model-facing federation Grid tool schemas."""
+    return [
+        function_tool(
+            "get_nodes",
+            "Return all available SuperNodes, or a random sample if requested.",
```
panh99: Do we provide context like what's SuperNode/SuperLink somewhere to the agent?

framework/py/flwr/supercore/task_process/agent/grid.py
```diff
@@ -0,0 +1,232 @@
...
+            properties={
+                "messages": {
+                    "type": "array",
+                    "minItems": 1,
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "dst_node_id": string_property(
+                                "Destination SuperNode ID as a decimal string."
+                            ),
```
panh99, line 79: Curious why this is a string property.

framework/py/flwr/supercore/task_process/agent/grid.py
panh99: I see, thanks for the clarification! Just one quick question: do we need to tell the agent what the returned result represents? For example, should we specify the schema of the returned JSON object and mention that it contains the message IDs of the pushed messages, which can then be used to pull the corresponding replies?

jafermarq: i think we don't need to specify is a `JSON` object. the tool already says that `Pass accepted message IDs to pull_message...`

framework/py/flwr/supercore/task_process/agent/grid.py
jafermarq: done

framework/py/flwr/supercore/task_process/agent/grid.py
```diff
@@ -0,0 +1,237 @@
...
+                                }
+                            )
+                        }
+                    ),
+                    dst_node_id=int(cast(str, item["dst_node_id"])),
+                    message_type="query",  # Replace with an AgentGrid message type.
+                    group_id="",
+                    ttl=ttl,
+                )
+            )
```
panh99: Could we create the `ConfigRecord` first, then the `Message`, and finally append the message to `outgoing`? The current nesting feels a bit like a tower 🗼  Wdyt?

```suggestion
            payload_json = strict_json_dumps(payload, compact=True)
            config_record = ConfigRecord(
                {AGENT_GRID_MESSAGE_PAYLOAD_JSON_KEY: payload_json}
            )
            content = RecordDict(
                {AGENT_GRID_MESSAGE_PAYLOAD_RECORD_KEY: config_record}
            )

            message = Message(
                content,
                dst_node_id=int(cast(str, item["dst_node_id"])),
                message_type="query",  # Replace with an AgentGrid message type.
                group_id="",
                ttl=ttl,
            )
            outgoing.append(message)
```

framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
panh99: Do we actually need this function? It seems more like a check enforcing task-type-based access control than something that `_get_authenticated_grid_run_id` describes.

Should we rename it to something like `_check_task_type` or `_require_serverapp_or_agentapp`, and then access the run ID directly via `task.run_id`?

That said, I’d also be happy to remove this check entirely. I don’t see much benefit in keeping this access control: `flwr-model` and `flwr-connector` cannot load user code anyway, and while we previously wanted to prevent AgentApps from accessing nodes, we’re intentionally removing that restriction now.

framework/py/flwr/supercore/task_process/agent/grid.py
```diff
@@ -0,0 +1,237 @@
...
+            if not pending or remaining <= 0:
+                break
+            time.sleep(min(0.25, remaining))
+        return {
+            "messages": [
+                cast(JSONObject, MessageToDict(message_to_proto(message)))
+                for message in replies
+            ],
+            "pending_message_ids": sorted(pending),
+        }
```
panh99, line 343: Could we return a simpler agent-facing response here instead of serializing the full protobuf `Message`? Since this API is effectively for agent-to-agent messaging, I think returning a JSONObject including fields like `src_node_id`, `reply_to_message_id`, and a plain string `message`/`response` would be easier for the model to consume without having to learn about Flower infra details.

framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
jafermarq: looking into removing it here: https://github.com/flwrlabs/flower/pull/8170

panh99 (dismissed):
LGTM!

framework/py/flwr/superlink/servicer/runtime/runtime_handlers.py
jafermarq: this is now gone

panh99 (approved):
LGTM!

PR #8144, opened 2026-09-14
refactor(framework): Scope runtime message pulls to authenticated run

jafermarq:
Validates requested message IDs against the authenticated run before retrieving replies, preventing cross-run messages from being consumed or deleted. Adds regression coverage for cross-run pull attempts.


panh99 (approved):
LGTM!

PR #8148, opened 2026-09-14
feat(framework): Add connector support to flwr chat CLI

danielnugraha:
## Summary

  - Add /connector to select connected connectors in flwr chat.
  - Support additive selection and display selected connectors in the context bar.
  - Pass selected connector refs to chat runs.
  - Add /connector clear to clear selections.
  - Direct connector setup to WebUI; OAuth is not handled by the CLI.


framework/py/flwr/cli/chat/chat_connector.py
```diff
@@ -0,0 +1,78 @@
...
+    with flwr_cli_exc_handler():
+        response = stub.ListConnectors(ListConnectorsRequest(federation=federation))
+    return [connector for connector in response.connectors if connector.connected]
+
+
+def complete_connectors(
+    query: str, connectors: list[Connector]
+) -> Iterable[Completion]:
+    """Yield connected connectors matching a completion query."""
+    ref_width = max(
```
dstripelis, line 45: When connectors is empty, the expansion produces `max(len(CHAT_CONNECTOR_CLEAR))`, which will be interpreted as a single integer argument and not an iterable. Better to use an explicit iterable.

framework/py/flwr/cli/chat/chat_app.py
```diff
@@ -501,8 +538,62 @@ def _handle_command(self, event: KeyPressEvent, prompt: str) -> bool:
...
+            event.app.invalidate()
+            return True
+
+        try:
+            connector = select_connector(prompt, connectors)
+        except click.ClickException as exc:
+            self._append_transcript("class:error", f"{exc.format_message()}\n\n")
+            return True
+
+        if connector.connector_ref not in self.connector_refs:
```
dstripelis, line 592: Refreshed connector state does not remove disconnected selections. If a selected connector is disconnected in WebUI, /connector fetches the current list but leaves its reference in `self.connector_refs`, see also line [834](https://github.com/flwrlabs/flower/pull/8148/changes#diff-5ed4e7dd5e61496ebf536761588f2e2719d41ed1780d0e22b8340e20cd53f5f6R834)

framework/py/flwr/cli/chat/chat_connector.py
danielnugraha: I think this is not needed. It's just to indent the text correctly.

framework/py/flwr/cli/chat/chat_app.py
danielnugraha: This is not needed since disconnecting requires webui and we don't expect flwr chat and webui to be used simultaneously. It will still cause error so the user needs to remove this on their own.

dstripelis (approved):

panh99 (approved):

PR #8151, opened 2026-09-14
refactor(framework): Disable `--plugin-type` for `SuperExec`

jafermarq:
Passing `--plugin-type` to `flower-superexec` is no longer needed because the `flwr-xyz` process it spawns is based on the `TaskType` . Therefore, we are deprecating its use. Passing a plugin has no effect and throws a warning.


panh99 (approved):
LGTM!

PR #8166, opened 2026-09-15
feat(framework): Add output schemas to function tools

jafermarq:
Allow `function_tool` to include an optional `output_schema` but leave existing tool schemas unchanged when it is omitted


danielnugraha (approved):

panh99 (approved):
LGTM!

PR #8170, opened 2026-09-16
refactor(framework): Remove server-app task type check from runtime handlers

jafermarq:
`get_authenticated_serverapp_run_id` guarded several Runtime API handlers to reject non-SERVER_APP tasks with `RUNTIME_ENDPOINT_UNAVAILABLE`, but task-token authentication already scopes these calls to the task's own run. Remove the redundant check and use task.run_id directly, dropping the now-unneeded endpoint-unavailable constant and its integration test coverage.


panh99 (dismissed):
LGTM!

panh99 (approved):

PR #8175, opened 2026-09-16
feat(framework): Introduce strict task identity

panh99:
### Description

Task processes currently populate message identity metadata in multiple push paths. This creates scattered sources of truth, and infrastructure processes can accidentally construct task-bound messages with incomplete identity metadata.

### Related issues/PRs

Supersedes #8173.

### Explanation

Introduce a process-local `TaskIdentity` for `flwr-*` task processes and initialize it before loading user code, including ServerApp simulation. Its class properties raise `RuntimeError` when accessed before initialization, so task-bound `Message` and `JSONMessage` construction cannot silently fall back to zero or missing IDs.

Use explicit metadata factories only for non-reply messages created by SuperLink/SuperNode infrastructure, where no task identity is available. Reply messages continue to use `Message(..., reply_to=...)` directly because their complete metadata is derived from the original message. Task processes, user code, and simulation construct ordinary messages from the initialized `TaskIdentity`.

Add a typed `make_json_message` factory for infrastructure JSON messages. The Responses runtime validates and normalizes the payload before creating its child task, then constructs the request once with the final task and node identity.

Remove redundant identity mutation from `InMemoryGrid.push_messages`. Existing tests are updated only where they bypass task-process startup and therefore need explicit identity or metadata; no new unit tests are added.


framework/py/flwr/app/message/message.py
```diff
@@ -445,7 +447,9 @@ def inflate(
 
 
 def make_message(
-    metadata: Metadata, content: RecordDict | None = None, error: Error | None = None
+    metadata: Metadata,
+    content: RecordDict | None = None,
+    error: Error | None = None,
```
panh99:
```suggestion
    metadata: Metadata, content: RecordDict | None = None, error: Error | None = None
```

jafermarq (approved):

danieljanes (approved):

PR #8177, opened 2026-09-16
feat(framework): Align Notion connector actions

danielnugraha:
## Summary

Align the Notion connector actions with the native Notion API contract.

  - Rename notion_get_page_content to notion_get_page.
  - Update notion_search inputs:
      - Replace limit with page_size.
      - Replace cursor with start_cursor.
      - Add native filter and sort objects.

  - Update notion_get_page to return:
      - Page metadata.
      - All direct child blocks.

  - Follow next_cursor internally until all child-block pages are retrieved.
  - Preserve Notion error codes and messages in connector errors.
  - Add focused tests for action registration, request forwarding, pagination, and error handling.


framework/py/flwr/supercore/task_process/connector/notion/actions.py
```diff
@@ -14,44 +14,63 @@
...
+                    "description": "The filter object to narrow results.",
+                },
+                "sort": {
+                    "type": "object",
+                    "description": "The sort object to order results.",
+                },
+                "page_size": _PAGE_SIZE,
+                "start_cursor": _CURSOR,
             },
             "required": ["query"],
```
panh99: ChatGPT spotted that Notion's `search` doesn't seem to require `"query"` field: https://developers.notion.com/reference/post-search . Is it true?

framework/py/flwr/supercore/task_process/connector/notion/executors.py
```diff
@@ -34,45 +34,72 @@ class NotionApiError(ConnectorApiError):
...
+    while block_children.get("has_more") is True:
+        cursor = block_children.get("next_cursor")
+        if not isinstance(cursor, str) or not cursor or cursor in cursors:
+            raise NotionApiError("invalid_response")
+        cursors.add(cursor)
+        block_children = _call_notion_api(
+            "GET",
+            f"/blocks/{page_id}/children",
+            context.credentials,
+            params={"start_cursor": cursor},
```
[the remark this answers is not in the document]
panh99: This sounds to me like the `get_page` will fetch all child pages internally. It can make an arbitrary number of sequential HTTP requests and return an arbitrarily large tool result. Plus, these sequential requests may hit Notion's rate limit. Do you think we need to handle this now, or this is rather an edge case?

framework/py/flwr/supercore/task_process/connector/notion/actions.py
```diff
@@ -14,44 +14,63 @@
...
+_CURSOR: JSONObject = {
+    "type": "string",
+    "description": "The cursor for pagination.",
+}
+_PAGE_SIZE: JSONObject = {
+    "type": "integer",
+    "minimum": 1,
+    "maximum": 100,
+    "description": "The number of results per page.",
+}
```
panh99: Could we use `integer_property` and `string_property` here?

framework/py/flwr/supercore/task_process/connector/notion/actions.py
```diff
@@ -14,44 +14,63 @@
...
+                    "description": "The search query text.",
+                },
+                "filter": {
+                    "type": "object",
+                    "description": "The filter object to narrow results.",
+                },
+                "sort": {
+                    "type": "object",
+                    "description": "The sort object to order results.",
+                },
```
panh99, line 107: Are the descriptions for `filter` and `sort` sufficient for model tool calls, or are they a bit too opaque? I'm asking because I personally wouldn't know what values to provide here by reading the descriptions, especially since both are typed as `object`.

framework/py/flwr/supercore/task_process/connector/notion/actions.py
danielnugraha: If no query is given, then it will return all pages and data source from the connection which is undesired behavior. I would create a separate action like recent_search with no query for this.

framework/py/flwr/supercore/task_process/connector/notion/executors.py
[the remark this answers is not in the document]
danielnugraha: We don't specify any limit for the child blocks. So only Notion default pagination 100 will be given. Unless its a very big document, I think it's okay. I would rather get a rate limited error than the model says I can't find your document for now.

framework/py/flwr/supercore/task_process/connector/notion/actions.py
danielnugraha: After implementing separate action, it seems to duplicate a lot from this action and we can merge the description into one.

panh99 (approved):
LGTM!

PR #8181, opened 2026-09-17
refactor(framework): Extend `push_messages` tool schema

jafermarq:
Added optional `reply_to_message_id` input argument


framework/py/flwr/supercore/task_process/agent/grid.py
```diff
@@ -282,6 +285,9 @@ def _push_messages(self, messages: list[JSONObject]) -> JSONObject:
                 group_id="",
                 ttl=ttl,
             )
+            reply_to_message_id = cast(str | None, item.get("reply_to_message_id"))
+            if reply_to_message_id is not None:
+                message.metadata.__dict__["_reply_to_message_id"] = reply_to_message_id
```
panh99, line 292: I just checked `HttpGrid.push_messages`, and it seems to always check if the message's `reply_to_message_id` field is empty: https://github.com/flwrlabs/flower/blob/main/framework/py/flwr/superlink/grid/http_grid.py#L282

I think we have the check when the Grid only sends instruction messages, but since now we are allowing SuperNode to send reply messages via the Grid, we could prolly remove the check. Wdyt?

framework/py/flwr/supercore/task_process/agent/grid.py
```diff
@@ -94,6 +94,9 @@ def _grid_tools() -> list[JSONObject]:
                                 "to preserve precision."
                             ),
                             "payload": string_property("String payload to send."),
+                            "reply_to_message_id": string_property(
+                                "Optional ID of a message this message replies to."
```
panh99: Wdyt?
```suggestion
                                "ID of the message being replied to. Required when replying to  "
                                "another message; otherwise, this field must not be set."
```

panh99 (approved):
LGTM!

PR #8184, opened 2026-09-17
feat(framework): Start model worker process on pod startup

charlesbvll:
## Summary

- Prestart a single-use Model worker in warm Model Pods to remove Python startup from task dispatch
- Preserve stdin-only task authority, fresh Runtime state, graceful shutdown, and bounded/redacted output forwarding
- Keep cold Model, AgentApp, and Connector behavior unchanged

## Performance

A matched local Kubernetes benchmark reduced median time to first text from 9.283s to 8.301s (-0.982s, -10.6%).


framework/py/flwr/supercore/model_worker.py
```diff
@@ -0,0 +1,378 @@
...
+from flwr.supercore import log
+from flwr.supercore import model_worker_protocol as protocol
+from flwr.supercore.warm_executor_constants import (
+    WARM_EXECUTOR_BUSY_FILE,
+    WARM_EXECUTOR_READY_FILE,
+    WARM_MODEL_EXECUTOR_SOCKET,
+)
+
+_OUTPUT_SEND_TIMEOUT_SECONDS = 0.1
+_RunModelOnce = Callable[
```
danielnugraha: Can we make this an explicit protocol to make it readable?

framework/py/flwr/supercore/model_worker.py
```diff
@@ -0,0 +1,378 @@
...
+class ModelInvocation:
+    """Describe one token-scoped invocation for the prestarted Model worker."""
+
+    token: str
+    runtime_api_address: str
+    insecure: bool
+    root_certificates_path: str | None
+
+    @classmethod
+    def from_json(cls, raw: str) -> ModelInvocation:
```
danielnugraha: I think this is unused? Ctrl+F returned 1 match only.

PR #8185, opened 2026-09-17
refactor(framework): Move `AgentApp` constant keys

jafermarq:


framework/py/flwr/supercore/constant.py
```diff
@@ -216,6 +216,9 @@ class InvitationStatus(StrEnum):
 
 FLOWER_AGENT_APP_ID = "@flwrlabs/flwr-agent"
 
+AGENT_GRID_MESSAGE_PAYLOAD_RECORD_KEY = "payload"
+AGENT_GRID_MESSAGE_PAYLOAD_JSON_KEY = "json"
```
panh99: Just an idea: since we’re not actually using JSON atm, and these constants will also be used outside `Grid`—for example, in SuperLink when creating the initial instruction message—should we consider something like:

```python
AGENT_MESSAGE_CONTENT_RECORD_KEY = "agent"
AGENT_MESSAGE_TEXT_KEY = "text"
```

Then we’d have:

```python
message.content["agent"]["text"] = "This is an agent message"
```

If we want to extend this in the future, we could introduce additional keys, for example:

```python
message.content["agent"]["text"]
message.content["agent"]["json"]
message.content["agent"]["image"]
```

Wdyt?

panh99 (approved):
LGTM!

PR #8191, opened 2026-09-17
feat(framework): Add filesystem connector with `list_directory` and `read_file` tools

jafermarq:
Adds a read-only filesystem connector under py/flwr/supercore/task_process/connector/filesystem/ with separate filesystem_list_directory and filesystem_read_file tools.

Access is limited to absolute paths inside directories configured via FLWR_FILESYSTEM_ALLOWED_DIRS (a path-separator-separated list). Requested paths and configured roots are resolved before containment is checked, which rejects traversal and symlinks pointing outside an allowed directory. The resolved target is opened with O_NOFOLLOW to reject a last-component symlink substitution between validation and open. Parent-directory rename races require hostile filesystem control and are outside this connector’s threat model.

The connector follows the credential-free built-in path used by web_search, requires no OAuth flow, and is available on POSIX platforms with O_NOFOLLOW. File reads and directory scans are bounded, special files are not read, and focused tests cover core operations, containment, limits, credential-free execution, and runtime acceptance.


framework/py/flwr/supercore/task_process/connector/filesystem/executors.py
```diff
@@ -0,0 +1,213 @@
...
+        if fd is not None:
+            os.close(fd)
+    if len(entries) > _MAX_DIRECTORY_ENTRIES:
+        raise FilesystemApiError("too_many_entries")
+    raw = sorted(entries)
+    items: list[JSONObject] = []
+    for name in raw:
+        entry_path = os.path.join(resolved, name)
+        try:
+            entry_st = os.stat(entry_path, follow_symlinks=False)
```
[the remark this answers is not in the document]
jafermarq: Fixed in ae1ac76: the directory descriptor is now kept open through the metadata loop and stat lookups use `os.stat(name, dir_fd=fd)`, so they cannot escape the sandbox via a swapped intermediate directory. Path-based stat only remains on the Windows fallback branch.

framework/py/flwr/supercore/task_process/connector/filesystem/executors.py
```diff
@@ -0,0 +1,213 @@
...
+    rejected by the kernel instead of followed. Without O_NOFOLLOW
+    (Windows), falls back to a plain open; _safe_resolve remains the
+    sandbox control there.
+    """
+    if not _O_NOFOLLOW:
+        return os.open(resolved, flags)
+    components = resolved.split(os.sep)
+    parts = [c for c in components if c]
+    if not parts:
+        raise FilesystemApiError("access_denied")
```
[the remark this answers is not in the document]
jafermarq: Fixed in ae1ac76: when the resolved path has no components (root `/`), `_open_sandboxed` now opens `os.sep` directly instead of rejecting it, so an explicitly configured root directory can be listed.

framework/py/flwr/supercore/task_process/connector/filesystem/executors.py
```diff
@@ -0,0 +1,214 @@
...
+    stat fall back to path-based forms and _safe_resolve remains the
+    sandbox control.
+    """
+    path = require_string(arguments.get("path"), "Filesystem", "path")
+    allowed = _allowed_dirs(context)
+    resolved = _safe_resolve(path, allowed)
+    fd = None
+    try:
+        if _O_NOFOLLOW:
+            fd = _open_sandboxed(resolved, os.O_RDONLY)
```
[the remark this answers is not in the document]
jafermarq: Fixed in d99e3e90: `list_directory` now opens the target with `O_DIRECTORY | O_NONBLOCK`, so FIFOs and other non-directory targets fail promptly instead of blocking on the open.

framework/py/flwr/supercore/task_process/connector/filesystem/filesystem.py
```diff
@@ -0,0 +1,242 @@
...
+    for root in allowed:
+        normalized_root = os.path.normpath(root)
+        try:
+            if os.path.commonpath((normalized, normalized_root)) != normalized_root:
+                continue
+        except ValueError:
+            continue
+        root_fd = None
+        try:
+            root_fd = os.open(normalized_root, _O_SEARCH | _O_NOFOLLOW | _O_DIRECTORY)
```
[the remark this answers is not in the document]
jafermarq: Fixed in 3fd4feaa9: configured roots are now opened one component at a time from `/` with `O_NOFOLLOW`, so a replaced ancestor symlink is never followed. Added a regression that swaps the ancestor after request canonicalization and before the root walk.

framework/py/flwr/supercore/task_process/connector/filesystem/filesystem.py
```diff
@@ -0,0 +1,242 @@
...
+    """Open a relative path without following any of its components."""
+    parts = [] if relative == "." else relative.split(os.sep)
+    fd = os.dup(root_fd)
+    try:
+        for part in parts[:-1]:
+            new_fd = os.open(part, _O_SEARCH | _O_NOFOLLOW | _O_DIRECTORY, dir_fd=fd)
+            os.close(fd)
+            fd = new_fd
+        name = parts[-1] if parts else "."
+        new_fd = os.open(name, flags | _O_NOFOLLOW, dir_fd=fd)
```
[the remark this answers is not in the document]
jafermarq: Fixed in 3fd4feaa9: after the final open, the still-open parent descriptor is walked through `..` and compared by device/inode to the pinned root. If an intermediate directory moved outside during traversal, the target fd is closed and access is denied. Added a deterministic rename-during-open regression.

framework/py/flwr/supercore/task_process/connector/filesystem/filesystem.py
```diff
@@ -0,0 +1,293 @@
...
+
+def _open_sandboxed(path: str, allowed: list[str], flags: int) -> tuple[int, str]:
+    """Open an absolute path relative to a pinned allowed-directory fd."""
+    if not os.path.isabs(path):
+        raise FilesystemApiError("access_denied")
+    normalized = os.path.realpath(path)
+    for root in allowed:
+        normalized_root = os.path.normpath(root)
+        try:
+            if os.path.commonpath((normalized, normalized_root)) != normalized_root:
```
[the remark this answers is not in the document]
jafermarq: Fixed in 343c0fae8: configured roots are now explicitly required to be canonical absolute paths. Symlinked roots such as `/tmp` or `/var/run` fail with `invalid_config` instead of appearing configured and then denying every request. Added a regression for a symlinked configured root.

framework/py/flwr/supercore/task_process/connector/filesystem/filesystem.py
```diff
@@ -0,0 +1,293 @@
...
+    fd = os.dup(root_fd)
+    try:
+        for part in parts[:-1]:
+            new_fd = os.open(part, _O_SEARCH | _O_NOFOLLOW | _O_DIRECTORY, dir_fd=fd)
+            os.close(fd)
+            fd = new_fd
+        name = parts[-1] if parts else "."
+        new_fd = os.open(name, flags | _O_NOFOLLOW, dir_fd=fd)
+        try:
+            beneath = _is_beneath(fd, root_fd)
```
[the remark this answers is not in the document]
jafermarq: Fixed in 343c0fae8 without adding a Linux-only syscall dependency: after confirming the parent descriptor is beneath the pinned root, the code now compares the opened target fd device/inode with a fresh no-follow stat of its directory entry. The swap-out/open/restore sequence therefore fails because the opened fd no longer matches the in-root entry. The regression now performs that exact sequence.

framework/py/flwr/supercore/task_process/connector/filesystem/filesystem.py
```diff
@@ -0,0 +1,177 @@
...
+    if not os.path.isabs(path):
+        raise FilesystemApiError("access_denied")
+    resolved = os.path.realpath(path)
+    if not any(_is_beneath(resolved, root) for root in allowed):
+        raise FilesystemApiError("access_denied")
+    try:
+        # O_NOFOLLOW closes the common race where the resolved target is replaced
+        # with a symlink before it is opened. Parent-directory rename races require
+        # hostile filesystem control and are outside this connector's threat model.
+        return os.open(resolved, flags | _O_NOFOLLOW), resolved
```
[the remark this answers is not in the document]
jafermarq: This is intentionally outside the connector threat model after simplifying the PR. The sandbox protects against ordinary traversal mistakes, symlinks resolving outside configured roots, and final-component symlink substitution via O_NOFOLLOW. It does not attempt to defend against an actor concurrently replacing intermediate directories; doing so requires the substantially larger descriptor-walk/identity machinery removed in d42ee2f8c (or a platform-specific atomic primitive). The limitation is documented beside the open and in the PR description.

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -256,12 +261,18 @@ def __init__(  # pylint: disable=too-many-arguments
...
         self, *, name: str, call_id: str, arguments: JSONObject
     ) -> JSONValue:
-        """Create one connector response through a child connector task."""
+        """Create one connector response."""
         name = name.strip().lower()
-        create_res = self._stub.CreateTask(
-            CreateTaskRequest(
-                type=TaskType.CONNECTOR, connector_ref=get_connector_ref(name)
+        connector_ref = get_connector_ref(name)
+        if connector_ref == FILESYSTEM_CONNECTOR_NAME:
```
panh99: I wonder if we should rename it to `FILESYSTEM_CONNECTOR_REF` instead of `*_NAME`. Wdyt?

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -256,12 +261,18 @@ def __init__(  # pylint: disable=too-many-arguments
     def create_connector_response(
         self, *, name: str, call_id: str, arguments: JSONObject
```
panh99, line 261: Should we rename this to `tool_name`? @danielnugraha

framework/py/flwr/supercore/task_process/connector/registry.py
```diff
@@ -54,6 +54,12 @@
 _CREDENTIAL_CONNECTOR_REFS: dict[str, str] = {
     name: connector.ref for connector in CONNECTORS for name in connector.handlers
 }
+_BUILTIN_CONNECTOR_REFS = {
+    filesystem.FILESYSTEM_LIST_DIRECTORY_TOOL_NAME: (
+        filesystem.FILESYSTEM_CONNECTOR_NAME
+    ),
+    filesystem.FILESYSTEM_READ_FILE_TOOL_NAME: filesystem.FILESYSTEM_CONNECTOR_NAME,
+}
```
panh99: I wonder if we should merge the two dicts. It seems like we need a single mapping from `tool_name` to `connector_ref`.
```suggestion
_CONNECTOR_REFS = {
    filesystem.FILESYSTEM_LIST_DIRECTORY_TOOL_NAME: (
        filesystem.FILESYSTEM_CONNECTOR_NAME
    ),
    filesystem.FILESYSTEM_READ_FILE_TOOL_NAME: filesystem.FILESYSTEM_CONNECTOR_NAME,
    **_CREDENTIAL_CONNECTOR_REFS,
}
```

framework/py/flwr/supercore/task_process/connector/registry.py
```diff
@@ -95,12 +101,14 @@ def requires_connector_credentials(name: str) -> bool:
 
 
 def get_connector_ref(name: str) -> str:
-    """Resolve a connector tool name to its OAuth connector reference."""
-    return _CREDENTIAL_CONNECTOR_REFS.get(name, name)
+    """Resolve a connector tool name to its connector reference."""
+    return _CREDENTIAL_CONNECTOR_REFS.get(name, _BUILTIN_CONNECTOR_REFS.get(name, name))
```
panh99: Related to comment https://github.com/flwrlabs/flower/pull/8191/changes#r4061402186
```suggestion
    return _CONNECTOR_REFS.get(name, name)
```

framework/py/flwr/supercore/task_process/connector/registry.py
```diff
@@ -95,12 +101,14 @@ def requires_connector_credentials(name: str) -> bool:
...
 
 
 def get_connector_tools(connector_ref: str) -> list[JSONObject]:
     """Return model-facing tools for one built-in or OAuth connector."""
+    if connector_ref == filesystem.FILESYSTEM_CONNECTOR_NAME:
+        return filesystem.make_filesystem_tools()
```
panh99: Could we add this to `_BUILTIN_CONNECTOR_TOOL_FACTORIES` instead of creating a `if` block here?

framework/py/flwr/supercore/task_process/agent/session.py
panh99: Nvrm. We can do it in a separate PR.

framework/py/flwr/supercore/task_process/connector/filesystem/filesystem.py
```diff
@@ -0,0 +1,201 @@
...
+class FilesystemApiError(ConnectorApiError):
+    """Secret-safe file system access failure."""
+
+    provider = "Filesystem"
+
+
+def make_filesystem_tools() -> list[JSONObject]:
+    """Return the filesystem function tool schemas."""
+    if not os.getenv(FILESYSTEM_ALLOWED_DIRS_ENV):
+        return []
```
panh99: Let's not advertise the tool on unsupported platforms.
```suggestion
    if not _PLATFORM_SUPPORTED or not os.getenv(FILESYSTEM_ALLOWED_DIRS_ENV):
        return []
```

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -256,12 +261,18 @@ def __init__(  # pylint: disable=too-many-arguments
...
         name = name.strip().lower()
-        create_res = self._stub.CreateTask(
-            CreateTaskRequest(
-                type=TaskType.CONNECTOR, connector_ref=get_connector_ref(name)
+        connector_ref = get_connector_ref(name)
+        if connector_ref == FILESYSTEM_CONNECTOR_NAME:
+            return invoke_filesystem(
+                name,
+                arguments,
+                usage_recorder=TaskUsageRecorder(self._stub),
```
danielnugraha: Do we need `TaskUsageRecorder`? It seems to be discarded in `invoke_filesystem`.

framework/py/flwr/supercore/task_process/agent/session.py
```diff
@@ -256,12 +261,18 @@ def __init__(  # pylint: disable=too-many-arguments
     def create_connector_response(
         self, *, name: str, call_id: str, arguments: JSONObject
     ) -> JSONValue:
-        """Create one connector response through a child connector task."""
+        """Create one connector response."""
         name = name.strip().lower()
-        create_res = self._stub.CreateTask(
-            CreateTaskRequest(
-                type=TaskType.CONNECTOR, connector_ref=get_connector_ref(name)
+        connector_ref = get_connector_ref(name)
```
danielnugraha, line 265: Can we run filesystem with flwr-connector? Is there any drawbacks for this?

framework/py/flwr/supercore/task_process/connector/registry_test.py
```diff
@@ -30,3 +30,13 @@ def test_connector_tool_names_are_unique() -> None:
...
+
+
+def test_filesystem_registry() -> None:
+    """Filesystem should be local and require no OAuth credentials."""
+    assert "filesystem" not in registry.OAUTH_FLOWS
+    assert registry.get_connector_ref("filesystem_list_directory") == "filesystem"
+    assert registry.get_connector_ref("filesystem_read_file") == "filesystem"
+    assert not registry.has_builtin_connector("filesystem")
+    assert not registry.has_builtin_connector("filesystem_read_file")
+    assert not registry.has_builtin_connector("start_automation")
```
panh99: What's the purpose of this test?

framework/py/flwr/supercore/task_process/connector/registry.py
panh99: I think the underlying issue is that the current built-in connector abstraction implicitly assumes a 1:1 mapping between `connector_ref` and tool, since all existing built-in connectors only expose a single tool. OAuth connectors, on the other hand, already support one connector ref with multiple tools.

Filesystem is the first built-in connector that breaks that assumption, which is why we now need special handling here.

Would it make sense to generalize the built-in connector abstraction so that a connector ref can map to multiple tools as well? That would also remove the filesystem-specific branch. We can also handle this in a separate PR if we want to keep this one focused.

jafermarq: yeah, currently it assumes a 1:1; i'll take note of this and let's do it in a follow up PR so we remove the `if`

framework/py/flwr/supercore/task_process/connector/registry_test.py
jafermarq: ah, residual from an earlier stage of this branch (it went through a lot of chagnes). I'll remove it

framework/py/flwr/supercore/task_process/agent/session.py
jafermarq: it was a legacy from a previous state of this PR (where i in fact had a `flwr-connector`-like design). But the diff was quite big, so i opted for something more minimal as a v1. I've remove passing `TaskUsageRecorder`

danielnugraha (approved):

panh99 (approved):

PR #8195, opened 2026-09-18
feat(framework): Add ListAppAssociations proto scaffold

danielnugraha:
Add proto scaffold to list all of the federations an app is associated with.


panh99 (approved):
LGTM!

PR #8196, opened 2026-09-18
feat(framework): Add ListAppAssociations implementation

danielnugraha:
Implement `ListAppAssociations` across the Control API.

  - Adds reverse app-association lookup to the in-memory and SQL CoreState implementations.
  - Returns only federations accessible to the authenticated account.
  - Reports every accessible federation for the built-in Flower Agent, matching its implicit availability in `ListApps`.
  - Adds the protobuf-over-HTTP route and Control HTTP client method.
  - Adds coverage for state lookup, authorization filtering, routing, and the HTTP client.

Associations are returned newest first as federation IDs.


framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -1669,8 +1669,18 @@ def list_app_associations(
...
 ) -> ListAppAssociationsResponse:
     """List the caller's federations associated with an app."""
-    _ = request, account, state
-    raise NotImplementedError("ListAppAssociations is not implemented.")
+    flwr_aid = account.flwr_aid
+    state.federation_manager.ensure_default_federations_exist(flwr_aid=flwr_aid)
+    accessible_federation_ids = [
+        federation.id
+        for federation in state.federation_manager.get_federations(flwr_aid)
+    ]
```
panh99, line 1678: Will this include archived federations? Should this RPC return all federations including archived ones or not?

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -1669,8 +1669,18 @@ def list_app_associations(
...
+    if request.app_id == FLOWER_AGENT_APP_ID:
+        return ListAppAssociationsResponse(federation_ids=accessible_federation_ids)
```
panh99: Let's move this logic inside `state.list_app_associations`. This is a SuperLink domain invariant and we can centralise it inside `LinkState/CoreState`.

panh99 (approved):
LGTM!

PR #8203, opened 2026-09-18
feat(framework): Align GitHub connector actions

danielnugraha:
## Summary

  Align the GitHub connector actions with GitHub’s native API contract and return provider responses without custom
  transformation.

  - Rename get_file_content to get_file_contents.
  - Update search_code to accept GitHub search syntax directly.
  - Add native search options:
      - sort
      - order
      - per_page from 1–100
      - Positive page values without an artificial upper bound

  - Return the Contents API response unchanged instead of decoding or replacing its Base64 content.
  - Reject . and .. repository path segments to prevent paths from escaping the Contents endpoint.
  - Preserve GitHub error messages in connector errors.
  - Add focused coverage for raw file responses, pagination, invalid paths, parameter bounds, and API diagnostics.


framework/py/flwr/supercore/task_process/connector/github/actions.py
```diff
@@ -27,31 +27,51 @@
...
+                },
+                "order": {
+                    "type": "string",
+                    "enum": ["asc", "desc"],
+                    "description": "Sort direction.",
+                },
+                "per_page": {
+                    "type": "integer",
+                    "minimum": 1,
+                    "maximum": 100,
```
charlesbvll: Should we move this into a constant?

framework/py/flwr/supercore/task_process/connector/github/executors.py
```diff
@@ -43,53 +39,50 @@ class GitHubApiError(ConnectorApiError):
...
+        if string_value := optional_string(arguments.get(name), "GitHub", name):
+            params[name] = string_value
+    for name in ("per_page", "page"):
+        if name in arguments:
+            integer_value = arguments[name]
+            if (
+                isinstance(integer_value, bool)
+                or not isinstance(integer_value, int)
+                or integer_value < 1
+                or (name == "per_page" and integer_value > 100)
```
charlesbvll: And update this

framework/py/flwr/supercore/task_process/connector/github/executors.py
```diff
@@ -43,53 +39,50 @@ class GitHubApiError(ConnectorApiError):
...
+            ):
+                constraint = "between 1 and 100" if name == "per_page" else "positive"
```
charlesbvll: And this

charlesbvll (approved):

PR #8210, opened 2026-09-21
feat(framework): Update connector table to federation-scoped

danielnugraha:
## Summary
Updated connector table to federation-scoped.

  - Auto-incrementing connector_id primary key
  - federation_id
  - created_at
  - deleted_at
  - created_by
  - deleted_by
  - Unique constraint on (flwr_aid, connector_ref)

  The lifecycle fields remain nullable because the current OAuth flow does not yet supply them.

  Updated schema docs and connector lookup behavior. Migration upgrade/downgrade preserves existing connector rows.


framework/py/flwr/server/superlink/linkstate/linkstate_test.py
```diff
@@ -2615,21 +2615,30 @@ def test_get_connector_refreshes_cached_row_in_shared_session(self) -> None:
...
             state.upsert_connector(
-                flwr_aid="account-a",
+                federation_id="fed-a",
                 connector_ref="calendar",
                 credentials_json='{"token":"new"}',
                 config_json='{"calendar":"work"}',
+                created_by="account-a",
+            )
+            second = state.get_connector(
+                federation_id="fed-a", connector_ref="calendar"
```
panh99, line 2640: Should we use a valid `federation_id` like `@bob/fed-a`?

framework/py/flwr/supercore/corestate/corestate_test.py
panh99, line 1: Same as above. Should we still follow the federation ID format `@<account>/<federation-name>` in the tests?

framework/py/flwr/supercore/state/schema/corestate_models.py
```diff
@@ -157,14 +158,22 @@ class FederationApp(FlwrBase):
 
 
 class Connector(FlwrBase):
-    """Represent connector configuration for an account."""
+    """Represent connector configuration for a federation."""
 
     __tablename__ = "connector"
+    __table_args__ = (UniqueConstraint("federation_id", "connector_ref"),)
```
panh99, line 164: Do we need this constraint? I thought the goal is to allow one connector to be added multiple times, e.g., connecting to multiple Slack accounts.

framework/py/flwr/superlink/servicer/control/control_handlers.py
```diff
@@ -253,7 +264,7 @@ def disconnect_connector(
     account: AccountInfo,
     state: LinkState,
 ) -> DisconnectConnectorResponse:
-    """Delete one account-scoped connector connection."""
+    """Delete one connector from the account's personal federation."""
```
panh99, line 267: Since now the connector is federation-scoped, should we require a `federation_id` field in `DisconnectConnectorRequest` instead of always assuming the caller's personal federation?

framework/py/flwr/supercore/state/schema/corestate_models.py
danielnugraha: I wanted to remove the constraint when we start implementing multiple connections for a single connector. Or at least when we change the other table to store `connector_id` instead of `connector_ref`.

panh99 (approved):
LGTM!

PR #8217, opened 2026-09-21
feat(framework): Add `get_nodes` method to `Grid`

jafermarq:
This enables getting access to the full `NodeInfo` object when the `Grid` samples nodes.


panh99 (approved):
LGTM!

PR #8218, opened 2026-09-21
refactor(framework): Update `get_nodes` tool for `AgentApp`

jafermarq:
The tool uses the new `get_nodes` method in `HttpGrid`. Responses to calling the `tool` are in the format:

```python
{
  "nodes": [
    {"id": "123", "name": "Node A", "location": "London"},
    {"id": "456", "name": null, "location": null}
  ],
  "num_available": 2
}
```


panh99 (approved):
LGTM!

